Issue #639, opened 2015-01-22
Malformed value for push.default: simple error: Must be one of nothing, matching, tracking or current


nblock:
During creation of a new subproject on a fresh weblate 2.1 installation, I came accross the following issue: 
```
We couldn't find any code matching 'Malformed value for push.default: simple error: Must be one of nothing, matching, tracking or current'
```

PR #640, opened 2015-01-22
Change push.default to current (Fixes: #639).

nblock:
A quick and dirty fix for issue #639. It works for my setup, but I haven't tested it extensively.

OS: Debian Wheezy
Git: 1.7.10.4


nijel:
Merged, thanks. This indeed seems to proper solution for supporting older Git versions, which do not support simple push strategy.

PR #655, opened 2015-02-17
Basic Gerrit Support

madmuffin1:
Gerrit is a very popular Code Review Tool. It is a git under the hood but uses a specific review command instead of pushes.

To make use of it, the repository needs to contain a .gitreview file as defined at https://pypi.python.org/pypi/git-review#gitreview-file-format

I am rather unsure about the tests, since Gerrit is just with a different push workflow, I just added a class overriding the `VCSGitTest` that basically does the same as the Git Tests but uses a different push method.

### DIFF
--- a/weblate/trans/tests/test_vcs.py
+++ b/weblate/trans/tests/test_vcs.py
@@ -19,7 +19,7 @@
 #
 
 from weblate.trans.tests.test_models import RepoTestCase
-from weblate.trans.vcs import GitRepository, HgRepository, RepositoryException
+from weblate.trans.vcs import GitRepository, HgRepository, RepositoryException, GitWithGerritRepository
 
 import tempfile
 import shutil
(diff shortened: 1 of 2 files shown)


weblate/trans/tests/test_vcs.py
```diff
@@ -228,6 +228,12 @@ def test_configure_branch(self):
         )
 
 
+class VCSGerritTest(VCSGitTest):
+    _class = GitWithGerritRepository
+    _vcs = 'git'
+    _branch = 'default'
```
nijel: I think the correct branch for Git is master, default is probably copied from Mercurial testcase...

madmuffin1: you are right. corrected with subsequent commit.

nijel:
Merged, thanks.

PR #664, opened 2015-03-03
allow creating new languages when using Java Property files

madmuffin1:
It's as simple as this. The repository maintainer needs to make sure, that the framework in use is using fallback to some base language, but that is a very common behaviour with Java Frameworks

### DIFF
--- a/weblate/trans/formats.py
+++ b/weblate/trans/formats.py
@@ -792,6 +792,21 @@ class PropertiesUtf8Format(FileFormat):
     loader = ('properties', 'javautf8file')
     monolingual = True
 
+    @classmethod
+    def supports_new_language(cls):
+        '''
+        Checks whether we can create new language file.
+        '''
+        return True
+
+    @staticmethod
+    def add_language(filename, code, base):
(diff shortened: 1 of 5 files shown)


weblate/trans/formats.py
```diff
@@ -809,6 +824,21 @@ class PropertiesUtf8Format(FileFormat):
...
+        return True
+
+    @staticmethod
+    def add_language(filename, code, base):
+        '''
+        Adds new language file.
+        '''
+        with open(filename, 'w') as output:
+            output.write('\n')
+
```
nijel: I think it's better idea to share the code, so can you please rebase your changes on current master where I've changed the properties classes to inherit?

madmuffin1: Sure. I thought about something similar aswell but wanted to keep it as simple as possible. Let's see if I can fix this just now.

nijel:
Thanks, I've cherry picked your changes and fixed testsuite.

madmuffin1:
Thanks!

PR #672, opened 2015-03-11
Changed method for getting config data for Mercurial.

buzbohdan:
*config* command is not part of vanilla Mercurial, so usage of *hg config* causes error. Working with Mercurial config file was written by analogy to set_config method of HgRepository class. Also, renamed *key* variable to more canon *option*

### DIFF
--- a/weblate/trans/vcs.py
+++ b/weblate/trans/vcs.py
@@ -681,28 +681,29 @@ class HgRepository(Repository):
         """
         Reads entry from configuration.
         """
-        try:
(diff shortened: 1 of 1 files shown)


nijel:
Looks like it does not corectly handle unicode strings, see test failures.

buzbohdan:
I forgot to remove print, will fix it tomorrow

nijel:
Thanks, looks good now.

buzbohdan:
And thank you for this great project!

PR #682, opened 2015-03-26
Add permission to save template (source) translation

notz:
In some cases it will be very helpful to let only managers save template (source) translations. Regular  users (external) should be able to contribute to all other languages only.


nijel:
Further comments:

* Please follow coding style, you can use pep8 tool to check locally or see above reports generated by codacy/landscape.
* It would be great to have this new permission covered by testsuite, so that we can be sure it works properly.

notz:
I have no idea why, but on my machine most of the checks are failing (~ 500).  Therefore i'm not able to add some new checks for this feature. 

Codestyle should be ok now.

nijel:
I've adjusted the code for current master and merged, thanks.

PR #699, opened 2015-04-07
Fixed login hint if user is authenticated

notz:
Fixed login hint if user is authenticated but has no permission to save translation


nijel:
Merged, thanks.

PR #708, opened 2015-04-12
Openshift

dtschan:
Hi Michal

Since commit 7ba24ec Weblate can no longer be installed on OpenShift. This pull request contains the necessary adjustments to the OpenShift configuration to fix this. Some explanations:
As far as I can see setuptools doesn't support URLs in extra_requires. The `dependency_links` settings must be used instead. Because the version given in extra_requires must match the one in `setup.py` behind the given URL (currently still 1.0), I pinned the git commit to the current HEAD of the master branch of pyuca. Also see: https://pythonhosted.org/setuptools/setuptools.html#dependencies-that-aren-t-in-pypi
The other commit fixes the generation of the admin password on OpenShift which was broken (but didn't abort the installation) since commit 5758d22. 

Kind regards
  Daniel

### DIFF
--- a/openshift/install.sh
+++ b/openshift/install.sh
@@ -100,7 +100,7 @@ sh "python ${OPENSHIFT_REPO_DIR}/openshift/manage.py collectstatic --noinput"
 if [ ! -s $OPENSHIFT_DATA_DIR/.credentials ]; then
   echo "Generating Weblate admin credentials and writing them to ${OPENSHIFT_DATA_DIR}/.credentials"
   sh "python ${OPENSHIFT_REPO_DIR}/openshift/manage.py createadmin"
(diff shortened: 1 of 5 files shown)


openshift/install.sh
```diff
@@ -51,20 +51,20 @@ source $OPENSHIFT_HOMEDIR/python/virtenv/bin/activate
 
 cd ${OPENSHIFT_REPO_DIR}
 
-sh "python ${OPENSHIFT_REPO_DIR}/setup_weblate.py develop"
-
 # Pin Django version to 1.7 to avoid surprises when 1.8 comes out.
```
nijel: This is not needed, Django 1.8 is supported.

dtschan: Updated to Django 1.8. I'm pinning the Django version in the OpenShift configuration so that stable Weblate releases can still be installed on OpenShift when a newer, incompatible Django version is released.

nijel:
Merged, thanks.

Issue #709, opened 2015-04-13
Openshift and static files


nijel:
Currently the Openshift setup does not properly handle static files placed in the web server root - ``favicon.ico`` and ``robots.txt``.

Seen at https://github.com/nijel/weblate/issues/679#issuecomment-85029278

Might be something for @dtschan...

dtschan:
I'll look into this

nijel:
Thanks @dtschan

paulpach:
Just to confirm this worked for me.   Tx.

PR #712, opened 2015-04-14
Solution for issue #709: Openshift and static files

dtschan:
Hi Michal

Please note that the current master head doesn't install on OpenShift as the newly released python-social-auth 0.2.5 doesn't install. This doesn't seem to be related to OpenShift as I'm also unable to install python-social-auth 0.2.5 into a local virtualenv (IOError: [Errno 2] No such file or directory: 'social/tests/requirements.txt').

Kind regards
  Daniel


nijel:
Thanks.

PR #723, opened 2015-04-18
Add GitHub VCS provider.

### LINKED ISSUE #656: Github integration
It would be nice to have better integration with Github, so even when users don't give Weblate access to push directly to the repository, it would create pull request for the translations.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8674890-github-integration?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

fwalch:
Tries to fix #656. Uses the [`hub`](https://github.com/github/hub) command line tool and requires that is has been set up (either by creating `~/.config/hub` or setting `GITHUB_USERNAME` and `GITHUB_PASSWORD` environment variables). The most complicated part is probably getting the Github username configured for hub out of the config, which is needed because it's also the name of the remote created for the fork (`hub fork`). Maybe there's an easier way?

I thought about using a Github API module for Python instead of `hub`, but this would probably require some admin functions to manage Github API tokens.. using `hub` should be simpler.

I didn't test this too much (and didn't yet add any automated tests), but wanted to get some feedback.. so let me know what you think!

You can see an example pull request at https://github.com/fwalch/neovim/pull/11.

### DIFF
--- a/weblate/trans/vcs.py
+++ b/weblate/trans/vcs.py
@@ -21,6 +21,7 @@
 Minimal distributed version control system abstraction for Weblate needs.
 """
 import subprocess
+import os
 import os.path
 import email.utils
 import re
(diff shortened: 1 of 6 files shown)


nijel:
Looks good, but I need a bit more to decide if this is approach we want to use.

As for managing OAuth tokens, it would be just matter of requiring more privileges when authenticating through github (what we already support using python-social-auth).

weblate/trans/vcs.py
```diff
@@ -630,6 +631,75 @@ def push(self, branch):
 
 
 @register_vcs
+class GithubRepository(GitRepository):
+
+    name = 'GitHub'
+
+    _cmd = 'hub'
+    _hub_config_file = os.path.join(os.path.expanduser('~'),
+                                    '.config/hub')
```
nijel: Placing this to DATA_DIR might be better approach, as user home is not really usable location for most of hostings.

weblate/trans/vcs.py
```diff
@@ -630,6 +631,75 @@ def push(self, branch):
...
+    _hub_user_prefix = '- user: '
+
+    _hub_user = None
+    if 'GITHUB_USER' in os.environ:
+        _hub_user = os.environ['GITHUB_USER']
```
nijel: This should rather go to settings.py instead to have just one place to configure stuff. (See wammu/appsettings.py for handling default values and so on).

weblate/trans/vcs.py
```diff
@@ -630,6 +631,75 @@ def push(self, branch):
...
+    elif os.path.isfile(_hub_config_file):
```
nijel: ...and when having the setting in settings.py, I think this branch can be completely removed.

weblate/trans/vcs.py
fwalch: I think we can't change the location of this config file:

```
GitHub OAuth authentication
    Hub  will  prompt  for  GitHub  username  & password the first time it needs to access the API and exchange it for an OAuth token, which it saves in
    "~/.config/hub".

    To avoid being prompted, use GITHUB_USER and GITHUB_PASSWORD environment variables.
```

weblate/trans/vcs.py
fwalch: Thanks, I'll have a look.

weblate/trans/vcs.py
nijel: Then we should pass it by environment only and not use config file at all...

fwalch:
Updated to use `appsettings`.

I've been thinking whether using `hub` or an API client is easier.. for the implementation, it shouldn't make much of a difference (`hub fork` and `hub pull-request` would just be replaced with direct API calls).

Here's what I think a Weblate admin would need to configure for each case:

 * When using `hub`:
    * Install `hub` on server and configure for a GitHub user.
    * Set the same GitHub user as `GITHUB_USERNAME` in the appsettings.

 * Using [API client](https://developer.github.com/libraries/).
    * Set `GITHUB_USERNAME` in settings.
    * Set `GITHUB_TOKEN` in appsettings or use token of Weblate admin account (?)

Weblate could also use the GitHub token of each user (instead of a "global" token) to send the PR to GitHub (i.e. author, committer and PR author would be the same), but that would require each user authenticating through Github.

nijel:
Thanks, now it looks good. Can you please also document how to configure this to make it work?

fwalch:
Sure. Is the following enough?

* Add info about how sending PRs to Github works to http://docs.weblate.org/en/latest/admin/continuous.html
* Add a description of the config variable to http://docs.weblate.org/en/latest/admin/config.html
* Add notes about installing/configuring `hub` to http://docs.weblate.org/en/latest/admin/install.html

nijel:
Yes, that should do it.

fwalch:
Added a few sentences to the docu. Suggestions for improvements are welcome!

If this looks good, I'll squash the commits for merging.

nijel:
Merged, thanks!

fwalch:
Thanks! Do you plan to make this available on https://hosted.weblate.org/? I think it would be interesting for Neovim (ref https://github.com/neovim/neovim/issues/2150).

nijel:
Most likely yes, but it will take some time.

PR #724, opened 2015-04-18
Work in openshift with mysql and postgresql

paulpach:
This pull request changes the openshift configuration so weblate uses mysql or postgresql if the cartridges are installed.

To install weblate in openshift using mysql,  use the following command:

```rhc -aweblate app create -t python-2.7 -t mysql-5.5 --from-code https://github.com/paulpach/weblate.git --no-git```

to install weblate in openshift using postgresql, use the following command:
```rhc -aweblate app create -t python-2.7 -t postgresql-9.2 --from-code https://github.com/paulpach/weblate.git --no-git```

What this will _not_ do:
* if you currently have a sqllite based install and add the mysql cartridge,  the database migration needs to be done by hand
* the secure_db.py file is sqllite specific,  so this will leave the default admin password in mysql and postgresql
* if no database cartridge is specified,  it will use sqllite like before.

Another change it now does, is create the admin with a random password in a db independent way wether in openshift or not.


paulpach:
going to create a new one that fixed code style problems

PR #727, opened 2015-04-18
Use Mysql and postgresql in openshift if available

paulpach:
This pull request changes the openshift configuration so weblate uses mysql or postgresql if the cartridges are installed.

To install weblate in openshift using mysql,  use the following command:

```rhc -aweblate app create -t python-2.7 -t mysql-5.5 --from-code https://github.com/paulpach/weblate.git --no-git```

to install weblate in openshift using postgresql, use the following command:
```rhc -aweblate app create -t python-2.7 -t postgresql-9.2 --from-code https://github.com/paulpach/weblate.git --no-git```

What this will _not_ do:
* if you currently have a sqllite based install and add the mysql cartridge,  the database migration needs to be done by hand
* if no database cartridge is specified,  it will use sqllite like before.

This also changes how the admin password is created.   secure_db.py is not db independent and the password was actually not random.   Instead in this pull request,  whenever the admin account is created (wether in openshift or not),  a random password is assigned instead of 'admin'


nijel:
Thanks, however I had to make some changes. Most importantly the fixed password in createadmin is needed for appliance build, so I've added this option back. I've also added some brief documentation, but maybe it could use some clarification from somebody who actually knows openshift :-).

paulpach:
In that case,  I suggest a secure-by-default approach:

if you call:

python manage.py createadmin

then a random password is created and displayed.

if you call:

python manage.py createadmin --password 'Admin'

then whatever password you pass to it will be used.   This can be used for
the appliances.

nijel:
That's exactly what I've implemented :-).

paulpach:
Lol,  I should have looked before speaking :).   Tx Nijel

paulpach:
The documentation is pointing to my fork.

nijel:
@paulpach Already fixed 220b2aa6af4883826ed1d9e0ae682ac2b4b04e2a (I just to forgot reference this PR)

PR #728, opened 2015-04-20
Modifed google web translate to changed api

notz:
Seems that google has changed their API. I now always get error code 503. This a different approach to get only one translation.


nijel:
Merged, thanks.

PR #729, opened 2015-04-20
Allow adding new language on resx and osx strings (utf-8 only) file formats

notz:


nijel:
Thanks, I've refactored the code to have one generic implementation of this simple case of creating new translation in eb39f05.

PR #762, opened 2015-06-17
Pull-request is not working

c-geek:
Trying to use pull-request feature for https://github.com/ucoin-io/cutecoin did not work.

After few debugging, it turns out that `hub` was throwing an error because main branch of the fork (`somebranch`) was not updated, whereas fork branch (`somebranch-weblate`) of the fork was.

### DIFF
--- a/weblate/trans/vcs.py
+++ b/weblate/trans/vcs.py
@@ -678,6 +678,7 @@ class GithubRepository(GitRepository):
         """
         self.fork()
         fork_branch = '{0}-weblate'.format(branch)
+        self.push_to_fork(branch, branch)
         self.push_to_fork(branch, fork_branch)
         try:
             self.create_pull_request(branch, fork_branch)


nijel:
I don't really see need to have those two synchronized, what was the exact error you got?

Maybe @fwalch can comment on this as well..

c-geek:
When I try to do manually the same command as Weblate do:

```
$ git pull-request -m "Test pull-request" -b translations -h ucoin-bot:translations-weblate
Aborted: 1 commits are not yet pushed to ucoin-bot/translations
(use `-f` to force submit a pull request anyway)
```

That's probably what Weblate actually encounters, since when I add the code: `self.push_to_fork(branch, branch)` then I don't get any issue anymore.

nijel:
I think github is just confused in way which Weblate setups local repository here - it expects to have same push and pull URLs. I don't see need to maintain two identical branches in the GitHub clone, so I think adding -f to the command will do the right thing without need to push repository twice to GitHub. Can you please try that?

nijel:
For now I've fixed this with -f, please reopen this in case it won't work.

PR #774, opened 2015-07-09
add post_commit_script and post_push_script config

mihu:
Useful for notifying remote parties about changes done by weblate.


nijel:
The patch looks good. Can you please also add documentation for these new options?

It should be covered in:

* configuration directives - https://docs.weblate.org/en/latest/admin/config.html#std:setting-PRE_COMMIT_SCRIPTS
* component configuration - https://docs.weblate.org/en/latest/admin/projects.html#component
* maybe document some use cases in continuous translation as well - https://docs.weblate.org/en/latest/admin/continuous.html#processing

Thanks.

mihu:
I've updated the config and project doc pages. But I'm not totally sure what should be put in continuous.

nijel:
Thanks a lot! Merged.

PR #777, opened 2015-07-13
Openshift: do not change existing groups on update

phw:
This patch prevents updating existing user groups when running an update on Openshift. This preserves any changes made to the default group. Of course this means the administrator might need to manually assign new permissions to the default group, but I think that makes some sense.

I have also included a fix to the documentation regarding the `setupgroups --no-privs-update` option which of course could be applied separately.


nijel:
Thanks, I think it makes sense.

PR #778, opened 2015-07-14
Checks can be disabled by default

phw:
Adds a `default_disabled` field to the `Check` class which can be set to disable a check by default. This generalizes the current implementation for the format checks and allows custom checks that are disabled by default but can be explicitly enabled using the proper flag.

This can be used in custom checks as follows:

```
class FooCheck(TargetCheck):

    check_id = 'foo'
    name = _('Foo check')
    description = _('Your translation is foo')
    default_disabled = True

    # Real check code
    def check_single(self, source, target, unit, cache_slot):
        return 'foo' in target
```

If this checks gets included it will not be run by default, but one can explicitly set the `foo` tag for a unit. The existing checks in weblate/trans/checks/format.py have been migrated to use this new functionality.

I will write some tests for my own use using this functionality and if the tests have a wider use I will provide pull requests for them, too.

### DIFF
--- a/weblate/trans/checks/base.py
+++ b/weblate/trans/checks/base.py
@@ -32,17 +32,22 @@ class Check(object):
     target = False
     source = False
     ignore_untranslated = True
+    default_disabled = False
     severity = 'info'
 
     def __init__(self):
         id_dash = self.check_id.replace('_', '-')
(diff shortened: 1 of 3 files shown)


nijel:
Merged, thanks!

PR #794, opened 2015-08-04
fix typo

PeterTheOne:


docs/admin/continuous.rst
```diff
@@ -11,7 +11,7 @@ The complete process can be described in following steps:
 
 1. Developers make some changes and push them to the VCS repository.
 2. Optionally the translation files are updated (this depends on the file format, see :ref:`translations-update`).
-3. Weblate pulls changes from the VCS repository, see :ref:`update-vcs`.
+3. Weblate pulls changes from the VCS repository, see :ref:`update-vcs`.The set the base URL of y
```
nijel: This seems wrong...

PeterTheOne: true.

PeterTheOne: fixed in #795

PR #818, opened 2015-08-24
Make urls in comments clickable

lainbsd:
We often leave links to documentation or screenshots during discussion with translators in the comments.

This makes those links clickable so they don't have to be copy-pasted anymore.


nijel:
Merged, thanks!

Issue #605, opened 2014-11-26
Properly use static files framework in Django


nijel:
Weblate currently still uses MEDIA dirs to serve it's static files. This was true in some ancient Django version and we should really switch to static files usage. That would also make our setup much more standard.

The data files could be by default collected in DATA_DIR as defined in #604.

nijel:
Also while doing this, we should also use some solution for compression/concatenation of javascript and css files, for example https://github.com/evansd/whitenoise

nijel:
https://github.com/django-compressor/django-compressor might be better alternative as it's already packaged quite widely.

PR #827, opened 2015-08-31
Include static folder in .egg

quinox:
When installing the current git repo as a built package I ran into the exception `UncompressableFileError while rendering: 'bootstrap/css/bootstrap.css' could not be found in the COMPRESS_ROOT '...' or with staticfiles.`

This fixes it. (related: #605)


nijel:
Thanks!

PR #836, opened 2015-09-03
Update checks.rst

paour:
Commas -> dots


nijel:
Thanks for spotting this!

PR #851, opened 2015-09-17
OSX strings utf-16 format

pawel-furmaniak:
As outlined at https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/LoadingResources/Strings/Strings.html, apple recommends to use UTF-16 for *.strings files

### DIFF
--- a/weblate/trans/formats.py
+++ b/weblate/trans/formats.py
@@ -933,7 +933,7 @@ class StringsUtf8Format(FileFormat):
 
 
 @register_dialect
-class DialectStringsUtf8(DialectStrings):
+class DialectStringsUtf16(DialectStrings):
     name = "strings-utf16"
     default_encoding = "utf-16"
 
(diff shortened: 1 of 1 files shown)


nijel:
Thanks for the patch. Would you please consider submitting the dialect to [translate-toolkit][1]? I'd really like to see this upstream rather than in Weblate.

[1]:https://github.com/translate/translate/

weblate/trans/formats.py
```diff
@@ -932,6 +932,32 @@ class StringsUtf8Format(FileFormat):
     new_translation = '\n'
 
 
+@register_dialect
+class DialectStringsUtf8(DialectStrings):
```
nijel: I guess the name should be different.

pawel-furmaniak: Yes

pawel-furmaniak:
@nijel https://github.com/translate/translate/pull/3352

nijel:
The problem here is that we actually enforce iso-8859-1 on the properties. This is due to bug in ttkit (already fixed by me), which turned all iso-8859-1 files into utf-8, breaking the other tooling.

I've fix for this ready in git, but I'm unable to push it right now.

PR #871, opened 2015-10-12
Fix static file path in nginx configuration file

fengkaijia:
There's no folder named ```static``` under the ```data``` directory.

### DIFF
--- a/examples/weblate.nginx.conf
+++ b/examples/weblate.nginx.conf
@@ -4,17 +4,17 @@ server {
 	root /path/to/weblate/weblate;
 
 	location /favicon.ico {
-		alias /path/to/weblate/data/static/favicon.ico;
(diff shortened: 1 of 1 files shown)


nijel:
That folder would exist if you would follow installation instructions (it collects static files from both Weblate and Django), with your patch the files from Django would be missing making the admin interface look bad. See https://docs.weblate.org/en/latest/admin/install.html#serving-static-files

fengkaijia:
Found out. That command is too small and could be easily skipped in a very long Installation instructions. It should be highlighted.

nijel:
Indeed the installation chapter covers quite a lot of things, patches are welcome. Also there is quick guide not going much into details: https://docs.weblate.org/en/latest/admin/quick.html

PR #881, opened 2015-10-20
Skip migration for profile without email address

quinox:
## Desired situation

Using Weblate with a custom auth backend that doesn't provide email addresses.

## Current problem

The first user ever can login just fine. The second user triggers a crash during the login process:

```python
Internal Server Error: /cas/login/
  File "virtual/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 111, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "virtual/local/lib/python2.7/site-packages/cas_consumer/views.py", line 49, in login
    auth_login(request, user)
  File "virtual/local/lib/python2.7/site-packages/django/contrib/auth/__init__.py", line 103, in login
    user_logged_in.send(sender=user.__class__, request=request, user=user)
  File "virtual/local/lib/python2.7/site-packages/django/dispatch/dispatcher.py", line 198, in send
    response = receiver(signal=self, sender=sender, **named)
  File "virtual/local/lib/python2.7/site-packages/weblate/accounts/models.py", line 708, in post_login_handler
    uid=user.email,
...
  File "virtual/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 65, in execute
    return self.cursor.execute(sql, params)
IntegrityError: duplicate key value violates unique constraint "social_auth_usersocialauth_provider_2f763109e2c4a1fb_uniq"
DETAIL:  Key (provider, uid)=(email, ) already exists.
```

### DIFF
--- a/weblate/accounts/models.py
+++ b/weblate/accounts/models.py
@@ -717,7 +717,8 @@ def post_login_handler(sender, request, user, **kwargs):
 
     # Migrate django-registration based verification to python-social-auth
     if (user.has_usable_password() and
-            not user.social_auth.filter(provider='email').exists()):
+            not user.social_auth.filter(provider='email').exists() and
+            user.email):
         social = user.social_auth.create(
             provider='email',
             uid=user.email,


nijel:
Please note that Weblate uses the email address in VCS commits, so you might actually want to have the emails there. But otherwise the fix makes sense (I'll just wait for CI to see if it doesn't trigger some error in testsuite).

nijel:
...also I'm not sure if user preferences saving will work without valid email, have you tried that?

quinox:
You are correct, preferences can't be saved without a valid email address, which can't be added through the preferences dialogue either.

I haven't fully thought out how to deal with it, I think I'll just show a msg on screen if there's no valid email address urging the user to visit the admin to configure their user object correctly (they are all trustworthy).

nijel:
What auth backend are you actually using? Isn't there way to get email out of that?

quinox:
That would indeed be a nicer solution. Currently everything runs on CAS1 since we didn't really need anything more complicated but from the looks of it I can switch it over to CAS2 and provide the email address that way.

quinox:
Thanks for the feedback, I've solved my problem by upgraded our CAS server and implementation to work with CAS2 and putting the mail address in the extended attributes.

nijel:
Great. Maybe you could write down some documentation for CAS users? It would nicely fit to our authentication documentation: https://github.com/nijel/weblate/blob/master/docs/admin/auth.rst

PR #897, opened 2015-10-26
Document CAS authentication

quinox:
Documentation on how to get CAS working with Weblate (from #881).

I'm not sure how much of a disclaimer the section should get: while we're perfectly happy with our CAS setup it's hard to find mature CAS implementations and even proper, readable documentation about it. The whole CAS ecosystem feels ad-hoc and hacky. Also, using `django-cas-ng` with a CAS v3 server should work just fine but talking to a v2 server currently requires running `django-cas-ng` from git to extract the attributes (and the server needs to support the __right style__ of attributes. From my research there are at least 3 known ways of including them inside the XML).


nijel:
Thanks!

PR #899, opened 2015-10-27
Render Python blocks in the CAS documentation

quinox:
The python blocks are not showing up at https://docs.weblate.org/en/latest/admin/auth.html#cas-authentication - I suspects its because of these two missing newlines (is there an easy way to test this stuff with readthedocs?)


nijel:
Thanks for the fix. You can build the docs locally: ``make -C docs html`` (you need to have sphinx-doc installed)

PR #902, opened 2015-10-27
Remove duplicate groups from list_display

quinox:
Since [Django 1.5](https://github.com/django/django/commit/69ff1b7390e140f332a5aa55c44a091c838923fb) `groups` is part of `UserAdmin`'s `list_display` - the filter now shows up twice in the admin.


nijel:
Good catch. I've noticed that the groups filter is double listed there, but never got to figuring out why...

PR #907, opened 2015-10-29
Allow to add new languages with Qt TS files using lupdate command

ejjoman:
Allow to add new languages with Qt TS files using lupdate command (similar to PO file handling)


nijel:
Thanks!

PR #919, opened 2015-11-04
Make multiselects full-width for better look-and-feel

ejjoman:
* Make multiselects full-width for better look-and-feel
* Center the exchange icon between those two lists
* Fix: Change exchange icon from **\<li\>** to **\<span\>** because \<li\> needs \<ul\> or \<ol\> as parent

### DIFF
--- a/weblate/static/loader-bootstrap.js
+++ b/weblate/static/loader-bootstrap.js
@@ -751,7 +751,7 @@ $(function () {
         $(this).multiSelect({
             afterInit: function (target) {
                 $(target.children()[0]).after(
-                    '<div class="fa-multiselect"><i class="fa fa-exchange"></i></div>'
+                    '<div class="fa-multiselect"><span class="fa fa-exchange"></span></div>'
                 );
             }
         });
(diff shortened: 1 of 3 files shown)


nijel:
The icon was originally <i>, not <li>, so it should be perfectly valid there...

weblate/static/multi-select.css
```diff
@@ -1,6 +1,6 @@
 .ms-container{
   background: transparent url('../img/switch.png') no-repeat 50% 50%;
-  width: 370px;
+  width: 100%;
```
nijel: Please make the change rather in style-bootstrap.css. This file is taken verbatim from multi-select upstream and will be replaced on possible upgrades.

weblate/static/loader-bootstrap.js
```diff
@@ -751,7 +751,7 @@ $(function () {
         $(this).multiSelect({
             afterInit: function (target) {
                 $(target.children()[0]).after(
-                    '<div class="fa-multiselect"><i class="fa fa-exchange"></i></div>'
+                    '<div class="fa-multiselect"><span class="fa fa-exchange"></span></div>'
```
nijel: I don't see what is wrong on using <i> here....

ejjoman: Sorry, it was late last night :) somehow I read "li" instead of "i"...

nijel:
I've made the changes in a661389, thanks for suggestions.

Issue #864, opened 2015-10-06
enhanced length check


asoliverez:
we need to add a check for maximum length based on the style used, ie check the length for a specific font and size.
In order to do that, we would add the font, size, and max length as metadata in the source code, and it should be able to calculate if the translated string is the right length.
##

nijel:
What file format do you use? (I ask to see possible options for passing metadata.)

asoliverez:
The source code is mostly C code.
For the style, we use CSS style formatting.
This is GTK+ code, with CSS stylesheets.

I'll check

nijel:
I meant what file format do you use for translations :-).

asoliverez:
Sorry :)
We use gettext .po files

asoliverez:
Do you think this feature is worth implementing? Do you see it as useful for other projects?
Of course I'm open to award a bounty on it, if you think it's upstream-worthy

nijel:
I'm not really sure with font width calculation. It's really hard to do, especially for non latin fonts. Also to make it work Weblate would need the font locally.

Having text lenght limitation would be much easier to implement.

asoliverez:
We could start with text length check while I figure out a way for the font width calculation

El 15 de octubre de 2015 8:09:23 PM GMT+01:00, "Michal Čihař" <notifications@github.com> escribió:
>I'm not really sure with font width calculation. It's really hard to
>do, especially for non latin fonts. Also to make it work Weblate would
>need the font locally.
>
>Having text lenght limitation would be much easier to implement.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/nijel/weblate/issues/864#issuecomment-148491118

-- 
Sent from my mobile device. Please excuse my brevity.

asoliverez:
We do have a way to check font width length, but it's not cross-platform, and as you said, the server should have that font installed.
I'd split this into 3 tasks:
- add the metadata for styles (upstream worthy?)
- add the char length check (upstream worthy)
- add font width length check, based on previous check, we would add a call to a font renderer, and check for actual rendered length. In our case, it would be Pango on a Gnome-based system. This work is probably not worth upstreaming.

I can get someone to work on this next week, but would need some guidance to modify in the right places.

nijel:
[...]

PR #936, opened 2015-11-17
Additional flag to check max-length

justinjoy:
This patch is about handling 'maximum length' of translation which discussed in #864.
However, I am afraid that I got the point exactly and my approach is not graceful to implement initial version. Although basic function works, more exception handling is required.

To enable checking 'max-length', the follow steps are required.
 1. add 'weblate.trans.checks.chars.MaxLengthCheck' to CHECK_LIST variable at settings.py
 2. choose 'Maximum Length of Translation' to Check flags of "source / review " UI
 3. then edit value like "max-length:5"

@nijel, could you review my patch and advise me?

### DIFF
--- a/weblate/trans/checks/chars.py
+++ b/weblate/trans/checks/chars.py
@@ -19,7 +19,10 @@
 #
 
 from django.utils.translation import ugettext_lazy as _
-from weblate.trans.checks.base import TargetCheck, CountingCheck
+from weblate.trans.checks.base import (
+    TargetCheck, TargetCheckWithFlag, CountingCheck
+)
+import re
 
 
 class BeginNewlineCheck(TargetCheck):
@@ -360,3 +363,26 @@ class ZeroWidthSpaceCheck(TargetCheck):
         if self.is_language(unit, ('km', )):
             return False
         return (u'\u200b' in target) != (u'\u200b' in source)
+
+
+class MaxLengthCheck(TargetCheckWithFlag):
+    '''
+    Check for maximum length of translation.
+    '''
+    check_id = 'max-length'
+    name = _('Maximum Length of Translation')
+    description = _('Translation should not exceed given length')
+    severity = 'danger'
+    default_disabled = True
+
+    FLAGS_PAIR_RE = re.compile(r'\b([-\w]+):(\w+)\b')
+
+    def check_target_unit_with_flag(self, sources, targets, unit):
+        check_pair = set(self.FLAGS_PAIR_RE.findall('\n'.join(unit.all_flags)))
+        if len(check_pair) > 0:
+            check_value = max(
+                {(x) for x in check_pair if x[0] == self.check_id},
+                key=lambda v: int(v[1])
+            )[1]
+            return len(targets[0]) > int(check_value)
+        return False
--- a/weblate/trans/tests/test_chars_checks.py
+++ b/weblate/trans/tests/test_chars_checks.py
@@ -22,6 +22,7 @@
 Tests for quality checks.
 """
 
+from unittest import TestCase
 from weblate.trans.checks.chars import (
     BeginNewlineCheck, EndNewlineCheck,
     BeginSpaceCheck, EndSpaceCheck,
@@ -30,8 +31,9 @@ from weblate.trans.checks.chars import (
     EndEllipsisCheck,
     NewlineCountingCheck,
     ZeroWidthSpaceCheck,
+    MaxLengthCheck,
 )
-from weblate.trans.tests.test_checks import CheckTestCase
+from weblate.trans.tests.test_checks import CheckTestCase, MockUnit
 
 
 class BeginNewlineCheckTest(CheckTestCase):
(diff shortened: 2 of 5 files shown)


weblate/trans/checks/chars.py
```diff
@@ -360,3 +360,19 @@ def check_single(self, source, target, unit):
         if self.is_language(unit, ('km', )):
             return False
         return (u'\u200b' in target) != (u'\u200b' in source)
+
+class MaxLengthCheck(TargetCheck):
```
nijel: Codacy Issue found: [Empty class docstring](https://www.codacy.com/app/michal_2/weblate/file/1347549267/issues/source?bid=2038971&fileBranchId=2968054#l364)

weblate/trans/models/unit.py
```diff
@@ -899,11 +902,20 @@ def run_checks(self, same_state=True, same_content=True, is_new=False):
...
         )
 
+        check_pair = set(FLAGS_PAIR_RE.findall('\n'.join(self.all_flags)))
+
         # Run all checks
         for check in checks_to_run:
             check_obj = CHECKS[check]
+
+            if hasattr(check_obj, 'check_value'):
+                check_value = max({(x) for x in check_pair if x[0] == check}, key=lambda v:v[1])[1]
```
nijel: Codacy Issue found: [Exactly one space required after :](https://www.codacy.com/app/michal_2/weblate/file/1347549371/issues/source?bid=2038971&fileBranchId=2968054#l912)

weblate/trans/models/unit.py
```diff
@@ -899,11 +902,22 @@ def run_checks(self, same_state=True, same_content=True, is_new=False):
...
         # Run all checks
         for check in checks_to_run:
             check_obj = CHECKS[check]
+
+            if len(check_pair) > 0 and hasattr(check_obj, 'check_value'):
+                check_value = max(
+                    {(x) for x in check_pair if x[0] == check},
+                    key=lambda v: v[1]
+                )[1]
+                setattr(check_obj, 'check_value', check_value)
```
nijel: Passing the parameters as attributes is IMHO bad idea (the check objects are global, so they are shared between threads). I'd rather do the parsing in the check itself....

weblate/trans/checks/chars.py
```diff
@@ -360,3 +360,21 @@ def check_single(self, source, target, unit):
...
+    '''
+    check_id = 'max-length'
+    name = _('Maximum Length of Translation')
+    description = _('Translation should not exceed given length')
+    severity = 'danger'
+    default_disabled = True
+    check_value = 0
+
+    def check_target(self, sources, targets, unit):
+        return self.check_target_unit(sources, targets, unit)
```
nijel: Using parent check_target is better here, or at least you should keep option to ignore the check...

weblate/trans/models/unit.py
justinjoy: > I'd rather do the parsing in the check itself....

Could you tell me more details? I think I missed something and I haven't realized yet.

For me, the problem is that given flag string is like 'max-length:100' - key:value. To load check object by name, I tried to extract  key part from the string and the key part is used for loading a proper check object.

To make check object parse itself, each object should keep key:value string as parameter or attribute. If so, as I understand, it will be needed to create the 'max-length' check object per value because one check object can handle one parameter. e.g, "max-lengh:50" and "max-length:100",  might be different objects.

nijel: There is no object creation at this point (it would slow down the processing). The objects are created on application startup and used since then. So there will be just one length check and it should parse the limitation from the flags.

nijel:
Thanks, it looks great now! The only missing bit are tests for the new check...

weblate/trans/tests/test_chars_checks.py
```diff
@@ -233,3 +235,62 @@ def setUp(self):
         self.test_good_matching = (u'str\u200bing', u'str\u200bing', '')
         self.test_failure_1 = (u'str\u200bing', 'string', '')
         self.test_failure_2 = ('string', u'str\u200bing', '')
+
+
+class MaxLengthCheckTest(TestCase):
+    def setUp(self):
+        self.check = MaxLengthCheck()
+        self.test_good_matching = (
```
nijel: Codacy Issue found: [Trailing whitespace](https://www.codacy.com/app/nijel/weblate/file/1414943940/issues/source?bid=2038971&fileBranchId=2968054#l244)

weblate/trans/tests/test_chars_checks.py
```diff
@@ -233,3 +235,62 @@ def setUp(self):
...
+            'strings',
```
nijel: Codacy Issue found: [Trailing whitespace](https://www.codacy.com/app/nijel/weblate/file/1414943940/issues/source?bid=2038971&fileBranchId=2968054#l245)

justinjoy:
Whew, finally, I made it pass all test. @nijel could you review again? Thank you!

nijel:
Thanks, it looks great now. Can you please add docs as well for the new check (to [docs/user/checks.rst](https://github.com/nijel/weblate/blob/master/docs/user/checks.rst))? Thanks.

nijel:
Thanks a lot!

PR #944, opened 2015-12-03
Prevent lxml 3.5 or later to be used on OpenShift because its compila…

dtschan:
…tion needs more memory than small gears can provide.

You can only create small gears on free accounts. This may not be the best solution, but I don't want to spend to much time on the current OpenShift 2 support since OpenShift 2 is soon to be replaced by the new Docker-based OpenShift 3.


nijel:
Merged, thanks!

PR #951, opened 2015-12-10
Consider PyLibMCCache as a good cache backend

quinox:
[Since Django 1.3](https://docs.djangoproject.com/en/1.4/topics/cache/#memcached) Django ships with a pylibmc cache backend, so we can assume it's a good backend.


nijel:
Merged, thanks!

PR #957, opened 2015-12-17
MaxLengthCheck: allow spaces around colon

justinjoy:
It would make more sense to allow space characters around delimiter.


nijel:
I don't think this is a good idea, sticking to strict syntax seems better to me. And definitely it's not a good idea with newlines.

nijel:
Also this code handles only splitting strings already stored in the database and not parsing them out of translation files...

PR #981, opened 2016-01-27
Openshift optimize install

phw:
Even more changes to the Openshift deployment. When deploying on small gears, the deployment still can quickly run out of memory and this is another attempt to improve this. I wanted to keep this separate from #980 since theses fixes might need some discussion.

The changes:
* Before pip install stop both the mysql and the cron cartridges to free up resources. Keep the python cartridge running, so Weblate can display the update message
* Restart mysql for the execution of the management commands
* For restarting only stop the python cartridge, but keep mysql running. As the mysql cartridge can take quite some time to restart this reduces the full downtime of the system (where openshift reports just a 503 due to weblate not running)
* In addition the duplicate call to collectstatic was removed. I hope this was not intentional, but it looks ok to me ;)

### DIFF
--- a/openshift/install.sh
+++ b/openshift/install.sh
@@ -49,6 +49,10 @@ touch $OPENSHIFT_DATA_DIR/.install
 export PYTHONUNBUFFERED=1
 source $OPENSHIFT_HOMEDIR/python/virtenv/bin/activate
 
+# Stop unneeded cartridges to save memory
+gear stop --cart cron
+gear stop --cart mysql
+
 cd ${OPENSHIFT_REPO_DIR}
 
 # Pin Django version to 1.8 to avoid surprises when 1.9 comes out.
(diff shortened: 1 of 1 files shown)


nijel:
Looks good to me, but as I don't use OpenShift, I'd welcome other review as well. Maybe  @dtschan  can comment on this?

dtschan:
As far as I can see this implementation will break the default Weblate OpenShift installation which uses SQLite (which is recommended for small gears due to the memory constraints) as well as installations that use PostgreSQL. I recommend to improve this into a generic solution which works with all supported databases before merging.
Collectstatic is required as far as I remember. Haven't tested it now though.

phw:
> As far as I can see this implementation will break the default Weblate OpenShift installation which uses SQLite (which is recommended for small gears due to the memory constraints) as well as installations that use PostgreSQL. I recommend to improve this into a generic solution which works with all supported databases before merging.

Right, forgot about those. I will see how this can be solved generically.

> Collectstatic is required as far as I remember. Haven't tested it now though.

It's needed of course, but no need to run it two times ;)

nijel:
Any progress on making this work with SQLite?

phw:
Not yet, and it will take me some time until I come around dealing with Openshift again.

nijel:
Okay, for now I've cherry-picked https://github.com/nijel/weblate/commit/453afae15e969ed5eae3820a5bb45fb05301de32

nijel:
Closing this for now due to lack of progress.

PR #949, opened 2015-12-10
Add keyboard shortcuts and check highlights

lln-ijinus:
A proposition to add new shortcuts :
- Alt+v to copy source to target
- Esc then a number to remove a check
- Alt then a number to copy a machine translation
- Alt+(Char) to copy an highlighted check from source to cursor position

All these modifications are done to help translator to be efficient by not having to use the mouse.

To be able to add the last shortcuts, <span> are created around parts of the source string that matches a check. For now around source parts that checks formats( %s, {0}, etc.) and markup.

To do that, a new method "check_highlight" was added to the base check.

Most of the work is in the javascript to handle shortcuts when they are needed and have a little impact on the code.
For the shortcuts to remove checks, to copy a machine translation and to copy highlight checks where the shortcut is "variable", infos are given on the number on the given elements (<sup>1</sup>, etc.) and these informations are also added only in the javascript.

PR #982, opened 2016-01-27
Angularjs interpolation check

phw:
Adds a check that tries to validate AngularJS interpolation strings like e.g. `"Hello {{username}}"` (see https://docs.angularjs.org/api/ng/service/$interpolate). The check makes sure that any interpolation string used in the source is also used again in the target.

The check is disabled by default and can be added with the `angularjs-format` flag.

I actually found the checks API to be a bit confusing, so I hope I got it right. But this check is working fine for me in production for several months now.

### DIFF
--- a/weblate/trans/checks/angularjs.py
+++ b/weblate/trans/checks/angularjs.py
@@ -1,0 +1,69 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © 2012 - 2015 Michal Čihař <michal@cihar.com>
+# Copyright © 2015 Philipp Wolfer <ph.wolfer@gmail.com>
+#
+# This file is part of Weblate <http://weblate.org/>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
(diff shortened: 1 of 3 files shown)


phw:
I just saw that #949 got merged. I would actually like to add this to the angularjs-format check, too. I will update this PR as soon as I get it ready.

weblate/trans/checks/angularjs.py
```diff
@@ -0,0 +1,74 @@
...
+    '''
+    Check for AngularJS interpolation string
+    '''
+    check_id = 'angularjs_format'
+    name = _('AngularJS interpolation string')
+    description = _('AngularJS interpolation strings do not match source')
+    default_disabled = True
+    severity = 'danger'
+
+    def check_single(self, source, target, unit, cache_slot):
```
phw: Caching was removed from all checks in a0009086e4ead59e2c35bd6c18bd7975e8e853f9, this test should be updated.

weblate/trans/checks/angularjs.py
```diff
@@ -0,0 +1,74 @@
...
+        # Verify unit is properly flagged
+        if self.enable_string not in unit.all_flags:
```
phw: Is it right I can just remove this, as this is handled by the default implementation of `should skip`?

nijel:
Thanks, looks good. It would be nice to add highlighting for it as well.

phw:
> Thanks, looks good. It would be nice to add highlighting for it as well.

Yes, I actually looked into it. The highlighting API currently does not provide exactly what would be needed, as it has only access to the source. What I would like to do is to actually highlight the tags that are different in source and target.

I have put this on my to-do list, but not sure when I will find the time

nijel:
No, it should highlight all tags. Translators then can simply copy them from source (either by clicking or by Alt+NUMBER).

phw:
> No, it should highlight all tags. Translators then can simply copy them from source (either by clicking or by Alt+NUMBER).

I see, I misunderstood the feature. I originally thought it would highlight if the validation fails. But now I see that it is used to highlight the tags in the source and is used for the shortcut functionality. I will provide a PR providing the highlighting in the next couple of days.

PR #989, opened 2016-02-08
Custom views

### LINKED ISSUE #865: Provide custom views
The dashboard view should be configurable, we need at least following options:

* show all (accessible) translations (current way)
* show all translations in projects user has joined (see #568, should be default)
* show components matching given pattern
* show projects matching given pattern
* show selected translations within project/component

All of these will still filter for user language.

This should be configured in the user profile, the views would be defined in the admin interface.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27194466-provide-custom-views?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jitka:
This is our first attempt at writing custom views (should fix #865 when done).
We have extended the "Your translations" button. Now it is a drop-down that contains Projects, Your translations (all projects with your languages), Your subscriptions (subscribed projects, possibly filtered by language), and component lists.
We created a new model for component lists and the corresponding admin interface.
Component selection in the admin view is somewhat ugly (in Django 1.8 at least) -- do you know of a way to make the widget better? Now it's a plain `select` box where you have to ctrl+click everything. Maybe that thing from Users config for selecting user groups would be better.

Still missing is:
- user preference for the default view, plus a correct label for the dropdown :)
- filtering component lists by language. If that is necessary.
- maybe we should load component lists on demand, instead of hard-including into the index page, what do you think?
- cleanups, pylint, tests and stuff like that. This is mergeable, but we will move the migration into a single file, update to master, etc.

Please have a look at this and let us know whether this is a good way of doing it :)


weblate/html/index.html
```diff
@@ -42,43 +52,55 @@
 </li>
 </ul>
 
+<script>
+$("#views-menu li a").click(function(){
+  
+  $("#views-title").html($(this).text()+' <span class="caret"></span>');
+    
+  });
+</script>
```
nijel: Please do not use inline javascript, put the code to weblate/static/loader-bootstrap.js

nijel:
Generally this looks good.

As for the admin interface, you're probably looking for [ModelAdmin.filter_horizontal](https://docs.djangoproject.com/en/1.9/ref/contrib/admin/#django.contrib.admin.ModelAdmin.filter_horizontal) (I've not tested it, but sounds like what you're looking for).

Lazy loading of component list will be probably necessary if there is more of them, with few entries, it will probably just make things appear slower with additional request...

weblate/trans/context_processors.py
```diff
@@ -23,8 +23,10 @@
 import weblate
 from weblate import appsettings
 from weblate.trans.site import get_site_url
+from weblate.trans.models.componentlist import ComponentList
 from weblate.trans.models.project import Project
 from weblate.trans.models.translation import Translation
+from weblate.accounts.models import Profile
```
nijel: Codacy Issue found: [Trailing whitespace](https://www.codacy.com/app/nijel/weblate/file/2005162130/issues/source?bid=2038971&fileBranchId=3121328#l29)

weblate/trans/context_processors.py
```diff
@@ -40,17 +42,41 @@ def weblate_context(request):
...
+    userlanguages = None
+    active_tab_slug = Profile.DASHBOARD_ALL 
+
     if request.user.is_authenticated():
-        usertranslations = Translation.objects.filter(
+        active_tab_slug = request.user.profile.dashboard_view
+        if active_tab_slug == Profile.DASHBOARD_COMPONENT_LIST:
+            clist = request.user.profile.dashboard_component_list
+            active_tab_slug = clist.tab_slug()
+            dashboard_choices[active_tab_slug] = clist.name
```
nijel: Codacy Issue found: [Trailing whitespace](https://www.codacy.com/app/nijel/weblate/file/2005162130/issues/source?bid=2038971&fileBranchId=3121328#l60)

weblate/trans/context_processors.py
```diff
@@ -40,17 +42,41 @@ def weblate_context(request):
...
+    componentlists = ComponentList.objects.all()
+
+    dashboard_choices = dict(Profile.DASHBOARD_CHOICES)
 
     # Load user translations if user is authenticated
-    usertranslations = None
+    subscribed_projects = None
+    usersubscriptions = None
+    userlanguages = None
+    active_tab_slug = Profile.DASHBOARD_ALL
```
nijel: Codacy Issue found: [Trailing whitespace](https://www.codacy.com/app/nijel/weblate/file/2005162130/issues/source?bid=2038971&fileBranchId=3121328#l53)

weblate/accounts/models.py
```diff
@@ -496,6 +496,30 @@ class Profile(models.Model):
...
+        (DASHBOARD_LANGUAGES, _('Your languages')),
+        (DASHBOARD_ALL, _('All projects')),
+        (DASHBOARD_COMPONENT_LIST, _('Component list')),
+    )
+    dashboard_view = models.CharField(
+        max_length=100,
+        choices=DASHBOARD_CHOICES,
+        verbose_name=_('Default dashboard view'),
+        default=DASHBOARD_SUBSCRIPTIONS,
+    )
```
nijel: Wouldn't be better to store these choices as integers rather than strings? We will anyway use symbolic names instead of them and storing integer in the database is cheaper than string...

jitka: We use integer before and decided switched to string when we need to get slug from user.profile. I don't want to have mapping between object/integers and slug in two different files (index.html and that where I set active_tab_slug). 
Do you have idea how to avoid two mapping and use integers in models.py?

nijel: Alternative would be to use the number in slug like ``dashboard-1``.

nijel:
I think this looks good. Can you please also add documentation (both user and admin) for this feature?

weblate/accounts/models.py
matejcik: The problem is that these strings are listed explicitly in index.html, and also show up in URLs. "dashboard-1" would be pretty ugly for one, and confusing in the html code too. So we actually need the symbolic names _and_ the slug values. That is why we chose to put the slug values in the database directly; there is actually no place where numeric id is of any use.

We could decide to ignore the "ugly URL" thing, and use symbolic names in the html source, but that doesn't seem to lead to clarity either: `<a href="#dashboard-{{ Profile.DASHBOARD_SUBSCRIPTIONS}}">` is worse than `<a href="#subscriptions">` IMHO.

A possibility is to use a translation table between ids and slugs, or some sort of function to return slug given the ID, or label given the ID, and a list of slugs from which to generate the drop-down menu... I don't have a clear idea how to do this and where to put the resulting code though.
Hints? :)

Note that we expect the `active_tab` properties to be reusable on other pages, namely settings (#990). That's why we chose the generic names.

nijel: Indeed having readable URLs seems better.

How about following?

```py
    DASHBOARD_SUBSCRIPTIONS = 1
    DASHBOARD_LANGUAGES = 2
    DASHBOARD_ALL = 3
    DASHBOARD_COMPONENT_LIST = 4

    DASHBOARD_CHOICES = (
        (DASHBOARD_SUBSCRIPTIONS, _('Your subscriptions')),
        (DASHBOARD_LANGUAGES, _('Your languages')),
        (DASHBOARD_ALL, _('All projects')),
        (DASHBOARD_COMPONENT_LIST, _('Component list')),
    )

    DASHBOARD_SLUGS = {
        DASHBOARD_SUBSCRIPTIONS: 'your-subscriptions',
        DASHBOARD_LANGUAGES: 'your-languages',
        DASHBOARD_ALL: 'project',
        DASHBOARD_COMPONENT_LIST: 'list',
    }

    DASHBOARD_SLUGMAP = {
        d[1]: d[0] for d in DASHBOARD_SLUGS.items()
    }
```

weblate/trans/models/componentlist.py
```diff
@@ -0,0 +1,60 @@
...
+    )
+
+    components = models.ManyToManyField('SubProject')
+
+    def tab_slug(self):
+        return "list-" + self.slug
+
+    def clean(self):
+        if not self.name:
+            raise ValidationError(_('Name must be specified'))
```
nijel: Not needed, this is done automatically unless the field has ``blank=True``

weblate/trans/tests/test_views.py
```diff
@@ -399,6 +400,42 @@ def test_home_without_whiteboard(self):
         response = self.client.get(reverse('home'))
         self.assertNotContains(response, 'whiteboard')
 
+    def test_component_list(self):
+        clist = ComponentList(name="TestCL", slug="testcl")
+        clist.save()
+
+        response = self.client.get(reverse('home'))
+        self.assertContains(response, 'TestCL')
+        self.assertEquals(len(response.context['componentlists']), 1)
```
nijel: Please use assertEqual, the one with s is deprecated for quite some time (and in Python 3 it can cause noisy warnings).

matejcik:
documentation will come in a separate PR, both for this and for Group ACL
(as soon as I write it ;) )

nijel:
This looks almost good, though I still have few comments:

1. it defaults to no selection, I don't think that dashboard_component_list should allow nonevalues and should come with reasonable default
2. if there is nothing in the current list, I think should fall back to next one which contains something, it doesn't make much sense to present user empty page by default
3. there is minor merge conflict

As for (1), it's caused by intermediate migration and implicit conversions at the database (this seems to be true at least for MySQL, while sqlite will happily keep string value in the integer column). I've fixed this by squashing the migrations. For your test installation, you will have to revert to 0013 prior pulling changes (``./manage.py migrate accounts 0013``) and then apply new migration.

The (2) should be probably discussed separately if the fallback makes sense or it's better to force users to configure subscriptions.

And I've solved the merge conflict (3).

Anyway this is now merged, thanks for your work!

PS: Still the documentation for this is needed.

Issue #875, opened 2015-10-14
validate po file before commit/push


stanislav-brabec:
Weblate checks po files for some errors, reports them to the user, but it allows to ignore some of them.

If would be nice to mark checks that cause error in "msgfmt -c" as fatal, both in web interface and file upload.

Otherwise translator can easily break automated build process of the package.

Developer then must solve it in a way that is not friendly to thanslators: e. g. remove the file from the GIT repository.
##

stanislav-brabec:
Here is an example of such error:
https://github.com/openSUSE/libstorage/blob/7e556675fcdd0c88a28f51ef3d159e278c0d00ed/po/lt.po

Translator translated %2$s as %2$X and Weblate pushed broken translation.

matejcik:
would it be helpful (or sufficient), for the first step, to display a list of failing checks next to the "push" button?

scarabeusiv:
As a starter it might be good. But overall it needs to not allow the push if those checks fail ever.
On that note it should not allow the upload either. So if  you upload .po file the checks should be fatal too.

scarabeusiv:
I would guess as a check it would be good start. We still need to disallow the push if those checks fail.

Also the same apply for uploading .po file. If there are bugs, the update should be discarded.

aplanas:
AFAIS there are three elements in this issue:

1. Add an extra `msgfmt -c` check (some `weblate.trans.checks.format.MsgFmtCheck` class, disabled by default
2. Detect the placeholder bug (`%2$s` as `%2$X`)
3. Disable push if the PO still contains errors where `severity == 'danger'`

The feature in 2. looks like that is currently implemented in one of the `XXXFormatCheck` class, and is working in my local examples.

The feature 1 is a bit dangerous: on one side there is overlapping with 2, and `msgfmt -c` do not usually categorize the errors, even enumerate the total number of fatal errors. I guess that this can be fixed with some heuristics. Not sure that to do with the overlapping cases.

3 can make sense, but again we have 2, we forbid the push / upload from errors in the local checkers, or only for the ones found in msgfmt?

aplanas:
After meeting: the critical part is to avoid commits that are critical, not using the `msgfmt -c` check

stanislav-brabec:
aplanas: Exactly: Never allow to even commit broken translation that does not pass "msgfmt -c" check.

Once committed, it would be automatically pushed with the next commit (e. g. another language).

[...]

PR #1006, opened 2016-03-08
Do not proceed to the push if there are errors

aplanas:
If the translation contains important errors, the project,
subproject and translation push are forbiden.

Partial fix for #875

### DIFF
--- a/weblate/trans/models/project.py
+++ b/weblate/trans/models/project.py
@@ -433,6 +433,20 @@ class Project(models.Model, PercentMixin, URLMixin, PathMixin):
             ret |= component.can_push()
         return ret
 
+    def get_failing_translations(self):
+        """Return list of failing suproject and translations
+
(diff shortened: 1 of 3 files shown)


nijel:
At least you should check only for critical errors, failing checks can have different severities and most of them are really not that important to prevent push.

Not sure if it's intentional, but your change disables only explicit push using UI, but not when push happens automatically (push on commit is enabled).

aplanas:
@nijel but in the database critical fails are stored in the same field, aren't?

I need to check the code path for the automatic push and address it too.

aplanas:
* Fix PEP8 complains
* Update `push_if_needed` in subproject to disable push is `push_on_commit` is set in the project

Is this enough to disable the automatic push?

nijel:
Yes, that should do it...

PR #1039, opened 2016-03-23
Add new manager dashboard

aplanas:
- Add field in `language` model to signalize important languages
- Add field in `project` model to signalize projects where the
  important languages are tracked
- Update percentage mixin to allow filtering by important langs
- Add new manager dashboard


aplanas:
Still WIP, but submitted for early reviews.

weblate/trans/templatetags/translations.py
```diff
@@ -506,6 +506,15 @@ def translation_progress(translation):
 
 
 @register.inclusion_tag('progress.html')
+def important_translation_progress(translation):
```
nijel: I'd rather add parameter to existing tag than duplicating it's code.

weblate/html/list-important-projects.html
```diff
@@ -0,0 +1,25 @@
...
+<td class="percent">{{ percent }}%</td>
+</tr>
+{% endwith %}
+{% endfor %}
+</tbody>
+</table>
+
+<p>
+{% include "legend.html" %}
+</p>
```
nijel: This is almost same as list-projects.html, it would be better if the code can be shared...

weblate/trans/templatetags/translations.py
aplanas: Right.

aplanas:
* Remove duplicate code
* Use list-projects to list all and important projects

rvanlaak:
What would it take to finish the manager dashboard? :rocket:

nijel:
Not sure what is the code status, I really didn't look into that since last review. There is currently one minor merge conflict, but that is not hard to resolve.

PR #1067, opened 2016-04-11
Windows fix for popen and unicode_literals + import_project CLI

mchubby:
Fixes subprocess.call() on Windows due to use of future unicode_literals (see https://github.com/nephila/djangocms-installer/pull/228 and https://github.com/nvie/pip-tools/issues/209), when invoking manage.py migrate.

Also fixes `get_matching_files()` in import_project.py with glob.glob() returning paths with forward slashes.

EDIT: squashed two more commits for another glob() and os.path.realpath() calls.

### DIFF
--- a/weblate/trans/management/commands/import_project.py
+++ b/weblate/trans/management/commands/import_project.py
@@ -183,7 +183,7 @@ class Command(BaseCommand):
         Returns relative path of matched files.
         '''
         matches = glob(os.path.join(repo, self.filemask))
-        return [f.replace(repo, '').strip('/') for f in matches]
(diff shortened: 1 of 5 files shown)


nijel:
Thanks for your patch, however I've decided to fix this differently. Main problems with your approach:

- using ``str()`` all over the code makes it harder to read and also makes it easy to forget on this with future code changes
- unconditionally replacing ``\`` in the path can possibly break it for other platforms, where this is valid character in the path

My commits to address this issue:

- 7c0e724 Convert environment to string objects on Python 2 on Windows
- ab20f7f Use get_clean_env to process full environment for scripts
- 04cdc63 Use / as path separator

PR #1101, opened 2016-05-03
Spelling correction.

marcoslater:
Simple typing mistake I noticed whilst using this documentation for an install. :)


nijel:
Thanks!

PR #1104, opened 2016-05-04
Users admins actions to adding users to group.

jitka:
At current version there is no easy way to add many users to group. We add actions to users admin page to add selected users to group.
At the moment is created action for every group in model. Do you thing that is ok? 
If you want this we will add tests and finalize the pull request.


weblate/accounts/admin.py
```diff
@@ -92,6 +94,29 @@ def user_groups(self, obj):
         """
         return ','.join([g.name for g in obj.groups.all()])
 
+    def __init__(self, *args, **kwargs):
+        super(WeblateUserAdmin, self).__init__(*args, **kwargs)
+
+        groups = Group.objects.all()
```
nijel: Apparently you can not access database in admin __init__ as this is called on startup even before database migrations are executed, so it has to work completely without database. With your changes it's impossible to run initial migration (see errors in tests).

nijel:
Otherwise I think it's okay to have actions like this. 

BTW: You can also configure Weblate to add users to groups based on their email: https://docs.weblate.org/en/latest/admin/access.html#automatic-group-assignments

weblate/accounts/admin.py
jitka: I didn't expected that admin __init__ is called before database. I need run this code when the database is ready but before admin page is rendered. Do you have idea how to arrange it?

nijel:
It seems you've done something strange to the git repository, the diff and number of commits in this PR is huge...

jitka:
Hi, sorry It was unintentional. I get lost in all git branges. Please
ignore it.

nijel:
I think it's time to close this as it's not working and not mergeable. Also with current usage of groups it's probably not practical to add all of them to admin menu.

PR #1126, opened 2016-05-23
Fix missing typefaces in font-linux.css

paurullan:
There are no `font-linux.eot` typeface in the repo. While executing `collectstatic` and `compress` this raises an error.

This is a simple fix removing the missing font from the CSS.


nijel:
This should be already fixed since a4138834d779a572b97aab5b8e71ac6e09c4fbeb.

PR #1127, opened 2016-05-23
Simple health check urls

paurullan:
In order to be able to deploy weblate on a kubernetes we needed to add a health check URL.  

This is a simple check and I would love to get some feedback on the style and options.

Please tell me if you prefer to manage this on a issue and latter move it to a pull request.


nijel:
Is there some standard defined for this? Anyway I really don't see much need to have this configurable as I don't see when having this enabled would be problematic and the configurability just adds lot of complexity to the code...

Also I'd like to see this in documentation.

nijel:
Implemented without configuration.

PR #1128, opened 2016-05-23
[RFC - not to directly merge] First try on commands from API

paurullan:
In the mailing list we discussed on launching some of the `manage.py` maintenance commands from the REST API. This schema has come to light in order to be able to deploy weblate on a cluster where we can only have a single writer worker.

This is a first quick approach and I would love feedback to know if I should go on this way. I will obviously do the tests and documentation but first I wanted some kind of feedback.

There is a missing feature of enabling tokenized auth on the `/api/maintenance/x` methods.


nijel:
The authentication is there, all you need to do is check permissions.

Also I think these operations should use POST request rather than GET...

PR #1134, opened 2016-05-25
Doc: Add libjpeg as required package for installations using pip

nblock:
On Debian Jessie libjpeg-dev was required to install Pillow (as specified in requirements.txt). I don't have an OpenSUSE box at hand to test whether it is really called libjpeg-devel.


nijel:
Thanks, it seems correct.

PR #1135, opened 2016-05-25
Github - Fixes error when forking with multiple subprojects

erictrade:
### Scenario:
When a project has multiple subprojects (for example, multiple resource files), pushing the second subproject after pushing a first one fails. This happens because the forked branch has the same name for multiple subprojects (every branch is called master-weblate).

### Possible Solution:
These proposal changes makes the branch different for each subpreject, with this name pattern: "subproject-weblate" instead of "master-weblate".

I've tested in our enviroment and it works nicely. I don't know if this is the best solution, but I hope it can at least give a hint for fixing this problem.

### DIFF
--- a/weblate/trans/vcs.py
+++ b/weblate/trans/vcs.py
@@ -733,7 +733,7 @@ class GithubRepository(GitRepository):
         on fork and creates pull request against original repository.
         """
         self.fork()
-        fork_branch = '{0}-weblate'.format(self.branch)
+        fork_branch = '{0}-weblate'.format(self.name)
         self.push_to_fork(self.branch, self.branch)
         self.push_to_fork(self.branch, fork_branch)
         try:


weblate/trans/models/subproject.py
```diff
@@ -786,9 +786,10 @@ def do_push(self, request, force_commit=True, do_update=True):
 
         # Do actual push
         try:
+            fork_branch = '{0}-weblate'.format(self.name)
```
nijel: This really should live in the GitHub VCS backend, there is no need to put this to generic code.

weblate/trans/vcs.py
```diff
@@ -732,8 +732,11 @@ def push(self):
         Forks repository on Github, pushes changes to *-weblate branch
         on fork and creates pull request against original repository.
         """
-        self.fork()
         fork_branch = '{0}-weblate'.format(self.branch)
```
erictrade: maybe a change in just to this line, to something like

```
fork_branch = '{0}-weblate'.format(self.name)
```

I don't know the code so much to guess if this would put the subproject name instead of the branch name.

weblate/trans/models/subproject.py
erictrade: I agree. See my comment below.

weblate/trans/vcs.py
```diff
@@ -732,8 +732,11 @@ def push(self):
...
+        self.push(self, fork_branch)
+
+    def push(self, fork_branch):
```
nijel: Codacy Issue found: [method already defined line 730](https://www.codacy.com/app/nijel/weblate/file/2633791456/issues/source?bid=2038971&fileBranchId=3319319#l738)

erictrade:
Hi @nijel, what about this? I think it simplifies a lot. Do you think it might work?

weblate/trans/vcs.py
```diff
@@ -733,7 +733,7 @@ def push(self):
         on fork and creates pull request against original repository.
         """
         self.fork()
-        fork_branch = '{0}-weblate'.format(self.branch)
+        fork_branch = '{0}-weblate'.format(self.name)
```
nijel: This will not really improve situation as name in this case is VCS class name (GitHub).

erictrade: did you understand the scenario? for multiple subprojects, the fork branch name is created as master-weblate. For the first one, ok. But when resources are translated for the second one, it creates the same branch master-weblate and fails to push.
With the purposed fix, the branch name would be subproject1-weblate and subproject2-weblate, so it would not be the same branch name anymore.
Again, I don't know if it's the best solution at all, but do you have another idea to prevent this scenario error?
Just trying to help for future users, though. In our company we just edited the source code and it's working as we need. =)

nijel: The problem is that with this code you always get ``GitHub-weblate`` as branch name as you are using name of VCS class, not name of the component :-).

erictrade: I understand! Thanks for the explanation!!

nijel:
I've fixed this use case in git, however you might still do better job with sharing underlaying repository using weblate:// URL as repository, see https://docs.weblate.org/en/latest/vcs.html#weblate-internal-urls

erictrade:
Great! Thanks for the fixes and the tip to use internal URLs!

PR #1171, opened 2016-07-16
Subversion (git-svn) support

roidelapluie:


roidelapluie:
TODO:
- [ ] tests
- [ ] doc
- [ ] Use rebase instead of merge ; explicitely do not support merge
- [ ] Improve  configure_remote

docs/vcs.rst
```diff
@@ -229,3 +229,35 @@ Mercurial is another VCS you can use directly in Weblate.
...
+is a perl script that enables the usage of subversion witha git client, enabling
+users to have a full clone of the internal repository and commit locally.
+
+.. note:: 
+    Weblate supports only subversion repositories with standard layout (branches/,
+    tags/ and trunk/).
+
+.. _git-svn: https://git-scm.com/docs/git-svn
+
+.. _subversion: https://git-scm.com/docs/git-svn
```
nijel: This should probably point to http://subversion.apache.org/...

weblate/trans/tests/test_vcs.py
```diff
@@ -39,6 +39,9 @@ class GithubFakeRepository(GithubRepository):
     _version = None
     _cmd = get_test_file('hub')
 
+class GitSvnFakeRepository(GithubRepository):
```
nijel: Why basing on GithubRepository?

roidelapluie:
I will take a look at this tonight.

roidelapluie:
(and fix all the tests)

weblate/trans/vcs.py
```diff
@@ -674,6 +674,53 @@ def push(self):
...
+    name = 'Subversion (git-svn)'
+
+    _cmd_update_remote = ['svn', 'fetch']
+    _cmd_push = ['svn', 'dcommit']
+
+    def configure_remote(self, pull_url, push_url, branch):
+        '''
+        Initializes the repository.
+        '''
+        try:
```
roidelapluie: @nijel how can I replace that try..except? Do we have such a function already?

nijel: You should just catch specific exception in except, not everything.

roidelapluie:
@nijel Can you help me with the testing here?

We can use https://github.com/weblateorg/test also because github supports the svn protocol as well.

nijel:
Thanks, I've heavily improved the code and merged it.

glensc:
This is nice, but how to actually use this?

The [documentation](https://docs.weblate.org/en/latest/vcs.html#subversion) only mentions that git-svn is super, nothing else

in admin `/admin/trans/subproject/add/`) `Version control system` dropdown i have only `Git` and `Mercurial`

image


so i thought maybe i should pick `Git` and use some special url like `git+svn://foo.bar`, but there is no info whatsoever what the url should be like, the inline help only mentions `weblate://` schema.

webs86:
@glensc wich version of git did you use? the problem is yhat with git versions prior 2.10 there was a problem with command to test if git-svn was installed in your system. From 2.10 the problem is fixed and subversion repositories can be used in weblate.

glensc:
my versions,

```
# docker-compose exec --user root weblate svn --version
svn, version 1.8.10 (r1615264)
   compiled Apr 28 2016, 10:47:07 on x86_64-pc-linux-gnu
...
# docker-compose exec --user root weblate git --version
git version 2.1.4
```

glensc:
@webs86: and how the usage will look like? `Subversion` will appear to dropdown? i need to use some url prefix? how?

webs86:
@glensc you must use the url of your repository (http, https, svn prefix). But before you have to upgrade git version (git-svn is not git and svn. Git-svn is a script in perl to work with svn repositories through git like commands).

Git 2.10 was released few days ago and I don't think it is yet in official ubuntu or debian repositories.

webs86:
@glensc  If I 'm not mistaken git - svn is installed with git

webs86:
@glensc  If I 'm not mistaken git - svn is installed with git

glensc:
@webs86, no, `git-svn` is separate package:
https://github.com/WeblateOrg/docker/blob/1da0095fab68eff3f25fbedcefe47495fd6c6eef/weblate/Dockerfile#L9

but it doesn't install `subversion` package, had to install it manually. imho debian `git-svn` package is broken in that manner.

anyway i modified the `vcs.py` i referenced earlier to have `def is_supported()` to return `True`, but it still returns `False`

```
# /app/bin/python -c "from pprint import pprint;from weblate.trans.vcs import SubversionRepository; print(SubversionRepository.is_supported())"
False
```

and `Subversion` is not appearing in add subproject dropdown

glensc:
and `git-svn` is distributed with git:

```
# dpkg --status git|grep Version
Version: 1:2.1.4-2.1+deb8u2
# dpkg --status git-svn|grep Version
Version: 1:2.1.4-2.1+deb8u2
```

webs86:
@glensc  my last answer had an error and i have updated it.

Your version is old: the version you need to work correctly is 2.10.

glensc:
yes, you already said git version is old. no need to repeat that.

anyway, i compared /usr/lib/git-core/git-svn with 2.10.0 version, and it seemed safe just to update that script. so i did that. now `Subversion` appeared to dropdown after container restart

Issue #1160, opened 2016-06-30
Add query-level caching to enhance performance


Vince300:
### Context

I have encountered performance issues when dealing with a project containing 35k strings in 13 translations. The project is using monolingual PO files.
### Actual behaviour

The import process took around 25 minutes (2min per translation) to complete. Deleting the newly imported component was in the 70-75 minutes range.
### Performance analysis

During the import process, the python process was using more than 95% CPU time, while the postgres process used the remaining 5%. Latency of disk I/O operations did not excess 10ms, so storage didn't seem to be the bottleneck. The same behavior was observed when deleting the component.

Running PostgreSQL with the pg_stat_statements extension enabled revealed that when deleting the project, the same 3 queries to fetch project, language and component properties were run multiple times per translation unit being deleted. See [project_delete_request_stats.csv.txt](https://github.com/nijel/weblate/files/340930/project_delete_request_stats.csv.txt)

We can also guess that the 4th request to fetch translation info is called for every deleted unit, which should not be necessary.
### Possible fix

Adding query-level caching through [django-cache-machine](https://github.com/django-cache-machine/django-cache-machine) improved performance by 50% when deleting a project (between 30 and 35 minutes). Although django-cache-machine provides automatic cache invalidation when updating models, more in-depth testing is required to ensure this doesn't break anything.

Adding django-cache-machine to Weblate can be done according to the docs:

``` diff
diff --git a/requirements.txt b/requirements.txt
index a6c659e..38003cb 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,3 +10,4 @@ django-crispy-forms>=1.4.0
 oauthlib>=0.6.3
 django_compressor>=1.5.0,!=2.0
 djangorestframework>=3.3
+django-cache-machine>=0.9
\ No newline at end of file
diff --git a/weblate/lang/models.py b/weblate/lang/models.py
index 5a163f4..ebb7f4e 100644
--- a/weblate/lang/models.py
+++ b/weblate/lang/models.py
@@ -39,6 +39,7 @@ from weblate.trans.site import get_site_url
 from weblate.appsettings import SIMPLIFY_LANGUAGES
 from weblate.logger import LOGGER

+from caching.base import CachingManager, CachingMixin

 def get_plural_type(code, pluralequation):
     '''
@@ -82,7 +83,7 @@ def get_english_lang():
         return 65535


-class LanguageManager(models.Manager):
+class LanguageManager(CachingManager):
[...]

PR #1173, opened 2016-07-19
Added support for django-cache-machine for most used models

Vince300:
This PR is a potential fix for #1160. A few things to note:
- django-cache-machine requires a cache to be setup
- .values and .values_list is broken on Django 1.7+ with django-cache-machine, as described in https://github.com/django-cache-machine/django-cache-machine/issues/116. requirements.txt has been updated to use https://github.com/Vince300/django-cache-machine instead of the released version, but this fix is not generic.


nijel:
The caching seems to bring some problems, at least to the testsuite. Isnt' there some other caching module which would work correctly with current Django versions?

PR #1243, opened 2016-10-02
Rename templates directory

### LINKED ISSUE #918: Rename html to templates folder, use subfolder
I load Weblate from my virtual. This works great, three suggestions to make it even better:

* Weblate's template folder should be called `templates` so it's automatically picked up by Django (saves me the trouble of having to use the `TEMPLATE_DIRS` variable)

* More blocks in templates to override. For example, blocks inside the navbar's and tools' dropdown elements would make it easier to add links to non-weblate pages without having to override the whole `index.html`/`project.html` files. On the login page it would be nice to have a block in the right column for custom elements. F.E.:
```html
{% extends "weblate/index.html" %}
{% block tools_menu %}
    <li>My custom first link</li>
    {{ block.super }}
    <li>My custom last link</li>
{% endblock %}
```
 
* The templates should be inside a subfolder, fe `weblate/templates/weblate/index.html`. This way it's clear when I'm overriding a template where it originally came from.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27963906-rename-html-to-templates-folder-use-subfolder?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

galaxyfeeder:
This pull request fixes #918 

- Renamed templates directory from html/ to templates/.
- Changed the example settings removing the template.dir setting.
- Changed MANIFEST.in


nijel:
Can you please also adjust MANIFEST.in so that the templates get properly installed and included in the package? This is reason why the Travis test fails (you can ignore AppVeyor failure for now, we're in progress of making the tests work on Windows as well).

Also can you please exclude the po file updates? This is easiest way to avoid merge conflicts once this gets changed by translators meanwhile....

galaxyfeeder:
Passing Travis test but now failing on landscape, any idea?

Updated also the manifest.in

nijel:
Hmm, probably some version upgreade on their side made all errors appear as both fixed and new, probably not worth looking into...

nijel:
Merged, thanks!

PR #1244, opened 2016-10-03
Avoid displaying 0.0% or 100.0% stats when they're actually rounded (…

marcperrin:

…artificially display 0.1 and 99.9)


nijel:
Thanks!

marcperrin:
Thank you! (I'll review my personal changes on Weblate to check if I have other changelists that are worth a pull-request; for now I don't think so, but in the future I won't hesitate creating pull-requests)

PR #1250, opened 2016-10-10
duplicate version checks

HassenPy:
this is already done by [`weblate/requirements.py`](https://github.com/HassenPy/weblate/blob/master/weblate/requirements.py)

Do you prefer opening issues first for this kind of small changes?


nijel:
It's fine for open pull requests directly for such things, thanks!

PR #1302, opened 2016-12-20
Make configured component branch available in scripts as WL_BRANCH

crisp2u:


nijel:
Can you please add it to the documentation as well?

crisp2u:
Done (sorry, I was not sure from the start were is the source for the docs)

weblate/trans/scripts.py
```diff
@@ -80,7 +80,7 @@ def run_hook(component, translation, script, env=None, *args):
             'WL_PATH': target.get_path(),
             'WL_FILEMASK': component.filemask,
             'WL_TEMPLATE': component.template,
-            'WL_FILE_FORMAT': component.file_format,
```
nijel: I guess that the removal is not intentional...

nijel:
Thanks for the docs! However I've not noticed another problem with the patch before...

nijel:
Merged with dd65d4702 to fix removed variable.

PR #1308, opened 2017-01-03
Improve hg support

andrewshadura:
Hi,

These are a couple of changes which are supposed to improve the state of Mercurial support in Weblate. I haven’t tested them yet, but I will and I also hope I will come with some real tests to test this functionality in more or less real situations, especially those when it’s failing for me at the moment.

Thanks for considering this PR.

### DIFF
--- a/weblate/trans/vcs.py
+++ b/weblate/trans/vcs.py
@@ -1088,21 +1088,24 @@ class HgRepository(Repository):
         self.set_config('extensions.rebase', '')
         if abort:
             self.execute(['rebase', '--abort'])
-        else:
-            self.configure_merge()
-            try:
-                self.execute(['rebase', '-d', 'remote(.)'])
-            except RepositoryException as error:
-                if error.stdout:
-                    message = error.stdout
-                else:
-                    message = error.stderr
-                # Mercurial 3.8 changed error code and output
-                if (error.retcode in (1, 255) and
-                        'nothing to rebase' in message):
-                    self.execute(['update', '--clean', 'remote(.)'])
-                    return
-                raise
+        elif self.needs_merge():
+            if self.needs_ff():
+                self.execute(['update', '--clean', 'remote(.)'])
+            else:
+                self.configure_merge()
+                try:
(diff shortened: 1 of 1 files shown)


nijel:
Does any of the changes depend on some particular mercurial version? The tests just hang on Travis CI which has 2.0.2. It works fine on 4.0 I have locally installed.

nijel:
Hmm, actually Travis has both 2.0.2 (which should not be supported by Weblate at all) and 4.0.1 installed from pip, I've fixed this mess in 371a0d02e2ac420c67c7404d3a2a7a7963b936e7.

weblate/trans/vcs.py
```diff
@@ -991,6 +991,25 @@ def check_config(self):
...
+
+        merge_driver = os.path.abspath(
+            os.path.join(
+                os.path.dirname(os.path.dirname(os.path.dirname(__file__))),
+                'examples',
+                'git-merge-gettext-po'
+            )
+        )
+
+        if os.path.exists(merge_driver):
```
nijel: I've factored out this part from Git in 816e3f270, can you please update your code to use new method as well?

nijel:
The build still hangs on Travis, so it's probably not really working as expected...

nijel:
The documentation is here: https://docs.weblate.org/en/latest/contributing.html#testsuite

The particular problematic test (hanging on Travis CI) can be executed by:

```
./manage.py test --settings weblate.settings_test weblate.trans.tests.test_remote.MercurialMultiRepoTest
```

It really looks to me like the `ui.merge` configuration didn't work and it's interactively waiting for input.

andrewshadura:
Hmm, it still fails on Travis, even though it succeeded locally.

andrewshadura:
G
​reat, thanks!

Can you please also update the hosted instance so that I can test it in
action? I
​t's been in a very confused state since about middle of Se​ptember.

-- 
Cheers,
  Andrew

nijel:
I've just updated it :-)

PR #1309, opened 2017-01-05
Translation services' names fixes in docs

ihoru:
Prefix `weblate.` was missed all these days!


nijel:
Thanks, this comes from ages where modules were not prefixed by weblate and nobody did change it here...

PR #1350, opened 2017-02-03
Apertium API keys not used any more

unhammer:
The old api (see http://wiki.apertium.org/wiki/Old_Apertium_web_service ) required keys, but the new one doesn't (it might in the future, but we haven't had any abuse so far). The old service might still run, sometimes, but it's recommended to either run your own APY server or use https://www.apertium.org/apy/ for single-users


unhammer:
Was using the github web interface to edit here – should I re-open with squashed commits?

nijel:
Thanks, merged squashed.

Issue #1327, opened 2017-01-19
Reconsider checksums usage


nijel:
Currently we use md5 and char(32) fields to speed up strings lookup. This works well for long strings, but for short ones it is not that beneficial as the lookup costs about same as the actual string lookup.

Maybe it would be better to use some shorter hash (we really don't care that much about collisions for lookups) and store the value as integer in the database.

Things to consider:

* Which hash function to use? [SipHash](https://en.wikipedia.org/wiki/SipHash) seems to be option, though the pure Python implementation will probably lack desired performance. There is [csiphash](https://github.com/zacharyvoase/python-csiphash) though.
* Storing in [BigIntegerField](https://docs.djangoproject.com/en/1.10/ref/models/fields/#bigintegerfield) is possible since Django 1.10.
* Migration path would mean that all units in the database are updated, what would be quite slow migration. The benefits really have to outweigh this.

nijel:
This is quite tricky as the siphash implementations for Python 2.7 are far from being good replacement:

* Pure python siphash is 20x slower than md5
* The csiphash is 3x slower than md5
* With removed return value conversion csiphash is 2x slower than md5
* Python 3.5 builtin `hash()` (which uses siphash24) is about 4x faster than md5 (though there is no way to provide the key here, so it's not really usable for our case).

So performance wise it would be good replacement, if there was usable implementation, which does not seem to exist right now.

nijel:
On further look, there is also [siphashc3](https://pypi.python.org/pypi/siphashc3/3) which performs well (3x faster than md5 in this particular benchmark), so that might be way to go.

nijel:
Actually [siphashc3](https://pypi.python.org/pypi/siphashc3/3) is better choice due to compatibility with Python 3. The overall benefits of such move still should be measured before doing such change (and adding another dependency).

nijel:
Anyway this is something to decide for 2.12.

nijel:
Merged #1360

PR #1360, opened 2017-02-09
Change hash algorithm to siphash

nijel:
Change hash algorithm to siphash, see https://github.com/WeblateOrg/weblate/issues/1327.


nijel:
The results on Travis indicate that this is about 10% faster to run testsuite than original md5 based code, results on real workloads might vary, but still should be noticeable.

PR #1435, opened 2017-04-07
[FIX] serializers: Delete unnecesary True

JesusZapata:
The code deleted `or True` causes that ever the response return an empty answer

```python
result['vcs'] = None
result['repo'] = None
result['branch'] = None
result['filemask'] = None
```
The comprobation of the permision is made on `not can_see_git_repository(user, instance.project)`

So the code `or True` is unnecessary

### DIFF
--- a/weblate/api/serializers.py
+++ b/weblate/api/serializers.py
@@ -177,7 +177,7 @@ class ComponentSerializer(RemovableSerializer):
         """Remove VCS properties if user has no permission for that"""
         result = super(ComponentSerializer, self).to_representation(instance)
         user = self.context['request'].user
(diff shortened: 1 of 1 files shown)


nijel:
Thanks for spotting, this was left from testing the code...

JesusZapata:
@nijel 
But this PR will be accepted?

nijel:
Yes, it will be, I'm just waiting for tests to complete (just in case some tests depend on this broken code).

JesusZapata:
@nijel 
In which version will this PR be released?

nijel:
It will be in 2.13, probably released next Wednesday.

PR #1463, opened 2017-04-28
allow multiple commits with gerrit

Blaimi:
'git review' needs --yes as an argument to push commits silently if there is more than one commit.

From ‘man git-review’: “-y, --yes : Indicate that you do, in fact, understand if you are submitting more than one patch.”


nijel:
Looks sane, will merge it after tests complete.

nijel:

PR #1471, opened 2017-05-05
Include examples in data files.

### LINKED ISSUE #1468: Provide example hooks in pip installations
### Steps to reproduce
1. `pip install Weblate`
2. `ls path/to/site-packages/weblate`

### Actual behaviour
Example hooks are missing. We have to download the examples manually and copy them where we want on the server.

### Expected behaviour
Example hooks should be in site-packages, packaged as data inside Weblate package. This would allow to use examples, even when Weblate is installed using pip.

gilbsgilbs:
Fixes #1468

- [X] Every commit has message which describes it
- [X] Every commit is needed on it's own, if you have just minor fixes to previous commits, you can squash them
- [X] Any code changes come with testcase
- [X] Any new functionality is covered by user documentation

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,6 +32,11 @@ with open(os.path.join(os.path.dirname(__file__), 'README.rst')) as readme:
 with open('requirements.txt') as requirements:
     REQUIRES = requirements.read().splitlines()
 
+DATA_FILES = [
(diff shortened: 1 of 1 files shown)


nijel:

PR #1485, opened 2017-05-11
Fix type on Screenshots

saily:
Fix a typo in docs.


nijel:
Thanks, I should really pay more attention to reviewing that - either one e is missing or there is one e extra...

PR #1520, opened 2017-06-13
differentiate between “Weblate” and “Weblate similarity” in machine translation table

Blaimi:
to get the difference of the two machine-translation-engines, show their names in the table instead of a static “Weblate”


nijel:

PR #1539, opened 2017-07-01
docs: docker: missing _HOST in variables

dachary:


nijel:
Well even though the variable doesn't match Django, this is how it's implemented: https://github.com/WeblateOrg/docker/blob/docker/settings.py#L726

dachary:
@nijel I realize that now, my bad for assuming systematic naming convention and thanks for your comment.

nijel:
Well not using systematic naming was a mistake. Now the question is whether to fix that (and break some existing setups) or keep it (and stay inconsistent)...

nijel:
I'd probably fix the naming with compatibility for the old ones for some time...

nijel:
Well, even Django is pretty inconsistent here - there is EMAIL_PORT, but EMAIL_HOST_USER ;-).

nijel:
Merged, I've changed the Docker settings to accept both options.

PR #1556, opened 2017-07-14
Fix merge-pot-in-po hook. 

### LINKED ISSUE #1553: hook-merge-pot-in-po does not work
Yeah, I know, I've written [this](https://github.com/WeblateOrg/weblate/blob/weblate-2.15/examples/hook-merge-pot-in-po) a few months ago, but it cannot work since it expects to be called multiple times with the `.po` file as argument. This is [not going to happen](https://github.com/WeblateOrg/weblate/blob/701d8f420ed356ab65a36ab6ed3da8871c9bee3f/weblate/trans/scripts.py#L32) in `post_update` scripts.

I don't know what to do exactly. Correct me if I'm wrong, but I don't think we can handle the general in a `post_update` scripts currently, unfortunately. We may want to remove this hook, but lacking from such a script out of the box obviously makes things a lot more complicated.

gilbsgilbs:
Fixes #1553

As mentioned in #1553, I wont be able to test this before monday. So be sure to test this on your side if you're willing to merge.

Thanks, and sorry again for having written the broken hook in the first place. 😕

### DIFF
--- a/examples/hook-merge-pot-in-po
+++ b/examples/hook-merge-pot-in-po
@@ -2,6 +2,6 @@
 # Post-update hook to merge POT translations into PO.
 # Requires gettext.
 
-[ -z "${WL_NEW_BASE}" ] && exit 0
-msgmerge -U "$1" "${WL_NEW_BASE}"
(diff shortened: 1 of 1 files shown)


nijel:
Looks good and it will certainly work better than previous one :-).

PR #1577, opened 2017-07-26
use exec in ssh wrapper to free up os resources

glensc:
minor optimization


nijel:

PR #1597, opened 2017-08-14
Enforce INSTALLED_APPS order in settings_example

nschonni:
Maybe this isn't the best way in python, but I was just looking through the changes and thought it might might make it easier for end-users when modifying the list.


nijel:
Thanks, probably it's better way than just mentioning in comments...

nschonni:
Np, not sure if there is a more Django-y way of doing this somewhere in the application side

nijel:
AFAIK there is nothing better.

PR #1608, opened 2017-08-22
Changed hash algorithm from md5 to sha256

anurag1212:
The b64decode function throws a binaascii.Error exception if the key isn't 4-padded. Since you haven't run into issues yet, I assumed the key is guaranteed to be correctly padded and didn't add extra checks.

Among the myriad of ways to hash a key using sha256, the one that works perfectly and makes the least possible changes to the existing code would be to merely replace hashlib.md5 with hashlib.sha256.
I've done so and unittested the function with varying key lengths (multiples of 4 obviously) and nothing breaks.

Also, sorry about the 3 commits. This is my first contribution to OSS and I didn't know about the sign-off necessity.

### DIFF
--- a/weblate/trans/ssh.py
+++ b/weblate/trans/ssh.py
@@ -77,7 +77,7 @@ def is_key_line(key):
 def parse_hosts_line(line):
     """Parse single hosts line into tuple host, key fingerprint."""
     host, keytype, key = line.strip().split(None, 3)[:3]
-    fp_plain = hashlib.md5(b64decode(key)).hexdigest()
+    fp_plain = hashlib.sha256(b64decode(key)).hexdigest()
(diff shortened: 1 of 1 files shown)


nijel:
Thanks, however you've addressed only one part of the problem - calculating the hash, however it's also rendered differently, See output from ssh-keygen for both SHA256 and MD5 fingerprints:

```
256 SHA256:O4i2PPFELuj49wYZSwLt+a2r356sB19KMCFhrUKkYiM push.dgit.debian.org (ED25519)
256 MD5:f2:a0:e8:3d:fd:03:8a:5c:7b:15:62:fa:c4:3c:22:bc push.dgit.debian.org (ED25519)
```

PR #1620, opened 2017-09-03
The terms of service, email address, indefinetely.

comradekingu:
If not terms of service plural, it should be pointed out that it is a document.


nijel:

PR #1621, opened 2017-09-03
The terms of service, no "purpose", honours

comradekingu:
As with #1620 
If not terms of service plural, it should be pointed out that it is a document.


nijel:

PR #1629, opened 2017-09-12
Split shell commands to two lines

ibennetch:
Signed-off-by: Isaac Bennetch <bennetch@gmail.com>

I believe these won't run correctly if copied and pasted on a single line.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -397,7 +397,8 @@ as ``TEMP`` variable:
 
 .. code-block:: sh
 
-    TEMP=/path/to/big/temp ./manage.py rebuild_index --clean
+    TEMP=/path/to/big/temp
+    ./manage.py rebuild_index --clean
 
 .. seealso:: 
    
--- a/docs/vcs.rst
+++ b/docs/vcs.rst
@@ -225,7 +225,8 @@ action with `hub`_ to finish the configuration, for example:
 
 .. code-block:: sh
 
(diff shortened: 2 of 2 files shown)


Pawamoy:
Yes it works, even in `sh`. The difference with setting the variable THEN running the command is that the variable will have the defined value only inside the subprocess.

```bash
$ cat script.sh
echo hello $VAR

$ export VAR=universe

$ ./script.sh
hello universe

$ VAR=world ./script.sh
hello world

$ ./script.sh
hello universe
```

ibennetch:
I learned something new, then. In fact, when I first saw the code written like that, I thought that looked interesting and wanted to learn what happens in that case, so I did a test very similar to what you posted.

```
$VAR='world' echo $VAR

$
```

Which did not work. I tried a few other things, like

```
$ echo 'foo' ./script.sh
foo .script.sh
```

Which also did not work. In hindsight, I suppose it is just coincidence that the tests I tried weren't exactly right; trying your example code above of course works perfectly and `echo` of course takes the entire line and concatenates it all together so that was a poor test.

So yeah, of course you're right and in the process you taught me something new. Sorry to waste your time :)

nijel:
The problem with first command (`VAR='world' echo $VAR`) is that the same shell is evaluating the variable and setting it, what doesn't work - setting the variable this way is only valid for child processes.

So following code does work:

```sh
VAR=foo bash -c 'echo $VAR'
```

PR #1639, opened 2017-09-24
Update django.po

bovirus:
- Update Italian language
revise of already translated strings
revised fuzzy strings
General check


nijel:

PR #1646, opened 2017-10-02
Fix handler for csrf errors

bcanyelles:
* Add new view that wraps 'denied' view
* Add setting for handle this error

### DIFF
--- a/weblate/trans/views/basic.py
+++ b/weblate/trans/views/basic.py
@@ -512,6 +512,13 @@ def server_error(request):
         return django.views.defaults.server_error(request)
 
 
+def csrf_error(request, **kwargs):
+    """
+    Error handler for csrf errors.
+    """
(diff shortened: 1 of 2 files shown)


weblate/trans/views/basic.py
```diff
@@ -512,6 +512,13 @@ def server_error(request):
         return django.views.defaults.server_error(request)
 
 
+def csrf_error(request, **kwargs):
+    """
+    Error handler for csrf errors.
+    """
+    return denied(request)
```
nijel: This function seems to be unused, you're using directly denied view in CSRF_FAILURE_VIEW...

nijel:
I've added custom template for CSRF error, what should be sufficient to better handle these: a050956fd87b18999db557929af4d36d0c84049f

PR #1650, opened 2017-10-05
Convert makdown image to restructured text image

Jibec:
Oups, I'm not used to restructuredtext...


nijel:
I should have noticed this when merging, thanks for the fix!

Jibec:
wow, are you really automatically pushing your doc at each commit? Impressive :p

nijel:
Well https://readthedocs.org/ does it ;-).

PR #1662, opened 2017-10-17
add path for favicon

Jibec:
when installing weblate in a subpath, favicon still uses root's one, which is a bug ;)
I hope I used the correct way to fix this, I won't mind seeing this PR canceled if you wish to use another way.


nijel:
Well, when specifying any path for favicon, it might be  better to just use `{% static %}` tag instead.

Jibec:
does it means we can remove this?
https://github.com/WeblateOrg/weblate/blob/master/examples/weblate.nginx.conf#L6:L10

nijel:
AFAIK some browsers did hardcode favicon path (that's why it was implemented this way), but I'm not sure they are still relevant these days. It's probably sane to still have it, see https://github.com/audreyr/favicon-cheat-sheet

PR #1664, opened 2017-10-18
fix the compatibility issue of the new version of git for git svn

pptime:
Redefine the function configure_branch for SubversionRepository such that it uses the correct command line to switch to the requested branch. This command line has been changed since v1.8.3.2 if the post https://stackoverflow.com/questions/19712735/git-svn-cannot-setup-tracking-information-starting-point-is-not-a-branch is accurate enough. The fix help solve the issue on my Debian linux.

### DIFF
--- a/weblate/trans/vcs.py
+++ b/weblate/trans/vcs.py
@@ -608,7 +608,7 @@ class GitRepository(Repository):
         # Add branch
         if branch not in branches:
             self.execute(
-                ['branch', '--track', branch, 'origin/{0}'.format(branch)]
+                ['checkout', '-b', branch, 'origin/{0}'.format(branch)]
             )
         else:
             # Ensure it tracks correct upstream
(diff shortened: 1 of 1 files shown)


weblate/trans/vcs.py
```diff
@@ -759,6 +759,30 @@ def get_remote_branch_name(self):
...
+        # Get List of current branches in local repository
+        # (we get additional * there indicating current branch)
+        branches = self.execute(['branch']).splitlines()
+        if '* {0}'.format(branch) in branches:
+            return
+
+        # Add branch
+        if branch not in branches:
+            self.execute(
+                ['checkout', '-b',  branch, 'origin/{0}'.format(branch)]
```
nijel: The referenced SO question mentions refs/remotes/ here, are you sure origin is correct here?

pptime: Yes, I am sure, on line 700, Subversion version just uses --prefix=origin/

pptime: And this setting is particular for git svn, I don't think it can be propagated to more generic git version.

PR #1685, opened 2017-11-12
use msgcat for .po/.pot files diff

glensc:
adds diff filter for .po and .pot files.

before it can be used git needs to be configured:

```
git config diff.msgcat.textconv "msgcat --no-location --no-wrap --sort-output"
```

if you wish to cache callouts of msgcat in `notes/textconv/msgcat`:
```
git config diff.msgcat.cachetextconv true
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,3 +3,10 @@
 .gitattributes export-ignore
 scripts export-ignore
 appliance export-ignore
+
+# make .po/.pot files less noisy, need to configure git first:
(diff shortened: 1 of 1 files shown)


glensc:
i just played with this in other project: https://github.com/eventum/eventum/pull/325

nijel:
I usually prefer git to show real diff, not cleaned up link this, so I'm not really sure this is good fit to merge...

glensc:
it has no effect without configuring the diff filter (those `git config` calls). and there's `--no-textconv` option as well to reverse action.

nijel:

PR #1723, opened 2017-12-13
Ge'ez, spelling fixes aplenty :)

comradekingu:


weblate/templates/i18n.html
```diff
@@ -19,26 +19,26 @@
...
+{% blocktrans %}Avoid dates and years associated with you.{% endblocktrans %}
+{% blocktrans %}This is among the 10 most common passwords.{% endblocktrans %}
+{% blocktrans %}This is among the 100 most common passwords.{% endblocktrans %}
 {% blocktrans %}This is a very common password.{% endblocktrans %}
 {% blocktrans %}This is similar to a commonly used password.{% endblocktrans %}
-{% blocktrans %}A word by itself is easy to guess.{% endblocktrans %}
+{% blocktrans %}A word by itself is easily guessed.{% endblocktrans %}
 {% blocktrans %}Names and surnames by themselves are easy to guess.{% endblocktrans %}
-{% blocktrans %}Common names and surnames are easy to guess.{% endblocktrans %}
+{% blocktrans %}Common names and surnames are easily guessed.{% endblocktrans %}
```
nijel: As noted in the comment, all above messages are coming from django-zxcvbn-password, the strings are extracted here for localization only. If you want to submit fixes to these, please do that at https://github.com/dwolfhub/zxcvbn-python/blob/master/zxcvbn/feedback.py

weblate/templates/i18n.html
```diff
@@ -144,7 +144,7 @@
 {% trans "Galician" %}
 {% trans "Ganda" %}
 {% trans "Georgian" %}
-{% trans "Geez" %}
+{% trans "Ge'ez" %}
```
nijel: Thanks I've addressed this in 3503114f0a2921dae81dcca508f38250f5b52d4e.

nijel:
Thanks for pull request. I've applied part of the fix, however most of the messages are coming from third party module and are included for localization only.

PR #1725, opened 2017-12-13
No self referencing or sentient software

comradekingu:


weblate/templates/about.html
```diff
@@ -2,27 +2,27 @@
...
 <div class="panel panel-primary">
-<div class="panel-heading"><h4 class="panel-title">{% trans "About Weblate" %}</h4></div>
+<div class="panel-heading"><h4 class="panel-title">{% trans "About" %}</h4></div>
 <div class="panel-body">
 
-<p>{% blocktrans %}Weblate is a web based translation tool with tight version control integration. It features simple and clean user interface, propagation of translations across components, quality checks and automatic linking to source files.{% endblocktrans %}</p>
+<p>{% blocktrans %}Weblate is a web based translation tool with tight version control integration. It features a simple and clean user interface, propagation of translations across components, quality checks and automatic linking to source files.{% endblocktrans %}</p>
 
-<p>{% blocktrans %}If you are looking for support for Weblate or want to file bug reports check its website at <a href="{{ weblate_url }}">weblate.org</a>.{% endblocktrans %}</p>
+<p>{% blocktrans %}If you are looking for support, or want to file bug reports, go to <a href="{{ weblate_url }}">weblate.org</a>.{% endblocktrans %}</p>
```
nijel: This wording was intentional to differentiate support for Weblate and for given instance. If somebody is running Weblate, we certainly don't want users to ask for instance specific support (eg. about translations or projects available there) in our issue tracker. Still it could be probably written better...

comradekingu:
Seeing as weblate.org isn't actually where to file bugreports, and weblate is introduced already, I shortened it.

nijel:

It might be still place to file bug reports, it's just might not be a place where to look for support.

Also I've given back some of the self referencing in ac0f9019cb9d200dd6a8764fea325b5e0334411b to make it clear in case the site is branded differently.

PR #1736, opened 2017-12-13
The translation has come to an end

comradekingu:


nijel:
Merged with fix for testsuite and other location of the same message in 38d865f9d, thanks for your contribution!

PR #1744, opened 2017-12-13
Insufficient, comment on, to save

comradekingu:


nijel:

I've applied same change to the code and testsuite in b12abe250.

PR #1750, opened 2017-12-13
No !, ID, e-mail, when marked as such

comradekingu:


nijel:

I've adjusted the testsuite to this change in 483d07cbf.

PR #1771, opened 2017-12-29
Fix merge-pot-in-po hook.

gilbsgilbs:
I apologize for having pushed such clumsy code in the first place. I should have tested it first, seriously. 🙄  Still, I finally found some time to test it and it worked at least once.


nijel:

Issue #1791, opened 2018-01-15
Achieve better results with WeblateTranslation and WeblateSimilarTranslation


mlaggner:
When using the machine translation engines by Weblate, I've discovered some minor things with the following example:

Source string: Abfuhrschein

Results:

| Translation | Source string | Engine | Quality index |
|---|---|---|---|
|Should the advice note be saved before?|Soll der Abfuhrschein vorher gesichert werden?|Weblate (SAP/SAP-EFO) | 100|
|Change documents for advice note &1. | Änderungsbelege zu Abfuhrschein &1 | Weblate (SAP/SAP-EFO) | 100|
|Advice note | Abfuhrschein | Weblate (SAP/SAP-EFO) | 100|
|Advice note | Abfuhrschein | Weblate (SAP/SAP-EFO) | 100|
|Advice note | Abfuhrschein | Weblate (SAP/SAP-EFO) | 100|
|EFO advice note | [OBSOLETE!] EFO Abfuhrschein | Weblate (SAP/SAP-EFO) | 100|
|Change documents for delivery note & | Änderungsbelege zu Abfuhrschein & | Weblate (SAP/SAP-EFO) | 100|

two points could be improved here:

- create some sort of string distance (Levensthein?) for the _source string_ to compare it to the desired result rather than give it 100% (or 50% via WeblateSimilarTranslation)
- remove duplicates

what do you think?

mlaggner:
what about using the Dice Coefficient algorithm here?
https://en.wikibooks.org/wiki/Algorithm_Implementation/Strings/Dice%27s_coefficient#Python

have been using this algorithm for string scoring in another project and has been returned rather good results

nijel:
I've addressed this partly in 1ca5c60fae58e7fa6bf3d8ae789ee144f229c34d - now WeblateTranslation behaves are originally designed (to find only exact matches).

I guess the duplicates are coming from strings using different IDs but same source, right?

mlaggner:
yes, that's one nature of SAP-internal translations..
we have strings in different lengths to display (5 different texts for one dictionary element) and sometimes the text is rather short that there is the same in 5 different IDs

nijel:
Thank you for your report, the issue you have reported has just been fixed. 

* In case you see problem with the fix, please comment on this issue. 
* In case you see similar problem, please open separate issue.
* If you are happy with the outcome, consider supporting Weblate by [donating](https://weblate.org/donate/).

PR #1793, opened 2018-01-15
adopted score from Weblate machine translation

mlaggner:
- used the dice coefficent algorithm (from https://en.wikibooks.org/wiki/Algorithm_Implementation/Strings/Dice%27s_coefficient#Python)
to calculate the quality score in WeblateTranslation and WeblateSimilarTranslation
- also removed duplicates from the result list

adresses WeblateOrg/weblate#1791

### DIFF
--- a/weblate/trans/machine/weblatetm.py
+++ b/weblate/trans/machine/weblatetm.py
@@ -35,11 +35,11 @@ class WeblateBase(MachineTranslation):
         """Any language is supported."""
         return True
 
-    def format_unit_match(self, unit, quality):
+    def format_unit_match(self, unit, text):
         """Format unit to translation service result."""
         return (
             unit.get_target_plurals()[0],
-            quality,
+            int(self.dice_coefficient(text, unit.get_source_plurals()[0]) * 100),
             '{0} ({1})'.format(
                 self.name,
                 force_text(unit.translation.subproject)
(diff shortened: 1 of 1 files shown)


weblate/trans/machine/weblatetm.py
```diff
@@ -56,11 +94,11 @@ def download_translations(self, source, language, text, unit, user):
         """Download list of possible translations from a service."""
         matching_units = Unit.objects.same_source(unit)
 
-        return [
-            self.format_unit_match(munit, 100)
+        return list(set([
+            self.format_unit_match(munit, text)
```
nijel: It can stay 100 here - this service returns only exact matches.

weblate/trans/machine/weblatetm.py
```diff
@@ -56,11 +94,11 @@ def download_translations(self, source, language, text, unit, user):
         """Download list of possible translations from a service."""
         matching_units = Unit.objects.same_source(unit)
 
-        return [
-            self.format_unit_match(munit, 100)
+        return list(set([
```
nijel: Why `list(set(` here? The set of units already should be unique (it might be same source, but different project, so this won't reduce it anyway).

weblate/trans/machine/weblatetm.py
```diff
@@ -71,8 +109,8 @@ def download_translations(self, source, language, text, unit, user):
         """Download list of possible translations from a service."""
         matching_units = Unit.objects.more_like_this(unit)
 
-        return [
-            self.format_unit_match(munit, 50)
+        return list(set([
```
nijel: Same as above, did you see need for `list(set())`?

weblate/trans/machine/weblatetm.py
mlaggner: the example from WeblateOrg/weblate#1791 have shown me, that under certain circumstances the WeblateTranslation offered duplicates - this was my try to remove this duplicates (I am still a python noob)

weblate/trans/machine/weblatetm.py
mlaggner: that's not true (at least with Weblate 2.18 as shown in WeblateOrg/weblate#1791)

weblate/trans/machine/weblatetm.py
mlaggner: to remove potential duplicates

nijel:
I've tried to reuse scoring from the fulltext backend doing the matches in 90c2d8edcff1c984f3d953fe29f0e9a927b2feaf, can you please check how that works in your case?

mlaggner:
just tried to include your patch, but I always get: `MachineTranslationError: ValueError: too many values to unpack` from the WeblateSimilarTranslation service

logs:

```
ERROR Handled exception ValueError: too many values to unpack
ERROR Failed to fetch translations from Weblate similarity
```

nijel:
Was the part of the patch on weblate/trans/search.py correctly applied?

mlaggner:
you're absolutely right -> I copied the wrong file :/

the results look something different:

|Translation|Source string|Engine|Quality score|
|---|---|---|---|
|Advice note | Abfuhrschein | Weblate (SAP/SAP-EFO) |100|
|Advice note | Abfuhrschein | Weblate (SAP/SAP-EFO) |100|
|Advice note | Abfuhrschein | Weblate (SAP/SAP-EFO) |100|
|Change documents for advice note &1. | Änderungsbelege zu Abfuhrschein &1 | Weblate similarity (SAP/SAP-EFO) |71|
|EFO advice note | [OBSOLETE!] EFO Abfuhrschein | Weblate similarity (SAP/SAP-EFO) |71| 
|Change documents for delivery note & | Änderungsbelege zu Abfuhrschein & | Weblate similarity (SAP/SAP-EFO) |71|
|Should the advice note be saved before? | Soll der Abfuhrschein vorher gesichert werden? | Weblate similarity (SAP/SAP-EFO)|50|

As you see the duplicates are still here.
And I think the score for the similarity is rather high

nijel:
This was not supposed to address duplicates, only scoring. I've used whatever comes out of whoosh as it internally already does some scoring, so it might be good to reuse it. It works  reasonably well for long phrases I've tested. Here I'm not sure, in the end "Abfuhrschein" and "Änderungsbelege zu Abfuhrschein &" are quite similar, it's just one word extra....

nijel:
The uniqueness was addressed in b6118357bc3c6152920f22535d9f03f65676ea63.

Anyway I'm closing this PR in favor of changes which are already in the weblate repository. For me the current scoring results look reasonable, to change that we should really come with some way to measure how scoring works....

PR #1808, opened 2018-01-22
add a human-readable alias for Korean

burner1024:
Hi, this is a human-readable alias for Korean language

### DIFF
--- a/weblate/lang/data.py
+++ b/weblate/lang/data.py
@@ -346,7 +346,7 @@ LOCALE_ALIASES = {
     'schinese': 'zh_Hans',
     'chinese_zh': 'zh_Hant',
     'tchinese': 'zh_Hant',
-    'korean': 'ko_KR',
+    'korean': 'ko',
     'dutch_be': 'nl_BE',
     'english': 'en',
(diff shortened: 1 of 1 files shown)


weblate/lang/data.py
```diff
@@ -346,6 +346,7 @@
     'schinese': 'zh_Hans',
     'chinese_zh': 'zh_Hant',
     'tchinese': 'zh_Hant',
+    'korean': 'ko_KR',
```
nijel: Wouldn't it be better to alias it to `ko` only? The country specification is usually not needed and the alias should be consistent with others which do not use country unless necessary...

burner1024: Will it work like that? I see no definition for "ko" in data.py

nijel: The language definitions have been recently moved to separate file:

https://github.com/WeblateOrg/weblate/blob/4ded14d535b02c306f1cfaa5268557ef4711e3e5/weblate/langdata/languages.py#L943-L948

burner1024: changed to ko

nijel:

PR #1816, opened 2018-01-29
Spelling: Good news, The user, contribution to

comradekingu:
I got one e-mail just now that had no username, from "Weblate/Application: Languages"…


nijel:

As for missing name, it can indeed happen, we will be slowly fixing data by 627650c655e9a5284c04f5b540b7a965fa84039a.

PR #1830, opened 2018-02-04
Articles and phrasing

jhertel:


jhertel:
Regarding "Some checks were not successful":

About the "DCO — The sign-off is missing.": Well, I made the change, if you are unsure about that; when I commit something, I made the change. The whole signed-off-by-thing seems rather silly to me when all I want is to contribute – why add a manual step to each commit? Let's focus on the code. Of course I don't steal, of course I don't commit other people's work, and of course my code is meant to go into the project under the current license.

About the "continuous-integration/appveyor/pr — AppVeyor build failed": These massive amounts of fails seem unrelated to my change. Maybe you should remove these checks until they actually pass?

nijel:
I've dissabled the AppVeyor tests. The idea was to focus on making tests pass on Windows for this release, but it turned out that some of the libraries we use are not yet there, so there is no chance to make it work right now.

nijel:

jhertel:
You're welcome, nijel!

PR #1840, opened 2018-02-08
automatic translation via machine translation

mlaggner:
Added this feature into the automatic translation view. Calling the function from the view works as expected.

I touched/modified all files I am aware of, but there has been a few places where I did not know what to do:

- urls.py (ehance the regexp `r'^projects/' + TRANSLATION + 'auto/$',` by the -mt one): I did not find out if that extension is needed
- test_autotranslate.py: I did add some basic unit tests, but to do some more detailed testing I could not find out how the test model looks like
- command line execution: I did not understand how this exactly works and how I should implement this feature to the command line functions

beside that I changed everything I was aware of

### DIFF
--- a/docs/user/translating.rst
+++ b/docs/user/translating.rst
@@ -242,3 +242,20 @@ This feature can be useful in several situations like consolidating translation
 between different components (eg. website and application) or when
 bootstrapping translation for new component using existing translations
 (translation memory).
+
+.. _auto-translation-mt:
+
+Automatic translation via machine translation
+---------------------------------------------
+
+Weblate can also be used to automatically fetch translations from activated
+machine translation services. This tool is called :guilabel:`Automatic translation` 
+and is accessible in the :guilabel:`Tools` menu.
+
+You can choose which machine translations engines you want to use as source and how to handle
+conflicts.
+
+This feature can be useful in several situations like consolidating translation
+between different components (eg. website and application) or when
+bootstrapping translation for new component using existing translations
+(translation memory).
--- a/weblate/static/robots.txt
+++ b/weblate/static/robots.txt
@@ -22,6 +22,7 @@ Disallow: /zen/
 Disallow: /download/
 Disallow: /upload/
 Disallow: /auto-translate/
+Disallow: /auto-translate-mt/
 Disallow: /new-lang/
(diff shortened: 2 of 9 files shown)


nijel:
Thanks, it looks good. What I wonder if it could not share more code with current autotranslate as it seems to be quite similar in the end and you seem to have copied quite amount of existing code (in both implementation, views and tests)...

mlaggner:
just checked that again, but since we have some different entry points (one with a "source" translation project and one with a list of MT engines & threshold), I do not see that there should be shared code..
And according to the "separation of concerns" paradigm I think we should not mix the implementations

nijel:
Okay, I have just quickly looked and didn't really deeply compare the code, but it looked similar. I'll take deeper look after 2.19 is out what should be probably on Wednesday.

mlaggner:
just had an issue with this feature:
If the automatic translation returns a match which is longer than the `max-length` it should be inserted as a suggestion rather than as a _too long_ translation to avoid the maximum length error (https://docs.weblate.org/en/weblate-2.19.1/user/checks.html#check-max-length)

nijel:
Thanks, I've merged your code, but I've decided to share the basics of the auto translation code, see https://github.com/WeblateOrg/weblate/commit/a6bc9cb80e808bf43e78243678716c8e9c5c8aae

mlaggner:
thanks! 
will test it as soon as possible

mlaggner:
just tried it in my test environment and I must say it works perfectly!
Thanks!

mlaggner:
just found an issue with the combination of Weblate and WeblateSimilar:
see the line https://github.com/WeblateOrg/weblate/blob/a6bc9cb80e808bf43e78243678716c8e9c5c8aae/weblate/trans/autotranslate.py#L126 

my assumption of the initial pull request was, that the `weblate` machine translation only offers 100% results; but the weblate engine now combines the two engine from before and thus offers _good_ and _bad_ results.
I think we should only call other engines, if the weblate engine did not respond with a 100% result; if the result from weblate is lower than 100% (even above 80% e.g.), other engines could still offer _better_ results..

what do you think?

nijel:
That sounds reasonable.

PR #1848, opened 2018-02-09
Spelling: in bold, the object, the Weblate interface,

comradekingu:
As per https://ell.stackexchange.com/questions/99635/in-bold-letters-vs-in-bold
Not 100% sure about "the object". Could the sentence be explained, what happens?


nijel:
It's help text above administrative form where required fields are **bold**.

nijel:

PR #1850, opened 2018-02-09
Add all iso-codes gettext domains to blacklist

toddy15:
Hi Michal,

you might want to complete the iso-codes gettext domains for this blacklist.

Regards,
Tobias


nijel:
I originally did not want to include those as they did contain quite a lot of generic words, however I've added ignore list for some of these meanwhile anyway, so now it makes sense. Thanks.

Issue #1879, opened 2018-02-28
Download/Upload in XLSX


mlaggner:
when working with external translation agencies (or external translators), it would be helpful to download/upload translations in XLSX since many users work with Excel.

I've tried to use the CSV download, but Excel has giant problems with creation of a compatible CSV for the upload in Weblate.

I am thinking of the following workflow:

- export the translations from weblate via "Files -> Download translations as XLSX"
- send this XLSX file to our external translators
- receive the modified XLSX file
- upload the XLSX file via "Files -> Upload translations"

nijel:
There was fork of Weblate adding Excel export/import, but it was never submitted for merging and quickly looking at it I found several specific hacks specific to their usage (but I might have been wrong). Anyway it probably could be used as starting point for implementing something what could be merged: https://github.com/dontnod/weblate/commits/dev

mlaggner:
thanks for that hint - I'll try to create an exporter/importer next week

nijel:
Fixed by https://github.com/WeblateOrg/weblate/pull/1888 and https://github.com/WeblateOrg/weblate/pull/2164.

* In case you see problem with the fix, please comment on this issue. 
* In case you see similar problem, please open separate issue.
* If you are happy with the outcome, consider supporting Weblate by [donating](https://weblate.org/donate/).

PR #1888, opened 2018-03-07
provide Excel workbook format for download/upload

mlaggner:
This feature provides an external format (Excel workbook) for download
and upload. The upload is solved by converting the Excel workbook into
a csv file which is used for translation merging

addresses WeblateOrg/weblate#1879

Signed-off-by: Manuel Laggner <manuel.laggner@gmail.com>


mlaggner:
just saw that my test-installation and the new code base differ - will cleanup the code to the latest git revision

requirements.txt
```diff
@@ -17,3 +17,4 @@ defusedxml>=0.4
 django-appconf>=1.0
 user-agents>=1.1.0
 filelock>=3.0.1
+openpyxl>=2.5.0
```
nijel: I'd prefer this to be optional dependency and the xlsx to be available only if it's installed.

weblate/trans/models/translation.py
```diff
@@ -899,6 +900,19 @@ def merge_upload(self, request, fileobj, overwrite, author=None,
...
+        # check if the provided file is in an external format
+        # We are able to import different external file formats, but
+        # that needs to convert the external format to a translate-toolkit
+        external_format = external_formats.detect_filename(fileobj.name)
+        if external_format is not None:
+            try:
+                name, content = external_format.convert_to_internal(fileobj.name, filecopy)
+                if name is not None and content is not None:
+                    fileobj.name = name
+                    filecopy = content
```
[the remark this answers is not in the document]
nijel: It really looks weird to silently ignore all errors here.

weblate/utils/validators.py
```diff
@@ -61,7 +61,7 @@
     '.rtf',
     '.xls',
     '.docx',
-    '.xlsx',
+    #'.xlsx',   #allowed due to xlsx translation upload
```
nijel: Please remove the line instead of commenting it out.

nijel:
Thanks, it looks quite good. What I'm really missing are tests for this feature.

weblate/trans/models/translation.py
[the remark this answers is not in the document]
mlaggner: I've no clue which errors can occur in `external_format.convert_to_internal` (since that may execute 3rd party code like openpyxl), so I'd rather swallow all exceptions than let weblate crash.
I could log the exception at least (but did not implement that here since there is nothing else logged in the same file...)

as seen in the `try_load` beneath, you also catch all errors from loading the file formats and only raise the last catched exception if nothing could be loaded..
At my code I catch all exceptions because if no external format can be loaded, the internal format could still load that file...

requirements.txt
mlaggner: sorry, I do not have a clue how to implement optional dependencies in python

weblate/utils/validators.py
mlaggner: sure, I will do

mlaggner:
you are right - in the first place I did not find unit tests for the upload feature, so I did not implement unit tests.. Now I found them and will add some tests

weblate/trans/models/translation.py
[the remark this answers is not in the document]
nijel: At least it should use `weblate.utils.errors.report_error` so that the error appears in Weblate logs and in error monitoring (if enabled).

miloshavlicek:
@mlaggner @nijel Hi, could I ask in what state is the current implementation? Can I help somehow with anything? Thank you very much!

nijel:
The tests are in `weblate/trans/tests/test_files.py`, adding basic upload test should be matter of creating test file and defining the test like:

https://github.com/WeblateOrg/weblate/blob/a065f372a3e1160aba1621803c0bd660f86c3b6b/weblate/trans/tests/test_files.py#L278-L280

It might need some additional tweaks though (see other classes around in the file), but it seems you have already figured it out most of that.

What I currently see missing:

- The tests fail on Travis, so the test case probably needs adjusting.
- There is a merge failure now due to change in the template (7affa84f11fbfb5a65688f7f06d630b5e4e6dd05).
- Make the feature optional, as probably not everybody is going to need this and the dependency chain is quite big here.

requirements.txt
nijel: What would be needed for this:

- Move the dependency to the requirements-optional.txt
- Make the import of `openpyxl` conditional, register the excel format only if it is available

However I can do this later if other issues are addressed.

mlaggner:
Just had a look at the unit test in the CI build:
all pass on python 2.7 but on 3.6 there are several tests failing (but not the new test as far as I could see).. so I am unsure what to fix?!?

according to the optional requirement: I don't know how to do that. Could you give me a hint how to implement an optional requirement in python so I can implement this?

nijel:
I see there is failure in XlsxImportTest, so that seems to be the new code, not old one:

```
FAIL: test_import (weblate.trans.tests.test_files.XlsxImportTest)
  File "/home/travis/build/WeblateOrg/weblate/weblate/trans/tests/test_files.py", line 364, in test_import
    self.assertEqual(translation.stats.translated, 1)
AssertionError: 0 != 1
```

mlaggner:
sorry, I totally misunderstood the error log (and did not even see your comment about the optional requirement above).
Will fix it asap

nijel:
I didn't notice you've fixed the failing tests, I'll review this next week once I'll be back from vacation...

nijel:
Thanks, I've taken this and turned into https://github.com/WeblateOrg/weblate/pull/2164. It more relies on the standard operations and removes special handling of external formats.

PR #1892, opened 2018-03-08
Add "Copy" button to secondary languages

raphaelm:
I just did my first translations with Weblate.

In our scenario, we do translate to two very similar language variants -- informal German and formal German. As a wild guess, they differ in ~20% of our strings. Before we used Weblate, I used translate one of them first with POEdit and then open the other, copy all strings from the first as fuzzy translations and start from there.

With Weblate, I do see the other language variant as a secondary language, which is great. However, I currently need do manually copy the text down into the input field, which needs lots of mouse movement. I thought it would be great to have a "Copy" button for the secondary languages similar to the one already there for the source language.

I implemented this in this PR. The code is a bit less clean and short than I expected it to be because I wanted to support plural forms as well. It works well now, but if you think it should be implemented another way, please let me know.

2018-03-08-155508_1421x543_scrot

2018-03-08-155515_1420x808_scrot

---

  - Unsure how to do a test case for this, happy for any guidance
  - Is this necessary here?

### DIFF
--- a/weblate/templates/format-translation.html
+++ b/weblate/templates/format-translation.html
@@ -1,3 +1,4 @@
+{% load i18n %}
 {% if simple %}
 <span lang="{{ language.code }}" dir="{{ language.direction }}">{{ items.0.content }}</span>
 {% else %}
@@ -5,7 +6,18 @@
   {% for item in items %}
   <div class="list-group-item">
     {% if item.title %}
-    <h5 class="list-group-item-heading"><strong>{{ item.title }}</strong></h5>
+    <h5 class="list-group-item-heading">
+      <strong>{{ item.title }}</strong>
+      {% if copy_to %}
+        <button class="btn btn-xs btn-default copy-text"
+                title="{% trans "Use as translation" %}"
+                data-checksum="{{ copy_to }}"
+                data-index="{{ forloop.counter0 }}"
+                data-content="{{ item.encoded }}">
+          <i class="fa fa-clipboard"></i> {% trans "Copy" %}
+        </button>
+      {% endif %}
+    </h5>
     {% endif %}
     <div class="list-group-item-text" lang="{{ language.code }}" dir="{{ language.direction }}">{{ item.content }}</div>
   </div>
--- a/weblate/templates/translate.html
+++ b/weblate/templates/translate.html
@@ -98,15 +98,26 @@
     {% if secondary %}
     {% for secondary_unit in secondary %}
     <div class="form-group">
-    <label>{{ secondary_unit.translation.language }}</label>
+    <label>
+      {{ secondary_unit.translation.language }}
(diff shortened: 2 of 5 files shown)


weblate/trans/templatetags/translations.py
```diff
@@ -439,6 +441,12 @@ def naturaltime_future(value, now):
         ) % {'count': count}
 
 
+@register.filter
+def json_dumps(value):
+    print(repr(value))
```
nijel: The print should not be here.

weblate/trans/templatetags/translations.py
```diff
@@ -209,12 +210,13 @@ def format_translation(value, language, plural=None, diff=None,
         # Join paragraphs
         content = mark_safe(newline.join(paras))
 
-        parts.append({'title': title, 'content': content})
+        parts.append({'title': title, 'content': content, 'encoded': json.dumps(content)})
```
nijel: The content here is after highlighting whitespace and similar changes, you should probably use raw_value here.

weblate/templates/translate.html
```diff
@@ -98,15 +98,26 @@
     {% if secondary %}
     {% for secondary_unit in secondary %}
     <div class="form-group">
-    <label>{{ secondary_unit.translation.language }}</label>
+    <label>
+      {{ secondary_unit.translation.language }}
+      {% if unit.translation.plural.number == 1 %}
```
nijel: You should do this based on `unit.is_plural`, this way it's never shown for languages having plural, but current string does not having one.

weblate/trans/forms.py
```diff
@@ -230,7 +230,8 @@ def get_toolbar(self, language, fieldname, unit, idx):
         extra_params = COPY_TEMPLATE.format(
             escape(ugettext('Loading…')),
             unit.checksum,
-            escape(json.dumps(source))
+            escape(json.dumps(source)),
+            str(idx)
```
nijel: str() is not necessary here.

weblate/templates/format-translation.html
```diff
@@ -1,11 +1,23 @@
...
 {% else %}
 <div class="list-group">
   {% for item in items %}
   <div class="list-group-item">
     {% if item.title %}
-    <h5 class="list-group-item-heading"><strong>{{ item.title }}</strong></h5>
+    <h5 class="list-group-item-heading">
+      <strong>{{ item.title }}</strong>
+      {% if copy_to %}
+        <button class="btn btn-xs btn-default copy-text"
```
nijel: To be consistent, this should have also `pull-right flip` classes. Similarly for the other button code, maybe they could use shared template?

raphaelm:
Thank you for the feedback, I'll try to make some time in the next days to follow up on this!

PR #1927, opened 2018-03-23
Spelling: Marked noun, MO, PO, POT, package names

comradekingu:
As per https://toolkit.translatehouse.org/
and https://www.gnu.org/software/gettext/


nijel:

PR #1942, opened 2018-03-29
Spelling: Gettext → gettext

comradekingu:
As per https://www.gnu.org/software/gettext/


nijel:

Issue #1965, opened 2018-04-13
Feature Request: Add machine translation support for AWS translate


chocopowwwa:
Product: [AWS Translate](https://aws.amazon.com/translate)

nijel:
It should be pretty easy to implement, there is [AWS Translate API docs](https://docs.aws.amazon.com/translate/latest/dg/API_TranslateText.html), the implementation will be probably quite similar to DeepL (list of supported languages is hardcoded and there is just one API call to get the translation):

https://github.com/WeblateOrg/weblate/blob/3ed316e305193d1800bc034c1feba6f3876ea2ab/weblate/trans/machine/deepl.py#L30-L61

josephvalencia:
Hello I would like to help with this issue and would like to submit an implementation. The configuration with the AWS api is still pending.

nijel:
Thank you for your report, the issue you have reported has just been fixed. 

* In case you see problem with the fix, please comment on this issue. 
* In case you see similar problem, please open separate issue.
* If you are happy with the outcome, consider supporting Weblate by [donating](https://weblate.org/donate/).

PR #1976, opened 2018-04-19
Create aws.py

josephvalencia:
Adds support for AWS.  #1965


nijel:
Thanks! It would be great if you could add some tests for the code...

nijel:
Have you tested this at all? IMHO this could have never worked...

Issue #1820, opened 2018-01-30
Restructure code


nijel:
Historically we have most of the code in the `weblate.trans` module, while some of the things would deserve to be separate top level modules under `weblate`:

* [x] checks
* [x] machine translations
* [x] file formats
* [x] version control
* [x] rename subproject to component to avoid using historical name

There might be more of these and associated tests should moved as well.

BhaaLseN:
Those file formats are a pain point for me as well at this time. I got a new format hacked in until I can get around to clean it up and properly submit it to translate-toolkit; and even then it needs an addition in there.
Right now I have to make sure to reapply my customizations on every update, with the chance of conflicts.

I did start a local branch to seperate out the individual file formats into their own file, but kinda got stuck on how to sensibly load all formats in that module without having to update a file/variable all the time as well (like the `__all__` variable in the `__init__.py` file).
My eventual goal with this was to simply drop my custom format (hacky or not) in there and it-just-works™ (at least until the point where some API changes).

If you have thoughts and ideas on how to go about this, please share them; depending on your 3.0 schedule you may see a PR from me (once i got the tt store sorted out).

nijel:
The alternative approach to registration might be listing them in settings similarly as we do with addons, checks and so on...

BhaaLseN:
Sounds like a sensible choice, and would get of the registration code too (which was one of the issues I had...since I didn't want to split everything up just to see it not work, I started with one format and ended up in an include loop for those functions).
Plus it would clean up the drop down a little; right now we got like 20 or 30 formats in there and use effectively 3.

nijel:
Thank you for your report, the issue you have reported has just been fixed. 

* In case you see problem with the fix, please comment on this issue. 
* In case you see similar problem, please open separate issue.
* If you are happy with the outcome, consider supporting Weblate by [donating](https://weblate.org/donate/).

Issue #1897, opened 2018-03-11
Parameters for File formats


BhaaLseN:
There are a few file formats in translate-toolkit that accept parameters and/or allow for customization, for example the JSON format which at this point has pre-commit hooks to change indentation etc.
I'm also working on a XML format that would currently require hardcoding certain parameters and duplicating the `FileFormat`/`FileUnit` (and to some degree even the translate-toolkit `TranslationStore` depending on the kind of parameter/customization) for every specialization.

It would be nice if the `FileFormat` could expose optional/required parameters as list/map, which would then be presented to the admin when configuring a component (by showing/hiding new textboxes below the format combobox). Internally they could either be passed/unpacked into the `FileFormat.loader.__init__` or pass through a dedicated factory method in case `__init__` is not enough.

There is one caveat I can think of: Autodetection wouldn't really work depending on the kind of parameter/customization. But to be fair, I never used autodetection (primarily because I know what format it should be; but also because it may not find a unique/correct match in some cases); so this might be a non-issue.

nijel:
In (upcoming) Weblate 2.20, this is handled through addons:

![screenshot-2018-3-12 languagetool browser addon](https://user-images.githubusercontent.com/212189/37271282-71710068-25d2-11e8-9e83-0228f18c361a.png)

This is probably not perfect place, but was easiest to integrate.

BhaaLseN:
Nice! You've commented on translate/translate#3776 so I suppose you also saw the `__init__?` methods there. If addons allow me to customize those, I can totally live with that!

nijel:
Currently it's rather post init where the settings are applied (because this is way it's implemented for current formats as that only controls saving and not load).

The problem for passing args to __init__ there is chicken-egg problem: the addon can be configured only after component is created, so it really can not influence settings for first loading...

BhaaLseN:
Hm, that would be a little difficult, especially since I have a few cases where I want to force one specific element name but ignore (and also preserve!) others - so I cannot really go and do some parameter autodetection during load (which might be wrong) and change them later on.
That would also mean my flat xml format is basically useless at this point (unless they go and define a copy of the format inside Weblate with the right parameters).

I can personally live with that for my own on-premise installation, but hosted Weblate users won't have that luxury.

But...what about receiving back a map from a pre-init event and unwrapping it into `self.subproject.file_format_cls.parse`? All before the actual store is created.
[...]

PR #1985, opened 2018-04-23
Refactor formats

nijel:
See #1820


nijel:
@BhaaLseN can you please check if this works well for your use case?

BhaaLseN:
From a quick skim it should help me reduce friction on updates (because I can now add my customizations to a folder than isn't under weblates' version control and simply modify `WEBLATE_FORMATS` to import them).
It doesn't solve the initialization parameter pain from #1897, but it wasn't meant to; so I'm ok with that (again, since I can just dump my specific formats that do the initialization magic in a separate file that isn't touched during upgrades unless necessary).

So in general 👍 from me

PR #2004, opened 2018-05-08
fix sap translation hub to run on python 3

mlaggner:
the translation hub translation engine was not able to be run on python 3


nijel:
Wouldn't the decoded string work for Python 2 as well? Also it would be great to have this covered by the tests, so that the fix can be verified automatically.

mlaggner:
according to the docs of `base64.b64encode` this method accepts/returns str/str in python 2.7, but byte/byte in python 3
does a decode of a string with a following concatenation to a string work in python 2?

nijel:
It will return unicode string, I'm not sure whether it will work fine in `add_header`. Anyway the tests would give an instant answer for that :-).

nijel:

PR #2036, opened 2018-05-23
Make development easier for newcomers

GeorgeBekh:
This ecosystem is based on Docker, it can run project locally without developer worrying about setting up the environment.

Checklist:


nijel:
The dependencies can be an issue, though I have no clue how to address that. Other thing is that something should discourage users using this in production as it will not perform well.

GeorgeBekh:
Well, it will not work in production, because ports are open on localhost only, unless user changes it by hand but in that case we can assume he knows what he is doing.

Also, changing docker-compose.yml to docker-compose.dev.yml may help because it will not work with docker-compose right out of the box: only with `-f` flag

GeorgeBekh:
About dependencies
One of the solutions is that we can use production docker image as base and build development image (put Dockerfile in this repo) on top of that: update some packages, install new, etc. And then, when a new stable weblate is coming out, just move dependencies from development dockerfile to production one.

On of the disadvantages of this solution is that you can forget to update the production image. Although  if you update it manually in current workflow anyway, then it's not a problem, i guess :)

nijel:
Yes, the production image is updated manually right now (it usually anyway needs configuration changes as well).

docker-compose.dev.yml
```diff
@@ -0,0 +1,22 @@
...
+# https://github.com/WeblateOrg/docker
+#########
+
+services:
+  weblate:
+    build:
+      context: .
+      dockerfile: dev.Dockerfile
+    ports:
+      - 127.0.0.1:8080:8080
```
GeorgeBekh: works only on localhost

docker-compose.dev.yml
```diff
@@ -0,0 +1,22 @@
...
+    volumes:
+      - $PWD:/srv
+    working_dir: /srv
+    entrypoint: sh -c "./manage.py migrate && ./manage.py createadmin --update --password admin && ./manage.py runserver 0.0.0.0:8080"
```
GeorgeBekh: @nijel can we make this more pretty?
Fewer commands to get it running would be nice

dev.Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+##########
+# WARNING!
+# This dockerfile is meant to be used in the development process
+# and WILL perform very poorly in production.
+#
+# For production-ready dockerfile see:
+# https://github.com/WeblateOrg/docker/tree/docker
+#########
+
+# TODO: put some new dependencies here
```
GeorgeBekh: You can put new development dependencies here

@nijel I should probably add `pip install -r requirements.txt` here if it installs packages globally, may be something else?

docker-compose.dev.yml
nijel: This is probably minimum to make it work reasonably...

dev.Dockerfile
nijel: `pip install` should IMHO work fine here.

GeorgeBekh:
@nijel I added dev dependencies installation, what do i do next?
Should i add docs, or we can just give it a try and see how it works out?
Would you try to use this configuration in your development workflow so we could know what to improve?

nijel:
Having some docs is certainly needed, otherwise nobody will find this. Probably should be part of https://github.com/WeblateOrg/weblate/blob/master/docs/contributing.rst

As for my development workflow it's not practical as I typically use locally patched translate-toolkit to have latest fixes, so it would mean getting that into the container as well.

GeorgeBekh:
Ok, i'll add some docs.

Can you simply describe how you use it locally? May be just mounting it inside the docker container will be enough!

GeorgeBekh:
Also i've noticed that changing templates' code requires restart of the whole app, is there any other way to refresh them?

nijel:
I use it locally installed using `pip install -e`. Mounting it in the right place would probably do the trick.

The restart to refresh templates is only needed if templates caching is enabled, what is in the example settings:

https://github.com/WeblateOrg/weblate/blob/c861729915b5b1fd948b3b9259e39c6ea1f7a3cc/weblate/settings_example.py#L189-L194

Changing it to  following would remove that:

```python
            'loaders': [
                    'django.template.loaders.filesystem.Loader',
                    'django.template.loaders.app_directories.Loader',
            ],
```

PS: I currently do not use Docker at all and don't have it on my developer machine

GeorgeBekh:
Ok, thanks! i should probably create some kind of `settings_dev.py` (or patch) then with caching disabled, for environment to be more friendly for a developer?

So you probably won't be using this configuration? Or you are ready to install docker? :smiley:

nijel:
Probably creating `settings_dev.py` and changing just few values (similarly like `settings_test.py` does) is best for long term maintenance.

GeorgeBekh:
Fine, i'll try

nijel:

PR #2049, opened 2018-05-25
incosistent permissions in template fixed

GeorgeBekh:
Made rules of rendering the body same as rendering the button 
https://github.com/WeblateOrg/weblate/blob/master/weblate/templates/translation.html#L63-L65

### DIFF
--- a/weblate/templates/translation.html
+++ b/weblate/templates/translation.html
@@ -256,7 +256,7 @@
 </div>
 {% endif %}
 
-{% if object.is_template and object.component.file_format_cls.can_add_unit and user_can_add_unit %}
+{% if object.is_template and user_can_add_unit %}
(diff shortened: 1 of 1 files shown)


GeorgeBekh:
May be let's make separate variable for each tab to avoid such inconsistency in future?

nijel:
This is actually the wrong way, both things should be checked. IMHO it's better to do all the checks here:

https://github.com/WeblateOrg/weblate/blob/2e7bb66df33ef1798e1cd749e3a036ddca1f992c/weblate/auth/permissions.py#L167-L172

GeorgeBekh:
Can you give an example?

nijel:
I've implemented it in 45043043a31289101444185820a99c8735cb56e0.

Issue #2045, opened 2018-05-24
Date input selector twice


rugk:
### Steps to reproduce
1. Go to translation search.
2. Click into the date at the right.

### Actual behaviour
The browser's own and your own selector are shown:

![grafik](https://user-images.githubusercontent.com/11966684/40515173-17e9d420-5fac-11e8-8cc5-9b0fe4519ee3.png)

### Expected behaviour

Show one thing. Maybe you can use some feature detection or browser detection to not show your own one?

Also the upper one (I guess that is your custom one) is being overlapped by the header, as it can be seen in the screenshot, so this also is not how it should behave… 

### Server configuration
https://hosted.weblate.org/

Firefox 60

rugk:
It is, that's what I see…

GeorgeBekh:
Another option is to change input type from "date" to "text" which will work just fine with bootstrap datepicker, although it's a kind of hack from a html semantics point of view

nijel:
Thank you for your report, the issue you have reported has just been fixed by https://github.com/WeblateOrg/weblate/pull/2054

* In case you see problem with the fix, please comment on this issue. 
* In case you see similar problem, please open separate issue.
* If you are happy with the outcome, consider supporting Weblate by [donating](https://weblate.org/donate/).

PR #2054, opened 2018-05-28
Disable bootstrap datepicker if browser provides one

GeorgeBekh:
https://github.com/WeblateOrg/weblate/issues/2045


GeorgeBekh:
Testing needed

I've tested that it works with Firefox 58 and Chromium 62.
I've **not** tested it on browsers where native datepicker is not provided

nijel:
Tested in Firefox 60 and it works fine, thanks.

PR #2071, opened 2018-06-02
Show empty files as fully translated

pleonex:
Empty components (like .po files without translation units) shows in the Dashboard as 0%.
This was happening to us after importing automatically many .po files extracted from files that are expected to be empty in some cases.


nijel:

PR #2089, opened 2018-06-14
Mid sentence gettext

comradekingu:
Also, https://github.com/WeblateOrg/website/pull/34

As per https://www.gnu.org/software/gettext/ and https://en.wikipedia.org/wiki/Gettext


nijel:

PR #2098, opened 2018-06-24
Fix possible error on mass state change

atallahade:
I had a 500 error while performing a mass state change. I can't tell exactly why it happened but I've run the mass state change right after I've reverted some commits followed by a rebase, so I guess some index were invalid or something.
Anyway, I think it still a good idea to do some action if the `new_target` parameter is empty and not just throw an exception.

### DIFF
--- a/weblate/trans/models/unit.py
+++ b/weblate/trans/models/unit.py
@@ -1036,7 +1036,7 @@ class Unit(models.Model, LoggerMixin):
             self.target = new_target
         else:
             self.target = join_plural(new_target)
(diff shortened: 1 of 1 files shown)


nijel:
The code is indeed buggy, but your fix is wrong - it will allow to mark empty strings as translated. I've instead fixed it differently in dacb8e29a14262fa2195292c14096ec77e4e5676. Thanks for pointing out the issue.

atallahade:
Thanks for fixing it!

PR #2132, opened 2018-07-19
Mention health check URL in list of possible exceptions

Pitel:


nijel:

PR #2133, opened 2018-07-21
Add event `unit_post_save` for use with addons

pvsune:
This adds new event handler when unit is updated. One specific use case is to call an addon when translation (or unit) state is updated from "Waiting for Review" to "Approved". This uses built-in Django signal `django.db.models.post_save` and filters sender if of class `Unit`.

P.S. This is my first time contributing to open-source would be glad to know how to do it properly. I have missing tests and will be pushed in a while.

### DIFF
--- a/weblate/addons/models.py
+++ b/weblate/addons/models.py
@@ -24,6 +24,7 @@ from appconf import AppConf
 
 from django.db import models
 from django.db.models import Q
+from django.db.models.signals import post_save
 from django.dispatch import receiver
 from django.urls import reverse
 from django.utils.encoding import python_2_unicode_compatible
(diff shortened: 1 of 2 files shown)


weblate/addons/models.py
```diff
@@ -183,6 +184,17 @@ def unit_pre_create_handler(sender, unit, **kwargs):
         addon.addon.unit_pre_create(unit)
 
 
+@receiver(post_save)
```
nijel: You should filter here to avoid signal trigger for all models, see https://docs.djangoproject.com/en/2.0/topics/signals/#connecting-to-signals-sent-by-specific-senders, then the `if sender...` block is not needed.

pvsune: You're right. That looks cleaner. TIL

nijel:
Merged, thanks for your contribution! I've applied following fixes:

* 6391c26f1d Add migration for new addon signal
* 2a3fbf048f Filter sender at signal registration

pvsune:
Thanks @nijel! I was about to push today. Thanks for filling in the missing pieces. :)

PR #2143, opened 2018-07-26
Include instructions for virtualenv in Quick Install page

### LINKED ISSUE #2140: Documentation: add virtualenv install as an option in Quick Install
It would be nice to add a set of instructions for installation in virtualenv to the Quick start part. It's really just a series of quick commands, and people can do a quick local testing without messing with anything outside the virtualenv.

The reason why I'm using the virtualenv-with-pip way is because:
- I don't want to install the whole docker machinery
- I don't want to install all the dependencies system-wide, as they might interfere with other projects
- I don't want to try the development version
- I don't want to mess with my apache settings
- I just want to quickly test locally what kind of admin options are offered, as that isn't possible in the demo instance.

### Steps to reproduce

1. Follow the steps 1 to 5 in https://docs.weblate.org/en/latest/admin/install.html#installing-in-virtualenv , for reasons mentioned above
2. read step 7 - now what?

I did:
- copy /tmp/weblate/lib/python2.7/site-packages/weblate/settings-example.py to /tmp/weblate/lib/python2.7/site-packages/weblate/settings.py
- adjust a couple of values there in settings.py
- run `python /tmp/weblate/lib/python2.7/site-packages/weblate/wsgi.py`, get `django.db.utils.OperationalError: no such table: wladmin_configurationerror` (logically, as I haven't set up the database yet)
- not understand "Don’t forget to set Python search path to your virtualenv as well (for example using virtualenv = /tmp/weblate in uwsgi)." - please could you just put the command there or which file to edit, at least if it's relevant for testing? I think this isn't relevant when I'm just doing a test run, but not sure.
- run `weblate migrate`, and got `django.core.exceptions.ImproperlyConfigured: Requested setting GITHUB_USERNAME, but settings are not configured. You must either define the environment variable DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.`
- finally figured out the correct order of commands, fortunately I already had a bit of django experience.

### Actual behaviour

- instructions for virtualenv are not suitable for a quick start, but a virtualenv would be the perfect candidate for it.

### Expected behaviour

Quick start Instructions say something along these lines (after explaining the virtualenv setup as in https://docs.weblate.org/en/latest/admin/install.html#installing-in-virtualenv):
________________

If you just want to do a quick local test setup, to see if Weblate is for you, then:

(Insert steps 1 to 4 from above instructions)

5. Copy the file /tmp/weblate/lib/python2.7/site-packages/weblate/settings-example.py to /tmp/weblate/lib/python2.7/site-packages/weblate/settings.py
6. Adjust a couple of values in the new settings.py (are there any that are absolutely necessary? Or would it work with defaults?)
7. Run `export DJANGO_SETTINGS_MODULE=weblate.settings`
8. Run `weblate migrate`
9. Run `weblate createadmin`
10. Run `weblate runserver`
10. Open a webbrowser, go to http://localhost:8000/accounts/login/ and login with name and pw as was output by above command

### Server configuration and status

Version is 3.0.1., installed via pip in virtualenv

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/61295918-documentation-add-virtualenv-install-as-an-option-in-quick-install?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Moini:
Fixes #2140

Formatting might not be entirely correct, I've never used ReST before.  
I've put the virtualenv instructions first, because I find it the easiest and fastest way.  
Let me know about any changes you would like me to make, please.

-----------------------

- [ ] ~~Any code changes come with testcase~~
- [ ] ~~Any new functionality is covered by user documentation~~

### DIFF
--- a/docs/admin/quick.rst
+++ b/docs/admin/quick.rst
@@ -7,6 +7,105 @@ Quick setup guide
     testing purposes. Please check :ref:`install` for more real world setup
     instructions.
 
+Installing in a virtualenv
+--------------------------
+
+If you'd just like to do a quick installation locally on your device to find 
+out if Weblate is for you, you can install it using a virtual environment for 
+Python 2, a simple (and slow!) SQLite database, and the lightweight Django 
+development server.
+
+#. Install development files for libraries needed for building some
+   Python modules:
+
+   .. code-block:: sh
+
+        # Debian/Ubuntu:
+        apt install libxml2-dev libxslt-dev libfreetype6-dev libjpeg-dev libz-dev libyaml-dev python-dev
+
+        # openSUSE/SLES:
+        zypper install libxslt-devel libxml2-devel freetype-devel libjpeg-devel zlib-devel libyaml-devel python-devel
+
+        # Fedora/RHEL/CentOS:
+        dnf install libxslt-devel libxml2-devel freetype-devel libjpeg-devel zlib-devel libyaml-devel python-devel
+
+#. Install pip and virtualenv. Usually they are shipped by your distribution or
+   with Python:
+
+   .. code-block:: sh
+
+        # Debian/Ubuntu:
+        apt-get install python-pip python-virtualenv
+
+        # openSUSE/SLES:
+        zypper install python-pip python-virtualenv
+
+        # Fedora/RHEL/CentOS:
+        dnf install python-pip python-virtualenv
+
+#. Create the virtualenv for Weblate (the path in ``/tmp`` is really
+   just an example, you rather want something more permanent, even if this is
+   just for testing):
+
+   .. code-block:: sh
+
+        virtualenv --python=python2.7 /tmp/weblate
+     
+#. Activate the virtualenv for Weblate, so Weblate will look for Python libraries there first:
+        
+   .. code-block:: sh
+    
+        . /tmp/weblate/bin/activate
+
+#. Install Weblate including all dependencies. You can also use pip to install
+   the optional dependencies:
+
+   .. code-block:: sh
+        
+        pip install Weblate
+        # Optional deps
+        pip install pytz python-bidi PyYAML Babel pyuca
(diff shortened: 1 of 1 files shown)


Moini:
(Can't sign via web interface. Of course, I agree to the license.)

nijel:
You can signoff just by adding the signed-off-by line to the commit message, but it really doesn't matter.

docs/admin/quick.rst
```diff
@@ -70,6 +169,8 @@ Installing on OpenShift 2
    
     For more information, including on how to retrieve the generated admin password, see :ref:`openshift`.
 
+ .. _adding-translation:
+
```
nijel: docs/devel/translations.rst:7:duplicate label adding-translation, other instance in docs/admin/quick.rst, you should choose different label here.

Moini: Okay, so that label is auto-added for the heading? Will it work if I just remove it? 
I have read that auto-label-adding for headings requires a plugin in Sphinx, and I wasn't sure which plugins you may have installed.

nijel: AFAIK there are no automatic labels, so just choose something different in this document.

Moini: Thanks!

Moini: Done, hope it's going to work now. 

The reason for assuming auto-labelling is that there is no other instance of 'adding-translation' in the file, although the error message claims that there is.

Moini:
Thanks for merging!

PR #2145, opened 2018-07-26
Fix reference to numbers in QuickStart guide

Moini:
Sorry, the numbers changed after I had first written this - I had made a mental note to fix it, but forgot... :-/


nijel:

PR #2171, opened 2018-08-11
Open XML, Excel interpretation string

comradekingu:
The format is called "Office Open XML", as per https://en.wikipedia.org/wiki/Office_Open_XML

The Excel specific variant markup seems to be "SpreadsheetML", as per https://docs.microsoft.com/en-us/office/open-xml/how-to-parse-and-read-a-large-spreadsheet

Maybe "Excel Office Open XML" is correct? Though the tool should be irrelevant, I assume it is not.


nijel:

PR #2172, opened 2018-08-11
Invalidate caches just once per request

nijel:
There should be no need to do that several times over object lifetime
(what is typically request).


nijel:
Thinking more about this, we should probably invalidate cache at the end. Otherwise other request can calculate stats between and those would not be invalidated.

PR #2193, opened 2018-08-23
github/issue template: do not launch new containers to report versions

glensc:
using run would launch another instance of postgres and redis,

this not needed and not sure how well postgres handles concurrent writes from different process to same data store.


nijel:
Good point, thanks!

PR #2224, opened 2018-09-07
implement 'replace' tag for replacing a char

the-bionic:

### DIFF
--- a/weblate/utils/render.py
+++ b/weblate/utils/render.py
@@ -22,8 +22,10 @@ from __future__ import unicode_literals
 
 from django.template import Template, Context, Engine
 from django.utils.translation import override
-
+from django import template
 from weblate.utils.site import get_site_url
+
+register = template.Library()
 
 
 class RestrictedEngine(Engine):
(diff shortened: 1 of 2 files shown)


weblate/utils/render.py
```diff
@@ -36,6 +38,11 @@ def __init__(self, *args, **kwargs):
         super(RestrictedEngine, self).__init__(*args, **kwargs)
 
 
+@register.simple_tag()
```
nijel: AFAIK this won't work, you need to register the tag in the RestrictedEngine, otherwise it will not work there.

weblate/utils/tests/test_render.py
```diff
@@ -36,3 +36,8 @@ def test_float(self):
     def test_float_cs(self):
         with override('cs'):
             self.test_float()
+
+    def test_replace(self):
+        result = replace('a-string-with-dashes', '-', ' ')
+        self.assertEqual(result, 'a string with dashes')
+
```
nijel: Can you please test this through render_template, so that we know that the tag is registered there?

weblate/utils/render.py
the-bionic: I'll do that.

the-bionic: In your opinion, where do you think is best place to write the replace tag function?

PR #2299, opened 2018-10-08
Textual searchbox for glossaries

### LINKED ISSUE #1717: glossaries do not have a search box.
### Steps to reproduce
1. you have a large glossary
2. you need to check an entry if there is or not. 
3. you don't have a simple search box for entries.

### Expected behaviour
Put a searh box on every glossary page.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/52493566-glossaries-do-not-have-a-search-box?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Naruto0:
Fixes #1717

### DIFF
--- a/weblate/templates/dictionary.html
+++ b/weblate/templates/dictionary.html
@@ -84,6 +84,10 @@
 {% else %}
 <p>{% trans "No words found!" %}</p>
 {% endif %}
+<form class="searchform" action="{% url 'show_dictionary' project=project.slug lang=language.code %}">
+{{ searchform | crispy }}
+<button type="submit" class="btn btn-ok"> {% trans "Search" %} </button>
+</form>
 </div>
 
 {% if user_can_add_dictionary %}
--- a/weblate/trans/views/dictionary.py
+++ b/weblate/trans/views/dictionary.py
@@ -39,7 +39,7 @@ from weblate.lang.models import Language
 from weblate.utils.site import get_site_url
 from weblate.utils.errors import report_error
 from weblate.trans.util import render, redirect_next, redirect_param
-from weblate.trans.forms import WordForm, DictUploadForm, LetterForm
+from weblate.trans.forms import WordForm, DictUploadForm, LetterForm, OneWordForm
 from weblate.utils.views import get_project, import_message
 from weblate.utils.views import get_paginator
 
@@ -310,6 +310,16 @@ def show_dictionary(request, project, lang):
 
     letterform = LetterForm(request.GET)
 
+    searchform = OneWordForm(request.GET)
+
+    if searchform.is_valid() and searchform.cleaned_data['term'] != '':
+        words = words.filter(
+            source__icontains=searchform.cleaned_data['term']
+        )
+        search = searchform.cleaned_data['term']
+    else:
+        search = ''
+
     if letterform.is_valid() and letterform.cleaned_data['letter'] != '':
         words = words.filter(
             source__istartswith=letterform.cleaned_data['letter']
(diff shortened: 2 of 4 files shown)


weblate/trans/views/dictionary.py
```diff
@@ -310,11 +310,28 @@ def show_dictionary(request, project, lang):
...
+
+    # print(searchform.is_valid())
+
+    if searchform.is_valid() and searchform.cleaned_data['term'] != '':
+        # print(searchform.cleaned_data['term'])
+        words = words.filter(
+            source__icontains=searchform.cleaned_data['term']
+        )
+        search = searchform.cleaned_data['term']
+        query_string = 'term={}'.format(search)
```
nijel: You should probably urlencode the search term here.

weblate/trans/views/dictionary.py
```diff
@@ -310,11 +310,28 @@ def show_dictionary(request, project, lang):
...
+    else:
+        search = ''
+
     if letterform.is_valid() and letterform.cleaned_data['letter'] != '':
         words = words.filter(
             source__istartswith=letterform.cleaned_data['letter']
         )
         letter = letterform.cleaned_data['letter']
+        query_string = 'letter={}'.format(letter)
```
nijel: This would break when using search *and* letter together, what is probably not what users would expect.

weblate/trans/views/dictionary.py
```diff
@@ -334,9 +351,11 @@ def show_dictionary(request, project, lang):
             'language': lang,
             'page_obj': words,
             'form': form,
-            'query_string': 'letter={}'.format(letter) if letter else '',
+            'query_string': query_string,
             'uploadform': uploadform,
             'letterform': letterform,
+            'searchform': searchform,
+            'search': search,
```
nijel: Is this variable used somewhere in the template?

Naruto0: Yes, it is the value of the `term` input field.

weblate/trans/views/dictionary.py
Naruto0: What would you suggest? I tried it on running server and it either takes one or another. Previously i was asking if it's necessary to have both methods.
Thanks in advence!

weblate/templates/dictionary.html
```diff
@@ -84,6 +84,11 @@
 {% else %}
 <p>{% trans "No words found!" %}</p>
 {% endif %}
+<form class="searchform" action="{% url 'show_dictionary' project=project.slug lang=language.code %}">
+{% csrf_token %}
+<input name="term" value="{{ search }}"/>
```
nijel: You should render the form here instead of manually creating the markup. Probably just '{{ searchform }}' will do that.

weblate/trans/views/dictionary.py
nijel: Ah, but that should rendered though form, not manually.

weblate/trans/views/dictionary.py
nijel: You can build the query parameters gradually in a dict and then build whole query string in the end by django.utils.http.urlencode. That will address urlencoding as well.

weblate/trans/views/dictionary.py
```diff
@@ -39,7 +39,7 @@
 from weblate.utils.site import get_site_url
 from weblate.utils.errors import report_error
 from weblate.trans.util import render, redirect_next, redirect_param
-from weblate.trans.forms import WordForm, DictUploadForm, LetterForm
+from weblate.trans.forms import WordForm, DictUploadForm, LetterForm, OneWordForm
```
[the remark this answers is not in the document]
Naruto0: Should i pay attention to this @nijel? Or is it just cosmetic issue?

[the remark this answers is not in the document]
nijel: I try to wrap all lines at 80 chars, but maybe this limit is too strict these days...

nijel:

PR #2306, opened 2018-10-10
Vote suggestion up instead of adding the same

### LINKED ISSUE #1375: Vote for suggestion when adding duplicate
Currently duplicate suggestions are ignored, but it would be better if it would be accepted as upvote in case voting is enabled for given translation.

See also https://github.com/WeblateOrg/weblate/issues/1348#issuecomment-280706768

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/42236397-vote-for-suggestion-when-adding-duplicate?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Naruto0:
Fixes #1375

### DIFF
--- a/weblate/trans/models/suggestion.py
+++ b/weblate/trans/models/suggestion.py
@@ -35,6 +35,7 @@ from weblate.utils.antispam import report_spam
 from weblate.utils.fields import JSONField
 from weblate.utils.state import STATE_TRANSLATED
 from weblate.utils.request import get_ip_address
+from django.core.exceptions import ObjectDoesNotExist
 
 
 class SuggestionManager(models.Manager):
@@ -44,15 +45,22 @@ class SuggestionManager(models.Manager):
         """Create new suggestion for this unit."""
         user = request.user
 
-        same = self.filter(
-            target=target,
-            content_hash=unit.content_hash,
-            language=unit.translation.language,
-            project=unit.translation.component.project,
-        )
+        try:
+            same = self.get(
+                target=target,
+                content_hash=unit.content_hash,
+                language=unit.translation.language,
+                project=unit.translation.component.project,
+            )
 
-        if same.exists() or (unit.target == target and not unit.fuzzy):
-            return False
+            if same.user == user or not vote:
+                return False
+            else:
+                same.add_vote(unit.translation, request, True)
+                return False
+
+        except ObjectDoesNotExist:
+            pass
 
         # Create the suggestion
         suggestion = self.create(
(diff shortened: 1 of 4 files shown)


weblate/trans/models/suggestion.py
```diff
@@ -44,15 +45,22 @@ def add(self, unit, target, request, vote=False):
...
-        )
+        try:
+            same = self.get(
+                target=target,
+                content_hash=unit.content_hash,
+                language=unit.translation.language,
+                project=unit.translation.component.project,
+            )
+
+            if same.user == request.user:
```
nijel: Better use `user` instead of `request.user` here to avoid confusion.

weblate/trans/models/suggestion.py
```diff
@@ -44,15 +45,22 @@ def add(self, unit, target, request, vote=False):
...
+            same = self.get(
+                target=target,
+                content_hash=unit.content_hash,
+                language=unit.translation.language,
+                project=unit.translation.component.project,
+            )
+
+            if same.user == user:
+                return False
+            else:
```
nijel:
```suggestion
            elif vote:
```

It makes sense to vote only if user is allowed to vote (what also checks whether the component has voting enabled).

weblate/trans/models/suggestion.py
```diff
@@ -44,15 +45,22 @@ def add(self, unit, target, request, vote=False):
...
-        )
+        try:
+            same = self.get(
+                target=target,
+                content_hash=unit.content_hash,
+                language=unit.translation.language,
+                project=unit.translation.component.project,
+            )
+
+            if same.user == user:
```
nijel:
```suggestion
            if same.user == user or not vote:
```

This is actually better fix for voting disabled case.

weblate/trans/tests/test_suggestions.py
```diff
@@ -27,9 +27,11 @@
 
 from weblate.trans.models import Suggestion
 from weblate.trans.tests.test_views import ViewTestCase
+from weblate.auth.models import User
 
 
 class SuggestionsTest(ViewTestCase):
+
```
nijel: It seems tests are gone?

nijel:
I just need to find time to review and merge this. I'm quite busy these days, sorry.

Naruto0:
No problem, just wanted to bump.

nijel:

PR #2348, opened 2018-10-25
explicitly launching dev container with python3

HaroldGray:
Previously choosing python version ambiguously and defaulting to python2.7 as is configured in the container layer above. pip3 installed packages could not be found by python2.7 and it never found the django dependency.

I'm off to bed for now, but I will revisit tomorrow with the aim of getting the dev container to work. :)
Thanks for your patience as I slowly learn python.

Signed-off-by: Harold Gray <haroldgray3@gmail.com>


nijel:

PR #2350, opened 2018-10-25
Spelling: "All commits made with Weblate"

comradekingu:
This clarifies that it is a commit sent using Weblate, as opposed to on behalf of Weblate itself.


nijel:

PR #2351, opened 2018-10-25
setting module import to explicitly

HaroldGray:
Signed-off-by: Harold Gray <haroldgray3@gmail.com>

Was getting `ImportError: No module named 'settings_example’`. Set the import to explicitly pull from `weblate.settings_example`.

Dev docker container seems to be working now.


nijel:

PR #2357, opened 2018-10-30
Spelling: a unique, CAPTCHA, Leave empty

comradekingu:
Registered a new account and found the description of a field right next to it to be redundant.
It seems to me like this change is not a problem for blind/visually impaired users.
Also, accepting any password, and just disregarding it means it could be omitted altogether.

CAPTCHA, as per https://en.wikipedia.org/wiki/CAPTCHA


nijel:

nijel:
The motivation for showing the field even if no password is set is to avoid leaking information whether user has a password.

Issue #1040, opened 2016-03-23
Group + mark translations as reviewed on import


rvanlaak:
### Steps to reproduce
Import project `.xlf` translations from `.yml` format via `./manage.py import`

### Expected behaviour
We migrated our `.yml`files to `.xlf`, the translations are imported okay but we would expect the translations to be marked as reviewed.

* So, maybe we can add an extra parameter to mark everything as reviewed on import? Or maybe this should be the default behavior?
* How would the Weblate import handle [xliff groups](http://docs.oasis-open.org/xliff/xliff-core/xliff-core.html#group)? Our `.yml` files were nested, which is a really nice way of grouping, and if Weblate is able to handle these groups we could start using that to group the translations per page.

### Actual behaviour
The translations are marked as "need review".

### Server configuration
Latest `master`

nijel:
You mean import_project?

Anyway in XLIFF you need to have approved="yes", otherwise the string is considered as for review.

rvanlaak:
Okay great, then I'll add that to the `xlf` export script. See https://github.com/lexik/LexikTranslationBundle/pull/198 for that PR.

What about the grouping? This might be related to #263 ?

nijel:
No that's grouping of components in Weblate. I think groups should work, but you probably can't use same IDs in different groups (I've never tested this, but I think this won't work properly).

rvanlaak:
Okay that's clear, knowing that it never was tested already is a start :+1: 

Maybe @phlax can clarify what the functionality of xliff grouping within Weblate could provide? Is grouping meant to group small sets of translations? Like having translations per page?

pschneider:
> Anyway in XLIFF you need to have approved="yes", otherwise the string is considered as for review.

Hello,
is there a way to disable this behaviour? (I've already tried disabling all quality checks for this component) In our company we do not need this flag but it would still be great to see "not translated" translations counting into translation progress of the component.
Unfortunately the Xcode XLIFF export does not respect the `approved="yes"` attributes but overrides the file completely instead. Unfortunately I haven't found a way to automatically merge a Weblate file containing the `approved` attributes with an Xcode merged file. :-1: 

Any help appreciated! Thanks

rvanlaak:
My guess would be that the XLIFF standard states that `approved` should be `yes`, and weblate (read: `translation-toolkit`) follows this. Would be great if more tools (same for PHPStorm) would start supporting these attributes.

pschneider:
[...]

PR #2372, opened 2018-11-06
Clarify supported states for XLIFF format

PowerKiKi:
Previous documentation incorrectly mentioned "fuzzy" which is a concept that
does not exists in XLIFF and was also replaced in Weblate by "Needs editing".

More importantly it was very unclear that the `state` attribute is never used
at all. But instead only `approved` attribute is used. That lead to even more
confusion for the reader.

The following tables show all possibles cases in both directions:

## From XLIFF file to Weblate

| XLIFF file            | Weblate       |
|-----------------------|---------------|
| absence of `<target>` | Empty         |
| absence of `approved` | Needs editing |
| `approved="no"`       | Needs editing |
| `approved="yes"`      | Approved      |

## From Weblate to XLIFF file

| Weblate            | XLIFF file       |
|--------------------|------------------|
| Needs editing      | `approved="no"`  |
| Waiting for review | `approved="no"`  |
| Approved           | `approved="yes"` |

This should address the questions raised in
https://github.com/WeblateOrg/weblate/issues/1040#issuecomment-433366326

### DIFF
--- a/docs/formats.rst
+++ b/docs/formats.rst
@@ -210,17 +210,20 @@ is one of many standards in this area.
 
 XLIFF is usually used as bilingual, but Weblate supports it as monolingual as well.
 
-Translations marked for review
-++++++++++++++++++++++++++++++
+Translations states
++++++++++++++++++++
 
-.. versionchanged:: 2.18
+The ``state`` attribute in the file is completely ignored. It cannot be read and it
+cannot be written by Weblate. That means that a unit is considered translated as
+soon as a ``<target>`` element exists.
 
-    Since version 2.18 Weblate differentiates approved and fuzzy states, so
-    it should work as expected with Xliff. You still might apply note below in
-    cases where you don't want to use review process in Weblate.
+Also if the translation unit has ``approved="yes"`` it will be imported into Weblate
+as approved, anything else will be imported as waiting for review (which matches XLIFF
+specification).
 
-If the translation unit doesn't have ``approved="yes"`` it will be imported into
-Weblate as needing review (which matches XLIFF specification).
+That means that when using XLIFF format, it is strongly recommended to enable Weblate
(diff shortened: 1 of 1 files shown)


nijel:
Are you sure with that? The "Needs editing" state should not be set at all (for bilingual files, for monolingual it can be set if source string changes), see the code:

https://github.com/WeblateOrg/weblate/blob/67580c358c2c7e77dfded10e1019a881296863c4/weblate/formats/ttkit.py#L104-L116

PowerKiKi:
Mmmhhh, you are right. I mixed up "Needs editing" and "Translated"/"Waiting for review"...

I set up a git repo with [all-cases.xliff.txt](https://github.com/WeblateOrg/weblate/files/2551896/all-cases.xliff.txt) and the result in DB after a `./manage.py updategit --all` was as follow. Notice the SQL comment indicating when the screenshot was taken:

weblate

So we can have 0, 20 or 30 in DB which are "Empty", "Translated"/"Waiting for review" and "Approved", according to:

https://github.com/WeblateOrg/weblate/blob/67580c358c2c7e77dfded10e1019a881296863c4/weblate/utils/state.py#L25-L34

So I to amended the commit message and content to reflect this. The new tables are as follow:

## From XLIFF file to Weblate

| XLIFF file            | Weblate            |
|-----------------------|--------------------|
| absence of `<target>` | Empty              |
| absence of `approved` | Waiting for review |
| `approved="no"`       | Waiting for review |
| `approved="yes"`      | Approved           |

## From Weblate to XLIFF file

| Weblate            | XLIFF file       |
|--------------------|------------------|
| Needs editing      | `approved="no"`  |
| Waiting for review | `approved="no"`  |
| Approved           | `approved="yes"` |

nijel:
Thanks for review.

PR #2375, opened 2018-11-09
Checks and fixups language reworked

comradekingu:
BBCode, as per https://en.wikipedia.org/wiki/BBCode

### DIFF
--- a/docs/user/checks.rst
+++ b/docs/user/checks.rst
@@ -348,11 +348,11 @@ Java MessageFormat
 Missing plurals
 ~~~~~~~~~~~~~~~
 
-Checks that plural forms are present in translation.
-Check plural form definition to see the specifics of how each plural form is being used.
+Checks that all plural forms of a source string have been translated.
+Specifics on how each plural form is used can be found in the string definition.
 
 Failing to fill in plural forms will in some cases lead to displaying nothing when
-the plural is in use.
+the plural tense is in use.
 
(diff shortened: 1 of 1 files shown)


docs/user/checks.rst
```diff
@@ -372,59 +348,57 @@ Java MessageFormat
 Missing plurals
 ~~~~~~~~~~~~~~~
 
-Some plural forms are not translated. Check plural form definition to see for
-which counts each plural form is being used.
+Checks that plural forms are present in translation.
+Check plural form definition to see the specifics of how each plural form is being used.
```
comradekingu: Not 100% sure what is going on here.

nijel: This triggers in case the language has more plural forms, but the translation for some of them is empty.

comradekingu: Thank you :)

comradekingu: Is it possible to run the "Ellipsis" check on translated strings? It would make sense to move it up to the translation section, next to the "Trailing dots" check, which would also reduce some of the explanation, as it is shared between the two.

nijel: The ellipsis check is for source strings - when souce uses three dots where ellipsis would be probably more appropriate.

nijel:
Merged, thanks for your contribution! I had to fix markup a bit in 41ecc45f9d.

Issue #2380, opened 2018-11-12
When saving a translation I get Translation text too long! as error.


goetas:
**Describe the bug**
When saving a translation I get `Translation text too long!` as error.

**To Reproduce**
Try to translate a string. In our case 256 chrs.

**Expected behavior**
The string to be saved as usual

**Screenshots**
![screenshot from 2018-11-12 15-12-46](https://user-images.githubusercontent.com/776743/48352623-77d49d80-e68d-11e8-9f3a-9a48814e746e.png)

The issue seems to be triggered by https://github.com/WeblateOrg/weblate/blob/9e02f3c4f8ee9f70f6681e5a16e29735038d1b6d/weblate/trans/forms.py#L526

For some reason the filed is detected as max 160 chr, the textarea, in the html has 

```html
<textarea name="target_0" maxlength="160" id="id_d635adb6e6754d11_0" dir="ltr" lang="en" tabindex="100" cols="40" rows="3" class="translation-editor form-control" style="overflow: hidden; overflow-wrap: break-word; resize: horizontal; height: 54px;">...</textarea>
```

If looks that `get_max_length` return 160 in 
https://github.com/WeblateOrg/weblate/blob/9e02f3c4f8ee9f70f6681e5a16e29735038d1b6d/weblate/trans/models/unit.py#L1037

In my "Quality checks flags" I have:
```
ignore-end-question,ignore-end-exclamation,ignore-end-stop
```

Versions: 

Weblate	3.2.2
Python	3.5.3
Django	2.1.2
Celery	4.2.1
celery-batches	0.2
six	1.10.0
social-auth-core	1.7.0
social-auth-app-django	2.1.0
django-appconf	1.0.2
translate-toolkit	2.3.1
Whoosh	2.7.4
defusedxml	0.5.0
Git	2.11.0
Pillow	4.0.0
python-dateutil	2.5.3
lxml	3.7.1
django-crispy-forms	1.7.2
django_compressor	2.2
djangorestframework	3.9.0
user-agents	1.1.0
jellyfish	0.6.1
pytz	2018.7
pyuca	1.2
PyYAML	3.12
tesserocr	2.3.1
Mercurial	4.0
git-svn	2.11.0

nijel:
The problem is that maximal length is based on source string. And your source string doesn't contain actual source. You probably should use monolingual setup for your translations - the translation only contains keys and not source strings.

[...]

Issue #2240, opened 2018-09-14
"Needs editing" change to "Needs rewriting" or "Needs checking"


kyotaro-iijima:
Reason :
"neeeds editing" remembers that it was automatically flagged when a new translation was added. I think there is a demand to flag translations that need checking later. Also, it is confusing to be a record that wrote a new translation even though only flags are attached.

Solution :
Change the name of "Needs editing" to "Needs rewriting" or "Needs checking".
Just select the check box and it will be recorded.
This is enough.

Best Regards
Kyotaro

nijel:
It was changed to "needs editing" about year ago, see https://github.com/WeblateOrg/weblate/issues/1314 and https://github.com/WeblateOrg/weblate/issues/1616#issuecomment-326946263, I'd really like to come up with more acceptable name, but I'm not sure these two are actually better.

Maybe @comradekingu could share his opinion as well?

comradekingu:
I think the problem is at its core one term relating to many potential states

What is now _Needs work_ could either be
"upstream fuzzy", like XTM meaning the **string changed and is now possibly outdated**
or
"Translated tentatively", meaning the **translator isn't 100% sure the translation**
marking something like this brings into question other strings of similar nature, but that is another matter. "Marked for review" is what that essentially is, and it should show up first.

(I really like the translatewiki way of just getting many different projects in the same editor, it also has a "review-mode", where the same is done for checking things. But that is also another matter.)

Another state a string can effectively be in is "source error", where an improvement is likely to be found. It would help upstream devs, or people that try to help out, or change upstream strings to know what these are. Right now many obvious source errors just get translated correctly. Another matter too, and it is all about speed. I think a checkbox would be fine, and then one could go back and add a reason later. Making it "flagged as source error, but not explained".

I think the "another translation exists" should be system wide, and that one project should just be the various different strings, by complexity, sorted by frequency.

There the potential string variants are "auto-translated and no possible other way", (Like " or "pick one of many ways to translate this". That would auto-populate many easy Android projects.

With the way English works, sometimes one word has a duality, like "hip" can either be a body-part or something fashionable. In this case the potential states are "Duality undecided", where an admin can pick which variant it is (and mark it with an (auto)comment to the string, or effectively "Duality fixed", where the autotranslation kicks in.

[...]

PR #2384, opened 2018-11-13
Add basic support for XLIFF states

nijel:
We do not use translate-toolkit here, but rather parse it on our own as
translate-toolkit has a bit strange logic of handling this.

The following states are considered as fuzzy: new, needs-translation,
needs-adaptation, needs-l10n

Issue #2380
Issue #2372

Signed-off-by: Michal Čihař <michal@cihar.com>


nijel:
@PowerKiKi As you have played a bit with XLIFF states, I'd like to hear your input on this.

PowerKiKi:
Thanks for reaching out to me, and sorry about my late reply.

I think the documentation should be made more clear by listing exactly which `state` values are read or not, and which `state` values are written. And also how it combines (or not) with `approved` attribute.

I was wondering if a huge table (or several tables?) to describe those read/write mapping would be a nice thing to have for **all** formats. Hopefully it would make it easier to understand what happen between formats for end-users. Maybe something similar to https://github.com/WeblateOrg/weblate/pull/2372#issuecomment-436157111

Also, this is out of scope, but I feel the mixed terms in Weblate code and Weblate GUI are confusing. I think the worst is `STATE_TRANSLATED` that is declared as `_('Translated'))`, but it typically shows up as `"Waiting for review"` in the GUI. Also the obvious "fuzzy" vs `"Need editing"`. If we could standardize the terms used in Weblate code and Weblate GUI (and Weblate docs), it would probably make it easier to understand how to use the product, and to contribute to it.

Otherwise the code itself of this PR looks good to me, but I'd try to improve doc before merging  :)

nijel:
Well most of formats have no states at all so there is nothing to document for them :-). But it could be covered in the features table to make it clear.

The different naming comes from workflow settings (STATE_TRANSLATED is called Translated if review workflow is not used) and history (originally there was Fuzzy state in the UI as well, but the name was confusing, but it really didn't help, see https://github.com/WeblateOrg/weblate/issues/2240).

nijel:
I've extended the features table, that should make it easier to understand what is supported where. Also the states are listed in the xliff docs now.

PowerKiKi:
LGTM !

Issue #2393, opened 2018-11-17
Weblate can leak git passwords if push repo is misconfigured


Mataniko:
**Describe the bug**
When attempting to Push to a misconfigured Git repo, weblate will return the entire Push Url. However, this can leak the repository password if using a token/password.

**To Reproduce**
1. Go to 'Component Settings > Version Control'
2. Change the repo name to a nonexistent repo
3. Save changes
4. Go to 'Repository maintenance'
5. Click on Push

**Expected behavior**
A push error saying Repository not found. 

**Screenshots**
![image](https://user-images.githubusercontent.com/803371/48666439-2282fc80-ea90-11e8-8843-0fed64ff3476.png)
Note the token is revoked. Notice the actual org name is `scummvm` and not `scummvm-web`

**Server configuration and status**
```(python) www-data@11468-01:/opt/weblate/weblate$ ./manage.py list_versions
 * Weblate weblate-3.2.2
 * Python 2.7.13
 * Django 1.11
 * Celery 4.2.1
 * celery-batches 0.2
 * six 1.11.0
 * social-auth-core 1.7.0
 * social-auth-app-django 2.1.0
 * django-appconf 1.0.2
 * translate-toolkit 2.3.1
 * Whoosh 2.7.4
 * defusedxml 0.5.0
 * Git 2.11.0
 * Pillow 5.3.0
 * python-dateutil 2.7.3
 * lxml 4.2.5
 * django-crispy-forms 1.7.2
 * django_compressor 2.2
 * djangorestframework 3.9.0
 * user-agents 1.1.0
 * jellyfish 0.6.1
 * pytz 2018.5
 * PyYAML 3.13
 * Database backends: django.db.backends.mysql
 * Cache backends: default:RedisCache, avatar:FileBasedCache
 * Platform: Linux 4.9.0-6-amd64 (x86_64)
(python) www-data@11468-01:/opt/weblate/weblate$ ./manage.py check --deploy
System check identified some issues:

WARNINGS:
?: (security.W004) You have not set a value for the SECURE_HSTS_SECONDS setting. If your entire site is served only over SSL, you may want to consider setting a value and enabling HTTP Strict Transport Security. Be sure to read the documentation first; enabling HSTS carelessly can cause serious, irreversible problems.

System check identified 1 issue (0 silenced).
```

**Additional context**
HIDE_REPO_CREDENTIALS is set to True

nijel:
[...]

PR #2395, opened 2018-11-19
HIDE_REPO_CREDENTIALS in settings_example.py

Mataniko:
Ideally this should be True by default, but as this should be part of any sane config I think it's fit to have it here for now.

Related to #2393

### DIFF
--- a/weblate/settings_example.py
+++ b/weblate/settings_example.py
@@ -617,6 +617,9 @@ SECURE_HSTS_SECONDS = 0
 SECURE_HSTS_PRELOAD = False
 SECURE_HSTS_INCLUDE_SUBDOMAINS = False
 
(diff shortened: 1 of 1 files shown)


nijel:
It's already enabled by default:

https://github.com/WeblateOrg/weblate/blob/6f5922eb5b87dec989c10a754e17ae2734a262ad/weblate/trans/models/conf.py#L154-L155

I don't think most people will want to change this, that's why it's not part of example configuration.

Mataniko:
Thanks for the clarification @nijel and pushing a fix. Maybe the best solution is to update docs to clarify that it is in fact enabled.

nijel:
Good point, done in 8d2e2e23e3

PR #2428, opened 2018-12-03
Language Reworked

comradekingu:
Is "Units" really the word for string count?

How/is rendering done per language?


nijel:

nijel:
"strings" should be used instead of "units", I will change this.

PR #2431, opened 2018-12-04
Language reworked

comradekingu:
It is possible a program/library name or two didn't get corrected, I didn't check every single one in the first list.

### DIFF
--- a/docs/admin/install.rst
+++ b/docs/admin/install.rst
@@ -8,25 +8,25 @@ Looking for quick installation instructions? See :doc:`admin/quick`.
 Hardware requirements
 ---------------------
 
-Weblate should run on any contemporary hardware without problems, the following is
-the minimal configuration required to run Weblate on single host (Weblate, database
+Weblate should run on all contemporary hardware without problems, the following is
+the minimal configuration required to run Weblate on a single host (Weblate, database
 and web server):
 
-* 2 GB of RAM memory
+* 2 GB of RAM
 * 2 CPU cores
 * 1 GB of storage space
 
-The more memory you have, the better - it will be used for caching on all
-levels (file system, database and Weblate).
+The more memory you have, the better - it is used for caching on all
+levels (filesystem, database and Weblate).
 
-If you are going to have many concurrent users, you will need more CPU cores.
-For hundreds of translation components at least 4 GB of memory is
+Many concurrent users increases the amount of needed CPU cores.
+For hundreds of translation components at least 4 GB of RAM is
 recommended.
 
 .. note::
 
-    The actual requirements for your installation heavily vary based on the size of
-    translations managed by Weblate.
+    Actual requirements for your installation of Weblate vary heavily based on the size of
+    the translations managed in it.
 
 .. _requirements:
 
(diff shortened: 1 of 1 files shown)


docs/admin/install.rst
```diff
@@ -996,20 +996,20 @@ For production site, you want something like:
 Correctly configure HTTPS
 +++++++++++++++++++++++++
 
-It is strongly recommended to run Weblate using encrypted HTTPS protocol. After
-enabling you shoud sed :setting:`ENABLE_HTTPS` settings which also adjusts
+It is strongly recommended to run Weblate using the encrypted HTTPS protocol.
+After enabling it you should sed :setting:`ENABLE_HTTPS` settings, which also adjusts
```
comradekingu:
```suggestion
After enabling it, you should set :setting:`ENABLE_HTTPS` in the settings, which also adjusts
```
Or the tool "sed"?

docs/admin/install.rst
```diff
@@ -649,7 +649,7 @@ The :file:`settings.py` snippet for MySQL:
             'OPTIONS': {
                 # In case of older MySQL server which has default MariaDB
```
comradekingu:
```suggestion
                # In case of using an older MySQL server, which has MariaDB by default
```
Does it though?

docs/admin/install.rst
```diff
@@ -606,17 +606,17 @@ configuration file (usually :file:`/etc/mysql/my.cnf` on Linux):
...
 
-    # Grant all privileges to  weblate user
+    # Grant all privileges to the user ``weblate``
     GRANT ALL PRIVILEGES ON weblate.* TO 'weblate'@'localhost'  IDENTIFIED BY 'password';
 
-    # Create database on MySQL >= 5.7.7
+    # Create a database on MySQL >= 5.7.7
     CREATE DATABASE weblate CHARACTER SET utf8mb4;
 
     # Use utf8 for older versions
```
comradekingu:
```suggestion
    # Use utf-8 for older versions
```
Not sure here.

docs/admin/install.rst
```diff
@@ -243,19 +243,19 @@ To get latest sources using Git use:
...
++++++++++++++++++++++++++++
 
-If you decide to install Weblate using pip installer, you will notice some
-differences. Most importantly the command line interface is installed  to the
+If you decide to install Weblate using the pip installer, you will notice some
+differences. Most importantly the command line interface is installed to the
 system path as :command:`weblate` instead of :command:`./manage.py` as used in
 this documentation. Also when invoking this command, you will have to specify
-settings, either by environment variable `DJANGO_SETTINGS_MODULE` on the
+settings, either by the environment variable `DJANGO_SETTINGS_MODULE` on the
```
comradekingu: There is no "or" following the "either" here.

docs/admin/install.rst
```diff
@@ -215,26 +215,26 @@ system Python libraries.
         # Optional deps
         pip install pytz python-bidi PyYAML pyuca
 
-5. Create your settings (in our example it would be in
+5. Create your settings (in this example it would be in
    :file:`~/weblate-env/lib/python2.7/site-packages/weblate/settings.py`
-   based on the :file:`settings_example.py` in same directory).
+   based on the :file:`settings_example.py` in the same directory).
 6. You can now run Weblate commands using :command:`weblate` command, see
```
comradekingu: Something doesn't work using this in the text.

docs/admin/install.rst
nijel: It should be indeed "set"

docs/admin/install.rst
nijel: The original comment is wrong, it should be talking about MyISAM instead of MariaDB. Fixed in 418673d62a

docs/admin/install.rst
nijel:
```suggestion
settings by the environment variable `DJANGO_SETTINGS_MODULE` on the
```

There used to be more ways, but right now this one is only recommended one.

nijel:

PR #2437, opened 2018-12-05
Use 'fetch' instead of 'pull' in documentation

Kingdutch:
When talking about git the action 'pull' is essentially a fetch + merge. This has tripped me up when reading the documentation and was only made clear to me when reading related issues.

Weblate actually only does a nightly 'fetch' (updating the repository information) and requires additional config to do a merge as well.


nijel:

PR #2445, opened 2018-12-08
RFC: Use singular/plural for languages with just one plural

nijel:
Currenly Weblate consistetnly names plurals based on CLDR naming, so
depending on how many plurals language has each plural form will get one
of following labels: Zero, One, Two, Few, Many, Other

While this works fine for the more complex languages, showing just
Singular/Plural for simple languages with just one plural might make it
easier to understand.

Current displaying:

screenshot_2018-12-08 hello gettext - italian

This PR changes it to:

screenshot_2018-12-08 hello gettext - italian 1

### DIFF
--- a/weblate/lang/data.py
+++ b/weblate/lang/data.py
@@ -300,8 +300,8 @@ PLURAL_MAPPINGS = (
 PLURAL_NAMES = {
     PLURAL_NONE: ('',),
     PLURAL_ONE_OTHER: (
-        pgettext_lazy('Plural form description', 'One'),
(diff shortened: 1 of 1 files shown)


nijel:
I'd really like to hear input on this, for example from @comradekingu ;-).

comradekingu:
So I got what this is about now.
How about "Singular (1)" and "Plural (2+)"?

nijel:
The example of counts is shown in the popup over the question mark

PR #2465, opened 2018-12-20
add --recurse-submodules to git clone on install

alcalyn:
Actually, using only `git clone https://github.com/WeblateOrg/docker-compose.git weblate-docker` does not clone the submodule `docker-compose`.

We should add `--recurse-submodules` to avoid to have an empty `docker-compose` folder.

### DIFF
--- a/docs/admin/deployments.rst
+++ b/docs/admin/deployments.rst
@@ -26,7 +26,7 @@ The following examples assume you have a working Docker environment, with
 
    .. code-block:: sh
 
-        git clone https://github.com/WeblateOrg/docker-compose.git weblate-docker
+        git clone --recurse-submodules https://github.com/WeblateOrg/docker-compose.git weblate-docker
         cd weblate-docker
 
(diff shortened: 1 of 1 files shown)


docs/admin/deployments.rst
```diff
@@ -26,7 +26,7 @@ The following examples assume you have a working Docker environment, with
 
    .. code-block:: sh
 
-        git clone https://github.com/WeblateOrg/docker-compose.git weblate-docker
+        git clone --recurse-submodules https://github.com/WeblateOrg/docker-compose.git weblate-docker
```
nijel: https://github.com/WeblateOrg/docker-compose is the docker-compose module, you probably mean https://github.com/WeblateOrg/docker, which is on the other side (intentionally) not used in the install instructions.

alcalyn: Oh yes, totally !
I followed an old link and didn't seen that we now have to clone docker-compose.
Sorry, thank you !

PR #2469, opened 2018-12-22
Improve commit log in addons for linked components

SlavekB:
Commit log from addons is not sufficiently descriptive for projects with linked components. For example, after updating POT files, I get 50 individual commits that only contain in log: "Update translation files".

The attach patch adds the name of the component and, if it is a component linked from another project, also the name of the project. This makes commit log more useful – for example: "Update translation files tdegraphics / tdeiconedit".

### DIFF
--- a/weblate/addons/base.py
+++ b/weblate/addons/base.py
@@ -25,6 +25,8 @@ from django.utils.functional import cached_property
 
 from weblate.addons.events import EVENT_POST_UPDATE, EVENT_STORE_POST_LOAD
 from weblate.addons.forms import BaseAddonForm
+from weblate.utils.render import render_template
+from weblate.utils.site import get_site_url
 
 
 class BaseAddon(object):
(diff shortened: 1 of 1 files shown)


weblate/addons/base.py
```diff
@@ -148,20 +148,28 @@ class UpdateBaseAddon(BaseAddon):
     It hooks to post update and commits all changed translations.
     """
     events = (EVENT_POST_UPDATE, )
-    message = '''Update translation files
+    message = '''Update translation files{component}
 
-Updated by {name} hook in Weblate.'''
+Updated by "{name}" hook in Weblate.'''
```
nijel: Maybe it would be better to use same rendering engine and style as other Weblate commit messages?

```
Update translation files

Updated by "{{ hook_name }}" hook in Weblate.

Translation: {{ project_name }}/{{ component_name }}
Translate-URL: {{ url }}
```

weblate/addons/base.py
```diff
@@ -148,20 +148,28 @@ class UpdateBaseAddon(BaseAddon):
...
         repository = component.repository
         with repository.lock:
             if repository.needs_commit():
                 files = [t.filename for t in component.translation_set.all()]
                 repository.commit(
-                    self.message.format(name=self.verbose),
+                    self.message.format(
+                        name=self.verbose,
+                        component=component_description
+                    ),
```
nijel: This would then become `weblate.utils.render.render_template(self.message, component=component, hook_name=name)`...

nijel:

Issue #2476, opened 2019-01-07
Openshift3 deployment does not fully work


alesrosina:
For Weblate to fully work on openshift, there are still two things that needs to be fixed:
- [x] add redis cache
- [x] fix `Failed to generate key: b'No user exists for uid 1000140000\r\n'`error (*)

(*) This error is due to openshift nature of handling users (so `git clone` command will fail) and it is documented [here](https://docs.openshift.com/enterprise/3.1/creating_images/guidelines.html) and we need to use [nss wrapper](https://cwrap.org/nss_wrapper.html). 

Currently Weblate can be deployed to OpenShift, but it is not usable due to second issue, so this needs to be resolved.

alesrosina:
@nijel you can assign this to me, will try to solve it as soon as possible.

nijel:
Thanks! The current OpenShift3 integration is not updated for recent changes and probably had some missing bits even at time of merging see https://github.com/WeblateOrg/weblate/issues/1523 (but you've already found that).

nijel:
Should be now fixed by https://github.com/WeblateOrg/weblate/pull/2477, right?

alesrosina:
Yes, correct. Thanks!

Issue #1523, opened 2017-06-17
Add support for OpenShift 3


maxruby:
### Steps to reproduce
1. Set up new `python 2.7` project in OpenShift following docs
2. Allocate 1 GB in OpenShift
3. Build from source `https://github.com/WeblateOrg/weblate.git`

### Actual behaviour
Build  'lxml.etree' extension **failed**. See https://gist.github.com/maxruby/6ce93a1a1a4b94927102d9ecbc66a684

### Expected behaviour
Build should not fail.

### Server configuration
Please paste the output of command `./manage.py list_versions` over here

maxruby:
@nijel.  Thanks for the efforts towards fixing this. However, it still fails at the same place when using the latest commit.

Please see new [gist](a55260a24b68e0485ce3317363b7bf10) documenting the error.

Can you please leave this issue open until it is actually fixed?  Thanks again for the amazing work on this project.

nijel:
It seems that the installation instructions are not used, probably that needs somebody experienced with openshift to fix rather than me :-).

maxruby:
Thanks for the clarification. The trouble is I also tried the OpenSUSE appliance installation with VirtualBox with no success :(

Can you kindly point me to the most reliable way of getting weblate installed on Linux running on a VM? I would like to run it on my MacOSX with VirtualBox.

maxruby:
According to a couple posts in [_StackOverflow_](https://stackoverflow.com/questions/35529761/openshift-cant-install-lxml-for-python-app), a possible reason for the _lxml_ build failure might be that `OpenShift` does not support _lxml_ version > `v3.2.5`.

Currently, the `OpenShift` build log for _weblate_ shows the following:
```
Building lxml version 3.8.0. 
``` 

The post in StackOverflow say that:
```
Openshift is only able to compile/install 'lxml==3.2.5' the last time I checked.
Set your 'requires' in your setup.py to 'lxml==3.2.5'.
```

Can you please change this _lxml_ requirement so that `OpenShift` can successfully build _lxml_?

nijel:
If you want to avoid manual installation using pip, the best way for deployment is Docker as it's automatically tested for every release and updated frequently. Anything else is behind: I haven't found time to maintain the openSUSE appliance for some releases, the OpenShift was contributed some time ago and hasn't received any updates recently.

[...]

Issue #2267, opened 2018-09-22
Windows compatibility and tests


nijel:
Weblate should work pretty much fine on Windows. It would be great to have this covered by tests. There are currently tests running on AppVeyor, but there are some test failures which should be addressed. These might be both problems with AppVeyor setup or code issue. Best way to figure out is probably to reproduce the issues in local Windows build.

Besides others, there is some problem with parsing po files:

```
weblate.formats.ParseError: Invalid file format <translate.storage.pypo.pofile object at 0x081AC110>
```

See https://ci.appveyor.com/project/nijel/weblate for current build status.

Naruto0:
Hi @nijel 

It seems that windows cannot handle symbolic link, which in this case is `xx.po`. I am in the middle of solving another issue, which is dependent on this.
Do you have any idea, how to resolve this?

nijel:
We might just skip this particular test, it doesn't make sense to test behavior the platform does not support.

nijel:
One important thing to note here is that Celery no longer officially supports Windows, see https://docs.celeryproject.org/en/stable/faq.html#windows. We should probably make similar statement.

nijel:
I don't see enough interest in making Weblate work on Windows. Patches for it are always welcome, but I'm not going to invest any time into that besides reviewing pull requests. This is properly documented since 399dd5c.

PR #2477, opened 2019-01-07
Openshift3 deployment fixes

alesrosina:
There where some leftovers from @GrahamDumpleton 's fork, so I started with some small fixes, since in [last openshift issue](https://github.com/WeblateOrg/weblate/issues/1523) he mentioned he is finished with this. 

Otherwise this MR is a start of issue https://github.com/WeblateOrg/weblate/issues/2476

### DIFF
--- a/weblate/trans/util.py
+++ b/weblate/trans/util.py
@@ -171,7 +171,18 @@ def get_clean_env(extra=None):
     }
     if extra is not None:
         environ.update(extra)
-    variables = ('PATH', 'LD_LIBRARY_PATH', 'SystemRoot')
+    variables = (
+        'PATH',
+        'LD_LIBRARY_PATH',
+        'LD_PRELOAD',
+        'SystemRoot',
+        # below two are nedded for openshift3 deployment,
+        # where nss_wrapper is used
+        # more on the topic on below link:
+        # https://docs.openshift.com/enterprise/3.2/creating_images/guidelines.html
+        'NSS_WRAPPER_GROUP',
+        'NSS_WRAPPER_PASSWD',
+      )
     for var in variables:
         if var in os.environ:
             environ[var] = os.environ[var]
--- a/openshift3/ephemeral-sqlite.json
+++ b/openshift3/ephemeral-sqlite.json
@@ -21,7 +21,7 @@
         {
             "name": "WEBLATE_REPOSITORY_URL",
             "description": "The URL of the Weblate Git repository.",
-            "value": "https://github.com/GrahamDumpleton/weblate",
+            "value": "https://github.com/WeblateOrg/weblate",
             "required": true
         },
         {
--- a/openshift3/persistent-mysql.json
+++ b/openshift3/persistent-mysql.json
@@ -21,7 +21,7 @@
         {
             "name": "WEBLATE_REPOSITORY_URL",
             "description": "The URL of the Weblate Git repository.",
-            "value": "https://github.com/GrahamDumpleton/weblate",
+            "value": "https://github.com/WeblateOrg/weblate",
             "required": true
         },
         {
--- a/openshift3/persistent-mysql.json
+++ b/openshift3/persistent-mysql.json
@@ -101,6 +101,12 @@
             "description": "The version of the Python builder.",
             "value": "3.5",
             "required": true
+        },
+        {
+          "name": "REDIS_VERSION",
+          "description": "The version of the Redis cache.",
+          "value": "4-alpine",
+          "required": true
         }
     ],
     "objects": [
@@ -259,6 +265,14 @@
                                     {
                                         "name": "OPENSHIFT_MYSQL_DB_PORT",
                                         "value": "3306"
+                                    },
+                                    {
+                                      "name": "REDIS_HOST",
+                                      "value": "${APPLICATION_NAME}-cache"
+                                    },
+                                    {
+                                      "name": "REDIS_PORT",
+                                      "value": "6379"
                                     }
                                 ],
                                 "volumeMounts": [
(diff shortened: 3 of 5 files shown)


GrahamDumpleton:
FWIW. There are better ways than relying on ``nss_wrapper`` now.

alesrosina:
@GrahamDumpleton do you have any better suggestion? I didn't find anything else and this is my first time playing with openshift so would appreciate your input.

GrahamDumpleton:
Until I can show a tailored example, have a dig through what Jupyter notebook image does.

* https://github.com/jupyter/docker-stacks/blob/master/base-notebook/Dockerfile#L47
* https://github.com/jupyter/docker-stacks/blob/master/base-notebook/start.sh#L112

Their case is slightly different as they don't make things group ``root``. Having group ``root`` is better option.

Steps which are different from using ``nss_wrapper`` are:

* Make ``/etc/passwd`` group writable.
* In startup, add user entry to ``/etc/passwd`` for user ID if it doesn't exist.
* Add ``wheel`` group if doesn't exist.
* Update PAM so that only users in group ``wheel`` can use ``su``.

The ``su`` stuff is needed to prevent someone adding a password for ``root`` to the ``/etc/passwd`` file and then using ``su``. That restricts it to a user who is in group ``wheel`` which wouldn't normally be the case. This is more for use outside of OpenShift, as OpenShift doesn't allow ``su`` anyway. But then I just realised another potential problem with this if image is used outside of OpenShift, which I will need to look into.

So right now ``nss_wrapper`` is still the safest.

alesrosina:
@GrahamDumpleton ok, so I tried cupple of things, but I am failing badly :) Do you have any idea hot to install it? Do I really need to create my own docker image, that already includes this? Otherwise I'm always bumping into permission error, since `yum install nss_wrapper`needs to run under sudo ...

I would like to avoid creating my own docker image, since that means one maintenance more for weblate.

GrahamDumpleton:
Now I am confused. I was assuming that since you were playing around with ``nss_wrapper`` that you were trying to fix some existing ``Dockerfile`` type build. If you are using an OpenShift Python S2I build, you don't need to fiddle with ``nss_wrapper`` since the base image already includes it and does what is required for the application. So right now I have no idea what problem you are trying to solve. Can you point to the specific issue which explains in detail what you current problem is. The only issue I can find doesn't include a sufficient explanation of how/where ``git clone`` is called.

alesrosina:
Yes, so I used your implementation for openshift, so I ran this command:
`oc new-app https://raw.githubusercontent.com/WeblateOrg/weblate/master/openshift3/persistent-postgresql.json`

Application is running, but there is a functionality for generating SSH key in weblate, which returns `Failed to generate key: b'No user exists for uid 1000140000\r\n'`error. As far as I understood, this is a "feature" of openshift, so you need to use `nss_wrapper` for that.

But as far as I can see, you are using standard `python:3.5` docker image or am I missing sth? Where is this `nss_wrapper` already included?

GrahamDumpleton:
Where in the weblate code is the SSH key gen command called? That is what I need to know. Was there a Python stack trace showing where in the code it was being done from?

alesrosina:
I am guessing it is happening here: https://github.com/WeblateOrg/weblate/blob/e78de11db7b2beefb169d876a2b2034ceec52e21/weblate/wladmin/views.py#L78

But maybe @nijel can explain a bit more into detail what is going on and how this ssh thing works.

alesrosina:
ok, actually is this function that gets called: 
https://github.com/WeblateOrg/weblate/blob/e78de11db7b2beefb169d876a2b2034ceec52e21/weblate/vcs/ssh.py#L190

So it will execute `ssh-keygen` on openshift. And yeah, if I run this in console on openshit, got exactly that error.

GrahamDumpleton:
The problem is this line:

* https://github.com/WeblateOrg/weblate/blob/39ae3e37edeb6890fae619a559164602da7d7f58/weblate/vcs/ssh.py#L123

Because it clears environment, it isn't passing through the ``LD_PRELOAD`` and ``NSS_*`` environment variables. This means that the ``ssh-keygen`` program will not inherit the fiddle required to use the alternate ``passwd`` file.

GrahamDumpleton:
The question is whether a change to:

* https://github.com/WeblateOrg/weblate/blob/39ae3e37edeb6890fae619a559164602da7d7f58/weblate/trans/util.py#L166

is going to be acceptable, to also pass ``LD_PRELOAD`` and the ``NSS_*`` environment variables (can't remember names of these), given that is an OpenShift specific issue.

GrahamDumpleton:
These two variables:

* https://github.com/sclorg/s2i-python-container/blob/master/2.7/root/opt/app-root/etc/generate_container_user#L14

and ``LD_PRELOAD``.

alesrosina:
@GrahamDumpleton yes, this works now. Thank you very much for this help, I would struggle much longer if you would not point me to this direction.

@nijel i don't see a connection with failing AppVeyor, maybe you can check this? And also check if you agree with change.

For adding cashing, I will add a new comit, so let's keep this MR in WIP state.

nijel:
You can ignore failing AppVeyor, we're not yet there on passing tests there, see https://github.com/WeblateOrg/weblate/issues/2267.

weblate/trans/util.py
```diff
@@ -171,7 +171,14 @@ def get_clean_env(extra=None):
...
         environ.update(extra)
-    variables = ('PATH', 'LD_LIBRARY_PATH', 'SystemRoot')
+    variables = (
+        'PATH',
+        'LD_LIBRARY_PATH',
+        'NSS_WRAPPER_GROUP',
+        'NSS_WRAPPER_PASSWD',
+        'LD_PRELOAD',
+        'SystemRoot',
+      )
```
nijel: The list probably grew up too much, so it might be good to document why each variable is needed.

alesrosina: OK, added a comment. Do you find it readable enough?

nijel:

PR #2501, opened 2019-01-18
Require login setting for openshift deployments

alesrosina:
When deploying to openshift, you cannot set `LOGIN_REQUIRED_URLS` via openshift environment variables, since it require an array with regex. 

So I've used same solution as it is in docker container - introduced one more config varable `WEBLATE_REQUIRE_LOGIN` with boolean possible value.


nijel:

PR #2512, opened 2019-01-22
Enable social auth also for openshift enviroment

alesrosina:
Added possibility of using other auth providers via configuration.


nijel:

PR #2513, opened 2019-01-22
Added VScode configuration to .gitignore

alesrosina:
There are already other editors config in gitignore, I've added also VScode.


nijel:

PR #2565, opened 2019-02-20
Added also Auth0 provider for social-auth-core

alesrosina:


weblate/utils/requirements.py
```diff
@@ -183,14 +183,14 @@ def get_versions():
         'social-auth-core',
         'https://python-social-auth.readthedocs.io/',
         'social_core',
-        '2.0.0',
+        '3.1.0',
```
alesrosina: @nijel I am not sure if this change is actually required?

alesrosina:
@nijel I was also considering adding an option, that user can add Auth0 image & text via environment variables, since Auth0 is a provider, where you can completely customize login to your needs. Do you think this would be a good idea?

nijel:
This pull request **introduces 1 alert** when merging 1a4bbabdf986c466a8d4ba3b8abc601f0bfb6133 into 295c827ce1a8332cd8b7b7cc54825a716c4f989a - [view on LGTM.com](https://lgtm.com/projects/g/WeblateOrg/weblate/rev/pr-6e03d99a9e885715ae376cb65b1dd095ae944cd0)

**new alerts:**

* 1 for Unused import

---

*Comment posted by [LGTM.com](https://lgtm.com)*

alesrosina:
Fixed.

weblate/settings_openshift.py
```diff
@@ -161,6 +170,26 @@ def get_env_list(name, default=None):
...
+if 'WEBLATE_SOCIAL_AUTH_AUTH0_AUTH_EXTRA_ARGUMENTS' in os.environ:
+    auth0_args = parse_qs(ast.literal_eval(
+      Template(
+        os.environ.get('WEBLATE_SOCIAL_AUTH_AUTH0_AUTH_EXTRA_ARGUMENTS', '""')
+      ).substitute(os.environ)
+    ))
+    auth0_args = dict(
+      [(i, auth0_args[i][0]) if auth0_args[i] else (i, '') for i in auth0_args]
+    )
+    SOCIAL_AUTH_AUTH0_AUTH_EXTRA_ARGUMENTS = auth0_args
```
nijel: I've just moved helper functions we use in Docker to Weblate, so you might better use that as well instead of using ast (though I'm not sure what kind of args this expects, `get_env_map` deals with strings only):

https://github.com/WeblateOrg/weblate/blob/e002f976000ecdea1d8e3893520aaef12d54a56e/weblate/utils/environment.py#L32-L41

requirements.txt
```diff
@@ -8,8 +8,8 @@ backports.csv; python_version < '3.0'
 Pillow
 six>=1.7.0
 python-dateutil
-social-auth-core>=2.0.0
-social-auth-app-django>=3.0.0
+social-auth-core[openidconnect]>=3.1.0
```
nijel: I don't mind increasing the required versions, but pulling in openidconnect dependencies unconditionally seems over - most people won't use that. At most that should be in requirements-optional.txt or not listed at all. We don't list additional dependencies for other optional auth backends as well.

nijel:
I have no experience with this service, so it's hard to tell for me what would be actually used. You probably know better than me whether these options would be useful ;-).

weblate/settings_openshift.py
alesrosina: Well, it expects querystring style of params, so sth like `arg1=value1&arg2=something%20else`.

But if you decided to do it in format `arg1:value,arg2:something else` - its also fine by me, will refactor it. It makes sense it is same all around weblate.

requirements.txt
alesrosina: Sure thing. I am a rookie in python dev, so I pretty sure there are better solutions :) So if we add it to `requirements-optional.txt`, how would I then trigger this in my build?

alesrosina:
OK, I will try to make some sense in also adding an option to add image url and text of a service via params.

requirements.txt
alesrosina: OK, forget about my question - I see it is already installed by default for openshift - which is my choice of deployment.

Will move it there.

nijel:
This pull request **introduces 3 alerts** when merging 0640ac09af0bb39321a89a9b4deb1af337978562 into e002f976000ecdea1d8e3893520aaef12d54a56e - [view on LGTM.com](https://lgtm.com/projects/g/WeblateOrg/weblate/rev/pr-40f6e919530abfc3aeb959c56b718539ab31a567)

**new alerts:**

* 3 for Unused import

---

*Comment posted by [LGTM.com](https://lgtm.com)*

alesrosina:
@nijel I have fixed all the issues and also introduced possibility to add custom image/title for Auth0 provider.

Please do check it one more time, since as I already said, I am rookie in python, so probably did some weird stuff again :)

weblate/settings_openshift.py
alesrosina: @nijel actuall all strings are required to be in `"some_string"`, while this one does not expect this - aka I need to set variable as `some_string`, so without **""**. 

It just doesn't seems to be consistent with `import_env_vars` which is used in `openshhiftlib.py`. Maybe also refactor that one, so all strings does not need to be wrapped in **""**?

requirements-optional.txt
```diff
@@ -12,3 +12,4 @@ tesserocr>=1.2; platform_system != 'Windows'
 akismet>=1.0
 boto3
 zeep
+social-auth-core[openidconnect]>=3.1.0
```
nijel: Thinking about it more, this makes it harder to maintain - it's yet another place where version needs to be adjusted. So probably it's better to not list this at all.

weblate/accounts/templatetags/authnames.py
```diff
@@ -48,6 +49,15 @@
     'facebook': {'name': 'Facebook', 'fa_icon': 'facebook'},
     'twitter': {'name': 'Twitter', 'fa_icon': 'twitter'},
     'stackoverflow': {'name': 'Stack Overflow', 'fa_icon': 'stackoverflow'},
+    'auth0': {
+      'name': getattr(settings, 'SOCIAL_AUTH_AUTH0_TITLE', 'Auth0'),
+      'image': getattr(
+          settings,
+          'SOCIAL_AUTH_AUTH0_IMAGE',
+          'btn_auth0_badge.png'
+          ),
```
nijel: Please set default values using appconf and you can then use just `settings.SOCIAL_AUTH_AUTH0_IMAGE` here.

https://github.com/WeblateOrg/weblate/blob/54fa2346bb41cffdbc5173439acba93fd15b0c70/weblate/accounts/models.py#L566-L588

nijel:
Thanks, it now looks almost ready to merge :-). Please address two additional issues I've posted, rebase on current master (preferably squashing all Auth0 related commits to one) and I think we're ready to merge.

requirements-optional.txt
alesrosina: OK, but how can a user add this then? And how does he/she knows, if using a provider from social_auth that uses this functionality (currently OpenID and Auth0).

Is it really that bad to include also this in default requirements?

If yes, should I add _another_ requirements file, like `requirements-openidconnect.txt` and request it only on openshift implementation?

weblate/accounts/templatetags/authnames.py
alesrosina: OK, looks much nicer, thanks for this hint. 

Fixed.

alesrosina:
@nijel thanks for the review! One more question beside the one from above regarding `requirements.txt` - codeclimate is complaining about a too complex method ... maybe you can also help me refactor that one?

requirements-optional.txt
nijel: For Azure it's up to user as well. There might be other social-auth backends with additional dependencies, we really can't depend on everything user might potentially set up here.

Note that I'm talking about dependencies in the code, it might make sense to include these in Docker or OpenShift integrations though.

nijel:
Well codeclimate is sometimes too strict here, you can ignore this one.

.s2i/action_hooks/build
```diff
@@ -38,6 +38,11 @@ while read line; do
     esac
 done < requirements-optional.txt
 
+# For openshift deployment we will also install all requirements for
+# social-auth-core package, otherwise OpenID providers would not work
+
+pip install 'social-auth-core[openidconnect]>=3.1.0'
```
alesrosina: @nijel so i removed this from `requirements.txt` and added it directly here. Would that work for you?

alesrosina:
@nijel hopefully cleared all your concerns. Ready to merge?

requirements-optional.txt
```diff
@@ -12,3 +12,4 @@ tesserocr>=1.2; platform_system != 'Windows'
 akismet>=1.0
 boto3
 zeep
+social-auth-core>=3.1.0
```
nijel: Not needed at all here, it's alredy in `requirements.txt`.

alesrosina: My bad. Removed.

nijel:

PR #2569, opened 2019-02-22
Disable text selection on badges

jlowcs:
Goal is to avoid translator copying the '1' and '2' indicators from the source string when using text selection instead of 'Copy' button

I don't think this warrants a test case...


nijel:
Thanks, I was not aware of this possibility in CSS.

PR #2570, opened 2019-02-22
Improve audit log messages for authenticating

nijel:
The current form of method:name can be confusing (eg. Password:user), so
better describe it with words.

Signed-off-by: Michal Čihař <michal@cihar.com>

### DIFF
--- a/weblate/accounts/models.py
+++ b/weblate/accounts/models.py
@@ -60,28 +60,28 @@ ACCOUNT_ACTIVITY = {
         'Password reset has been confirmed and password has been disabled.'
     ),
     'auth-connect': _(
-        'Authentication ({method}:{name}) has been added.'
+        'You can now log in using {method} ({name}).'
     ),
     'auth-disconnect': _(
-        'Authentication ({method}:{name}) has been removed.'
+        'You can no longer log in using {method} ({name}).'
     ),
     'login': _(
-        'Authenticated ({method}:{name}).'
+        'Logged on using {method} ({name}).'
     ),
     'login-new': _(
-        'Authenticated ({method}:{name}) from new device.'
+        'Logged on using {method} ({name}) from a new device.'
     ),
     'register': _(
         'Somebody has attempted to register with your email.'
     ),
     'connect': _(
-        'Somebody has attempted to add your email to existing account.'
+        'Somebody has attempted to register using your email address.'
     ),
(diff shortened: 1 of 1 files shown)


nijel:
@comradekingu I'd really like to hear your opinion here on changing audit log strings.

comradekingu:
@nijel "Could not log in as {name} with {method}"?  What are the methods?

And
"Somebody has attempted to register using your email address"
or (if I understand it correctly)
"Confirm registration on Weblate for this e-mail address."

nijel:
The `{method}` is name of authentication method. It's either "Password" or name of third party service (eg. "GitHub", "Google", "Facebook", ...). The `{name}` is identifier from this service, it doesn't have to be username or email, it can be as well numerical ID (eg. for GitHub or Facebook) or URL (for OpenID based providers such as Fedora).

comradekingu:
Should be "password" to fit it into a sentence.
Changed to "with {name}, but one case doesn't really cover all here.
I imagine it is an active error in some language.
Maybe do one case for each service?

"You can now log in with {name} using {method}"?
"You can no longer log in with {name} using {method}"
"Logged on with {name} using {method} from a new device"
"Could not log in with {name} using {method}"
"Somebody has attempted to register using your email address"
/
"Confirm registration on Weblate for this e-mail address."
"Account locked due to many failed logins."

nijel:
It doesn't matter whether it's password or Password, it won't fit nicely when localizing anyway (that's why I used the '({{method}}:{{user}})` form in the beginning because it doesn't need declining).

Maybe it's better to special case the password authentication completely and avoid to compose the sentence in this case. For third party services it will stay though...

nijel:
@comradekingu I've changed it based on your suggestions, but looking at it again it probably doesn't work well. The problem is that `user` is more an internal identifier of the third party service than something user would be aware of and this way it gives it too much of attention:

* You can now log in with http://nijel.id.fedoraproject.org/ using Fedora
* You can now log in with 148674 using GitLab
* You can now log in with 10205554029310662 using Facebook

I've changed it to give more attention to the service name:

* You can now log in using Fedora (http://nijel.id.fedoraproject.org/
* You can now log in using GitLab (148674)
* You can now log in using Facebook (10205554029310662)

Also I've separated messages for password authentication to avoid confusion in that case.

Again, feedback is welcome ;-).

PR #2571, opened 2019-02-22
Add a configuration option to make the translation length limit based on the source string optional

wol-soft:
Use case:

When translating language files which are key value based with short keys like "home.description" and a long translation Weblate currently limits the length of the translation to max(100, source.length * 10), in this example 160 chars. To avoid adding a manual max-length constraint to all keys tapping into this case the option LIMIT_TRANSLATION_LENGTH_BY_SOURCE_LENGTH can be set to false. If it's set to false the calculation will be skipped and instead translations up to 10.000 characters are allowed.

Signed-off-by: Enno Woortmann <enno.woortmann@web.de>

- [ x ] Every commit has message which describes it
- [ x ] Every commit is needed on it's own, if you have just minor fixes to previous commits, you can squash them
- [ x ] Any code changes come with testcase
- [ x ] Any new functionality is covered by user documentation

### DIFF
--- a/docs/admin/config.rst
+++ b/docs/admin/config.rst
@@ -571,6 +571,15 @@ URL where your Weblate instance shows it's legal documents. This is useful if
 you host your legal documents outside Weblate for embedding inside Weblate
 please see :ref:`legal`.
 
+.. setting:: LIMIT_TRANSLATION_LENGTH_BY_SOURCE_LENGTH
+
+LIMIT_TRANSLATION_LENGTH_BY_SOURCE_LENGTH
+-----------------------------------------
+
+By default the length of a given translation is limited to the length of the source string * 10 characters. Set this option to False to allow longer translations (up to 10.000 characters) irrespective of the source length.
+
+Defaults to True.
+
 .. setting:: LOGIN_REQUIRED_URLS
 
 LOGIN_REQUIRED_URLS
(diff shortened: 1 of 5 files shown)


nijel:
While I don't mind adding such configuration option, for your case you should rather setup the translation as monolingual to get proper source strings, see https://docs.weblate.org/en/latest/formats.html#bimono and https://docs.weblate.org/en/latest/admin/admin.html#monolingual

wol-soft:
Thanks for the answer,

wasn't aware that defining a monolingual base file changes the behaviour which string is used to calculate the max length. I've added my source language file there (and many associated information like associated context data, eg. screenshots, or translation approvals have been removed, seems like the project was initialized again. Maybe a warning should be added to the documentation to be aware of this when changing the settings).

Now the max length is calculated by the length of the source language translation and not by the length of the key which is ok for me since it provides enough buffer for translation. But I'm still irritated by the length calculation and the use cases to limit the length to source string length * 10 (seems to me like: There should be no "reachable" limit for the translation length but anyway, let's set a not obvious limit somewhere).

Maybe an option to switch it off would still be a good thing for people facing this limitation.

Now goin to restore the context data of my translations.

nijel:

The reason for this is to have some limit for the translation to avoid problems when translators put there very long string. Originally the limit was way stricter (AFAIR it was 2x length of the source string), but that really doesn't work for many languages...

wol-soft:
Ah ok, to not reach the real technical length limitation I guess.

Thanks for merge!

PR #2586, opened 2019-02-28
Spelling: New strings reworked, Filemask, announcement text, turn off

comradekingu:


weblate/trans/models/component.py
```diff
@@ -1280,32 +1280,32 @@ def clean_lang_codes(self, matches):
...
             if len(code) > 20:
                 raise ValidationError({'filemask': _(
-                    'Language code "%s" is too long, please check filemask!'
+                    'The language code "%s" is too long, please check the filemask.'
                 ) % code})
             if code in langs:
                 raise ValidationError(_(
-                    'There are more files for single language (%s), please '
-                    'adjust the mask and use components for translating '
+                    'There are more files for the single language (%s), please '
```
comradekingu: Not 100% sure here.

weblate/trans/validators.py
```diff
@@ -41,19 +41,19 @@
 
 
 def validate_filemask(val):
-    """Validate file mask that it contains *."""
+    """Validate that filemask contains *."""
     if '*' not in val:
         raise ValidationError(
-            _('File mask does not contain * as a language placeholder!')
+            _('Filemask does not contain * as a language placeholder!')
```
comradekingu:
```suggestion
            _('The filemask does not contain * as a language placeholder.')
```

weblate/trans/validators.py
```diff
@@ -41,19 +41,19 @@
...
         )
 
 
 def validate_autoaccept(val):
     """Validate correct value for autoaccept."""
     if val == 1:
         raise ValidationError(_(
-            'Value of 1 is not allowed for autoaccept as '
-            'every user gives vote to his suggestion.'
+            'A value of 1 is not allowed for autoaccept as '
```
comradekingu:
```suggestion
            'A value of \"1\" is not allowed for autoaccept as '
```

weblate/templates/addons/squash_help.html
```diff
@@ -4,5 +4,5 @@
 {% trans "All Weblate commits from all components within this repository will be squashed." %}
 </p>
 <p>
-{% trans "The resulting commits will include all commit messages and the commit authorship will not be preserved for all but per author squash. If you want to include authorship information it is recommended to customize the commit message to include this." %}
+{% trans "The resulting commits will include all commit messages, but the commit authorship will be the granted to the one squashing, and not be preserved for all. If you want to include authorship info, it is recommended to customize the commit message to include this." %}
```
nijel: This is not what it means :-) (what is my fault that the original sentence is hard to understand). There are currently four types of squashing (configurable just above this) and all of them with exception of per author do drop commit authorship information.

See screenshot:

screenshot_2019-03-03 f-droid f-droid

nijel:
Thanks, looks good with exception of the squash addon, see my comment above. 

To be merged after 3.5 release.

weblate/templates/addons/squash_help.html
comradekingu: My bad. My assumption now is that "per author" possibly makes a meta-git user per Weblate user, so that it is Git-specific (git-blame works), and that OTOH adding it to the commit log means it is just historic, unless Weblate has some support for picking it back out?

nijel:

PR #2640, opened 2019-03-18
Fixed issue with wrong permissions for id_rsa on openshift

alesrosina:
By default, openshift storage has `chmod 660` for all files, but for git and ssh this level of permission on  `id_rsa` file is to wide, so we need to change it after openshift setup script is run.


nijel:

PR #2642, opened 2019-03-18
Fix a typo in openshipft deployment

alesrosina:
Just found another bug in deployment script. Missing `]`.

### DIFF
--- a/.s2i/action_hooks/deploy
+++ b/.s2i/action_hooks/deploy
@@ -23,12 +23,12 @@ else
     powershift image setup
 fi
 
-echo "-----> Set chmod 600 for id_rsa keys."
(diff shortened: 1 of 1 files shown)


nijel:
Do you think there is way to automatically test OpenShift integration in our CI to avoid such problems?

alesrosina:
Hm, yeah, this is tricky, since it is actually bash code ... So maybe setup also some bash linter or sth?

For the previous bug, CI would not be able to detect it, unless you would actually run application on OpenShift and run feature tests there. But I guess this is not really realistic, right?

nijel:
We do this with Docker image, so it's not unrealistic, the question is how to do that with OpenShift, I have zero knowledge there...

PR #2647, opened 2019-03-19
Allow custom location of MO files

netniV:
Signed-off-by: netniV <netniv@hotmail.com>

Unfortunately the documentation mentioned nothing about the latter two requests so I haven't done those.  This functionality allows you to customise the path to the MO files which may not always be in the same folder as the PO files.

For example, cacti stores it's MO files in /locales/LC_MESSAGES but stores the PO/POT files in /locales/po/.  This change allows you to customise the output by either entering a full path or a relative path (eg, ../LC_MESSAGES)

### DIFF
--- a/weblate/addons/forms.py
+++ b/weblate/addons/forms.py
@@ -46,6 +46,27 @@ class BaseAddonForm(forms.Form):
         return self._addon.instance
 
 
+class GenerateMoForm(BaseAddonForm):
+    path = forms.CharField(
+        label=_('Path of generated MO file'),
+        required=False,
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(GenerateMoForm, self).__init__(*args, **kwargs)
+        self.helper = FormHelper(self)
+        self.helper.layout = Layout(
+            Field('path'),
+        )
+
+    def test_render(self, value):
+        translation = self._addon.instance.component.translation_set.all()[0]
+        validate_render(value, translation=translation)
+
+    def clean_path(self):
+        self.test_render(self.cleaned_data['path'])
+        return self.cleaned_data['path']
+
 class GenerateForm(BaseAddonForm):
     filename = forms.CharField(
         label=_('Name of generated file'),
--- a/weblate/addons/gettext.py
+++ b/weblate/addons/gettext.py
@@ -32,6 +32,7 @@ from translate.storage.po import pofile
 from weblate.addons.base import BaseAddon, UpdateBaseAddon, StoreBaseAddon
 from weblate.addons.events import EVENT_PRE_COMMIT, EVENT_POST_ADD
 from weblate.addons.forms import GettextCustomizeForm
+from weblate.addons.forms import GenerateMoForm
 from weblate.formats.exporters import MoExporter
 
 
@@ -61,11 +62,25 @@ class GenerateMoAddon(GettextBaseAddon):
     description = _(
         'Automatically generates MO file for every changed PO file.'
     )
+    settings_form = GenerateMoForm
 
     def pre_commit(self, translation, author):
         exporter = MoExporter(translation=translation)
         exporter.add_units(translation.unit_set.all())
         output = translation.get_filename()[:-2] + 'mo'
+
+        if self.instance.configuration['path']:
+             mo_name = os.path.basename(output)
+             if mo_name:
+                 mo_path = os.path.dirname(output)
+                 if not mo_path:
+                    mo_path = ''
+
+                 output = os.path.join(
+                     mo_path,
+                     self.instance.configuration['path'],
+                     mo_name
+                 )
         with open(output, 'wb') as handle:
             handle.write(exporter.serialize())
         translation.addon_commit_files.append(output)
--- a/weblate/addons/gettext.py
+++ b/weblate/addons/gettext.py
@@ -70,17 +70,18 @@ class GenerateMoAddon(GettextBaseAddon):
         output = translation.get_filename()[:-2] + 'mo'
 
         if self.instance.configuration['path']:
-             mo_name = os.path.basename(output)
-             if mo_name:
-                 mo_path = os.path.dirname(output)
-                 if not mo_path:
-                    mo_path = ''
+            mo_name = os.path.basename(output)
+            if mo_name:
+                mo_path = os.path.dirname(output)
+                if not mo_path:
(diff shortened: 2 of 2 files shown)


weblate/addons/forms.py
```diff
@@ -46,6 +46,27 @@ def save(self):
...
+        self.helper.layout = Layout(
+            Field('path'),
+        )
+
+    def test_render(self, value):
+        translation = self._addon.instance.component.translation_set.all()[0]
+        validate_render(value, translation=translation)
+
+    def clean_path(self):
+        self.test_render(self.cleaned_data['path'])
```
nijel: Is this really needed? You don't seem to use rendering in the addon code, so there is no need to validate the markup here as it won't get used.

weblate/addons/forms.py
```diff
@@ -46,6 +46,27 @@ def save(self):
         return self._addon.instance
 
 
+class GenerateMoForm(BaseAddonForm):
+    path = forms.CharField(
+        label=_('Path of generated MO file'),
```
nijel: Some help what is expected to be set here (and what blank value means) could be useful.

weblate/addons/gettext.py
```diff
@@ -61,11 +62,26 @@ class GenerateMoAddon(GettextBaseAddon):
...
         exporter = MoExporter(translation=translation)
         exporter.add_units(translation.unit_set.all())
         output = translation.get_filename()[:-2] + 'mo'
+
+        if self.instance.configuration['path']:
+            mo_name = os.path.basename(output)
+            if mo_name:
+                mo_path = os.path.dirname(output)
+                if not mo_path:
+                    mo_path = ''
```
nijel: The `get_filename` returns absolute path, so this should not happen and we want to crash in case dirname doesn't work.

weblate/addons/gettext.py
```diff
@@ -61,11 +62,26 @@ class GenerateMoAddon(GettextBaseAddon):
...
+
+                output = os.path.join(
+                    mo_path,
```
nijel: Maybe it's better to use `self.instance.component.full_path` as root instead of trying to parse it from the filename?

weblate/addons/forms.py
nijel: On the other side it should use `validate_filename` to avoid use of `..` in the path.

weblate/addons/gettext.py
netniV: Well, that's a good point, I did consider that but I ultimately decided that since the current location is the directory of the filemask (locales/po/*.po, for example), using the component root may lead to confusion.  On the flip side, for those who were not used to the current method of working may expect that it is the component root being used as a base.

Should we include an option set for "Relative to component" / "Relative to PO file" ?  That way, we can use either method.

weblate/addons/gettext.py
netniV: If you are sure it's a full path, then I will remove the mo_path check.  The reason I left it in was if it was a relative path and in the root of the component (ie, filemask was just *.po) then there would be no directory path and since the path.join expects strings, I was ensuring it was a string.

weblate/addons/forms.py
netniV: I shall see what I can set.

weblate/addons/forms.py
netniV: I'm not sure about those two parts if I'm honest.  I just took them from one of the other addons and tweaked.  I totally expected that there may be something i've overlooked or a test that needs to be implemented that isn't there because I'm still learning how you code the system 👍

weblate/addons/gettext.py
nijel: I'd just document in the help text that it's relative path in the repository root. That should make it clear. But you will probably want to use template rendering then to allow using language code in the path...

weblate/addons/forms.py
netniV: Oh and the .. is required if you are relative to the PO files.  I guess we should check that you are not trying to go outside of the component root's absolute path though?

netniV: How would that work with linked_repo's as I have noticed those mentioned in the code a few times.

weblate/addons/gettext.py
nijel: Yes, I'm sure, it's built here:

https://github.com/WeblateOrg/weblate/blob/98876cd6c770720b5e1106c3a053d7eceb962524/weblate/trans/models/translation.py#L200-L202

weblate/addons/gettext.py
netniV: The thing is, if we just change it to relative to the repository root, it will move the location of existing .mo files.  For example, weblate generated a load for us in /locales/po/<lang>.mo to match the <lang>.po location.  If we just move that suddenly to repository root, it will make them /<lang>.mo whilst the po files are in /locales/po/<lang>.po.  That would surely break existing setups?

weblate/addons/forms.py
nijel: You don't have to care about linked repos here, it's transparent on this level.

netniV:
So, just to be clear I will be performing the following updates to this on my next push:

1. Adding additional help text
2. Ensuring we have a path always
3. Changing validate_render to validate_filename
4. Adding a relative to option, defaulting to the base filename but optionally from the component root
5. Checking that the path generated is not outside the component root.
6. Adding a help page to document better the options and what they mean.

nijel:
First of all sorry for not structuring the feedback better, I was just going through the patch and fired comments as I spot problems, but generic feedback would be probably more appropriate.

I would really like to make it simple for now. What is your actual use case you need to cover?

The path should be validated using `validate_filename` to catch malicious paths. I think using only repository based paths is way to go, it's most flexible approach. This leads to need to use templating in the path to allow using language code and other variables in the path. You can look at generator addon for both generating and validation part (I've added it today as I've realized it is needed when reviewing this patch). In case the path empty it should still behave like now - just replace extension for the `.mo`. And the behavior should be explained in the help text in the form.

netniV:
OK, that does make things clearer, thanks!

I have no issue with forcing things that way and documenting it as such since it keeps current behaviour and makes things consistent going forward.

nijel:
Thanks, I've merged your changes and applied some fixed in 856349bf707d98e5e465b51029bf08dd9c3b6759.

Issue #2675, opened 2019-04-04
Incorrect version of sqlite is required


alesrosina:
When deploying to Openshift, I got an error:
```
ImproperlyConfigured: SQLite 3.8.3 or later is required (found 3.7.17).
```

This happens in a step 
```
python manage.py collectstatic --noinput
```
of `.s2i/action_hooks/build`.

I did not change any configuration on OpenShift, I just wanted to rebuild image, to get the latest version deployed - this means pulling from git latest from master.

See full error stack:

```
Handled exception ImproperlyConfigured: SQLite 3.8.3 or later is required (found 3.7.17).
Traceback (most recent call last):
  File "manage.py", line 33, in <module>
    main()
  File "/opt/app-root/src/weblate/runner.py", line 34, in main
    execute_from_command_line(argv)
  File "/opt/app-root/lib/python3.5/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/opt/app-root/lib/python3.5/site-packages/django/core/management/__init__.py", line 357, in execute
    django.setup()
  File "/opt/app-root/lib/python3.5/site-packages/django/__init__.py", line 24, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/opt/app-root/lib/python3.5/site-packages/django/apps/registry.py", line 114, in populate
    app_config.import_models()
  File "/opt/app-root/lib/python3.5/site-packages/django/apps/config.py", line 211, in import_models
    self.models_module = import_module(models_module_name)
  File "/opt/app-root/lib64/python3.5/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 986, in _gcd_import
  File "<frozen importlib._bootstrap>", line 969, in _find_and_load
  File "<frozen importlib._bootstrap>", line 958, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 673, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 662, in exec_module
  File "<frozen importlib._bootstrap>", line 222, in _call_with_frames_removed
  File "/opt/app-root/lib/python3.5/site-packages/django/contrib/auth/models.py", line 2, in <module>
    from django.contrib.auth.base_user import AbstractBaseUser, BaseUserManager
[...]

PR #2676, opened 2019-04-04
Fix openshift deployment to not trigger sqlite on build

alesrosina:
This fixes https://github.com/WeblateOrg/weblate/issues/2675, where on build sqlite was actually triggered, instead of psql/mysql. 
This bug affected only build stage, on deployment correct database was used.


alesrosina:
@nijel i've found the error, so I fixed it at source instead of fixing required django version.

nijel:

Issue #2677, opened 2019-04-05
OpenShift and management commands


alesrosina:
When Weblate is deployed on OpenShift none of management commands (https://docs.weblate.org/en/latest/admin/management.html) that use `git`, does not work. You would always get an error:

```
No user exists for uid XXXXXXXXX
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

This is solved in actual app, but it seems that management commands use some kind of different logic for accessing git and it always fails.

alesrosina:
@GrahamDumpleton I would again need your help here, I've spent a day trying to figure it out but got completely lost. I thought `openshift/python` image, which is used to build weblate image for deployment on openshift has already `nss_wrapper` implemented, but this somehow does not work. In general - if i run `ssh whatever.example.com`, I will always get this error, even if I run it via
```
#!/bin/sh
    exec ssh \
        -o "UserKnownHostsFile=/opt/app-root/data/ssh/known_hosts" \
        -o "IdentityFile=/opt/app-root/data/ssh/id_rsa" \
        -o StrictHostKeyChecking=yes \
        -o HashKnownHosts=no \
        "$@"
```
which is generated by Weblate in `/opt/app-root/data/ssh/ssh-weblate-wrapper-[some-md5-hash]`.

What am I missing here?

nijel:
Maybe you execute management command under different user than is actually running Weblate?

alesrosina:
I am not sure actually what is going on here :)

That's why I mentioned Graham, since he is quite an expert on Openshift and also did the initial setup of it. I am probably missing out sth quite obvious, but right now I just cannot get my head around it.

GrahamDumpleton:
Try using:

```
#!/bin/bash
```

for your script, not normal ``/bin/sh``.

The ``nss_wrapper`` stuff relies on a shell initialisation trick. I can't remember if it works for ``sh``, or only ``bash``.

GrahamDumpleton:
BTW, how exactly are you executing this script inside of the container. That is probably the more important question. Is it executed from the web interface, or are you using ``oc rsh`` to execute it from outside?

alesrosina:
Hey @GrahamDumpleton thanks for this hint.
Sadlly I did not manage to make it work, even with `#!/bin/bash`. 

This error appears on web interface, when I'm doing `oc rsh` it's even worse - I got this error:
```
[...]

PR #2678, opened 2019-04-08
OpenShift configuration issues

alesrosina:
As discussed here:
https://github.com/WeblateOrg/weblate/issues/2677#issuecomment-480742285


alesrosina:
@nijel this is still WIP, so please do not merge it yet.

alesrosina:
@GrahamDumpleton can you quickly review this, when you will have time?

alesrosina:
@nijel i think this should actually cover the issue mentioned above - so from my side it is ready to merge. Maybe we can wait for a few more days, if @GrahamDumpleton will have time to quickly review it.

openshift3/persistent-postgresql.json
```diff
@@ -27,7 +27,7 @@
         {
             "name": "WEBLATE_VOLUME_SIZE",
             "description": "Size of the persistent volume for Weblate.",
-            "value": "1Gi",
+            "value": "40Gi",
```
nijel: Are these changes intentional? It heavily increases storage requirements...

alesrosina: Yes. `1gi` is OK for initial setup, but if you work on a project, that is bigger then 1GB (which is sadly today in `node` world quite often), this will obviously not work.

And 40GB today is really not a lot.

What do you think?

nijel: I have no clue what impact this has on OpenShift - force example if you can override that somehow or whether it just makes running Weblate unconditionally expensive. My guess is that most users won't need that much of space.

alesrosina: Fair enough, will lower it to 10Gi.

openshift3/persistent-mysql.json
```diff
@@ -69,19 +69,19 @@
...
           "description": "Size of the persistent volume for Redis Cache.",
-          "value": "1Gi",
+          "value": "10Gi",
           "required": true
         },
         {
             "name": "DATABASE_VOLUME_SIZE",
             "description": "Size of the persistent volume for the database.",
-            "value": "1Gi",
+            "value": "40Gi",
```
alesrosina: @nijel since we are talking about disk space - what is your experience, what would be average disk usage on database level?

openshift3/persistent-postgresql.json
GrahamDumpleton: The whole point of it being a template parameter though is that people can override it when instantiating the template. In that sense, is better to set it to small value so test installs will work, with people who are doing production installs overriding it to what they believe they need. The alternative is to leave the field empty value, but with the field being required. This forces everyone to specify size as template parameter.

Setting allow expansion as default for PVC will also break for most people as not typical that OpenShift clusters would allow that. So deploy would presumably fail.

Key with the template is that it really should only serve as an example. For production use people should be copying it and customising it to their needs. So I would suggest leave it as 1Gi and not enable volume expansion.

Similar problem with memory. Now will not work for test installs.

My benchmark would be can it be installed in OpenShift Online starter tier.

GrahamDumpleton: The allow expansion could be a template parameter also with default of false.

openshift3/persistent-mysql.json
nijel: The whole https://hosted.weblate.org/ service consumes 8 GB in the database, that's why I think that 40 GB is way too much for default installation. Is this value easy to override when deploying? If yes, I'd stick with original small values, which would allow users to scale up, what is typically way easier than scaling down.

alesrosina: OK, lowered it to 10Gi.

openshift3/persistent-postgresql.json
alesrosina: @GrahamDumpleton OK, lowered it to 10Gi, i think this is reasonable. Also set `allowVolumeExpansion` to false.

Would that work?

nijel:

PR #2740, opened 2019-05-11
Spelling: msgid of translated messages → strings

comradekingu:
If not, what messages are translated?


nijel:
The string was taken from gettext man page, which uses messages instead of strings.

Issue #2808, opened 2019-06-10
Openshift and Celery


alesrosina:
I am trying to run Celery on Openshift without success. I am using default configuration as provided in https://github.com/WeblateOrg/weblate/blob/master/openshift3/persistent-postgresql.json

Things I have tried: 
- adding `./weblate/examples/celery start` to `.s2i/action_hooks/run` just before `exec mod_wsgi-express start-server $ARGS`. It all looks fine, no errors, but when execting `ps aux -A` in terminal of a running pod, no celery service is running.
- Running celery with `--service-script` option of `mod_wsgi-express`:
  - add additional argument to `$ARGS` like so: `ARGS="$ARGS --service-script celery_starter scripts/openshift-celery-runner.py"`
  - add `openshift-cellery-runner.py`, where I've started processes:
```py
import os

string_command = "celery -A weblate multi start \
  search notify celery memory \
  --pidfile=weblate-%n.pid \
  --logfile=weblate-%n%I.log --loglevel=DEBUG \
  --beat:celery --concurrency:celery=4 --queues:celery=celery --prefetch-multiplier:celery=4 \
  --concurrency:notify=4 --queues:notify=notify --prefetch-multiplier:notify=4 \
  --concurrency:search=1 --queues:search=search --prefetch-multiplier:search=2000 \
  --concurrency:memory=1 --queues:memory=memory --prefetch-multiplier:memory=2000"

print(string_command)

os.system(string_command)
```
With this it looks like service is running (result of `ps aux`):
```
USER        PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
1000340+      1  0.0  0.0 107192  4208 ?        Ss   12:54   0:00 httpd (mod_wsgi-express)         -f /tmp/mod_wsgi-localhost:8080:1000340000/httpd.conf -DMOD_WSGI_WITH_PROXY_HEADERS -DMOD_WSGI_MPM_ENABLE_EVENT_MODULE -DMOD_WSGI_MPM
1000340+     61  0.1  0.6 599736 108296 ?       Sl   12:55   0:05 (wsgi:localhost:8080:1000340000) -f /tmp/mod_wsgi-localhost:8080:1000340000/httpd.conf -DMOD_WSGI_WITH_PROXY_HEADERS -DMOD_WSGI_MPM_ENABLE_EVENT_MODULE -DMOD_WSGI_MPM
[...]

Issue #372, opened 2013-10-01
Progress bar for loading from git [$50]


bahramhamid:
Make the "save" button of  subproject, asynch and show kind of progress bar for loading from the git

nijel:
What I think is needed to implement this:

- [ ] Make component creation just save the database model and trigger further update in Celery
- [ ] Store flag that the component update is in progress to properly report this in UI
- [ ] Once Celery gets to processing, it should log the status and progress
- [ ] UI should indicate that that the component is being updated using status icon
- [ ] Detailed view showing import progress (link from icon and redirect on component creation)

Unclear topics:

- Storage for the update flag and progress - cache or database?
- Should this be integrated into Django admin as well
- How to pipe updates from server to client? Websockets or polling?

rhofer:
Potentially to add:
- [ ] (Optional) If Email configured, drop email on success / failure to admin role (or similar)

nijel:
The notification will be already covered by https://github.com/WeblateOrg/weblate/issues/1956.

rhofer:
@nijel yes thanks. That should do it.

Another remark:

Error handling / exit strategy: Basically we should remove time-out based processing and really going for asynchronous based on state-event also leading to a progress observation. Additionally, we need to avoid to be caught in an "endless loop". Potentially to provide:
- [ ] Cancel / Abort - Button to stop importing (szenario: import started, thereafter, recognized it was the wrong thing, but not to wait for hours until finished...)

/cc @fennibay

rhofer:
Another scenario which should belong to CRUD with components (import is covered above)
- [ ] **Deletion** of large component to be handled asynchronously, as well

Facing similar problems as with import of large components.

nijel:
Thank you for your report, the issue you have reported has just been fixed. 

* In case you see problem with the fix, please comment on this issue. 
* In case you see similar problem, please open separate issue.
* If you are happy with the outcome, consider supporting Weblate by [donating](https://weblate.org/donate/).

nijel:
@rhofer @alesrosina Looking at #2721 I've realized that you're not running Celery (see https://docs.weblate.org/en/latest/admin/install.html#celery) in the OpenShift. Without that this feature will not work, so the OpenShift setup will need some adjustments...

alesrosina:
@nijel hm, thanks for pointing this out. I will try to fix that.

alesrosina:
[...]

PR #2809, opened 2019-06-11
Run celery also on openshift instance

alesrosina:
Covers issues https://github.com/WeblateOrg/weblate/issues/2808 and https://github.com/WeblateOrg/weblate/issues/372

### DIFF
--- a/weblate/settings_openshift.py
+++ b/weblate/settings_openshift.py
@@ -94,6 +94,15 @@ CACHES = {
     }
 }
 
+if 'REDIS_PASSWORD' in os.environ:
+    CELERY_TASK_ALWAYS_EAGER = False
+    CELERY_BROKER_URL = '"redis://:{0}@{1}:{2}"'.format(
+          os.environ.get('REDIS_PASSWORD'),
+          os.environ.get('REDIS_HOST'),
+          os.environ.get('REDIS_PORT')
+      )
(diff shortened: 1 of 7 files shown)


alesrosina:
@nijel & @GrahamDumpleton a code review would be appreciated.

But this solution seems to work.

weblate/settings_openshift.py
```diff
@@ -94,6 +94,29 @@
...
+    CELERY_BEAT_SCHEDULE_FILENAME = os.path.join(
+        DATA_DIR, 'celery', 'beat-schedule'
+    )
+    CELERY_TASK_ROUTES = {
+        'weblate.trans.search.*': {'queue': 'search'},
+        'weblate.trans.tasks.optimize_fulltext': {'queue': 'search'},
+        'weblate.trans.tasks.cleanup_fulltext': {'queue': 'search'},
+        'weblate.memory.tasks.*': {'queue': 'memory'},
+        'weblate.accounts.tasks.notify_change': {'queue': 'notify'},
+    }
```
nijel: You already get the above 4 settings from settings_example, so better not to copy them.

alesrosina: Right, I've removed this. Thanks.

nijel:

PR #2833, opened 2019-06-24
SSH keys directory made configurable

plmakas:
SSH keys directory made configurable.


nijel:
Thanks for submitting PR. Can you please explain motivation behind this? Please note that the directory currently also contains Weblate ssh wrapper, not only the keys.

When we agree having this is a good idea, I can see following issues with this PR:

* The new setting is not documented
* There is no default value in weblate/trans/models/_conf.py
* The SSHWrapper.filename in weblate/vcs/ssh.py will need adjustments
* Tests in weblate/vcs/tests/test_ssh.py and weblate/wladmin/tests.py will need adjustment

nijel:
Closing due to lack of feedback, updates and conflicts.

Issue #2840, opened 2019-06-26
Translations in YAML Format are not working in weblate==3.7


vanDonselaar:
**Describe the bug**
Translations in YAML Format are not working in weblate==3.7

**To Reproduce**
I managed to to run into this bug twice. Once after an upgrade from 3.6.1 to 3.7 and once after reinstalling Weblate from a clean slate on a new Debian EC2 instance.

**Expected behavior**
I expected the same YAML format translation behaviour as with 3.6.1.

**Screenshots**
Yaml format detection when adding a component seems to work:
![image](https://user-images.githubusercontent.com/883626/60177822-365ac200-981a-11e9-9b1f-f068d1fb9f20.png)
But then on the next page the File format falls back to 'App store metadata files'
![image](https://user-images.githubusercontent.com/883626/60177890-538f9080-981a-11e9-9259-3229dcc08bd1.png)

**Server configuration and status**
Output of `list_versions`:
```
$ weblate list_versions
 * Weblate 3.7
 * Python 3.5.3
 * Django 2.2.2
 * Celery 4.3.0
 * celery-batches 0.2
 * six 1.12.0
 * social-auth-core 3.2.0
 * social-auth-app-django 3.1.0
 * django-appconf 1.0.3
 * translate-toolkit 2.4.0
 * translation-finder 1.5
 * Whoosh 2.7.4
 * defusedxml 0.6.0
 * Git 2.11.0
 * Pillow 6.0.0
 * python-dateutil 2.8.0
 * lxml 4.3.4
 * django-crispy-forms 1.7.2
 * django_compressor 2.3
 * djangorestframework 3.9.4
 * user-agents 2.0
 * jellyfish 0.7.2
 * diff-match-patch 20181111
 * pytz 2019.1
 * pyuca 1.2
 * PyYAML 5.1.1
 * Database backends: django.db.backends.postgresql
 * Cache backends: default:RedisCache
 * Email setup: django.core.mail.backends.smtp.EmailBackend: smtp.postmarkapp.com
 * Celery: redis://localhost:6379, redis://localhost:6379, regular
 * Platform: Linux 4.9.0-9-amd64 (x86_64)
```

And `check --deploy`:
```
$ weblate check --deploy
SystemCheckError: System check identified some issues:

ERRORS:
?: (weblate.E008) Please configure separate avatar caching to reduce pressure on the default cache
        HINT: https://docs.weblate.org/en/weblate-3.7/admin/install.html#production-cache-avatar

[...]

PR #2841, opened 2019-06-26
Add optional dependency ruamel.yaml to install.rst

vanDonselaar:
Fixes https://github.com/WeblateOrg/weblate/issues/2840

### DIFF
--- a/docs/admin/install.rst
+++ b/docs/admin/install.rst
@@ -75,16 +75,14 @@ of them in :file:`requirements-optional.txt`.
     https://github.com/sirfz/tesserocr
 ``akismet`` (optional for suggestion spam protection)
     https://github.com/ubernostrum/akismet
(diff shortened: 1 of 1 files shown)


docs/admin/install.rst
```diff
@@ -83,6 +83,8 @@ of them in :file:`requirements-optional.txt`.
     https://python-zeep.readthedocs.io/
 ``aeidon`` (optional for :ref:`subtitles`)
     https://pypi.org/project/aeidon/
+``ruamel.yaml`` (optional for :ref:`yaml`)
+    https://pypi.org/project/ruamel.yaml/
```
nijel: Thanks, but it should actually replace PyYAML above. This has changed in translate-toolkit we use for parsing the files and where new version was released just after release of Weblate 3.7, so it is not covered by docs so far...

PR #2869, opened 2019-07-08
Spelling: to copy configuration from

comradekingu:
Does
"Specify configuration manually" mean 
"Configure manually?


nijel:
Merged thanks.

Yes, it does mean that.

PR #2881, opened 2019-07-15
docs: fix typo in LDAP auth example

zypA13510:
A very minor fix.


nijel:

zypA13510:
@nijel Actually, after taking a second look, I suspect that `dc=example,dc: net` is actually meant to be `dc=example,dc=net`?

nijel:
Hmm, you're probably right. I've probably messed up that when changing the syntax of the environment variables. Fixed in 4e78ef125e

PR #2895, opened 2019-07-19
Spelling: No global, preventiong translations, for uploaded translation,no disabled

comradekingu:


weblate/formats/ttkit.py
```diff
@@ -670,7 +670,7 @@ def target(self):
 
 
 class PoFormat(TTKitFormat):
-    name = _('Gettext PO file')
+    name = _('gettext PO file')
```
nijel: These are labels for the file format, all are starting with uppercase, so I'd rather keep this consistent.

weblate/lang/models.py
```diff
@@ -523,7 +523,7 @@ class Plural(models.Model):
         verbose_name=ugettext_lazy('Plural definition source'),
         choices=(
             (SOURCE_DEFAULT, ugettext_lazy('Default plural')),
-            (SOURCE_GETTEXT, ugettext_lazy('Gettext plural formula')),
+            (SOURCE_GETTEXT, ugettext_lazy('Plural gettext formula')),
```
nijel: It's plural formula coming from gettext, so I believe the original string was correct.

weblate/templates/dictionary.html
```diff
@@ -31,7 +31,7 @@
     </a>
     <ul class="dropdown-menu">
 <li><a href="{% url 'download_dictionary' project=project.slug lang=language.code %}?format=csv">{% trans "Comma separated values (CSV)" %}</a></li>
-<li><a href="{% url 'download_dictionary' project=project.slug lang=language.code %}?format=po">{% trans "Gettext (PO)" %}</a></li>
+<li><a href="{% url 'download_dictionary' project=project.slug lang=language.code %}?format=po">{% trans "gettext (PO)" %}</a></li>
```
nijel: Same as above - "Comma separated" is with capital letter, same should be done for Gettext.

weblate/lang/models.py
comradekingu: In my head that is a "Plural gettext formula", I could be wrong.

weblate/templates/dictionary.html
comradekingu: gettext supports MO and other things too, no?

weblate/formats/ttkit.py
comradekingu: Hmmm, I figured it was easier to always use it like upstream gettext uses it.
Is it a solution to go "GNU gettext"?
Also, I've seen "i18next", which isn't "I18next"

nijel: True, we should be consistent with i18next as well. However even upstream is not consistent on capitalization here when used as first word in a sentence, see https://www.i18next.com/. But in most case they also use lowercase...

nijel:

Issue #2210, opened 2018-09-03
Non-git based components


wichert:
We have translatable content that is not tied to code, for example CMS-style content and email campaign content. Since weblate is a really good tool to handle translations, and are translators have experience with it we would like to use weblate to handle translations instead of trying to replicate parts of it. Currently we are doing that by automatically managing a git repository with a copy of our content, essentially doing a data-export on every change, and re-importing when we receive a webhook from github/bitbucket. I am wondering if it is possible to take out the git part of this? For example use a weblate API to directly push texts to weblate, along with a webhook from weblate so we can automatically receive updated translations or mark content as unusable if a translators flags text as needing review.

nijel:
The biggest problem here is missing API to do all this. Implementing simple backend with not VCS integration would be quite easy.

wichert:
The celery work you are doing would certainly be very useful for webhook delivery.

dragosv:
@nijel What are the missing API?

nijel:
Not yet working, but I'd like to have it for 3.3. There needs to be way to upload new files to translate, that should be exposed in the API. The current upload interface *merges* content of uploaded file into current translation.

dragosv:
@nijel What about a File based repository instead of Source Control (Git, Hg, Svn) ? Just pushing files into a folder then running wlc push and pull? As adding a Source Control is not needed. Not sure how should this options be named.

dragosv:
@nijel This is same concept in Pootle
http://docs.translatehouse.org/projects/pootle/en/stable-2.8.x/features/pootle_fs.html

```
What is a filesystem
A filesystem is actually itself a Pootle FS plugin. Currently two exist:

localfs - allowing synchronization with the filesystem on which Pootle is running
git - synchronization with a Git repository
You can write a plugin for any version control system, Pootle FS will ensure that the same commands and operations are used to ensure Pootle and your filesystem stay synchronized.
```

nijel:
The wlc push and pull does push and pull from remote repository. There is no such thing for local one :-). That is really the missing part here - to implement support for replacing files by the ones uploaded (it doesn't matter whether by web or API).

I will look into this for the 3.6 release, the vcs requirement is really a blocker for some users.

dragosv:
So why not go for a pootle like approach? I mean instead of pushing to git, put the files onto the local filesystem. Same for git pull. I am missing something?

nijel:
There is no problem with such approach, it just needs to be implemented.

[...]

PR #2907, opened 2019-07-25
VCS less operation

nijel:
Implementation for #2210


Jibec:
Hello Michal, will we have an option to deactivate this on a whole Weblate installation?

I feel like having this feature may makes sense for some use case, but may be counterproductive in some community (like Fedora) where I want to make sure to always be in sync with upstream repository.

nijel:
Yes, it will be just another VCS backend which can be removed from the configuration (VCS_BACKENDS). When speaking about that it's not documented right now, I'll add it to the docs :-).

PR #2920, opened 2019-07-30
Spelling: Open Graph

comradekingu:
As per https://ogp.me/
Btw, what does "Legacy badge which used to render PNG" mean?


nijel:
Merged, thanks.

The It's replacement for badges present in some past Weblate releases.

PR #2921, opened 2019-07-30
Spelling: Translate Toolkit, TTKit, iOS strings, SubStation Alpha, media type

comradekingu:
As per 
https://en.wikipedia.org/wiki/SubStation_Alpha
https://toolkit.translatehouse.org/
https://en.wikipedia.org/wiki/Media_type (used to be called MIME type)

Not 100% sure abut TTKit, from both https://www.thumbtack.com/engineering/adopting-swift-migrating-frameworks-ios/ and https://github.com/CocoaPods/CocoaPods/issues/7584

### DIFF
--- a/weblate/formats/ttkit.py
+++ b/weblate/formats/ttkit.py
@@ -17,7 +17,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
-"""translate-toolkit based file format wrappers."""
+"""Translate Toolkit based file format wrappers."""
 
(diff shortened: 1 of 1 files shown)


nijel:
ttkit is used as shortcut for translate-toolkit.

comradekingu:
@nijel And the TTKit that I found is something else? I would prefer it to be "Translate Toolkit" everywhere, since it (otherwise) is very hard to search for.

nijel:
Yes, it's something else. Probably worth fixing to Translate Toolkit in all places.

nijel:

PR #2922, opened 2019-07-30
Spelling: colors indicate → color indicates

comradekingu:
My rationale is that the indicated state is not multicoloured, so colour indicates is always correct, whereas colours indicate can be misinterpreted.

### DIFF
--- a/weblate/templates/ratelimit.html
+++ b/weblate/templates/ratelimit.html
@@ -2,7 +2,7 @@
 {% load translations %}
 
 {% documentation_icon 'user/translating' 'user-rate' right=True %}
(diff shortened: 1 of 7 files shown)


weblate/templates/ratelimit.html
```diff
@@ -2,7 +2,7 @@
 {% load translations %}
 
 {% documentation_icon 'user/translating' 'user-rate' right=True %}
-{% trans "You have recently performed too many operations and your request has been rate limited." %}
+{% trans "You have recently performed too many operations, so your request is carried out a bit slower." %}
```
nijel: Actually when user sees this, the particular request was rejected.

nijel:

PR #2952, opened 2019-08-10
Spelling: Reworked. -the, containing strings, create a ZIP

comradekingu:
Is it always "ZIP", or should it be "compressed file"?


nijel:

PR #2971, opened 2019-08-21
Update django.po

mansil:
Add my name as a translator


nijel:
Thanks for your patch, however this field is updated automatically when edited in Weblate and is designed by gettext to contain name of last translator. All translators are listed on top of the file in the comment.

Issue #2889, opened 2019-07-19
Update OpenShift documentation


nijel:
**Is your feature request related to a problem? Please describe.**
Currently we have only docs covering [OpenShift 2](https://github.com/WeblateOrg/weblate/blob/master/docs/admin/install/openshift2.rst) but no docs for OpenShift 3.

**Describe the solution you'd like**
Add documentation for OpenShift 3, possibly dropping OpenShift 2 one (if it's no longer useful).

**Additional context**
https://github.com/WeblateOrg/docker/issues/298, https://github.com/WeblateOrg/docker/issues/292, https://github.com/WeblateOrg/weblate/pull/2977

alesrosina:
Yeah, let's write it down when docker is fixed, since this would bring also a bit of changes to actual openshift deployment.

But agree, this needs to be added.

nijel:
I've added stub article here: https://github.com/WeblateOrg/weblate/blob/master/docs/admin/install/openshift.rst

GrahamDumpleton:
Latest is actually OpenShift 4.2, but new documentation would work for for 3.X and 4.X at this point. Note that 4.2+ support is blocked by https://github.com/WeblateOrg/weblate/issues/3199

PR #2977, opened 2019-08-26
Chore/openshift from docker image deployment

alesrosina:
I've reconfigured openshift deployment to use official Weblate docker image and does not have it's own build and settings.

I have also removed options for mysql & sqllite, since official docker image runs only with postgres.


alesrosina:
@GrahamDumpleton @sylvainar I would appreciate your input, since you two guys also worked on openshift implementation.

openshift3/persistent-postgresql.json
```diff
@@ -256,7 +171,8 @@
                         "containers": [
                             {
                                 "name": "weblate",
-                                "image": "${APPLICATION_NAME}-img",
+                                "image": "weblate/weblate:${WEBLATE_VERSION}",
+                                "imagePullPolicy": "IfNotPresent",
```
nijel: This will lead to never updating if somebody chooses `latest`. I have no clue how common that approach is on OpenShift though, so it might be actually okay :-).

nijel:
Thanks, looks great (removes lot of code ;-)). I will wait with merging for some time to gather feedback from others as well.

openshift3/persistent-postgresql.json
alesrosina: actually, you are right, this is not really wanted. Will change it to `Always`, makes more sense.

nijel:

It would be great if you could write up short docs on how to deploy on OpenShift 3, see https://github.com/WeblateOrg/weblate/issues/2889

sylvainar:
Sorry guys i'm a bit late. That's great! I'm happy to see this is moving :smile:

Issue #2643, opened 2019-03-18
Advanced search


nijel:
**Is your feature request related to a problem? Please describe.**
In several situations it might be useful to search for more complex conditions than current search allows. Probably something similar like GitHub has (see https://github.com/search/advanced).

This at least should cover:

* Number of votes
* Last change (see https://github.com/WeblateOrg/weblate/issues/2455)
* Skipping own suggestions (see https://github.com/WeblateOrg/weblate/issues/2023)

**Describe the solution you'd like**
Possibility to specify more search parameters in both UI and text box.

**Describe alternatives you've considered**
Current search forms really don't scale well further.

nijel:
In the end, this might be blocked by #2831, so the question is whether to move all this to 3.10 or prioritize blocking issues for 3.9.

nijel:
The new search is currently used only for screenshots adding (where only substring search was available previously) and is documented here: https://docs.weblate.org/en/latest/user/search.html

Full features and replacement of normal search depends on https://github.com/WeblateOrg/weblate/issues/3007 and https://github.com/WeblateOrg/weblate/issues/2831.

nijel:
Partly addressed by https://github.com/WeblateOrg/weblate/pull/3059, rest will be implemented after #3007 and #2831.

ramon-b:
A dropdown to filter on context (groups) would also be nice. It is already possible to search in context, but translators do not always know the context (group(s)). We are using i18next, whereby we have contexts like .support.form.title. Our translators want to be able to select a group, so that they can search in specific (sub)groups without using regex.

Issue #2455, opened 2018-12-10
Translate(/search) recently added [$200]


Kingdutch:
We are just starting with translations for some of our projects which causes us to have quite a backlog of untranslated strings. One of the things we'd like to do is translate the most recently added strings (because they'll be most visible).

In search we can specify "Changed since" but this seems to work mostly on the translations.

Is there a way to order missing translations by the date on which their source string was added? Or alternatively to show only missing translations for source strings that were added after a certain date?

nijel:
Currently it's not possible, but it might be useful.

nijel:
You can use [Bountysource](https://www.bountysource.com/issues/67291562-question-translate-search-recently-added) for that :-). Other options for getting paid support are listed on https://weblate.org/support/

nijel:
In the end, this might be blocked by #2831, so the question is whether to move all this to 3.10 or prioritize blocking issues for 3.9.

nijel:
Thank you for your report, the issue you have reported has just been fixed. 

* In case you see a problem with the fix, please comment on this issue. 
* In case you see similar problem, please open separate issue.
* If you are happy with the outcome, consider supporting Weblate by [donating](https://weblate.org/donate/).

Issue #2831, opened 2019-06-24
Improve source strings handling and visibility


nijel:
**Is your feature request related to a problem? Please describe.**
Currently source strings are tracked by separate model what leads to several problems:

* UI wise the source strings and their checks are hidden somewhere in the navigation, it should be at same level as translations to get higher visibility
* Editing the priority or flags anyway leads to updating all other units, so the source really does not matter here
* The UI for source strings is completely different to translations and is ugly. This is to biggest extend caused by it's invisibility - nobody cares.

**Describe the solution you'd like**

* Track source strings as regular translations. This is already there for monolingual files with editable base, for rest it needs to be implemented.
* Adjust translation and translate UI to better accommodate source strings view (show source checks, source comments, make the string non editable if bilingual or configured so, probably depends on #743).
* Adjust comment notification to link to source string for source string comments (should address #2812)

**Steps needed**

* [x] Add support for non translatable strings, https://github.com/WeblateOrg/weblate/issues/743
* [x] Move Source model logic to Unit on source translation
* [x] Migrate Source metadata to Unit, creating source language Translation
* [x] Change the `edit_template` logic to add read-only flag on translation
* [x] Integrate Source view into current translate views (context, screenshots, ...)
* [x] Rethink `source.edit` and `unit.template` permissions

comradekingu:
I really like this, clicking into specific projects per language to get every instance of every check gets old quickly.

nijel:
Thank you for your report, the issue you have reported has just been fixed. 

* In case you see a problem with the fix, please comment on this issue. 
* In case you see similar problem, please open separate issue.
* If you are happy with the outcome, consider supporting Weblate by [donating](https://weblate.org/donate/).

PR #3001, opened 2019-09-07
Avanced search

nijel:
Query parser for advanced search (#2643, #2455)

Todo:

- [x] Complete Whoosh to SQL conversion
- [x] Fix term parsing for TEXT fields
- [x] Add Whoosh to SQL conversion tests
- [x] Improve timestamp handling (2012-02-02 should be parsed) (see https://github.com/whoosh-community/whoosh/issues/552)
- [ ] Use in unit search
- [x] Use in screenshot adding
- [ ] Change UI to generate query instead of submitting it directly
- [x] User documentation

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -14,6 +14,7 @@
     user/translating
     user/files
     user/checks
+    user/search
     devel/index
     workflows
     faq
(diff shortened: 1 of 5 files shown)


nijel:
In the end, this might be blocked by https://github.com/WeblateOrg/weblate/issues/2831, so the question is whether to move all this to 3.10 or prioritize blocking issues for 3.9.

PR #3024, opened 2019-09-15
Add support for pushing to GitLab using merge requests.

maxking:
This support is based on "lab" CLI, which is inspired by "hub" tool, which is
used for GitHub's pull request support.

"lab" also proxies a lot of commands to "hub", so parts of the Github and
Gitlab's requirements are abstracted out in a mixin class and then only the
basic functionality that is different is implemented via separate classes.

Creating a draft PR for early reviews, will update with test cases soon.

### DIFF
--- a/docs/vcs.rst
+++ b/docs/vcs.rst
@@ -344,3 +344,68 @@ monolingual translations).
 In the background Weblate creates Git repository for you and all changes are
 tracked in in. In case you decide later to use VCS to store the translations,
 it's already within Weblate and you can base on that.
+
+.. _vcs-gitlab:
+
+GitLab
+------
+
+.. versionadded:: 3.9
+
+This just adds a thin layer on top of :ref:`vcs-git` to allow pushing
+translation changes as merge requests instead of pushing directly to the
+repository. It currently uses the `lab`_ tool to do the push.
+
+There is no need to use this access Git repositories, ordinary :ref:`vcs-git`
+works the same, the only difference is how pushing to a repository is
+handled. With :ref:`vcs-git` changes are pushed directly to the repository,
+while :ref:`vcs-gitlab` creates merge request.
+
+.. _gitlab-push:
+
+Pushing changes to GitLab as merge request
+++++++++++++++++++++++++++++++++++++++++++
+
+If you are translating a project that is hosted on GitLab and don't want to
+push translations to the repository, you can have them sent as a merge request.
+
+You need to configure the `lab`_ command line tool and set
+:setting:`GITLAB_USERNAME` for this to work.
+
+.. seealso::
+
+   :setting:`GITLAB_USERNAME`, :ref:`lab-setup` for configuration instructions
+
+.. _lab-setup:
+
+Setting up lab
+++++++++++++++
+
+:ref:`gitlab-push` requires a configured `lab`_ installation on your
+server. Follow the installation instructions at
+https://github.com/zaquestion/lab#installation and perform and run it without
+any arguments to finish configuration, for example:
+
+.. code-block:: sh
+
+    # DATA_DIR is set in Weblate settings.py, set it accordingy.
+    # Is is /app/data in Docker
+    $ HOME=${DATA_DIR}/home lab
+    Enter GitLab host (default: https://gitlab.com):
+    Create a token here: https://gitlab.com/profile/personal_access_tokens
+    Enter default GitLab token (scope: api):
+    Config saved to ~/.config/lab.hcl
+
+
+The `lab`_ will ask you for your GitLab access token, retrieve a token and
+store it into :file:`~/.config/lab.hcl`. The file has to be readable by user
+running Weblate.
+
+
+.. note::
+
+    Use the username you configured :guilabel:`lab` with as
+    :setting:`GITLAB_USERNAME` (:envvar:`WEBLATE_GITLAB_USERNAME` for the
+    Docker image).
+
+.. _lab: https://github.com/zaquestion/lab
--- a/weblate/vcs/git.py
+++ b/weblate/vcs/git.py
@@ -517,25 +517,9 @@ class GitMergeRequestBase(GitRepository):
             return False
         return super(GitMergeRequestBase, cls).is_supported()
 
-    def create_pull_request(self, origin_branch, fork_branch):
-        """Create pull request to merge branch in forked repository into
-        branch of remote repository.
-        """
-        cmd = [
-            'pull-request',
-            '-f',
-            '-h',
-            '{0}:{1}'.format(settings.GITHUB_USERNAME, fork_branch),
-            '-b',
-            origin_branch,
-            '-m',
-            settings.DEFAULT_PULL_MESSAGE,
-        ]
-        self.execute(cmd)
-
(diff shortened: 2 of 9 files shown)


weblate/vcs/git.py
```diff
@@ -615,7 +590,39 @@ def push(self):
...
-        super(GithubRepository, self).configure_remote(pull_url, None, branch)
+        super(_GitRemoteRepositoryMixin, self).configure_remote(
+            pull_url, None, branch)
+
+
+class GithubRepository(_GitRemoteRepositoryMixin):
+
+    name = 'GitHub'
+    _cmd = 'hub'
+    _username = settings.GITHUB_USERNAME
```
nijel: This assigns a value to local variable at class initialization, what makes it impossible to override setting in the tests. It might be better to assign setting name here and use getattr to fetch actual value later.

Apparently running tests for GitHub pull requests was broken for some time, I've just fixed it in bbaadb62c696442c380b7c888173047ee7bc7e2b.

maxking: Ah, that makes sense.

How about I convert `_username` to a `@property` instead?  It is slightly smaller and cleaner than multiple `getattr(settings, self._username_setting_name)`.

nijel: Good idea.

nijel:
Looks like you've done something wrong in rebasing, the merge request contains lot of upstream commits...

maxking:
Yeah, I realized that. Will need to cherry-pick the commits out of this mess.

weblate/vcs/git.py
```diff
@@ -544,57 +544,32 @@ def push(self):
...
-            '-m', settings.DEFAULT_PULL_MESSAGE,
-        ]
-        self.execute(cmd)
+        return super(_GitRemoteRepositoryMixin, cls).is_supported()
 
     def push_to_fork(self, local_branch, fork_branch):
         """Push given local branch to branch in forked repository."""
-        if settings.GITHUB_USERNAME:
-            cmd_push = ['push', '--force', settings.GITHUB_USERNAME]
+        if self._username:
```
nijel: I know this code was originally there, but can you please remove this condition? It's not useful anymore (I'm too lazy to dig out original use case in git history). The class will not be used without username being available.

weblate/vcs/git.py
```diff
@@ -544,57 +544,32 @@ def push(self):
         self.execute(['svn', 'dcommit', self.branch])
 
 
-class GithubRepository(GitRepository):
+class _GitRemoteRepositoryMixin(GitRepository):
```
nijel: This is not really a mixin, but rather a base class, so I'd prefer cleaner name. How about following?
```suggestion
class GitMergeRequestBase(GitRepository):
```

docs/vcs.rst
```diff
@@ -340,3 +340,66 @@ monolingual translations).
 In the background Weblate creates Git repository for you and all changes are
 tracked in in. In case you decide later to use VCS to store the translations,
 it's already within Weblate and you can base on that.
+
+.. _vcs-gitlab:
+
+GitLab
+------
+
+.. versionadded:: 3.10
```
nijel: I think we can do it for 3.9 :-)

weblate/vcs/git.py
maxking: Sure, will do!

weblate/vcs/git.py
maxking: Sounds good, do you need a `_` in the beginning since this is kind of an Internal Implementation detail? :)

docs/vcs.rst
maxking: Ah, I didn't check the latest version. Will do.

maxking:
I haven't tested it out yet on a real environment, part of which is because last time I tried to start up a local instance using the `rundev.sh` script, it kept on failing due to missing container dependencies. I haven't had time to actually get to run Weblate locally yet. 

I probably need to read the docs more for the setup from sources and it should be good to go, but I hoped to skip the setup with containers if possible.

As soon as I can get this to run locally and verify against a repo, I'll remove the draft status. Thanks for patiently reviewing everything :)

nijel:
Hmm, the `rundev.sh` was contributed some time ago and I don't think anybody uses that now, so it can be broken. I fixed the most visible breakage in 365ee4680dd1e8e9e197532a6092c752a287afda

docs/vcs.rst
nijel:
```suggestion
.. versionadded:: 3.9
```

weblate/vcs/git.py
nijel: It's still base class and somebody else might want to build third party vcs backend on that.

weblate/vcs/git.py
nijel: I've done this in 3d7cc2ddf2, you will have to rebase, sorry.

maxking:
I think I was finally able to test the MR creation (https://gitlab.com/maxking/hyperkitty/merge_requests/19) and seems to work. There were a few problems with the original code, which I fixed.

Some interesting details I discovered:
- `lab` does not allow specifying which branch to use as the source and defaults to the current branch, because of which, I had to wrap around the MR creation with a checkout of the remote branch we want to use as the source.
- `lab` does not have a way to specify the configs via the environment variables, so it must be created in the `DATA_DIR/home/.config/lab.hcl`. I have updated the docs to setup lab as well, so it should reflect the same changes.

I think I have fixed all of your suggestions as well, but please let me know if you have any more.

I am really excited to be able to use this feature in hosted.weblate.org and retire my custom shell script running on cron :)

weblate/vcs/git.py
maxking: Done.

nijel:
Thanks, can you please check the test failure? There seems to be something passing None to the commandline:

```py-tb
  File "/home/travis/build/WeblateOrg/weblate/weblate/vcs/tests/test_vcs.py", line 234, in test_upstream_changes
    self.add_remote_commit()
  File "/home/travis/build/WeblateOrg/weblate/weblate/vcs/tests/test_vcs.py", line 151, in add_remote_commit
    repo.push()
  File "/home/travis/build/WeblateOrg/weblate/weblate/vcs/git.py", line 558, in push
    self.push_to_fork(self.branch, self.branch)
  File "/home/travis/build/WeblateOrg/weblate/weblate/vcs/git.py", line 539, in push_to_fork
    self.execute(cmd_push + ['{0}:{1}'.format(local_branch, fork_branch)])
  File "/home/travis/build/WeblateOrg/weblate/weblate/vcs/base.py", line 191, in execute
    args = [arg.encode('utf-8') for arg in args]
  File "/home/travis/build/WeblateOrg/weblate/weblate/vcs/base.py", line 191, in <listcomp>
    args = [arg.encode('utf-8') for arg in args]
AttributeError: 'NoneType' object has no attribute 'encode'
```

maxking:
@nijel I think the tests are passing now, there are some jobs which might need re-starting since they seems to have errored out due to something getting stuck.

nijel:
We're hitting the 50 minutes job limit on Travis CI, right now we're testing Drone CI as a replacement for that and the tests passed there.

nijel:

Issue #3022, opened 2019-09-15
Error when trying to add Mercurial repository with http access


tariver:
I have a repository that runs Kallithea 0.2.1 with Mercurial support and access via http
So I add new translation component to the Weblate Project with the following parameters:
Version control system - Mercurial
Source code repository - http://username:mypassword@domain:portnumber/repository
Repository branch - testing

But after pressing continue I get an error under the "Source code repository" field - "Could not update repository: abort: unknown revision 'testing'! (255)"

Weblate's log looks as following
2019-09-12 18:38:03,510 DEBUG hg: exec hg init /data_directory/vcs/project/component [retcode=0]
2019-09-12 18:38:03,510 DEBUG hg: exec hg init /data_directory/vcs/project/component [retcode=0]
2019-09-12 18:38:03,522 INFO project/component: updating repository
2019-09-12 18:38:27,016 DEBUG hg: exec hg pull --branch . [retcode=0]
2019-09-12 18:38:27,016 DEBUG hg: exec hg pull --branch . [retcode=0]
2019-09-12 18:38:27,017 INFO project/component: update took 23.49 seconds
2019-09-12 18:38:27,017 DEBUG project/component: update: pulling from http://username:***@domain:portnumber/repository
2019-09-12 18:38:27,018 DEBUG project/component: update: adding changesets
2019-09-12 18:38:27,018 DEBUG project/component: update: adding manifests
2019-09-12 18:38:27,018 DEBUG project/component: update: adding file changes
2019-09-12 18:38:27,018 DEBUG project/component: update: added 1326 changesets with 4606 changes to 1300 files
2019-09-12 18:38:27,019 DEBUG project/component: update: (run 'hg update' to get a working copy)
2019-09-12 18:38:27,106 DEBUG hg: exec hg branches --template {branch}
 [retcode=0]
2019-09-12 18:38:27,106 DEBUG hg: exec hg branches --template {branch}
 [retcode=0]
2019-09-12 18:38:27,190 DEBUG hg: exec hg update testing [retcode=255]
2019-09-12 18:38:27,190 DEBUG hg: exec hg update testing [retcode=255]

If I then connect to the Weblate server via ssh and go to the /data_directory/vcs/project/component and print out command `hg branch` it states that the current branch is "default". Moreover the output of command `hg branches --template {branch}"\n"` doesn't list the branch "testing" in available branches.
[...]

PR #3025, opened 2019-09-16
Hg fixes

nijel:

### DIFF
--- a/weblate/vcs/base.py
+++ b/weblate/vcs/base.py
@@ -73,7 +73,6 @@ class Repository(object):
     _cmd = 'false'
     _cmd_last_revision = None
     _cmd_last_remote_revision = None
-    _cmd_push = None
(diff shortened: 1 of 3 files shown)


patrickdepinguin:
From what I can see and judge, these changes are sound and should solve the pull issue.
I still wonder what will happen at the moment of switching to another branch. Initially, the weblate local repo will be at the old branch. The pull will pull in the new branch. However, I'm not sure if the rebase logic will do what it should. In fact, one could argue whether weblate should rebase anything at all at this point. In my point of view, it may be safer to not rebase anything when the branch switches, just update to the new head (and assume the repo owner has first pulled all translations from the old branch).

nijel:
Yes, the rebase is apparently problematic, see https://github.com/WeblateOrg/weblate/issues/3022#issuecomment-531827538

PR #3044, opened 2019-09-22
Watched components

comradekingu:


weblate/accounts/models.py
```diff
@@ -419,7 +419,7 @@ class Profile(models.Model):
     DASHBOARD_COMPONENT_LISTS = 6
 
     DASHBOARD_CHOICES = (
-        (DASHBOARD_WATCHED, _('Watched translations')),
+        (DASHBOARD_WATCHED, _('Watched components')),
```
nijel: Actually it's "Translations to your languages from watched components", so neither of these is accurate.

comradekingu: **Watched component string changes in your languages**
is not good
**New translations from watched components in your languages**
is somewhat better

Doesn't seem to upset anything length-wise

image

nijel: This is about *Dashboard* - what will be default tab shown there. The same string is used as label there.

nijel:
I've merged some minor changes here, but the biggest change is IMHO wrong - this is about what is shown on dashboard, not about notifications. Also "Watched components" sounds wrong label for list of translations from watched components.

PR #3062, opened 2019-09-28
Spelling: Turn on, for the project

comradekingu:
I don't understand "to get access to additional components." Should it be something like "to make use of automated translations from more projects"?


nijel:
Only components with this enabled are listed in the select box.

nijel:

PR #3075, opened 2019-09-30
Spelling: Download original translation files as ZIP file, -an

comradekingu:


weblate/templates/project.html
```diff
@@ -48,7 +48,7 @@
       {% trans "Files" %} <span class="caret"></span>
     </a>
     <ul class="dropdown-menu">
-      <li><a href="{% url "download_project"  project=object.slug %}?format=zip" title="{% trans "Download for an offline translation." %}">{% blocktrans %}Download original translation files in ZIP{% endblocktrans %}</a></li>
+      <li><a href="{% url "download_project"  project=object.slug %}?format=zip" title="{% trans "Download for offline translation." %}">{% blocktrans %}Download original translation files as ZIP file{% endblocktrans %}</a></li>
```
nijel: Was `an` removal intentional here? In that case it should be applied to all other occurrences as well.

comradekingu: Yes, done :)

nijel:

PR #3079, opened 2019-10-01
Spelling: Avoiding "translation keys". and "units".

comradekingu:
A lot of guesswork going on here.

### DIFF
--- a/weblate/addons/resx.py
+++ b/weblate/addons/resx.py
@@ -31,8 +31,8 @@ class ResxUpdateAddon(BaseCleanupAddon):
     verbose = _("Update RESX files")
     description = _(
         "Update all translation files to match the monolingual base file. "
-        "It removes stale translation keys and adds new ones with source "
-        "string as a translation."
+        "It removes unused strings, and adds new ones wherein the translation is "
+        "a copy of the source string."
     )
     icon = "refresh"
     compat = {"file_format": {"resx"}}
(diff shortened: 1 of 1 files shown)


weblate/addons/resx.py
```diff
@@ -31,8 +31,8 @@ class ResxUpdateAddon(BaseCleanupAddon):
     verbose = _("Update RESX files")
     description = _(
         "Update all translation files to match the monolingual base file. "
-        "It removes stale translation keys and adds new ones with source "
-        "string as a translation."
+        "It removes unused strings, and adds new ones wherein the translation is "
```
comradekingu: Removes from where, unused as in?

nijel: Short summary: It does pretty much same thing as msgmerge does for gettext PO files

It takes the source language file and removes from target language files any strings which are no longer present there, adds new ones and updates changed ones. This used to be part of the cleanup addon, but it turned out that only part of RESX users want the updating part, it depends how RESX is implemented in the application.

weblate/addons/resx.py
```diff
@@ -67,7 +67,7 @@ def update_resx(self, index, translation, storage, changes):
 
     @staticmethod
     def find_changes(index, storage):
-        """Find changed units in storage"""
+        """Find changed string IDs in upstream repository"""
```
nijel: In the code, we still call these units. Also it's not a repository, it's a file storage.

comradekingu: The difference is lost on me here. Does it just find them, or something more? How about "Find changed string IDs upstream"?

nijel: Unit is an object representation of what we call string in the interface. Mostly for historical reasons it's called differently in the code (actually it matches translate-toolkit terminology here).

nijel:

PR #3096, opened 2019-10-08
git: don't assume the last element is the version number

minusf:

This change makes weblate work with some custom git versions. The current implementation simply takes the last element from the output of `git --version`, which in the case of git on macOS returns:

```
git version 2.20.1 (Apple Git-117)
```

A more robust check might be to always take the 3rd element from the beginning.

The current VCS test infrastructure does not make it possible to mock the git version output, and the whole `VCSGitTest` class depends on a working `git` right from `setUp`...


nijel:

PR #3107, opened 2019-10-13
Add the ability to switch to Gitlab as a VCS type for a component.

maxking:
This change follows a previous PR (#3024 ) and adds the ability to switch translation components to Gitlab as a VCS type.


nijel:
Good catch. It was missing in few other places as well, I've fixed that in 3f2c036205.

maxking:
@nijel Thanks! How long does it usually take to land these changes in hosted.weblate.org? I would love to use these for GNU Mailman translation.

Issue #2787, opened 2019-05-30
Doublespace check


comradekingu:
Right now adding a doublespace is not an active check, but it is one of the most common errors.
https://hosted.weblate.org/checks/

Wrong:
```
"This is an  example"
```
Correct:
```
"This is an example"
```

Triggering it on doublespaces within first to last char would avoid false checks for spaces used as indentation, or line-endings. (For which there are already other checks)

MetalDent:
Hi, anyone working on this?

jschmidtlein:
I was planning to take a look, but hadn't actually started yet!

nijel:
Addressed by https://github.com/WeblateOrg/weblate/pull/3115

nijel:
Thank you for your report, the issue you have reported has just been fixed. 

* In case you see a problem with the fix, please comment on this issue. 
* In case you see similar problem, please open separate issue.
* If you are happy with the outcome, consider supporting Weblate by [donating](https://weblate.org/donate/).

PR #3115, opened 2019-10-15
Adding doublespace check

y0nd3rb0y:
Resolving issue #2787 

- [X] Every commit has message which describes it
- [X] Every commit is needed on it's own, if you have just minor fixes to previous commits, you can squash them
- [X] Any code changes come with testcase

### DIFF
--- a/weblate/checks/chars.py
+++ b/weblate/checks/chars.py
@@ -20,6 +20,8 @@
 
 from __future__ import unicode_literals
 
+import re
+
 from django.utils.translation import ugettext_lazy as _
 from six.moves.html_parser import HTMLParser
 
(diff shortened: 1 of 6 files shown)


weblate/checks/chars.py
```diff
@@ -137,6 +139,28 @@ def get_fixup(self, unit):
...
+
+    def check_single(self, source, target, unit):
+        # One letter things are usually decimal/thousand separators
+        if len(source) <= 1 and len(target) <= 1:
+            return False
+        if not source or not target:
+            return False
+
+        # Check if target contains double space
+        return bool(re.search(' {2,}', target))
```
nijel:
```suggestion
        return '  ' in target
```

I think this will perform way better than doing regexp search.

y0nd3rb0y:
Thanks Michal @nijel . I've added your suggestions accordingly
Please let me know if I can do anything else for the check

nijel:
Thanks. The code looks good now, the only missing piece is documentation in https://github.com/WeblateOrg/weblate/blob/master/docs/user/checks.rst

nijel:
Thanks, I think it's not ready to be merged after bugfix 3.9.1 is released. It would be great if you could meanwhile squash all commits into one on top of current master (if not I will do it when merging).

PR #3133, opened 2019-10-21
deepl: Switch to API version 2

aszlig:
Since October 2018, the DeepL API version 1 doesn't seem to be supported anymore and in fact when trying to send requests to their API via curl, we now subsequently get an `HTTP 403` error.

This is also noted in their upstream documentation:

https://www.deepl.com/en/docs-api.html?part=accessing

Fortunately the changes between API version 1 and 2 seem to be not very substantial and while in theory we could just change the URL and be done, I decided to improve the implementation a bit.

So first of all, there seems to now be an API endpoint, which will respond with the list of languages, which we now use instead of the hardcoded list.

Additionally, the response for the `/translate` endpoint not only returns the translation(s) but also the "detected source language". It looks something like this:

```json
{"translations": [
   {
     "detected_source_language": "DE",
     "text": "Hello world!"
   }
]}
```

When testing against the API I found that the detected source language seems to always match the source language we provide via the `source_lang` parameter, but to be absolutely sure things don't go wrong I added an additional check to make sure that the detected_source_language is equal
to the source language we provided.

### DIFF
--- a/weblate/machinery/tests.py
+++ b/weblate/machinery/tests.py
@@ -679,7 +679,7 @@ class MachineTranslationTest(TestCase):
     def test_deepl(self):
         machine = self.get_machine(DeepLTranslation)
         httpretty.register_uri(
-            httpretty.POST, 'https://api.deepl.com/v1/translate', body=DEEPL_RESPONSE
+            httpretty.POST, 'https://api.deepl.com/v2/translate', body=DEEPL_RESPONSE
         )
         self.assert_translate(machine, lang='de', word='Hello')
 
@@ -688,7 +688,7 @@ class MachineTranslationTest(TestCase):
     def test_cache(self):
         machine = self.get_machine(DeepLTranslation, True)
         httpretty.register_uri(
-            httpretty.POST, 'https://api.deepl.com/v1/translate', body=DEEPL_RESPONSE
+            httpretty.POST, 'https://api.deepl.com/v2/translate', body=DEEPL_RESPONSE
         )
         # Fetch from service
         self.assert_translate(machine, lang='de', word='Hello')
--- a/weblate/machinery/deepl.py
+++ b/weblate/machinery/deepl.py
@@ -24,7 +24,7 @@ from django.conf import settings
 
 from weblate.machinery.base import MachineTranslation, MissingConfiguration
 
-DEEPL_API = 'https://api.deepl.com/v1/translate'
+DEEPL_API_BASE = 'https://api.deepl.com/v2'
 
 
 class DeepLTranslation(MachineTranslation):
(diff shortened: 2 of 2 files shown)


nijel:
Thanks. I didn't go into this because of following:

> Version 2 (v2) of the DeepL API is not compatible with CAT tools and is not included in DeepL plans for CAT tool users.

Probably something which should be clarified with them prior to switching to v2 API.

Also the both API endpoints are expected to be same:

> Currently, the two DeepL API versions differ only in the URL used for access.

See https://www.deepl.com/en/docs-api.html?part=accessing

aszlig:
> Also the both API endpoints are expected to be same:

Yeah, that's what I also noted in the commit message:

> Fortunately the changes between API version 1 and 2 seem to be not very substantial and while in theory we could just change the URL and be done, I decided to improve the implementation a bit.

aszlig:
Hm, the [`flake8` error](https://travis-ci.com/WeblateOrg/weblate/jobs/247739849#L700) seems to be unrelated to this change.

aszlig:
@nijel: FYI: Just contacted them via the contact form.

weblate/machinery/tests.py
```diff
@@ -679,7 +679,7 @@ def test_saptranslationhub_invalid(self):
     def test_deepl(self):
         machine = self.get_machine(DeepLTranslation)
         httpretty.register_uri(
-            httpretty.POST, 'https://api.deepl.com/v1/translate', body=DEEPL_RESPONSE
+            httpretty.POST, 'https://api.deepl.com/v2/translate', body=DEEPL_RESPONSE
         )
```
nijel: You should mock the languaages response here as well.

aszlig:
@nijel: Okay, then I misread their [API documentation](https://www.deepl.com/docs-api.html?part=accessing), since I thought APIv1 would not be supported anymore since October last year. Rereading the documentation however does look like they don't intend to deprecate it for CAT plans, at least it seems so. We'll see once I get an answer from them.

nijel:
Yes, the v1 API still works fine for earlier purchased plans (we're using it at Hosted Weblate), while v2 API doesn't work for these.

aszlig:
Catching up on this, I got an answer from them (a while ago already, sorry for the delay):

> For Weblate, you will need either DeepL Pro Advanced or Ultimate. Communication with Weblate runs through v1 of the DeepL API, as it is a CAT tool.                                                                                           
>
> You currently have the DeepL Pro API subscription. This subscription does not allow you to integrate your DeepL Pro subscription into CAT tools, and communication through v1 of the DeepL API is not possible with this subscription. DeepL Pro API subscriptions only communicate through v2 of the DeepL API.                                                                                                                                                                          
>                                                                                                                                                                                                                                              
> If you want to use DeepL Pro within Weblate, you will have to book a DeepL Pro Advanced or Ultimate subscription. You can either cancel your current API subscription and then book a DeepL Pro Advanced or Ultimate subscription, or book an additional subscription.

So if I understand this correctly (I think their terms, calling them `v1` and `v2`, is quite confusing), Weblate should be using API v1 and not v2 as it is, hence I'm closing this PR.

nijel:

PR #3144, opened 2019-10-24
Fix Docker docs //EMAIL_HOST_* vars to EMAIL_*

XaviTorello:
It seems that there are some typos with `EMAIL_HOST_USER` and `EMAIL_HOST_PASSWORD` vars.

Reviewing the code it seems that should be `EMAIL_USER` and `EMAIL_PASSWORD`.

### DIFF
--- a/docs/admin/install/docker.rst
+++ b/docs/admin/install/docker.rst
@@ -36,8 +36,8 @@ The following examples assume you have a working Docker environment, with
               - 80:8080
             environment:
               WEBLATE_EMAIL_HOST: smtp.example.com
-              WEBLATE_EMAIL_HOST_USER: user
-              WEBLATE_EMAIL_HOST_PASSWORD: pass
+              WEBLATE_EMAIL_USER: user
+              WEBLATE_EMAIL_PASSWORD: pass
               WEBLATE_SERVER_EMAIL: weblate@example.com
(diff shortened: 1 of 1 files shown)


nijel:
Why do you think there is a typo? The both variables are correct, see https://docs.djangoproject.com/en/2.2/ref/settings/#email-host-user and https://docs.djangoproject.com/en/2.2/ref/settings/#email-host-password

nijel:
See also https://github.com/WeblateOrg/weblate/pull/1539

XaviTorello:
Sorry @nijel, I don't want to disturb you...

I'm trying to create a `helm` `Chart` for `weblate` (based from the docker-compose repo) and both env vars seems to be abled: 

https://github.com/WeblateOrg/docker-compose/blob/master/environment#L53-L56
 https://github.com/WeblateOrg/weblate/blob/ac8e2c62554150c11a83fba315e8325db0f153bf/weblate/settings_docker.py#L967

Maybe it's a little bit confusing and can be improved using just the normal `Django` vars. 

Provably it's something that should be fixed in the composition.

nijel:
The variables were misnamed, we've fixed them to follow Django ones, but for compatibility with existing setups the wrong names are also accepted, see https://github.com/WeblateOrg/weblate/pull/1539

There is helm chart being created in https://github.com/WeblateOrg/docker-compose/pull/8, maybe you can join the discussion there.

PR #3173, opened 2019-11-04
Remove unused code

alesrosina:
I think this is not used anywhere else, since we removed special openshift deployment config in favor of using docker instance for openshift.

Please review the code removal, but i think it is safe to remove it.


nijel:
There is still some OpenShift 2 files referring this in `.openshift` and `openshift`. In case these have no use, we should remove them as well.

nijel:
I've removed rest of OpenShift 2  in 9098bacf187c4c8d9b915a8e15831c3c8e6c2681

alesrosina:
Yeah, I dont have experince with openshift 2, but i guess those files are still needed for this platform. Maybe @GrahamDumpleton could tell more, he is more experinced in openshift then me.

nijel:
Well it has been broken since a9594ff58ecc637a5ad8b102a46d16398a36e602 (it used settings file which has been removed here), so my impression was that it's no longer useful.

alesrosina:
Ah, didnt know that. Then even better if it is removed.

PR #3174, opened 2019-11-04
Add new url to CSP request header

alesrosina:
When using Auth0 as authentication provider, admin can set also custom image hosted on some external server. When this occurs, it should be added to `Content-Security-Policy` http header.


nijel:
Thanks for review.

PR #3202, opened 2019-11-14
Reworked

comradekingu:


nijel:
Can you please include debug log in the report? See https://wammu.eu/support/bugs/ for information how to obtain it.

PR #3210, opened 2019-11-15
Update profile.rst

mayeulk:
Adding "by default" in "While by default everybody can browse projects", as access control settings can change this


nijel:

PR #3220, opened 2019-11-17
Update review.rst

mayeulk:
More details on activity reports. Rename page as "Reviewing strings" since it involves target strings.


nijel:

PR #3258, opened 2019-11-30
Spelling: in {{ component }}

comradekingu:
Consistent with https://github.com/WeblateOrg/weblate/blob/master/weblate/templates/mail/repository_error_subject.txt#L3


nijel:

PR #3269, opened 2019-12-03
Spelling: A user, -removed

comradekingu:


weblate/templates/manage-access.html
```diff
@@ -52,7 +52,7 @@
   </tbody>
   </table>
   <div class="panel-body">
-  <p class="helptext">{% trans "The user will be removed from the project once all user permissions are removed." %}</p>
+  <p class="helptext">{% trans "The user will be removed from the project once all user permissions are." %}</p>
```
nijel: Are you sure with this one? It sounds weird to me.

comradekingu: I think if I have to argue that I loose.

comradekingu:
```suggestion
  <p class="helptext">{% trans "How about "Once all its permissions are removed, the user will be removed from the project"?." %}</p>
```

nijel:

Issue #3200, opened 2019-11-13
Configurable SQL backups


surli:
**Is your feature request related to a problem? Please describe.**
We used to perform automatically backups of our database using pg_dump and gzip. Now Weblate perform backups automatically but they're not gzipped. It might be on purpose since Borg is doing diffs when saving, but when no Borg instance is configured it's a bit annoying. 

See the difference:
```
$ du -sh /var/weblate/backup/weblate_20191112.sql.gz 
2.3G	/var/weblate/backup/weblate_20191112.sql.gz
$ du -sh /var/weblate/data/backups/database.sql 
6.9G	/var/weblate/data/backups/database.sql
```

**Describe the solution you'd like**
When no borg instance is configured, at least the database backup should be compressed.

**Describe alternatives you've considered**
For now I'll disable the backups from Weblate and keep my old script to perform backups.

nijel:
The file is uncompressed to be easier to do incremental backups. With compressed files, it won't work regardless whether you use borg or some other tool.

mayeulk:
In my use case, I'd prefer to have compressed files than incremental backups.
If some users prefer the later, maybe letting them choose would be great.

nijel:
Patches are welcome to make that configurable, related code is here:

https://github.com/WeblateOrg/weblate/blob/88c30780f06b900dd55a73d37c6f47f950773cdf/weblate/utils/tasks.py#L66-L86

nijel:
Based on https://github.com/WeblateOrg/weblate/issues/3255, there should be probably three options:

* uncompressed backups (default)
* compressed backups
* none backups

PR #3271, opened 2019-12-04
feat: Add some new options for backups.

surli:
This PR aims at fixing #3200. 
Specifically it does the following: 
  * Add new options to skip backups or to compress DB backups
  * Improve documentation

### DIFF
--- a/weblate/settings_example.py
+++ b/weblate/settings_example.py
@@ -839,6 +839,9 @@ CELERY_TASK_ROUTES = {
     "weblate.wladmin.tasks.backup_service": {"queue": "backup"},
 }
 
+# Enable plain database backups
+DATABASE_BACKUP = "plain"
+
 # Enable auto updating
 AUTO_UPDATE = False
 
--- a/weblate/utils/tasks.py
+++ b/weblate/utils/tasks.py
@@ -71,8 +71,6 @@ def database_backup():
         return
     cmd = [
         "pg_dump",
-        "--file",
-        data_dir("backups", "database.sql"),
         "--dbname",
         database["NAME"],
     ]
@@ -82,6 +80,11 @@ def database_backup():
         cmd += ["--port", database["PORT"]]
     if database["USER"]:
         cmd += ["--username", database["USER"]]
+    if settings.DATABASE_BACKUP == "compressed":
+        cmd += ["--file", data_dir("backups", "database.sql.gz")]
+        cmd += ["--compress", "6"]
+    else:
+        cmd += ["--file", data_dir("backups", "database.sql")]
 
     subprocess.check_call(cmd, env=get_clean_env({"PGPASSWORD": database["PASSWORD"]}))
 
@@ -91,7 +94,8 @@ def setup_periodic_tasks(sender, **kwargs):
     sender.add_periodic_task(
         crontab(hour=1, minute=0), settings_backup.s(), name="settings-backup"
     )
-    sender.add_periodic_task(
-        crontab(hour=1, minute=30), database_backup.s(), name="database-backup"
-    )
+    if settings.DATABASE_BACKUP != "none":
+        sender.add_periodic_task(
+            crontab(hour=1, minute=30), database_backup.s(), name="database-backup"
+        )
     sender.add_periodic_task(60, heartbeat.s(), name="heartbeat")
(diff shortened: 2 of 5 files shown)


surli:
Regarding the testcase I checked the existing tests for backups, but AFAICS there's no test that covers the `setup_periodic_tasks`, and I'm really too rusty in Python to know how to test it properly right now :)

weblate/utils/tasks.py
```diff
@@ -82,16 +80,22 @@ def database_backup():
         cmd += ["--port", database["PORT"]]
     if database["USER"]:
         cmd += ["--username", database["USER"]]
+    if settings.BACKUP_COMPRESSED:
+        cmd += ["--file", data_dir("backups", "database.sql.gz")]
+        cmd += ["--format", "custom"]
```
nijel: Naming the file `sql.gz` with custom format is confusing.

weblate/settings_example.py
```diff
@@ -839,6 +839,11 @@
     "weblate.wladmin.tasks.backup_service": {"queue": "backup"},
 }
 
+# Enable to skip backups
+
+BACKUP_SKIPPED = False
+BACKUP_COMPRESSED = False
```
nijel: * I think it should include what kind of backup is this about, for example `DATABASE_BACKUP`.
* Maybe having just one option would make it easier (`DATABASE_BACKUP` with options (`none`, `plain`, `custom`).
* Please define new setting in [related AppConf](https://github.com/WeblateOrg/weblate/blob/664459eed0d7404de7e83e871b98e4305fdeb10d/weblate/utils/models.py#L32-L66) so that the setting is always present and the code doesn't have to deal with situation it is not present.

weblate/utils/tasks.py
```diff
@@ -82,16 +80,22 @@ def database_backup():
...
 
 @app.on_after_finalize.connect
 def setup_periodic_tasks(sender, **kwargs):
-    sender.add_periodic_task(
-        crontab(hour=1, minute=0), settings_backup.s(), name="settings-backup"
-    )
-    sender.add_periodic_task(
-        crontab(hour=1, minute=30), database_backup.s(), name="database-backup"
-    )
+    if not settings.BACKUP_SKIPPED:
```
nijel: Why disabling settings backup as well?

weblate/settings_example.py
surli: > Maybe having just one option would make it easier (DATABASE_BACKUP with options (none, plain, custom).

I hesitated but ok, but I'd prefer the values `none`, `plain`, `compressed` then I guess.

> Please define new setting in related AppConf so that the setting is always present and the code doesn't have to deal with situation it is not present.

Thanks, I was wondering about something like this.

weblate/utils/tasks.py
surli: > Naming the file sql.gz with custom format is confusing.

Hmm I guess I used the wrong set of options, should have used plain format with compress options since, according to the doc:
> For plain text output, setting a nonzero compression level causes the entire output file to be compressed, as though it had been fed through gzip

Or do you suggest something else?

weblate/utils/tasks.py
surli: Because at first I was doing the skip for the entire backup, but reading back your comments and the issue it should be indeed only for the DB, I'll change that.

weblate/settings_example.py
nijel: My idea was to stay consistent with pg_dump options and to be able to use the option directly there.

surli:
Looks like the travis failure are not related at all to my changes: 
```
apt-get install failed

The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install git-svn graphviz libcairo-dev libacl1-dev gir1.2-pango-1.0 libgirepository1.0-dev libenchant1c2a g++ chromium-browser" failed and exited with 100 during .
```

nijel:
I've triggered rebuild. That's problem with Travis, that it often fails on unrelated issues. I'm thingking moving to Drone.io (it's currently running some jobs as proof of concept) or GitHub Actions.

surli:
Apparently it's still failing on the same Travis CI issue, but only on one job now.

PR #3318, opened 2019-12-18
Use WEBLATE_GITLAB_USERNAME environment variable

eborned:
The WEBLATE_GITLAB_USERNAME environment variable is specified in the documentation. But it is never assigned.

### DIFF
--- a/weblate/settings_docker.py
+++ b/weblate/settings_docker.py
@@ -203,6 +203,10 @@ TEMPLATES = [
 # Please see the documentation for more details.
 GITHUB_USERNAME = os.environ.get("WEBLATE_GITHUB_USERNAME", None)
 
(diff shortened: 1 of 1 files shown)


weblate/settings_docker.py
```diff
@@ -203,6 +203,10 @@
 # Please see the documentation for more details.
 GITHUB_USERNAME = os.environ.get("WEBLATE_GITHUB_USERNAME", None)
 
+# GitHub username for sending pull requests.
```
nijel: Can you please adjust the comment as well?

nijel:
Good point, the other part which is missing is lab being installed in the Docker image, see https://github.com/WeblateOrg/docker/issues/467

Issue #3322, opened 2019-12-19
inconsistencies between store units with java properties and comments


surli:
**Describe the bug**
When translating a java properties containing headers, the internal units are inconsistent: the unit for the headers actually contains the data of the second unit. 
This has been reproduced on both Weblate 3.9.1 and current master.

**To Reproduce**
1. Import the [xwiki-commons-job.zip](https://github.com/WeblateOrg/weblate/files/3982785/xwiki-commons-job.zip) as a new component with default proposed configuration
2. Put a breakpoint on https://github.com/WeblateOrg/weblate/blob/master/weblate/trans/models/translation.py#L614
3. Make a new translation and commit it
4. On the breakpoint, check `self.store.all_units[0].unit` and `self.store.all_units[1].unit`: they shouldn't be the same, but they are.

![Capture d’écran_2019-12-19_10-38-06](https://user-images.githubusercontent.com/1478232/71162755-0bdb4600-224c-11ea-9029-414c86c6e04a.png)

I tried to created a dedicated unit test for it, but AFAICS I don't perform the right actions: see https://github.com/surli/weblate/commit/9ac1a452222889cab672c9427e01f9f54bdda985 (the data file is added on the same branch)

surli:
I have not been able to properly locate where the bug is: the problem is accessing the `all_units` property in the debugger trigger the code to gather the units and the breakpoints called from internal methods of the debugger are skipped by default. I'll continue to investigate on it.

nijel:
Maybe it's related to monolingual units? Try adding the same file as template in https://github.com/surli/weblate/commit/9ac1a452222889cab672c9427e01f9f54bdda985#diff-7528f5a66693be334235bd63f6b6519dR382

surli:
You mean to call `storage = self.FORMAT(self.FILE, is_template=True)`?

surli:
I still don't reproduce with that, but note that my test already fail on https://github.com/surli/weblate/commit/9ac1a452222889cab672c9427e01f9f54bdda985#diff-7528f5a66693be334235bd63f6b6519dR390 Looks like the comments are not taken into account for unit equality which might be part of the issue I guess.

nijel:
Sorry for not being clear: `storage = self.FORMAT(self.FILE, template=self.FORMAT(self.FILE), is_template=True)` (the is_template flag is not really needed for this, but just to be consistent)

surli:
[...]

PR #3323, opened 2019-12-19
Proposal for fixing #3322

surli:

### DIFF
--- a/weblate/formats/tests/test_formats.py
+++ b/weblate/formats/tests/test_formats.py
@@ -83,6 +83,7 @@ TEST_HE_CLDR = get_test_file('he-cldr.po')
 TEST_HE_CUSTOM = get_test_file('he-custom.po')
 TEST_HE_SIMPLE = get_test_file('he-simple.po')
 TEST_HE_THREE = get_test_file('he-three.po')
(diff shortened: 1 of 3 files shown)


surli:
@nijel Codefactor is complaining about the new tests since I duplicate some codes, but I didn't want to change the existing test resource. Let me know if I need to improve this.

surli:
Hi @nijel (and happy new year :) ) any feedback on that PR?

nijel:
I still think that this is not good place to address this. The root cause is IMHO API violation in translate-toolkit that it parses those as separate units. It is related to https://github.com/translate/translate/issues/3928. I think the parser should only emit actual strings and include all comments up to the string with it. This is how other formats behave.

surli:
> The root cause is IMHO API violation in translate-toolkit that it parses those as separate units. It is related to translate/translate#3928.

OK, on my side it's a blocker since I need this to implement properly my new formats with all the features we currently have. 
I'm ok to help on fixing this on translate-toolkit side but might need some guidance on what's the expected result then :) 
And for info does translate-toolkit has a specific release process?

nijel:
The fix for translate-toolkit should be here: https://github.com/translate/translate/pull/3931

There is no specific release process on translate-toolkit, usually releases happen when somebody (often it's me) asks the maintainer.

Issue #3253, opened 2019-11-28
Can't create new translation using API


loktionovam:
**Describe the bug**
I try to create translations via API, but can't create empty translation using API, so upload translation files leads to an error.

**To Reproduce**
1. Try to upload file:
  ```python
  print("Upload test translation file")
  files = {
      'file': open('/tmp/test.ts', 'rb')
  }
  weblate.request(path="translations/test/test/ru/file/", method="post", files=files)
  ```
2. See error:
  ```python
  requests.exceptions.HTTPError: 404 Client Error: Not Found for url: http://localhost/api/translations/test/test/ru/file/
  wlc.WeblateException: Object not found on the server (maybe operation is not supported on the server)
  ```

**Expected behavior**
Create new translation using API and then upload file:

```python
  lang = {
        'lang': 'ru'
  }
  weblate.request(path="translations/test/test/", method="post", params=lang)

  files = {
      'file': open('/tmp/test.ts', 'rb')
  }
  weblate.request(path="translations/test/test/ru/file/", method="post", files=files)
```

**Server configuration and status**
Using docker-compose:

```
docker exec weblate_weblate_1  weblate list_versions
 * Weblate 3.9.1
 * Python 3.7.3
 * Django 2.2.7
 * Celery 4.3.0
 * celery-batches 0.2
 * six 1.12.0
 * social-auth-core 3.2.0
 * social-auth-app-django 3.1.0
 * django-appconf 1.0.3
 * translate-toolkit 2.4.0
 * translation-finder 1.7
 * Whoosh 2.7.4
 * defusedxml 0.6.0
 * Git 2.20.1
 * Pillow 6.2.1
 * python-dateutil 2.8.0
 * lxml 4.3.2
 * django-crispy-forms 1.7.2
 * django_compressor 2.3
 * djangorestframework 3.10.3
 * user-agents 2.0
 * jellyfish 0.7.2
 * diff-match-patch 20181111
 * ruamel.yaml 0.16.5
 * tesserocr 2.5.0
 * Mercurial 4.8.2
 * git-svn 2.20.1
 * Database backends: django.db.backends.postgresql
 * Cache backends: default:RedisCache, avatar:FileBasedCache
 * Email setup: django.core.mail.backends.smtp.EmailBackend: 127.0.0.1
 * Celery: redis://cache:6379/1, redis://cache:6379/1, regular
 * Platform: Linux 5.0.0-36-generic (x86_64)


[...]

PR #3336, opened 2019-12-24
API: Creates new translation in the given component.

loktionovam:
Fix creation translation via REST API

Issue #3253

Signed-off-by: Aleksandr Loktionov <loktionovam@gmail.com>

### DIFF
--- a/weblate/api/views.py
+++ b/weblate/api/views.py
@@ -404,9 +404,36 @@ class ComponentViewSet(MultipleFieldMixin, WeblateViewSet, DestroyModelMixin):
 
         return self.download_file(obj.get_new_base_filename(), 'application/binary')
 
-    @action(detail=True, methods=['get'])
+    @action(detail=True, methods=['get', 'post'])
     def translations(self, request, **kwargs):
         obj = self.get_object()
+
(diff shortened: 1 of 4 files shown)


weblate/api/views.py
```diff
@@ -404,10 +404,36 @@ def new_template(self, request, **kwargs):
...
+
+            if 'language_code' not in request.data:
+                raise ParseError("Missing 'language_code' parameter")
+
+            language_code = request.data['language_code']
+
+            if not Language.objects.filter(code__in=(language_code,)):
+                raise Http404("No language code '%s' found!" % language_code)
+
+            language = Language.objects.get(code=language_code)
```
nijel: I'd rather use try/except here instead of checking for existence and then retrieving the object again from the database.

```suggestion
            try:
                language = Language.objects.get(code=language_code)
            except Language.DoesNotExist:
                raise Http404("No language code '%s' found!" % language_code)
```

loktionovam: Done. I rebased onto non-broken upstream master and use `try/except` here.

nijel:

PR #3351, opened 2020-01-03
New Functionality to allow for changing of application theme

obuya:
Application Theme can be modified by visiting `[application_home]/manage/tools/` url, and then selecting one of the themes in `Change Theme` form.

### DIFF
--- a/weblate/accounts/forms.py
+++ b/weblate/accounts/forms.py
@@ -789,3 +789,10 @@ class NotificationForm(forms.Form):
         ).exclude(
             pk__in=handled
         ).delete()
+
+class ThemeField(forms.Form):
+    """Theme field."""
+
+    def __init__(self, *args, **kwargs):
+        super(ThemeField, self).__init__(*args, **kwargs)
+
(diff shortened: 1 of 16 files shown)


nijel:
I see you closed the PR, but still few remarks:

* Using `global` variable for storing configuration is wrong. It doesn't persist server restart and with multiple workers each worker will have different configuration. This looks more like user setting which should go to the user profile.

* Not sure how you created the changed CSS files, but we need way to keep them in sync with master ones so this needs to be automated to allow automated updates once the master files change. Or  add extra CSS file changing the colors on top of current ones instead of copying all of them and doing color changes.

obuya:
> I see you closed the PR, but still few remarks:
> 
>     * Using `global` variable for storing configuration is wrong. It doesn't persist server restart and with multiple workers each worker will have different configuration. This looks more like user setting which should go to the user profile.
> 
>     * Not sure how you created the changed CSS files, but we need way to keep them in sync with master ones so this needs to be automated to allow automated updates once the master files change. Or  add extra CSS file changing the colors on top of current ones instead of copying all of them and doing color changes.

Thanks for your response, 
- I noticed the inconsistency in persistence, and thus was compelled to close the PR till I got it sorted, together with the PEP whistles. Was of the idea of placing the configuration in the admin settings, so as to have a site-wide effect across all users, something consistent with the corporate/organisation brand.

- The CSS files were created by modifying select values in the original values. I believe keeping them in sync with master won't be that difficult, just a matter of adding some configuration instructing the application to change the particular values to different hex values, or an equivalent

PR #3355, opened 2020-01-07
Remove trailing slash of repo URL if exists

aparcar:
Webhook detection (at least for GitHub) is broken if the repository URL contains a trailing slash.

This patch uses rstrip to remove a trailing slash if it exists. The validation logic isn't changed by this.

Signed-off-by: Paul Spooren <mail@aparcar.org>

### DIFF
--- a/weblate/trans/models/component.py
+++ b/weblate/trans/models/component.py
@@ -1451,7 +1451,7 @@ class Component(models.Model, URLMixin, PathMixin):
                 raise ValidationError(
                     {setting: _("Option is not available for linked repositories.")}
                 )
-        self.linked_component = Component.objects.get_linked(self.repo)
+        self.linked_component = Component.objects.get_linked(self.repo).rstrip("/")
 
     def clean_lang_codes(self, matches):
         """Validate that there are no double language codes."""


weblate/trans/models/component.py
```diff
@@ -1451,7 +1451,7 @@ def clean_repo_link(self):
                 raise ValidationError(
                     {setting: _("Option is not available for linked repositories.")}
                 )
-        self.linked_component = Component.objects.get_linked(self.repo)
+        self.linked_component = Component.objects.get_linked(self.repo).rstrip("/")
```
nijel: This looks wrong. The `get_linked` returns Component instance.

nijel:
Generally I don't think that unconditionally stripping trailing slash is a good idea. It might break some setups which require it. Better to fix the hooks code to accept such URLs as well. What exactly was the URL you used before?

aparcar:
Sorry for the slow response. I initially set the *Source code repository* to `https://github.com/openwrt/luci/` with a slash at the end, as GitHub provides it in the browser URL bar. However it has to be `https://github.com/openwrt/luci` to work, without the trailing slash.

nijel:
This should be better addressed by 16b249ece1b6db09b8e93c81cd9b09f136210ff6

PR #3361, opened 2020-01-09
Changed deepl api version to v2.

timesqueezer:
This simply changes the used deepl API version to v2. For the simple translation request that weblate performs the request and response format stayed exactly the same so no further changes are required.

### DIFF
--- a/weblate/machinery/deepl.py
+++ b/weblate/machinery/deepl.py
@@ -24,7 +24,7 @@ from django.conf import settings
 
 from weblate.machinery.base import MachineTranslation, MissingConfiguration
 
-DEEPL_API = 'https://api.deepl.com/v1/translate'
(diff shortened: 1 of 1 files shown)


nijel:
According to DeepL, Weblate is supposed to use v1 API, see https://github.com/WeblateOrg/weblate/pull/3133#issuecomment-552066743.

timesqueezer:
Oh, thanks. I didn't see the closed pull request and there are no open deepl issues. Would it be an option to have deepl v2 as a separate machine translation module or maybe as an optional setting? It would be great to get this into upstream and I would be willing to help with the implementation.

nijel:
It could be setting to make API version configurable. However, I'm not sure it's a good idea to include code that violates their terms of service.

nijel:

PR #3364, opened 2020-01-10
Weblate fix glossaries

Burzo:
Signed-off-by: Burzo <miha.drofenik92@gmail.com>

Hello there, 

I saw this project on up-for-grabs and saw that you needed help. I really want to do some good and help out and learn stuff on the way :)

What I did:

1. Added the project to the show_dictionaries render() method, so that dictionaries.html also gets the relevant information.

2. Copied the HTML from the project.html to dictionaries.html and fixed all the relevant URLs.

There is still the issue of the hard coded "active" class in the project.html. If you for example click Search or Languages or Info while on Glossaries, the Components will flash for a short time before "active" is applied to the correct element.

Any feedback is appreciated. Thank you.


weblate/templates/dictionaries.html
```diff
@@ -1,13 +1,108 @@
...
+        <li><a href="https://www.facebook.com/sharer.php?u={{ share_url }}">{% trans "Share on Facebook" %}</a></li>
+        <li><a href="https://twitter.com/share?text={% blocktrans %}Translate {{ object }} using %23Weblate at {{ share_url }}!{% endblocktrans %}">{% trans "Tweet this translation" %}</a></li>
+        <li><a href="{% url 'engage' project=object.slug %}">{% trans "Engage page" %}</a></li>
+        <li><a href="{{ widgets_url }}">{% trans "Status widgets" %}</a></li>
+  {% endwith %}
+        </ul>
+    </li>
+  {% endif %}
+</ul>
+
```
nijel: I'd really prefer if the menu would be in separate template and included in both pages. Copying it will make it easy to miss updates in one of these.

Burzo: Sure thing, I'm on it :)

weblate/templates/snippets/project-nav.html
```diff
@@ -0,0 +1,143 @@
+{% load i18n %}
+{% load translations %}
+
+{% url 'show_dictionaries' project=object.slug as show_dictionaries %}
+
+<ul class="nav nav-pills">
+    {% if request.path == show_dictionaries %}
```
nijel: It would be probably better to compare against project URL here to make the code prepared for inclusion in another pages. I wonder if we could make it without these conditions...

Burzo: Not sure how to tackle this one without major changes, since a different view is taking care of the `/dictionaries/project/` urls. I could try to improve the templates perhaps, but the `data-toggle="tab"` is a pain.

Any suggestions on how to proceed?

nijel: I think what you did is good enough for now. In case you have further ideas to improve, please do that :-).

nijel:

PR #3386, opened 2020-01-16
Allow extending instead of replacing LOGIN_REQUIRED_URLS_EXCEPTIONS

Kingdutch:
Quite a few of the URLs included in `LOGIN_REQUIRED_URLS_EXCEPTIONS` by default are needed for Weblate to function properly. When using the Weblate docker container, if you need to add a single path as exception then you suddenly need to replace them all, which may cause issues if they change during an update.

By allowing the extension next to the replacement it's trivial to add a single URL while still allowing Weblate to work as normal.

### DIFF
--- a/weblate/settings_docker.py
+++ b/weblate/settings_docker.py
@@ -999,6 +999,7 @@ if get_env_bool("WEBLATE_REQUIRE_LOGIN", False):
             rf"{URL_PREFIX}/legal/(.*)$",  # Optional for legal app
         ),
     )
+    LOGIN_REQUIRE_URLS_EXCEPTIONS.extend(get_env_list("WEBLATE_EXTEND_LOGIN_REQUIRED_URLS_EXCEPTIONS"))
 
 # Email server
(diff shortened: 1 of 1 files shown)


weblate/settings_docker.py
```diff
@@ -999,6 +999,7 @@
             rf"{URL_PREFIX}/legal/(.*)$",  # Optional for legal app
         ),
     )
+    LOGIN_REQUIRE_URLS_EXCEPTIONS.extend(get_env_list("WEBLATE_EXTEND_LOGIN_REQUIRED_URLS_EXCEPTIONS"))
```
nijel: Maybe better to use existing code to modify lists?

```suggestion
    modify_env_list(LOGIN_REQUIRE_URLS_EXCEPTIONS, "LOGIN_REQUIRE_URLS_EXCEPTIONS")
```

Also see https://docs.weblate.org/en/latest/admin/install/docker.html#changing-enabled-apps-checks-addons-or-autofixes

PR #3402, opened 2020-01-19
Added per component individual checks

Burzo:
Hey,

I just added another query parameter to the component.html template, that then filters the component__slug field on the existing queryset.

Let me know if it's alright.

---

Signed-off-by: Burzo <miha.drofenik92@gmail.com>

### DIFF
--- a/weblate/checks/tests/test_views.py
+++ b/weblate/checks/tests/test_views.py
@@ -29,30 +29,24 @@ from weblate.trans.tests.test_views import ViewTestCase
 
 class ChecksViewTest(ViewTestCase):
     """Testing of check views."""
-
-    def setUp(self):
-        super(ViewTestCase, self).setUp()
-        self.component = self.create_component()
-        self.project = self.component.project
-
     def test_browse(self):
         response = self.client.get(reverse('checks'))
-        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, '/same/')
 
         response = self.client.get(reverse('checks'), {'language': 'de'})
-        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, '/same/')
 
         response = self.client.get(
             reverse('checks'),
             {'project': self.project.slug}
         )
-        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, '/same/')
 
         response = self.client.get(
             reverse('checks'),
             {'project': self.project.slug, 'component': self.component.slug}
         )
-        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, '/same/')
 
     def test_check(self):
         response = self.client.get(
--- a/weblate/checks/tests/test_views.py
+++ b/weblate/checks/tests/test_views.py
@@ -29,24 +29,30 @@ from weblate.trans.tests.test_views import ViewTestCase
 
 class ChecksViewTest(ViewTestCase):
     """Testing of check views."""
+
+    def setUp(self):
+        super(ViewTestCase, self).setUp()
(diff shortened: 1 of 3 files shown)


nijel:
Thanks, the patch looks good. Can you please also add tests covering this? It's in `weblate/checks/tests/test_views.py`.

Burzo:
I wrote the tests, but they're behaving strangely. I run them by connecting to the docker container and then running `weblate test weblate.checks.tests.test_views` (also tried  `weblate test weblate.checks`) but even before writing any code the tests already there fail with the same error:
```
ERROR: test_component (weblate.checks.tests.test_views.ChecksViewTest)
  File "/usr/local/lib/python3.7/dist-packages/weblate/trans/tests/test_views.py", line 122, in setUp
    self.translation_url = self.get_translation().get_absolute_url()
  File "/usr/local/lib/python3.7/dist-packages/weblate/trans/tests/test_views.py", line 151, in get_translation
    return self.component.translation_set.get(language_code=language)
  File "/usr/local/lib/python3.7/dist-packages/django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python3.7/dist-packages/django/db/models/query.py", line 417, in get
    self.model._meta.object_name
weblate.trans.models.translation.Translation.DoesNotExist: Translation matching query does not exist.
```
I assume it's connected to the setUp() of ViewTestCase, where
```
    def get_translation(self, language='cs'):
        return self.component.translation_set.get(language_code=language)
```
can't find the translation.

Also, these assertions in other test cases don't make sense: `self.assertContains(response, '/same/')`.

If I'm missing something or if I skipped some setup method I should've run before, then I'm sorry for the complication!

For now I rewrote the relevant test method for this use case to only check the return status codes and added the setUp method to populate the needed fields so that the tests even run and don't throw an error.

weblate/checks/tests/test_views.py
```diff
@@ -29,18 +29,30 @@
 
 class ChecksViewTest(ViewTestCase):
     """Testing of check views."""
+
+    def setUp(self):
+        super(ViewTestCase, self).setUp()
```
nijel: Skipping one class in inheritance is not a good idea.

weblate/checks/tests/test_views.py
```diff
@@ -29,18 +29,30 @@
...
+        self.component = self.create_component()
```
nijel: The default setUp method already does this.

weblate/checks/tests/test_views.py
```diff
@@ -29,18 +29,30 @@
...
+        self.project = self.component.project
+
     def test_browse(self):
         response = self.client.get(reverse('checks'))
-        self.assertContains(response, '/same/')
+        self.assertEqual(response.status_code, 200)
```
nijel: The original tests was asserting that some check is present, why removing that?

nijel:
Sorry, I'm not sure how to effectively run test in Docker, I don't use that.

Burzo:
Sorry for the complication, managed to get the tests running. Reverted and tested my previous iteration, it went through now.

Thank a lot for the help and the tips :)

nijel:

Did you need some special steps to run tests in the Docker container? If so, it would be great if you could document these in https://github.com/WeblateOrg/weblate/blob/master/docs/contributing.rst

Burzo:
Honestly, I just deleted all the images and rebuilt everything and it went through. Not sure what was wrong. I did add a warning to the docs though.

PR #3403, opened 2020-01-20
Newer changelogs reworked

comradekingu:
As per
https://en.wikipedia.org/wiki/Markdown
https://borgbackup.readthedocs.io/en/stable
https://en.wikipedia.org/wiki/Webhook

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -68,7 +68,7 @@ Released on December 20th 2019.
 * Added support for automatic translation strings marked for editing.
 * Avoid sending duplicate notifications for linked component alerts.
 * Improve default merge request message.
-* Better indicate string state in zen mode.
+* Better indicate string state in Zen mode.
 * Added support for more languages in Yandex Translate.
 * Improved look of notification e-mails.
 * Provide choice for translation license.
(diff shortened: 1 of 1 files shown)


docs/changes.rst
```diff
@@ -54,21 +54,21 @@ Released on December 20th 2019.
...
 * Show context for nearby strings.
 * Added support for notifications on repository operations.
 * Improved translation listings.
 * Extended search capabilities.
-* Added support for automatic translation strings marked for edit.
+* Added support for automatic translation strings marked for editing.
 * Avoid sending duplicate notifications for linked component alerts.
 * Improve default merge request message.
-* Better indicate string state in Zen mode.
+* Better indicate string state in zen mode.
```
nijel: Not sure what is better here. See also https://hosted.weblate.org/translate/weblate/master/en/?checksum=7867c65b412a1631#comments

docs/changes.rst
```diff
@@ -150,7 +150,7 @@ Released on August 15th 2019.
 * Avoid sending multiple notifications for single event.
 * Add support for filtering changes.
 * Extended predefined periods for reporting.
-* Added support for Azure Repos webhooks.
+* Added support for Azure repository webhooks.
```
nijel: The service is actually called "Azure Repos", see https://azure.microsoft.com/cs-cz/services/devops/repos/

docs/changes.rst
```diff
@@ -119,7 +119,7 @@ Released on October 15th 2019.
 * Display addon parameters on overview.
 * Sentry is now supported through modern Sentry SDK instead of Raven.
 * Changed example settings to be better fit for production environment.
-* Added automated backups using Borg backup.
+* Added automated backups using BorgBackup.
```
nijel: Good point, fixed that in rest of the docs in 8d35a1ff489c358f7f2772f5aba4ee474ed3ff39.

PR #3415, opened 2020-01-23
Improve matomo support

minusf:
- PIWIK_URL was misnamed as it was expecting just a domain name really.
  Change this to use a full URL with trailing slash.
- Use latest tracking script from latest matomo, and use the URL
  correctly. Stop using `//`, https is everywhere.
- Update documentation to mention that Matomo Tag manager is not
  supported with this integration.  A more complicated custom solution
  will be put directly in the template anyway probably.
- Do not use a `<p>` for an invisible image.


nijel:
You're right that it was misnamed. However doing this change will break existing setup. While doing incompatible changes, it might be better to also rename it to MATOMO_URL.

minusf:
can you elaborate a bit please?  do you want to keep the old setting with the old tempalte bit and add new `MATOMO_*` settings with a new template bit?  this way there could be a deprecation warning too...

nijel:
My original intention was to drop PIWIK_* and introduce MATOMO_* with new semantics and document this in the upgrade notes. Having depreciation warning is nice, but IMHO most people won't find it :-).

minusf:
here is an updated diff to rename thing to MATOMO_

nijel:

PR #3432, opened 2020-01-28
API: Allow to use vcs push url during component creation via API.

loktionovam:
**Problem**: when a new translation component is created via API `Repository push URL` field became empty (pushing is turned off if this field is empty), so translated strings don't send back to the main VCS.

**Solution**: allow to setup a VCS push URL via API request during component creation.

### DIFF
--- a/weblate/api/serializers.py
+++ b/weblate/api/serializers.py
@@ -141,6 +141,11 @@ class RepoURLField(serializers.CharField):
         return instance.get_repo_url()
 
 
+class RepoPushField(serializers.CharField):
+    def get_attribute(self, instance):
+        return instance.get_push_url()
+
+
 class ComponentSerializer(RemovableSerializer):
     web_url = AbsoluteURLField(source='get_absolute_url', read_only=True)
     project = ProjectSerializer(read_only=True)
@@ -168,6 +173,8 @@ class ComponentSerializer(RemovableSerializer):
 
     repo = RepoURLField()
 
+    push = RepoPushField()
+
     serializer_url_field = MultiFieldHyperlinkedIdentityField
 
     class Meta(object):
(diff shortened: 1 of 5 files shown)


weblate/api/tests.py
```diff
@@ -292,9 +292,18 @@ def test_create_component(self):
                 'repo': self.format_local_path(self.git_repo_path),
                 'filemask': 'po/*.po',
                 'file_format': 'po',
+                'push': 'git@github.com:example/push.git',
             },
         )
         self.assertEqual(Component.objects.count(), 2)
+        self.assertEqual(
+            Component.objects.all()[1].push,
```
nijel: I don't think this is safe, no ordering is guaranteed here.

weblate/api/serializers.py
```diff
@@ -177,7 +177,7 @@ class Meta(object):
             'branch', 'filemask', 'template', 'new_base', 'file_format',
             'license', 'license_url', 'web_url', 'url',
             'repository_url', 'translations_url', 'statistics_url',
-            'lock_url', 'changes_list_url', 'new_lang',
+            'lock_url', 'changes_list_url', 'new_lang', 'push',
```
nijel: This needs more care to be consistent with VCS view permissions:

* Add wiping to `to_representation` below
* Apply credentials cleanup in case it is enabled (it is done in the `RepoURLField`)

weblate/api/tests.py
loktionovam: Done. I replaced this with more predictable behavior
```python
Component.objects.get(
                slug='api-project',
                project__slug='test'
            ).push
```

weblate/api/serializers.py
loktionovam: - Done. I add wiping push link to `to_representation`
- Done. I add credentials cleanup and rewrite `test_create_component` to check the response does not contain secret data in 'push' field

loktionovam:
@nijel Thanks for the review. I guess I've resolved all your comments. Is the patch looks good for you now?

nijel:

PR #3456, opened 2020-02-09
Typo

tdelmas:


tdelmas:
I'm not sure I understand why the coverage will decrease according to @codecov ... I think that bot have a bug, the only difference between master and this PR is **one letter**, in the documentation.

nijel:
The `0.01%` difference is probably just rounding error...

nijel:

PR #3463, opened 2020-02-12
Fix icon file handling.

minusf:
There are 2 issues with the current code:

1. `DATA_DIR` being ignored for icons (use `settings.STATIC_ROOT`)

2. in case of an exception `icon()` was called with an empty `name`. I could not track where this happens yet, but regardless file reading should handle errors (python2 might need `IOError`, but sorry I don't use python2 anymore, cannot test).

For example:
```
...
django.core.exceptions.DisallowedHost: Invalid HTTP_HOST header: 'xxx.xxx.xxx.xxx'. You may need to add 'xxx.xxx.xxx.xxx' to ALLOWED_HOSTS.

During handling of the above exception, another exception occurred:

...
  File "/export/apps/weblate/venv/lib64/python3.6/site-packages/weblate/utils/templatetags/icons.py", line 42, in icon
    with open(os.path.join(PATH, name), "r") as handle:
IsADirectoryError: [Errno 21] Is a directory: '/export/apps/weblate/venv/lib64/python3.6/site-packages/weblate/static/icons/'

During handling of the above exception, another exception occurred:

...
```

### DIFF
--- a/weblate/utils/templatetags/icons.py
+++ b/weblate/utils/templatetags/icons.py
@@ -23,13 +23,13 @@ from __future__ import unicode_literals
 import os
 
 from django import template
+from django.conf import settings
 from django.utils.safestring import mark_safe
 
-register = template.Library()
+from weblate.logger import LOGGER
 
-PATH = os.path.join(
-    os.path.dirname(os.path.dirname(os.path.dirname(__file__))), "static", "icons"
-)
+
+register = template.Library()
 
 CACHE = {}
 
@@ -38,10 +38,21 @@ SPIN = '<span class="icon-spin" {} {}>{}</span>'
 
 @register.simple_tag()
 def icon(name, fallback=None):
-    name = name or fallback
+    if fallback is not None:
+        name = fallback
+
+    if not name:
+        return ""
+
     if name not in CACHE:
-        with open(os.path.join(PATH, name), "r") as handle:
-            CACHE[name] = mark_safe(handle.read())
+        icon_file = os.path.join(settings.STATIC_ROOT, "icons", name)
+        try:
+            with open(icon_file, "r") as handle:
+                CACHE[name] = mark_safe(handle.read())
+        except OSError as e:
+            LOGGER.error(e)
+            return ""
+
     return CACHE[name]
 
 
--- a/weblate/utils/templatetags/icons.py
+++ b/weblate/utils/templatetags/icons.py
@@ -27,6 +27,7 @@ from django.conf import settings
 from django.utils.safestring import mark_safe
 
 from weblate.logger import LOGGER
+from weblate.utils.errors import report_error
 
 register = template.Library()
 
@@ -41,6 +42,7 @@ def icon(name, fallback=None):
         name = fallback
 
     if not name:
+        LOGGER.error("Empty icon name")
         return ""
 
     if name not in CACHE:
@@ -48,8 +50,8 @@ def icon(name, fallback=None):
         try:
             with open(icon_file, "r") as handle:
                 CACHE[name] = mark_safe(handle.read())
-        except OSError as e:
-            LOGGER.error(str(e))
+        except OSError as error:
+            report_error(error, prefix="Failed to load icon")
             return ""
 
     return CACHE[name]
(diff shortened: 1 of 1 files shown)


nijel:
The error you saw was most likely fixed by c6b9d20806c10edb99ad03d68ac64a9c0c3d2229.

I don't see much reasons to use STATIC_ROOT here - we don't need all collected files, we need just the icons and we know where they are (and from where they would be copied to STATIC_ROOT).

minusf:
i think this `G200` check is bogus.  It's totally fine to log parts of the exception (esp `OSError`).

weblate/utils/templatetags/icons.py
```diff
@@ -38,10 +37,21 @@
...
+
     if name not in CACHE:
-        with open(os.path.join(PATH, name), "r") as handle:
-            CACHE[name] = mark_safe(handle.read())
+        icon_file = os.path.join(settings.STATIC_ROOT, "icons", name)
+        try:
+            with open(icon_file, "r") as handle:
+                CACHE[name] = mark_safe(handle.read())
+        except OSError as e:
+            LOGGER.error(str(e))
```
nijel: Using `report_error` would be more appropriate here, that way the error would correctly show up in Sentry/Rollbar if these are configured. It takes care of the logging as well.

```suggestion
            report_error(error, prefix="Failed to load icon")
```

weblate/utils/templatetags/icons.py
```diff
@@ -38,10 +37,21 @@
 
 @register.simple_tag()
 def icon(name, fallback=None):
-    name = name or fallback
+    if fallback is not None:
+        name = fallback
+
+    if not name:
+        return ""
```
nijel: This is a bug, it should not fail silently.

minusf:
> I don't see much reasons to use STATIC_ROOT here - we don't need all collected files, we need just the icons and we know where they are (and from where they would be copied to STATIC_ROOT).

perhaps, but i prefer consistency.  `STATIC_ROOT` might be mounted somewhere else, maybe even in memory, etc.

what's the point using `collectstatic` at all then, your argument could be applied to all static resources that don't need overriding.

nijel:
> i think this G200 check is bogus. It's totally fine to log parts of the exception (esp OSError).

Logging *only* exception message is not really useful because it will give you no clue where the failure has happened. Besides that, Weblate already has handler for capturing such errors, see my previous comments.

nijel:
> what's the point using collectstatic at all then, your argument could be applied to all static resources that don't need overriding.

The main point is to have all static files in single location to be served by a web server. It collects them from all apps and puts them into single location, otherwise you would have to specify files from each app in the webserver configuration.

But you're right that with using `STATIC_DIR`, it makes it easier to override the icons by third party app.

minusf:
> The error you saw was most likely fixed by [c6b9d20](https://github.com/WeblateOrg/weblate/commit/c6b9d20806c10edb99ad03d68ac64a9c0c3d2229).

was `error.svg` added in a different commit?

nijel:
> was error.svg added in a different commit?

Yes, in d78d4b713392cf60d71464bd2b8872fec3198c26

minusf:
> > i think this G200 check is bogus. It's totally fine to log parts of the exception (esp OSError).
> 
> Logging _only_ exception message is not really useful because it will give you no clue where the failure has happened. Besides that, Weblate already has handler for capturing such errors, see my previous comments.

here is an example of what the log would have looked like (`str(e)`):

```
2020-02-12 09:46:36,042 ERROR: icons.py:57: [Errno 21] Is a directory: '/export/apps/weblate/data/static/icons/'
```

i think it's fairly spot on.  but i dont mind changing to whatever you want.

minusf:
the new log message:

```
2020-02-12 11:06:34,993 ERROR: errors.py:68: Failed to load icon: IsADirectoryError: [Errno 21] Is a directory: '/export/apps/weblate/data/static/icons/'
```

weblate/utils/templatetags/icons.py
nijel: I was thinking about raising ValueError here, this would be really a programming error and such should be visible while running tests, not just logged.

nijel:

PR #3466, opened 2020-02-12
Settings: add more options to AUTO_UPDATES

### LINKED ISSUE #3341: Ability to disable auto update
For Tails we are using Weblate with a set of scripts to integrate Weblate into the rest of our infrastructure [1]. Therefore the `update-remotes` cronjob should not be run. Currently we use a broken git url to make sure, that Weblate can't update the repository on it's own. But it is quite annoying to see always the alerts:

    The VCS repository has many changes compared to the upstream. Please merge the changes manually or set up push to automate this.
    --- 
    Weblate could not fetch upstream changes when updating the repository.

I searched around to lean about celery, but I haven't found a way to disable task, that's why  I think it needs to be supported by Weblate itself. If it is possible to disable cronjobs outside touching the Weblate code I would also be happy. 

[1] https://git.tails.boum.org/puppet-tails/tree/files/weblate/scripts/cron.sh

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/86133717-ability-to-disable-celery-jobs?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Naruto0:
(True,False)-> (full,remote,none)
	attempt to resolve #3341

### DIFF
--- a/docs/admin/config.rst
+++ b/docs/admin/config.rst
@@ -85,9 +85,9 @@ do not use :ref:`hooks` to update Weblate repositories automatically.
     There are both boolean options or string due to backwards compatibility.
 
 Options are:
-* `"none"` - no daily updates
-* `"remote"` also `False` - update remotes only
-* `"full"` also `True` - update remotes and merge working copy
+* ``"none"`` - no daily updates
+* ``"remote"`` also ``False`` - update remotes only
+* ``"full"`` also ``True`` - update remotes and merge working copy
 
 .. note::
 
--- a/docs/admin/config.rst
+++ b/docs/admin/config.rst
@@ -77,8 +77,17 @@ AUTO_UPDATE
 
 .. versionadded:: 3.2
 
-Automatically update all repositories on daily basis. This can be useful if you
+Update all repositories on daily basis. This can be useful if you
 do not use :ref:`hooks` to update Weblate repositories automatically.
+
+.. note::
+
+    There are both boolean options or string due to backwards compatibility.
+
+Options are:
+* `"none"` - no daily updates
+* `"remote"` also `False` - update remotes only
+* `"full"` also `True` - update remotes and merge working copy
 
 .. note::
 
--- a/weblate/trans/tasks.py
+++ b/weblate/trans/tasks.py
@@ -71,11 +71,6 @@ def perform_update(cls, pk, auto=False):
             obj.do_update()
         elif settings.AUTO_UPDATE in ("remote", False):
             obj.update_remote_branch()
-        elif settings.AUTO_UPDATE == "none":
-            pass
-        else:
-            # TODO: raise some warning error.
-            pass
     except FileParseError:
         # This is stored as alert, so we can silently ignore here
         return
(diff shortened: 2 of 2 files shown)


weblate/trans/tasks.py
```diff
@@ -67,11 +67,15 @@ def perform_update(cls, pk, auto=False):
...
+        if settings.AUTO_UPDATE in ("full", True) or auto:
             obj.do_update()
-        else:
+        elif settings.AUTO_UPDATE in ("remote", False):
             obj.update_remote_branch()
+        elif settings.AUTO_UPDATE == "none":
+            pass
+        else:
+            # TODO: raise some warning error.
+            pass
```
nijel: It would be better to to the "none" / invalid check in `update_remotes` and handle here only the "remote" / "full" case.

docs/admin/config.rst
```diff
@@ -77,9 +77,18 @@ AUTO_UPDATE
...
 do not use :ref:`hooks` to update Weblate repositories automatically.
 
+.. note::
+
+    There are both boolean options or string due to backwards compatibility.
+
+Options are:
+* `"none"` - no daily updates
+* `"remote"` also `False` - update remotes only
+* `"full"` also `True` - update remotes and merge working copy
```
nijel: Please use double backticks for literals, see https://www.sphinx-doc.org/en/latest/usage/restructuredtext/basics.html

weblate/trans/tasks.py
```diff
@@ -228,8 +227,14 @@ def cleanup_suggestions():
 def update_remotes():
     """Update all remote branches (without attempt to merge)."""
     non_linked = Component.objects.with_repo()
-    for component in non_linked.iterator():
-        perform_update.delay("Component", component.pk, auto=True)
+    if settings.AUTO_UPDATE == "none":
+        pass
+    elif settings.AUTO_UPDATE not in ("none", "full", "remote", True, False):
```
nijel: I'd probably simplify all this to just:

```suggestion
    if settings.AUTO_UPDATE not in ("full", "remote", True, False):
        return
```

Naruto0: thanks for suggestion

weblate/trans/tasks.py
```diff
@@ -228,8 +227,12 @@ def cleanup_suggestions():
 def update_remotes():
     """Update all remote branches (without attempt to merge)."""
     non_linked = Component.objects.with_repo()
-    for component in non_linked.iterator():
-        perform_update.delay("Component", component.pk, auto=True)
+
+    if settings.AUTO_UPDATE not in ("full", "remote", True, False):
+        return
+    else:
```
nijel: The else is no longer needed here as there is return in the if clause, see also https://deepsource.io/gh/WeblateOrg/weblate/run/3b96cf45-4966-4da0-a4ef-29440c86121d/python/PYL-R1705

weblate/trans/tasks.py
```diff
@@ -67,10 +67,9 @@ def perform_update(cls, pk, auto=False):
             obj = Project.objects.get(pk=pk)
         else:
             obj = Component.objects.get(pk=pk)
-
-        if not auto or settings.AUTO_UPDATE:
+        if settings.AUTO_UPDATE in ("full", True) or auto:
```
nijel: This inverts the original logic, should be: 
```suggestion
        if settings.AUTO_UPDATE in ("full", True) or not auto:
```

weblate/trans/tasks.py
```diff
@@ -67,10 +67,9 @@ def perform_update(cls, pk, auto=False):
...
             obj.do_update()
-        else:
+        elif settings.AUTO_UPDATE in ("remote", False):
```
nijel: Using just `else` should be enough here, other values are handled in update_remotes.

nijel:
I think it looks good now, let's see if the tests pass.

nijel:

Naruto0:
welcome

PR #3468, opened 2020-02-12
Spelling: amount of newlines matches, characters

comradekingu:


weblate/checks/chars.py
```diff
@@ -327,10 +327,10 @@ class EscapedNewlineCountingCheck(CountingCheck):
 
 
 class NewLineCountCheck(CountingCheck):
-    r"""Check whether there is same amount of new lines"""
+    r"""Check whether the amount of newlines matches"""
     string = '\n'
     check_id = 'newline-count'
-    name = _('Newline charater count mismatch')
+    name = _('Mismatching newline character count')
```
nijel: I was thinking about calling this rather "Mismatching lines count". What do you think?

comradekingu: It is a bit hard to decipher. "Mismatching line-count" would do.
Is it always technically correct vs. number of newline chars?

nijel: It just counts newline chars. When translating to Czech I realized that it's easier to understand when talking about number of lines than number of newline chars...

comradekingu:
```suggestion
    name = _('Mismatching line breaks')
```
I think this (typographic) variant makes sense over "Mismatching number of new lines".
"Added line breaks" does away with the idea of wrapping, but alas.

nijel:

PR #3514, opened 2020-02-22
Addons documentation reworked

comradekingu:
A bit tired, but I don't get all the "This way you can easily filter" in the
"Flag * as “Needs editing" sections.
Edit: In case anyone wants the web admin panel to review:

bilde

### DIFF
--- a/docs/admin/addons.rst
+++ b/docs/admin/addons.rst
@@ -22,8 +22,7 @@ Automatic translation
 
 .. versionadded:: 3.9
 
-This addon automatically translates strings using machine translation or other
-components.
+Automatically translates strings using machine translation or other components.
 
 .. seealso::
 
@@ -135,7 +134,7 @@ translations created by the developers.
 Statistics generator
 --------------------
 
-This addon generates a file containing detailed info about the translation.
+Generates a file containing detailed info about the translation.
 You can use Django template in both filename and content, see :ref:`markup`
 for a detailed markup description.
 
@@ -181,8 +180,8 @@ The PO file header will contain a list of contributors and years contributed:
 Update ALL_LINGUAS variable in the "configure" file
 ---------------------------------------------------
 
-Updates the ALL_LINGUAS variable in :file:`configure`, :file:`configure.in` or the
-:file:`configure.ac` file, when a new translation is added.
+Updates the ALL_LINGUAS variable in :file:`configure`, :file:`configure.in` or any
+:file:`configure.ac` files, when a new translation is added.
 
 .. _addon-weblate.gettext.customize:
 
@@ -257,7 +256,8 @@ Allows adjusting JSON output behavior, for example indentation or sorting.
 Formats the Java properties file
 --------------------------------
 
-This addon sorts the Java properties file, if present.
+Sorts the Java properties file.
+Available only if applicable.
 
 .. _addon-weblate.removal.comments:
 
@@ -288,8 +288,9 @@ Update RESX files
 
 .. versionadded:: 3.9
 
-Update all translation files to match the monolingual upstream base file, if present.
+Update all translation files to match the monolingual upstream base file.
 Unused strings are removed, and new ones added as copies of the source string.
+Available only if applicable.
 
 .. hint::
 
@@ -435,7 +436,7 @@ Post update repository processing
 ---------------------------------
 
 Post update repository processing can be used to update translation files when
-the source changes. To achieve this, please remember that Weblate only sees
+the VCS upstream source changes. To achieve this, please remember that Weblate only sees
 files committed to the VCS, so you need to commit changes as a part
 of the script.
 
--- a/docs/admin/addons.rst
+++ b/docs/admin/addons.rst
@@ -31,8 +31,8 @@ Automatically translates strings using machine translation or other components.
 
 .. _addon-weblate.cleanup.generic:
 
-Update PO files to match POT (msgmerge)
----------------------------------------
+Cleanup translation files
+-------------------------
 
 Update all translation files to match the monolingual base file. For most file
 formats, this means removing stale translation keys no longer present in the
--- a/docs/admin/addons.rst
+++ b/docs/admin/addons.rst
@@ -232,9 +232,11 @@ Squash Git commits
 
 Squash Git commits prior to pushing changes. You can choose one of following modes:
 
+.. versionadded:: 3.4
 * All commits into one
 * Per language
 * Per file
+.. versionadded:: 3.5
 * Per author
 
 Original commit messages are kept, but authorship is lost unless "Per author" is selected, or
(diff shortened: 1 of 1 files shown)


docs/admin/addons.rst
```diff
@@ -32,8 +32,8 @@ components.
 
 .. _addon-weblate.cleanup.generic:
 
-Cleanup translation files
--------------------------
+Update PO files to match POT (msgmerge)
+---------------------------------------
 
 Update all translation files to match the monolingual base file. For most file
 formats, this means removing stale translation keys no longer present in the
```
comradekingu: Can't remember if there is a newer word for "translation keys".

docs/admin/addons.rst
```diff
@@ -433,12 +429,12 @@ Additionally, the following environment variables are available:
 Post update repository processing
 ---------------------------------
 
-Post update repository processing can be used to update translation files on
-the source change. To achieve this, please remember that Weblate only sees
-files which are committed to the VCS, so you need to commit changes as a part
+Post update repository processing can be used to update translation files when
+the source changes. To achieve this, please remember that Weblate only sees
```
comradekingu: The source of?

docs/admin/addons.rst
```diff
@@ -302,25 +298,25 @@ Customize YAML output
...
+Allows adjusting YAML output behavior, for example line-length or newlines.
 
 
 Customizing list of addons
 ++++++++++++++++++++++++++
 
-List of addons is configured by :setting:`WEBLATE_ADDONS`, to add another addon
-simply include class absolute name in this setting.
+The list of addons is configured by :setting:`WEBLATE_ADDONS`.
+To add another addon, simply include class absolute name in this setting.
```
comradekingu: This isn't very clear…

docs/admin/addons.rst
```diff
@@ -184,8 +181,8 @@ The PO file header will contain list of contributors with years they have contri
 Update ALL_LINGUAS variable in the "configure" file
 ---------------------------------------------------
 
-Updates the ALL_LINGUAS variable in :file:`configure`, :file:`configure.in` or
-:file:`configure.ac` files, when a new translation is added.
+Updates the ALL_LINGUAS variable in :file:`configure`, :file:`configure.in` or the
+:file:`configure.ac` file, when a new translation is added.
```
comradekingu: Are there more configure.ac files, or just the one?

docs/admin/addons.rst
```diff
@@ -32,8 +32,8 @@ components.
 
 .. _addon-weblate.cleanup.generic:
 
-Cleanup translation files
--------------------------
+Update PO files to match POT (msgmerge)
+---------------------------------------
```
nijel: I guess this is an error, this is not a msgmerge.

docs/admin/addons.rst
```diff
@@ -236,26 +232,32 @@ Squash Git commits
...
-Allows to customize JSON output behavior, for example indentation or sorting.
+Allows adjusting JSON output behavior, for example indentation or sorting.
 
 .. _addon-weblate.properties.sort:
 
 Formats the Java properties file
 --------------------------------
 
-This addon sorts the Java properties file.
+This addon sorts the Java properties file, if present.
```
nijel: Why if present? The addon is installable only if the component is using Java properties.

docs/admin/addons.rst
nijel: probably better to speak about "upstream repository" here

docs/admin/addons.rst
nijel: We use keys elsewhere as well (recently introduced "Nearby keys"), so that's probably okay

docs/admin/addons.rst
nijel: There can be one or more of the listed files in the repository.

docs/admin/addons.rst
comradekingu: @nijel How is "nearby keys" different from "nearby strings". I have used it and can't really tell…

docs/admin/addons.rst
comradekingu: "It is "Update PO files to match POT (msgmerge)" is however what it is currently called in the list of plugins in the web admin UI.
What should it be called? (Will change it in both places)

docs/admin/addons.rst
nijel: "nearby keys" orders strings by keys and lists nearby ones, "nearby strings" uses ordering as is in the translation file

docs/admin/addons.rst
nijel: These are different addons. You probably don't see this one as it's not valid for gettext PO files.

docs/admin/addons.rst
```diff
@@ -236,26 +231,33 @@ Squash Git commits
...
+Allows adjusting JSON output behavior, for example indentation or sorting.
 
 .. _addon-weblate.properties.sort:
 
 Formats the Java properties file
 --------------------------------
 
-This addon sorts the Java properties file.
+Sorts the Java properties file.
+Available only if applicable.
```
nijel: This applies to many addons as they have some precoditions - in most cases based on file format (many of the Gettext based ones).

docs/admin/addons.rst
comradekingu: Ah, nice. Still a bit unclear, but will revert the error.

docs/admin/addons.rst
```diff
@@ -96,11 +93,11 @@ Bulk edit
 
 .. versionadded:: 3.11
 
-This addon allow to bulk edit flags, labels or state.
+Mass edit flags, labels or state for strings.
```
nijel: It used to be called mass editing, but was changed to bulk years ago, so the description probably should be consistent with it.

docs/admin/addons.rst
```diff
@@ -44,44 +43,42 @@ base file.
...
-discovery addon to find all translation components in the VCS. The matching is
-done using regular expression so it can be quite powerful, but it can be complex
-to configure. You can use examples in the addon help for some common use cases.
+The matching is done using regular expressions, where power is a tradeoff for
+complexity in configuration. Some examples for common use cases can be found in
+the addon help sectoin.
 
-Once you hit save, you will be presented with a preview of matched components,
-so you can check whether the configuration actually matches your needs:
+Once you hit "Save", a preview of matching components will be presented,
```
nijel:
```suggestion
Once you hit :guilabel:`Save`, a preview of matching components will be presented,
```

docs/admin/addons.rst
comradekingu: Can confirm.

comradekingu:
```suggestion
Bulk edit flags, labels or state for strings.
```

PR #3517, opened 2020-02-23
Fix button in removal confirmation mail

sou-y:


nijel:

PR #3518, opened 2020-02-23
Spelling: account name, or any e-mail address, -!

comradekingu:


weblate/trans/forms.py
```diff
@@ -1043,8 +1043,7 @@ class UserManageForm(forms.Form):
     user = UserField(
         label=_('User to add'),
         help_text=_(
-            'Please provide username or e-mail. '
-            'User needs to already have an active account in Weblate.'
+            'Please type in an account name, or any e-mail address. '
```
nijel: This removes information that this is only for existing users.

comradekingu:
```suggestion
            'Please type in a Weblate account name, or any e-mail address. '
```

weblate/trans/forms.py
```diff
@@ -1043,8 +1043,7 @@ class UserManageForm(forms.Form):
     user = UserField(
         label=_('User to add'),
         help_text=_(
-            'Please provide username or e-mail. '
-            'User needs to already have an active account in Weblate.'
+            'Please type in a Weblate account name, or any e-mail address. '
```
nijel:
```suggestion
            'Please type in an username or e-mail address of existing Weblate account. '
```

How about this? It should be clear that this is only for existing accounts.

comradekingu:
```suggestion
            'Please type in an existing Weblate account name or e-mail address. '
```

nijel:

I've removed trailing space in 6723b2f0935f11463036c3ee06495e8197364514.

PR #3538, opened 2020-02-29
Spelling: Newer strings reworked

comradekingu:


weblate/templates/trans/alert/brokenbrowserurl.html
```diff
@@ -1,7 +1,7 @@
 {% load i18n %}
 <p>{% trans "This component is set up incorrectly." %}</p>
 
-<p>{% trans "The repository browser URL points to non existing locations:" %}</p>
+<p>{% trans "The repository browser URL points to a non-existing location:" %}</p>
```
comradekingu: Either "URLs / locations" or "URL location" :)

weblate/utils/validators.py
```diff
@@ -184,28 +184,28 @@ def validate_email(value):
     if EMAIL_BLACKLIST.match(user_part):
         raise ValidationError(_('Enter a valid e-mail address.'))
     if not re.match(settings.REGISTRATION_EMAIL_MATCH, value):
-        raise ValidationError(_('This e-mail address is not allowed.'))
+        raise ValidationError(_('This e-mail address is already in use.'))
```
nijel: This is really about not allowed addresses, it can be configured by https://docs.weblate.org/en/latest/admin/config.html#registration-email-match

weblate/templates/trans/alert/brokenbrowserurl.html
nijel: True, but the repository browser is actually a template and it evaluates to several locations.

comradekingu:
```suggestion
<p>{% trans "The repository browser URL points to non-existing locations:" %}</p>
```
On the assumption this shouldn't be "one or more non-existing locations".

weblate/utils/validators.py
comradekingu:
```suggestion
        raise ValidationError(_('This e-mail address is disallowed.'))
```

PR #3550, opened 2020-03-03
WIP: Import ABCs with collections.abc

comradekingu:
https://docs.python.org/3/library/collections.html#collections.Counter

> Deprecated since version 3.3, will be removed in version 3.9: Moved Collections Abstract Base Classes to the collections.abc module. For backwards compatibility, they continue to be visible in this module through Python 3.8.

It isn't there in 2.7, but WL has deprecated that in WL 4.0?

### DIFF
--- a/weblate/accounts/notifications.py
+++ b/weblate/accounts/notifications.py
@@ -18,7 +18,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
-from collections import defaultdict
+from collections.abc import defaultdict
 from copy import copy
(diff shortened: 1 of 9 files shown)


nijel:
This looks wrong. I don't think we use ABC directly in Weblate.

ImportError: cannot import name 'namedtuple'

nijel:
The only things which has moved are [Abstract Base Classes](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes) and we don't use these in our code. Some third-party module does that and that emits warnings.

PR #3551, opened 2020-03-03
Use url_has_allowed_host_and_scheme

comradekingu:
https://django.readthedocs.io/en/latest/releases/3.0.html 

> To avoid possible confusion as to effective scope, the private internal utility is_safe_url() is renamed to url_has_allowed_host_and_scheme(). That a URL has an allowed host and scheme doesn’t in general imply that it’s “safe”. It may still be quoted incorrectly, for example. Ensure to also use iri_to_uri() on the path component of untrusted URLs.

### DIFF
--- a/weblate/accounts/pipeline.py
+++ b/weblate/accounts/pipeline.py
@@ -26,7 +26,7 @@ from django.conf import settings
 from django.shortcuts import redirect
 from django.urls import reverse
 from django.utils.encoding import force_str
-from django.utils.http import is_safe_url
+from django.utils.http import url_has_allowed_host_and_scheme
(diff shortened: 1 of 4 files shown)


nijel:
This needs Django 3.0, while we still support 2.2. At least for now.

nijel:
BTW: I already did and reverted this change in 0f61e31e056ce4f88e81dfa1ccd2f4248bb16de3 and 496f3b78bbdf321df34dbf09dd200714051d4087.

nijel:
See also https://github.com/WeblateOrg/weblate/issues/3552

PR #3553, opened 2020-03-03
patch for #3526 Change Microsoft API to V3

### LINKED ISSUE #3526: Upgrade Microsoft translate API to v3
The v2 API currently used by Weblate is deprecated, see https://docs.microsoft.com/en-us/azure/cognitive-services/Translator/reference/v2-0-reference.

Upgrade the API usage.

We can stay on v2 for now as it doesn't see to have a deadline set.

**Additional context**
* https://docs.microsoft.com/en-us/azure/cognitive-services/translator/migrate-to-v3
* https://github.com/WeblateOrg/weblate/blob/master/weblate/machinery/microsoft.py

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/89121914-upgrade-microsoft-translate-api-to-v3?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Naruto0:

This should resolve #3526

### DIFF
--- a/docs/admin/config.rst
+++ b/docs/admin/config.rst
@@ -914,6 +914,13 @@ Client key for Microsoft Cognitive Services Translator API.
     `Cognitive Services - Text Translation API <https://azure.microsoft.com/services/cognitive-services/translator-text-api/>`_,
     `Microsoft Azure Portal <https://portal.azure.com/>`_
 
+.. setting:: MT_MICROSOFT_REGION
+
+MT_MICROSOFT_REGION
+-------------------
+
+Region prefix as defined in `Multi service subscription <https://docs.microsoft.com/en-us/azure/cognitive-services/translator/reference/v3-0-reference#multi-service-subscription>`_.
+
 .. setting:: MT_MYMEMORY_EMAIL
 
 MT_MYMEMORY_EMAIL
--- a/docs/admin/machine.rst
+++ b/docs/admin/machine.rst
@@ -145,14 +145,23 @@ Microsoft Cognitive Services Translator
 Machine translation service provided by Microsoft in Azure portal as a one of
 Cognitive Services.
 
-You need to register at Azure portal and use the key you obtain there.
+Weblate implements Translator API V3.
 
 To enable this service, add ``weblate.machinery.microsoft.MicrosoftCognitiveTranslation`` to
 :setting:`MT_SERVICES` and set :setting:`MT_MICROSOFT_COGNITIVE_KEY`.
 
+Translator Text API V2
+``````````````````````
+The key you use with Translator API V2 can be used with API 3.
+
+Translator Text API V3
+``````````````````````
+You need to register at Azure portal and use the key you obtain there.
+With new Azure keys, you also need to set :setting:`MT_MICROSOFT_REGION` to locale of your service.
+
 .. seealso::
 
-    :setting:`MT_MICROSOFT_COGNITIVE_KEY`,
+    :setting:`MT_MICROSOFT_COGNITIVE_KEY`, :setting:`MT_MICROSOFT_REGION`,
     `Cognitive Services - Text Translation API <https://azure.microsoft.com/services/cognitive-services/translator-text-api/>`_,
     `Microsoft Azure Portal <https://portal.azure.com/>`_
 
--- a/weblate/machinery/microsoft.py
+++ b/weblate/machinery/microsoft.py
@@ -117,7 +117,7 @@ class MicrosoftCognitiveTranslation(MachineTranslation):
         if isinstance(payload, str):
             raise Exception(payload)
 
-        return payload
+        return list(payload['translation'].keys())
 
     def download_translations(self, source, language, text, unit, user):
         """Download list of possible translations from a service."""
--- a/weblate/machinery/microsoft.py
+++ b/weblate/machinery/microsoft.py
@@ -25,12 +25,11 @@ from django.utils import timezone
 
 from weblate.machinery.base import MachineTranslation, MissingConfiguration
 
-COGNITIVE_BASE_URL = 'https://api.cognitive.microsoft.com/sts/v1.0'
-COGNITIVE_TOKEN = COGNITIVE_BASE_URL + '/issueToken?Subscription-Key={0}'
-
-BASE_URL = 'https://api.microsofttranslator.com/V2/Ajax.svc/'
-TRANSLATE_URL = BASE_URL + 'Translate'
-LIST_URL = BASE_URL + 'GetLanguagesForTranslate'
+BASE_URL = 'https://api.cognitive.microsofttranslator.com'
+TRANSLATE_URL = BASE_URL + '/translate'
+LIST_URL = BASE_URL + '/languages?api-version=3.0'
+TOKEN_URL = 'https://{0}api.cognitive.microsoft.com/sts/v1.0/'\
+            'issueToken?Subscription-Key={1}'
 TOKEN_EXPIRY = timedelta(minutes=9)
 
 
@@ -56,6 +55,16 @@ class MicrosoftCognitiveTranslation(MachineTranslation):
         super().__init__()
         self._access_token = None
         self._token_expiry = None
+
+        # check settings for Microsoft region prefix
+        if settings.MT_MICROSOFT_REGION is None:
+            region = ""
+        else:
+            region = "{}.".format(settings.MT_MICROSOFT_REGION)
+
+        self._cognitive_token_url = TOKEN_URL.\
+            format(region, settings.MT_MICROSOFT_COGNITIVE_KEY)
+
         if settings.MT_MICROSOFT_COGNITIVE_KEY is None:
             raise MissingConfiguration('Microsoft Translator requires credentials')
 
@@ -73,7 +82,7 @@ class MicrosoftCognitiveTranslation(MachineTranslation):
         if self._access_token is None or self.is_token_expired():
             self._access_token = self.request(
                 "post",
-                COGNITIVE_TOKEN.format(settings.MT_MICROSOFT_COGNITIVE_KEY),
+                self._cognitive_token_url,
                 skip_auth=True,
             ).text
             self._token_expiry = timezone.now() + TOKEN_EXPIRY
@@ -113,13 +122,17 @@ class MicrosoftCognitiveTranslation(MachineTranslation):
     def download_translations(self, source, language, text, unit, user):
         """Download list of possible translations from a service."""
         args = {
-            'text': text[:5000],
+            'api-version': '3.0',
             'from': source,
             'to': language,
-            'contentType': 'text/plain',
             'category': 'general',
         }
-        response = self.request("get", TRANSLATE_URL, params=args)
+        response = self.request(
+            "post",
+            TRANSLATE_URL,
+            params=args,
+            json=[{'Text': text[:5000]}]
+        )
         # Microsoft tends to use utf-8-sig instead of plain utf-8
         response.encoding = response.apparent_encoding
         payload = response.json()
--- a/weblate/machinery/microsoft.py
+++ b/weblate/machinery/microsoft.py
@@ -117,7 +117,7 @@ class MicrosoftCognitiveTranslation(MachineTranslation):
         if isinstance(payload, str):
             raise Exception(payload)
 
-        return list(payload['translation'].keys())
+        return payload['translation'].keys()
 
     def download_translations(self, source, language, text, unit, user):
         """Download list of possible translations from a service."""
--- a/weblate/machinery/microsoft.py
+++ b/weblate/machinery/microsoft.py
@@ -28,8 +28,9 @@ from weblate.machinery.base import MachineTranslation, MissingConfiguration
 BASE_URL = 'https://api.cognitive.microsofttranslator.com'
 TRANSLATE_URL = BASE_URL + '/translate'
 LIST_URL = BASE_URL + '/languages?api-version=3.0'
-TOKEN_URL = 'https://{0}api.cognitive.microsoft.com/sts/v1.0/'\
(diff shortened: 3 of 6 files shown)


weblate/machinery/microsoft.py
```diff
@@ -28,7 +28,7 @@
 COGNITIVE_BASE_URL = 'https://api.cognitive.microsoft.com/sts/v1.0'
 COGNITIVE_TOKEN = COGNITIVE_BASE_URL + '/issueToken?Subscription-Key={0}'
 
-BASE_URL = 'https://api.microsofttranslator.com/V2/Ajax.svc/'
+BASE_URL = 'https://api.cognitive.microsofttranslator.com/translate?api-version=3.0/'
```
nijel: This looks wrong, check TRANSLATE_URL and LIST_URL below - these are called for individual API calls. Also, the trailing / looks wrong as it would be parsed as part of the parameter.

You should test this against live service, they have free tier for translate service, which is good enough for testing.

weblate/machinery/base.py
```diff
@@ -82,10 +82,11 @@ def request(self, method, url, skip_auth=False, **kwargs):
         headers = {
             'Referer': get_site_url(),
             'Accept': 'application/json; charset=utf-8',
+            'Content-Type': 'application/json',
         }
         if "headers" in kwargs:
             headers.update(kwargs.pop("headers"))
-        # Optional authentication
+        # Optional authenticationr
```
nijel: Seems like a typo...

weblate/machinery/microsoft.py
```diff
@@ -112,14 +120,15 @@ def download_languages(self):
...
         args = {
-            'text': text[:5000],
+            'api-version': '3.0',
             'from': source,
             'to': language,
-            'contentType': 'text/plain',
             'category': 'general',
         }
-        response = self.request("get", TRANSLATE_URL, params=args)
+        response = self.request("post", TRANSLATE_URL, params=args, data=payload)
```
nijel: You can use `json` parameter and avoid manual encoding.

weblate/machinery/tests.py
```diff
@@ -693,7 +696,7 @@ def test_aws(self):
             stubber.add_response(
                 'translate_text',
                 {
-                    'TranslatedText': 'Hallo',
+                    'TranslatedText'    : 'Hallo',
```
nijel: Please remove this change.

weblate/machinery/microsoft.py
```diff
@@ -18,19 +18,27 @@
...
 from datetime import timedelta
 
 from django.conf import settings
 from django.utils import timezone
 
 from weblate.machinery.base import MachineTranslation, MissingConfiguration
 
-COGNITIVE_BASE_URL = 'https://api.cognitive.microsoft.com/sts/v1.0'
+try:
+    if settings.MT_MICROSOFT:
```
nijel: Please avoid using settings here. This is executed at import time, when settings might not be yet initialized or might have wrong value. Move this logic to `__ini__` method and construct the url in the class.

docs/admin/machine.rst
```diff
@@ -145,14 +145,23 @@ Microsoft Cognitive Services Translator
...
 :setting:`MT_SERVICES` and set :setting:`MT_MICROSOFT_COGNITIVE_KEY`.
 
+Translator Text API V2
+``````````````````````
+The key you use with Translator API V2 can be used with API 3.
+
+Translator Text API V3
+``````````````````````
+You need to register at Azure portal and use the key you obtain there.
+With new Azure keys, you also need to set :setting:`MT_MICROSOFT_REGION` to locale of your service.
```
nijel: Please add matching description for `MT_MICROSOFT_REGION` into `docs/admin/config.rst` (also see above failures).

docs/admin/config.rst
```diff
@@ -916,6 +916,11 @@ Client key for Microsoft Cognitive Services Translator API.
 
 .. setting:: MT_MYMEMORY_EMAIL
 
+MT_MICROSOFT_REGION
+-------------------
+
+Region prefix as defined in `Multi service subscription <https://docs.microsoft.com/en-us/azure/cognitive-services/translator/reference/v3-0-reference#multi-service-subscription>`_.
+
 MT_MYMEMORY_EMAIL
```
nijel:
```suggestion
.. setting:: MT_MICROSOFT_REGION

MT_MICROSOFT_REGION
-------------------

Region prefix as defined in `Multi service subscription <https://docs.microsoft.com/en-us/azure/cognitive-services/translator/reference/v3-0-reference#multi-service-subscription>`_.

.. setting:: MT_MYMEMORY_EMAIL

MT_MYMEMORY_EMAIL
```

The `.. setting` is anchor used for linking from other places.

weblate/machinery/microsoft.py
```diff
@@ -18,19 +18,17 @@
...
 
-COGNITIVE_BASE_URL = 'https://api.cognitive.microsoft.com/sts/v1.0'
-COGNITIVE_TOKEN = COGNITIVE_BASE_URL + '/issueToken?Subscription-Key={0}'
-
-BASE_URL = 'https://api.microsofttranslator.com/V2/Ajax.svc/'
-TRANSLATE_URL = BASE_URL + 'Translate'
-LIST_URL = BASE_URL + 'GetLanguagesForTranslate'
+BASE_URL = 'https://api.cognitive.microsofttranslator.com'
+TRANSLATE_URL = BASE_URL + '/translate'
+LIST_URL = BASE_URL + '/languages?api-version=3.0'
```
nijel:
```suggestion
LIST_URL = BASE_URL + '/languages?api-version=3.0'
TOKEN_URL = 'https://{0}api.cognitive.microsoft.com/sts/v1.0/issueToken?Subscription-Key={1}'
```

weblate/machinery/microsoft.py
```diff
@@ -56,6 +54,15 @@ def __init__(self):
...
         self._access_token = None
         self._token_expiry = None
+
+        # check settings for Microsoft region prefix
+        if settings.MT_MICROSOFT_REGION is None:
+            cbs_url = 'https://api.cognitive.microsoft.com/sts/v1.0'
+        else:
+            cbs_url = 'https://{0}.api.cognitive.microsoft.com/sts/v1.0' \
+                .format(settings.MT_MICROSOFT_REGION)
+        self._cognitive_token_url = cbs_url + '/issueToken?Subscription-Key={0}'
```
nijel:
```suggestion
        # check settings for Microsoft region prefix
        if settings.MT_MICROSOFT_REGION is None:
            region = ""
        else:
            region = "{}.".format(settings.MT_MICROSOFT_REGION)

        self._cognitive_token_url = TOKEN_URL.format(region, settings.MT_MICROSOFT_COGNITIVE_KEY)
```

weblate/machinery/microsoft.py
```diff
@@ -73,7 +80,7 @@ def access_token(self):
         if self._access_token is None or self.is_token_expired():
             self._access_token = self.request(
                 "post",
-                COGNITIVE_TOKEN.format(settings.MT_MICROSOFT_COGNITIVE_KEY),
+                self._cognitive_token_url.format(settings.MT_MICROSOFT_COGNITIVE_KEY),
```
nijel:
```suggestion
                self._cognitive_token_url,
```

weblate/machinery/base.py
```diff
@@ -82,6 +82,7 @@ def request(self, method, url, skip_auth=False, **kwargs):
         headers = {
             'Referer': get_site_url(),
             'Accept': 'application/json; charset=utf-8',
+            'Content-Type': 'application/json',
```
nijel: I'm not sure all machineries use JSON, so this might be wrong. Also requests set this automatically when you pass `json` to request, so it's not needed in your case.

Naruto0: Yes, this is a mistake, i would not say it!s good idea to touch base.py unless changes needs to be reflected to all machines. Any changes there are mistake.

weblate/machinery/microsoft.py
```diff
@@ -112,14 +122,15 @@ def download_languages(self):
...
         args = {
-            'text': text[:5000],
+            'api-version': '3.0',
             'from': source,
             'to': language,
-            'contentType': 'text/plain',
             'category': 'general',
         }
-        response = self.request("get", TRANSLATE_URL, params=args)
+        response = self.request("post", TRANSLATE_URL, params=args, json=payload)
```
nijel:
```suggestion
        response = self.request("post", TRANSLATE_URL, params=args, json=[{'Text': text}])
```

`json` param accepts data directly. You probably haven't tested this at all, because this could not work with the real service (in the test we do not look at request bodies, so it will pass).

Naruto0: Yes i tested it and it does not work as suggested.

nijel: Because you have there `{'Text','Hello'}`, while my suggestion is `{'Text': text}`

nijel: But what I meant - did you test your code with real service? IMHO it does not work as it is now, because you pass it a string instead of json values.

weblate/machinery/tests.py
```diff
@@ -378,14 +380,15 @@ def test_microsoft_cognitive(self):
...
         )
         responses.add(
-            responses.GET,
-            'https://api.microsofttranslator.com/V2/Ajax.svc/Translate',
-            json="svět",
+            responses.POST,
+            'https://api.cognitive.microsofttranslator.com/'
+            'translate?api-version=3.0&from=en&to=cs&category=general',
+            json=MICROSOFT_RESPONSE,
         )
```
nijel: Can you please add second test where you would override region as well?  It should use different URL for issuing token.

weblate/machinery/microsoft.py
```diff
@@ -112,14 +122,15 @@ def download_languages(self):
 
     def download_translations(self, source, language, text, unit, user):
         """Download list of possible translations from a service."""
+        payload = _js.dumps([{'Text': text}])
+
         args = {
-            'text': text[:5000],
```
nijel: Is length limitation gone from the API? I see you now use full text.

weblate/machinery/microsoft.py
Naruto0: Yes, I try the service in shell with every commit.

weblate/machinery/tests.py
Naruto0: Sure, thanks for pointing out.

weblate/machinery/microsoft.py
Naruto0: You are correct, the limit for the translation is still [5000 chars](https://docs.microsoft.com/en-us/azure/cognitive-services/translator/request-limits).

Fixed now.

weblate/machinery/tests.py
```diff
@@ -378,14 +380,44 @@ def test_microsoft_cognitive(self):
...
         responses.add(
             responses.GET,
-            'https://api.microsofttranslator.com/V2/Ajax.svc/'
-            'GetLanguagesForTranslate',
+            'https://api.cognitive.microsofttranslator.com/'
+            'languages?api-version=3.0',
             json=["en", "cs"],
```
nijel: This doesn't match at all what is documented response here: https://docs.microsoft.com/en-us/azure/cognitive-services/translator/reference/v3-0-languages#response-body

That makes the service not working, because when executed with real service, the list of supported languages becomes: `{'dictionary', 'transliteration', 'translation'}`

weblate/machinery/microsoft.py
```diff
@@ -113,13 +122,17 @@ def download_languages(self):
...
-            'contentType': 'text/plain',
             'category': 'general',
         }
-        response = self.request("get", TRANSLATE_URL, params=args)
+        response = self.request(
+            "post",
+            TRANSLATE_URL,
+            params=args,
+            json=[{'Text': text[:5000]}]
+        )
```
nijel: Seems like some payload processing is missing here as it ends up being displayed this way:

image

I've extended the tests to cover this in f4226ace09545d8fb74f1867ad25115903e658fd so please rebase on current master.

weblate/machinery/tests.py
Naruto0: I also do not need to fetch supported languages in both tests, do I?

weblate/machinery/microsoft.py
Naruto0: Sure, can you point me somewhere how should i format the payload?

Thanks in avance!

weblate/machinery/tests.py
```diff
@@ -248,6 +248,14 @@
 
 DEEPL_RESPONSE = {"translations": [{"detected_source_language": "EN", "text": "Hallo"}]}
 
+MICROSOFT_RESPONSE = [{"translations": [{"text": "Svět.", "to": "cs"}]}]
+
+MS_SUPPORTED_LANG_RESP = {
+    "translation": {'Item': 'data'},
+    "transliteration": {'Item': 'data'},
+    "dictionary": {'Item': 'data'},
+}
```
nijel: You need to include real looking data, otherwise tests fail as the en cs translation seems unsupported.

weblate/machinery/microsoft.py
```diff
@@ -108,23 +117,24 @@ def download_languages(self):
         if isinstance(payload, str):
             raise Exception(payload)
 
-        return payload
+        return list(payload['translation'].keys())
```
nijel:
```suggestion
        return payload['translation'].keys()
```
Type casting is not needed here, the set is created from this in caller.

nijel:

PR #3563, opened 2020-03-04
fix hexlify()

minusf:
`hexlify()` returns `b''` in py3, resulting in usernames like `deleted-2-b'614b67018f'`

This is probably not the most elegant fix, but i dont use python2 anymore, so...

### DIFF
--- a/weblate/accounts/utils.py
+++ b/weblate/accounts/utils.py
@@ -21,6 +21,7 @@
 
 import binascii
 import os
+import six
 
 from social_django.models import Code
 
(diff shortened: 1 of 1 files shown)


minusf:
sorry, i was confused by setup.py and all the six imports :}

nijel:
There should be no six imports left in master branch. The 3.11 was last release supporting Python 2.

minusf:
right, then only `setup.py`:

```
        "Programming Language :: Python :: 2",
        "Programming Language :: Python :: 2.7",
```

nijel:
I forgot to remove that, done in f068e05d83. Anyway the most important thing was there:

https://github.com/WeblateOrg/weblate/blob/f068e05d839b6dfa98dbb9690178906bfbe31d94/setup.py#L50

nijel:

minusf:
now i know why i was so confused: 3.11.2 still has `six` imports and we use pypi releases in our project. when i debug stuff it's in these releases, but when i submit patches it's against master..

PR #3582, opened 2020-03-07
Mark string as translatable

urbalazs:
I found a string that are not marked as translatable. I tried to fix this, but if the fix does not work, feel free to modify it.


nijel:
It won't work this way, you have to use `blocktrans` to include variables. Anyway this was more or less intentional to keep this untranslated as it's hidden by CSS. The reason is that Gmail folds same content of emails from one sender and in case of Weblate it means it folds buttons to navigate to the translation. Adding unique invisble text to bottom of the email seems way to avoid that.

PR #3586, opened 2020-03-09
Checks and fixups documentation reworked

comradekingu:
https://docs.weblate.org/en/latest/admin/checks.html and https://docs.weblate.org/en/latest/user/checks.html are both called "Checks and fixups", that isn't ideal.

### DIFF
--- a/docs/admin/checks.rst
+++ b/docs/admin/checks.rst
@@ -12,13 +12,13 @@ include them in :setting:`AUTOFIX_LIST`.
 The automatic fixes are powerful, but can also cause damage;  be careful when
 writing one.
 
-For example, the following automatic fixup would replace every occurrence of string
-``foo`` in translation with ``bar``:
+For example, the following automatic fixup would replace every occurrence of the string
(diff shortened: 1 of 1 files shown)


docs/admin/checks.rst
```diff
@@ -208,30 +207,34 @@ flags (see :ref:`custom-checks`). You will probably need following:
...
+A wide range of quality checks are built-in, (see :ref:`checks`), though
+they might not cover everything you want to check. The list of performed checks
+can be adjusted using :setting:`CHECK_LIST`, and you can also add custom checks.
 
-To install custom checks, you need to provide a fully-qualified path to the Python class
+1. Subclass the `weblate.checks.Check`
+2. Set a few attributes.
+3. Implement either the ``check`` (if you want to deal with plurals in your code) or
+the ``check_single`` method, (which does it for you)
+you).
```
nijel:
```suggestion
3. Implement either the ``check`` (if you want to deal with plurals in your code) or
  the ``check_single`` method, (which does it for you).
```

PR #3595, opened 2020-03-11
Make the cached Loader check more robust.

minusf:
The check should be applied to the engine instances, not their (unprocessed) configuration. This fixes false positives such as when no `loaders` are defined (cached is added automatically in this case) and/or when `APP_DIRS` is enabled.

Don't encourage cached loader with `DEBUG`.

Get rid of the legacy TEMPLATE_LOADERS.


nijel:

PR #3602, opened 2020-03-13
Checks: Unchanged translation incorporate placeholders flag

### LINKED ISSUE #3490: Unchanged check should honor placeholders
It is possible to define placeholders using [`placeholder:` flag](https://docs.weblate.org/en/latest/admin/checks.html#customizing-behavior). The [Unchanged translation check](https://docs.weblate.org/en/latest/user/checks.html#unchanged-translation) should deal with those correctly.

The check strip custom placeholders from the string similarly like we do with format strings:

https://github.com/WeblateOrg/weblate/blob/ddfbd298a31a95229fa4487cfb2a5a4878da7b2b/weblate/checks/same.py#L107-L137

**Additional context**

Getting `placeholder:` flag value:

https://github.com/WeblateOrg/weblate/blob/256fa9370edc90c2404df6178d1daa002d441e2f/weblate/checks/base.py#L179

Building regexp for placeholders:

https://github.com/WeblateOrg/weblate/blob/256fa9370edc90c2404df6178d1daa002d441e2f/weblate/checks/placeholders.py#L57 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/88807457-unchanged-check-should-honor-placeholders?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Naruto0:

Resolves #3490 

### DIFF
--- a/weblate/checks/same.py
+++ b/weblate/checks/same.py
@@ -140,6 +140,17 @@ def test_word(word):
     return len(word) <= 2 or word in SAME_BLACKLIST or word in LANGUAGES
 
 
+def strip_placeholders(msg, unit):
+
+    return re.sub(
+        "|".join(
+            re.escape(param) for param in unit.all_flags.get_value("placeholders")
+        ),
+        "",
+        msg,
+    )
+
+
 class SameCheck(TargetCheck):
     """Check for not translated entries."""
 
@@ -168,6 +179,10 @@ class SameCheck(TargetCheck):
             return True
         # Strip format strings
         stripped = strip_string(source, unit.all_flags)
+
+        # Strip placeholder strings
+        if "placeholders" in unit.all_flags:
+            stripped = strip_placeholders(stripped, unit)
 
         # Ignore strings which don't contain any string to translate
         # or just single letter (usually unit or something like that)
--- a/weblate/checks/tests/test_same_checks.py
+++ b/weblate/checks/tests/test_same_checks.py
@@ -217,5 +217,5 @@ class SameCheckTest(CheckTestCase):
         self.do_test(True, ("Who is RMS?", "Who is RMS?", ""))
 
     def test_same_placeholders(self):
-        self.do_test(False, ("text $URL$", "text $URL$", "placeholders:$URL$"))
-        self.do_test(True, ("text $URL$", "text $URL$", ""))
+        self.do_test(False, ("DOMAIN", "DOMAIN", "placeholders:DOMAIN"))
+        self.do_test(True, ("PLACHOLDER", "PLACEHOLDER", "placeholders:DOMAIN"))
--- a/weblate/checks/tests/test_same_checks.py
+++ b/weblate/checks/tests/test_same_checks.py
@@ -217,5 +217,5 @@ class SameCheckTest(CheckTestCase):
         self.do_test(True, ("Who is RMS?", "Who is RMS?", ""))
 
     def test_same_placeholders(self):
-        self.do_test(False, ("DOMAIN", "DOMAIN", "placeholders:DOMAIN"))
-        self.do_test(True, ("PLACHOLDER", "PLACEHOLDER", "placeholders:DOMAIN"))
+        self.do_test(True, ("%location%", "%location%", ""))
+        self.do_test(False, ("%location%", "%location%.", "placeholders:%location%"))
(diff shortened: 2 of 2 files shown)


weblate/checks/same.py
```diff
@@ -140,6 +140,14 @@ def test_word(word):
     return len(word) <= 2 or word in SAME_BLACKLIST or word in LANGUAGES
 
 
+def strip_placeholders(msg, unit):
+    regexp = re.compile(
+        "|".join(re.escape(param) for param in unit.all_flags.get_value("placeholders"))
+    )
+    stripped = regexp.sub("", msg)
+    return stripped
```
nijel: Please simplify this:

* Pre-comiling regexp doesn't bring any benefits here as it is used just once. Better use directly `re.sub`.
* There is no need for local variable `stripped`, the value should be returned directly.

weblate/checks/tests/test_same_checks.py
```diff
@@ -215,3 +215,8 @@ def test_same_uppercase(self):
         self.do_test(False, ("RMS", "RMS", ""))
         self.do_test(False, ("<primary>RMS</primary>", "<primary>RMS</primary>", ""))
         self.do_test(True, ("Who is RMS?", "Who is RMS?", ""))
+
+    def test_same_placeholders(self):
+        self.do_test(False, ("string $URL$", "string $URL$", "placeholders:$URL$"))
+        self.do_test(False, ("string $URL$", "string ", "placeholders:$URL$"))
+        self.do_test(True, ("string $URL$", "string $URL$", "placeholders:"))
```
nijel: "string" is bad for testing as it's in same check blacklist. I'd probably just remove it as it's not really needed here and just adds confustion that it might be important.

Naruto0: Should there be another word instead?

nijel: I don't think any word is needed, this should test solely placeholders behavior.

Naruto0: In this case, we want to test if the `same check` fires if placeholder is missing. It fires with `self.do_test(True, ("text $URL$", "text $URL$", "")` not with `self.do_test(True, ("$URL$", "$URL$", "")` I am unable to track down the problem, but i imagine, there is some other same check, that ignores specifict patteren like that.
I hope you can help me there.

nijel: It might be that the placeholder is stripped out by other logic. I think there is something to trip out abbreviations, also it's likely that URL is in existing blacklist. So better choose something what works :-).

weblate/checks/tests/test_same_checks.py
```diff
@@ -217,6 +217,5 @@ def test_same_uppercase(self):
         self.do_test(True, ("Who is RMS?", "Who is RMS?", ""))
 
     def test_same_placeholders(self):
-        self.do_test(False, ("string $URL$", "string $URL$", "placeholders:$URL$"))
-        self.do_test(False, ("string $URL$", "string ", "placeholders:$URL$"))
-        self.do_test(True, ("string $URL$", "string $URL$", "placeholders:"))
+        self.do_test(False, ("DOMAIN", "DOMAIN", "placeholders:DOMAIN"))
+        self.do_test(True, ("PLACHOLDER", "PLACEHOLDER", "placeholders:DOMAIN"))
```
nijel: I still think this doesn't actually test the desired behavior (what should be purpose of the tests).

1. The first test does not fire regardless placeholders settings, because upper case strings are removed later:
https://github.com/WeblateOrg/weblate/blob/91be77f33b0e872bbc91afc5bdb89b9504a688a2/weblate/checks/same.py#L175-L176
2. The second test never fires for the same reason, but should not fire anyway as the string is different.

The test should cover the behavior that it works as intended. Basically it should cover:

1. The check does fire on a string when no placeholder is configured.
2. The check does not fire once placeholder is set.

I'd start with something like this (not tested):

```python
self.do_test(True, ("@location@", "@location@", ""))
self.do_test(False, ("@location@", "@location@", "placeholders:@location@"))
```

Naruto0: Thank you @nijel, i followed your suggestion and now it should work.

nijel:

Issue #3393, opened 2020-01-17
Add batch API for machinery


nijel:
**Is your feature request related to a problem? Please describe.**
The built-in translation memory is handled in same way as external services. This makes it quite slow as we're getting translations one by one.

**Describe the solution you'd like**
Perform one search on translation memory for all strings we want to translate instead of doing that one by one. This might be easier to implement once the memory is in same database as translated strings (see https://github.com/WeblateOrg/weblate/issues/2825).

**Describe alternatives you've considered**
Current approach works, but is slow for big components.

nijel:
Actually some of the machine translation services provide batch API as well, so it might make sense to have this in machinery and use it in that case. Brought up at https://github.com/WeblateOrg/weblate/pull/3609#discussion_r395191363

PR #3609, opened 2020-03-19
Implement google api v3

### LINKED ISSUE #2979: Google Translate API v3
There is new version of Google Translate API - [version 3](https://cloud.google.com/translate/docs/intro-to-v3).

The API itself is almost same, the biggest difference is that it does not support API keys:

> Cloud Translation v3 does not currently support API keys. We recommend that you create a service account for Cloud Translation v3beta1 requests. For information on creating a service account, see Creating and managing service accounts. Your service account must be added to one of the IAM roles added for Cloud Translation v3.

Add support for this new authentication. Most likely by using [existing Python binding](https://pypi.org/project/google-cloud-translate/).

**Additional context**
The v3 API supports for customization of used MT model, what might be useful feature for many use cases.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/79463614-google-translate-api-v3?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Naruto0:

Resolves #2979

### DIFF
--- a/docs/admin/config.rst
+++ b/docs/admin/config.rst
@@ -900,7 +900,7 @@ API key for Google Translate API v2, you can register at https://cloud.google.co
 
    :ref:`google-translate`, :ref:`machine-translation-setup`, :ref:`machine-translation`
 
-.. setting:: MT_MT_GOOGLE_CREDENTIALS
+.. setting:: MT_GOOGLE_CREDENTIALS
 
 MT_GOOGLE_CREDENTIALS
 ---------------------
--- a/docs/admin/config.rst
+++ b/docs/admin/config.rst
@@ -894,11 +894,37 @@ API key for the DeepL API, you can register at https://www.deepl.com/pro.html
 MT_GOOGLE_KEY
 -------------
 
-API key for the Google Translate API, you can register at https://cloud.google.com/translate/docs
+API key for Google Translate API v2, you can register at https://cloud.google.com/translate/docs
 
 .. seealso::
 
    :ref:`google-translate`, :ref:`machine-translation-setup`, :ref:`machine-translation`
+
+.. setting:: MT_MT_GOOGLE_CREDENTIALS
+
+MT_GOOGLE_CREDENTIALS
+---------------------
+
+API v3 json credentials file obtained in Google cloud console. Credentials are per service-account affiliated with certain project.
+See https://cloud.google.com/docs/authentication/getting-started
+
+.. setting:: MT_GOOGLE_PROJECT
+
+MT_GOOGLE_PROJECT
+-----------------
+
+Google cloud `project id` with activated Translation service having billing activated.
+See https://cloud.google.com/appengine/docs/standard/nodejs/building-app/creating-project
+
+.. setting:: MT_GOOGLE_LOCATION
+
+MT_GOOGLE_LOCATION
+------------------
+
+Your Google Cloud Application Engine may be specific to a location.
+Change accordingly if default ``global`` fallback does not work for you.
+
+See https://cloud.google.com/appengine/docs/locations
 
 .. setting:: MT_MICROSOFT_BASE_URL
 
@@ -911,6 +937,10 @@ Region base url domain as defined in the `Base URLs section
 Defaults to ``api.cognitive.microsofttranslator.com``.
 
 .. setting:: MT_MICROSOFT_COGNITIVE_KEY
+
+.. seealso::
+
+    :ref:`google-translate-advanced`
 
 MT_MICROSOFT_COGNITIVE_KEY
 --------------------------
--- a/docs/admin/config.rst
+++ b/docs/admin/config.rst
@@ -905,7 +905,8 @@ API key for Google Translate API v2, you can register at https://cloud.google.co
 MT_GOOGLE_CREDENTIALS
 ---------------------
 
-API v3 json credentials file obtained in Google cloud console. Credentials are per service-account affiliated with certain project.
+API v3 json credentials file obtained in Google cloud console. Please provide full OS path.
+Credentials are per service-account affiliated with certain project.
 See https://cloud.google.com/docs/authentication/getting-started
 
 .. setting:: MT_GOOGLE_PROJECT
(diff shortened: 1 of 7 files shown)


docs/admin/config.rst
```diff
@@ -892,14 +892,34 @@ API key for DeepL API, you can register at https://www.deepl.com/pro.html.
 MT_GOOGLE_KEY
 -------------
 
-API key for Google Translate API, you can register at https://cloud.google.com/translate/docs
+API key for Google Translate API v3, you can register at https://cloud.google.com/translate/docs
```
nijel: I think this one is actually for v2, isn't it?

weblate/machinery/google.py
```diff
@@ -99,3 +101,46 @@ def get_error_message(self, exc):
...
+        ]
+
+    def download_translations(self, source, language, text, unit, user):
+        """Download list of possible translations from a service."""
+        trans = self.client.translate_text(
+            [text], language, self.parent, source_language_code=source
+        )
+
+        yield {
+            "text": trans.translations[0].translated_text,
```
nijel: Are there more translations returned? In that case it might make sense to return them all.

weblate/machinery/google.py
```diff
@@ -99,3 +101,46 @@ def get_error_message(self, exc):
...
+        if settings.MT_GOOGLE_PROJECT is None:
+            raise MissingConfiguration(
+                "You have to specify Google Cloud project "
+                "affiliated with provided credentials"
+            )
+        credentials = service_account.Credentials.from_service_account_file(
+            settings.MT_GOOGLE_CREDENTIALS
+        )
+        self.client = TranslationServiceClient(credentials=credentials)
+        self.parent = self.client.location_path(settings.MT_GOOGLE_PROJECT, "global")
```
nijel: Shouldn't the second part be configurable as well? (I have no clue, just asking)

weblate/machinery/google.py
```diff
@@ -99,3 +101,46 @@ def get_error_message(self, exc):
                 pass
 
         return super().get_error_message(exc)
+
+
+class GoogleTranslationAdvanced(MachineTranslation):
```
nijel: I'd probably call it `GoogleTranslationV3`.

Naruto0: I didn't know if number in class name does not break any good practice. Yes, i will go with this suggestion.

weblate/machinery/google.py
Naruto0: Yes, it is possible to have a location in project. However, given that global is default, it works for anyone having empty translation project. In case of having the `parent` configurable, i'd suggest the whole location string than having. That again seems more complicated.
Or i can see option to add either project name or full path. We can determine which one it is by perhaps using a regular expression. This one is up to you.

docs/admin/config.rst
Naruto0: Yes, this is mistake.

weblate/machinery/google.py
Naruto0: Given that the method is fed by text, it returns one translation. Yes, there is possibility to translate multiple strings on one call, yet i do not know where to get all the string to translate. Should i override some machine function than?

weblate/machinery/google.py
Naruto0: I did some digesting, and i came to result. It seems from this error `InvalidArgument: 400 Invalid 'parent'.; Invalid location name. Unsupported location 'local'. Must be 'us-central1' or 'global'.
`
So the location can either be `'global'` or `'us-central1'`
We can assume that the location depends on cloud owner google user account. However ia assume we should try catch the correct location and save it. But i have no idea where.

Naruto0: Now, i do not know if only my account can either use `us-central` and `global`, but i can use both.

weblate/machinery/google.py
nijel: There is nothing "Advanced" on the new API, it's just evolution of it. There will come v4 and v5, and the naming should be prepared for that.

There is no problem in using numbers in class names, it just can't be first letter.

weblate/machinery/google.py
nijel: Let's make it configurable with "global" being default. I did similar change for Microsoft Translate in 037e50719718878217b74f9dc3b03be008db168a.

weblate/machinery/google.py
nijel: Okay, in case it matches number of strings passed to it, then it can stay as it is. We currently do not have batch API in machinery (though it will be probably added in future, see https://github.com/WeblateOrg/weblate/issues/3393).

weblate/machinery/google.py
Naruto0: It is not in any way advanced for us, i agree. They themselves call it in docs advanced, thus i attempted. I understand now, i will update as suggested :)

comradekingu:
https://cloud.google.com/translate/
I think it should be "Google Translation API v3"

Naruto0:
As of today i bumped a little. Hope it will have test soon.

weblate/machinery/tests.py
```diff
@@ -532,6 +537,38 @@ def test_google_invalid(self):
...
+            return_value=[
+                {
+                    "text": "Ahoj",
+                    "quality": 90,
+                    "service": "Google Translate API v3",
+                    "source": "Hello",
+                }
+            ]
+        ),
+    )
```
nijel, line 566: I think that patching google-cloud-translate would be more useful. In this case do simply skip testing download_languages and download_translations methods, what is quite big amount of untested code in the class.

Patching `TranslationServiceClient.translate_text` and `TranslationServiceClient.get_supported_languages` seems like a more useful approach.

Naruto0: Can't agree more. Make much more sense. Thanks for  pointing that out.

nijel:
Merged as 9a13ef5b1024bd4d976deb15724a578bf4e9beed, thanks for your contribution. I've made some cleanups on top of that and the tests improvements are on todo list...

PR #3613, opened 2020-03-20
Changed DeepL API endpoint to v2 in order to work with new API keys

FoSoZx:


nijel:
Thanks for your pull request. Unfortunately there are two issues with this:

1. It breaks existing setups which have API keys working only on API v1.
2. This is most likely violating their terms of service as you are supposed to use v1 API with "CAT tool" where Weblate probably belongs.

Please see https://github.com/WeblateOrg/weblate/pull/3133#issuecomment-552066743

nijel:

PR #3631, opened 2020-03-25
Managing translations documentation reworked

comradekingu:


docs/devel/translations.rst
```diff
@@ -6,54 +6,45 @@ Managing translations
...
 String shapings
 ---------------
 
-Shapings are useful to group several strings together so that translators can
-see all variants of the string at one place. You can define regular expression
-to group the strings in the :ref:`component`:
+To see variants of one string, it can be defined as belonging to a string shaping
+by using a regular expression to group the respective strings in the :ref:`component`:
 
 .. image:: /images/shapings-settings.png
```
comradekingu: The following paragraph is problematic, and I don't think I understood it.
I think it would help if the part of the Context called "Context" was instead called "description"(?).
My general view is that labels should be clickable to see them in the "shapings" view.
The ones that are of concern could be marked as a "things to consider" label. (This does away with the word shapings).
I don't understand why shapings have anything in particular to do with variants of a string differing in length either.

docs/devel/translations.rst
```diff
@@ -6,54 +6,47 @@ Managing translations
...
-incredibly easy:
+.. note::
 
-1. Add the translation file to VCS.
-2. Let Weblate update the repository (usually set up automatically, see
-   :ref:`update-vcs`).
+    Project admins can always start translation within Weblate directly.
 
+Language files added manually to the VCS are added when Weblate updates
+the repository if that is set up, see :ref:`update-vcs`).
```
comradekingu: When automatic addition or updating the repository is set up?

orangesunny: with every repo update. It depends when based on settings. Thanks for the heads up!

docs/devel/translations.rst
orangesunny: that would be the huge, overkill, change. The Context will stay Context. Shapings are new and subject to get a different name. We will keep an eye on improving its documentation.

Issue #3308, opened 2019-12-16
Inline editing of position and query


nijel:
**Is your feature request related to a problem? Please describe.**
With improved search based on query string instead of big forms, it seems better to allow editing the query string and position inline instead of poping up a dialog.

**Describe the solution you'd like**
Click on search term or position and input box appears in the position, enter submits it, escape cancels the edit.

**Describe alternatives you've considered**
At least proper popup placement would be great.

**Additional context**
<img width="442" alt="pozice" src="https://user-images.githubusercontent.com/212189/70911919-25457d80-2013-11ea-863b-9b61fd6aa420.png">

SaptakS:
@nijel I was trying to work on this. I still get a huge form dropdown on clicking in the query part. 
![image](https://user-images.githubusercontent.com/9530293/77858789-45e0be80-7223-11ea-93b9-e403c11615ed.png)

Should I remove the dropdown and replace it with inline editing completely? Because as in this the above screenshot, that area is showing `Strings needing action`. 

The position thing completely makes sense and I see that the dropdown is completely misplaced and I think inline editing totally makes sense there

nijel:
@SaptakS wrote on #3645:

> For making it inline editing, I think we need to make some design decisions. For example after clicking on the query field, do we show the applied filter (such as all queries) also or just replace that part by an input field? If we show the filter, I guess it needs to expand horizontally. Similarly for position, do we show an input field for the page number while still showing the total (I would prefer this)? or just an input box, with a max and min limit.
>
> I also think it will advantageous to move to bootstrap 4 because that gives us the scope to use flexbox directly. Flexbox really helps in designing grid-based UI.

nijel:
Switching to Bootstrap 4 is certainly an option, there were some issues with RTL support last time I looked, see https://github.com/WeblateOrg/weblate/issues/2916.

For the search field, I'd really like to make it more flexible to allow building more complex queries easily. I've mentioned some examples (GitLab and Sentry) in https://github.com/WeblateOrg/weblate/issues/3063#issuecomment-594458085 and https://github.com/WeblateOrg/weblate/issues/3063#issuecomment-606029363

SaptakS:
[...]

PR #3645, opened 2020-03-30
Fix the dropdown positions in the pagination tabs

SaptakS:

Refs: #3308 

I have fixed the position for the dropdowns in this PR.

For making it inline editing, I think we need to make some design decisions. For example after clicking on the query field, do we show the applied filter (such as `all queries`) also or just replace that part by an input field? If we show the filter, I guess it needs to expand horizontally. Similarly for position, do we show an input field for the page number while still showing the total (I would prefer this)? or just an input box, with a max and min limit.

I also think it will advantageous to move to bootstrap 4 because that gives us the scope to use flexbox directly. Flexbox really helps in designing grid-based UI.

cc: @nijel

### DIFF
--- a/weblate/static/style-bootstrap.css
+++ b/weblate/static/style-bootstrap.css
@@ -417,6 +417,12 @@ kbd {
 .unit-navi {
     min-width: 25em;
 }
+.unit-navi li {
+    float: left;
+}
+.unit-navi .dropdown-menu {
+    margin-top: 0;
+}
 .fullmodal .modal-dialog {
     width: auto;
 }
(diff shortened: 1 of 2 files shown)


nijel:

weblate/static/style-bootstrap.css
```diff
@@ -417,6 +417,12 @@ kbd {
 .unit-navi {
     min-width: 25em;
 }
+.unit-navi li {
+    float: left;
+}
```
nijel, line 422: JFYI: This broke rendering in RTL languages, see https://github.com/WeblateOrg/weblate/issues/3788, fixed by a3adbe9f58f173d222f7d0458758565696eb784f

SaptakS: Thanks for letting me know. Will keep in mind next time.

nijel: Basically anything asymmetric in the CSS (padding, margins, float, align,..) needs to be handled for RTL languages as well.

SaptakS: Got it. I should have thought of that.

Issue #3409, opened 2020-01-21
Search: Allow exact match


nijel:
**Is your feature request related to a problem? Please describe.**
With new search the exact match lookup is gone, all searches are substring (or regex).

**Describe the solution you'd like**
The exact match should be back in some form. Technically it's possible to perform it by regex (`r"^User$"`), but most people won't figure out and that's also worse performance wise than doing simple string comparison.

**Describe alternatives you've considered**
Regex works as a workaround, but it 's not a proper solution.

**Additional context**
_Originally posted by @Kingdutch in https://github.com/WeblateOrg/weblate/issues/3063#issuecomment-576602539_

nijel:
Fixed by https://github.com/WeblateOrg/weblate/pull/3653

PR #3653, opened 2020-04-01
Allows exact match in search

SaptakS:

Refs #3409 

I have added the code for allowing exact matches with a term `exact:true`. I haven't added tests or documentation yet. If this method looks good way to proceed with, I will add the corresponding tests and documentation updates.

cc: @nijel

### DIFF
--- a/weblate/utils/search.py
+++ b/weblate/utils/search.py
@@ -63,6 +63,25 @@ class QuotePlugin(whoosh.qparser.SingleQuotePlugin):
     expr = r"(^|(?<=\W))['\"](?P<text>.*?)['\"](?=\s|\]|[)}]|$)"
 
 
+class Exact(whoosh.query.Term):
+    """Class for queries with exact operator."""
+
+    pass
+
+
+class ExactPlugin(QuotePlugin):
+    """Exact match plugin with quotes to specify an exact term."""
+
+    class ExactNode(whoosh.qparser.syntax.TextNode):
+        qclass = Exact
+
+        def r(self):
+            return "Exact %r" % self.text
+
+    expr = r"\=(^|(?<=\W))['\"](?P<text>.*?)['\"](?=\s|\]|[)}]|$)"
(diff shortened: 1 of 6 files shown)


nijel:
Maybe it would be better to do the search using custom operator rather than introducing new lookup. I think it would be easier to use.

With your approach it is (please correct me if I misread the code, I didn't test it): `exact:1 AND source:something`

What I would find better: `source:=something`

See docs for existing set of operators: https://docs.weblate.org/en/latest/user/search.html#field-operators

SaptakS:
@nijel yes correct. I like the idea of the operator as well. Definitely gonna try and change the implementation to use operators over lookups.

weblate/utils/tests/test_search.py
```diff
@@ -89,6 +89,9 @@ def test_field(self):
             Q(source__substring="hello") & Q(target__substring="world"),
         )
         self.assert_query("location:hello.c", Q(location__substring="hello.c"))
+
```
nijel, line 92:
```suggestion

```

weblate/utils/search.py
```diff
@@ -282,6 +286,8 @@ def query_sql(obj):
             return has_sql(obj.text)
         if obj.fieldname == "is":
             return is_sql(obj.text)
+        if obj.text.startswith("="):
```
nijel, line 289: Looking into text is not the best approach (`source:"=value"` should probably search for the `=value` instead of doing exact match).

But it's probably good enough until the query parser is rewritten (see https://github.com/WeblateOrg/weblate/issues/3229).

SaptakS: I was also honestly wondering the same while implementing this. That it should probably search for `=value`. But since the parser always results in it being a Term instance, I couldn't figure a better way of segregating it apart from looking into the text. Let me know if you think there is a different better way I can do this.

nijel:

PR #3657, opened 2020-04-02
Command line for manual backup

mayeulk:
Suggests a command line using tar and xz to do the backup with a cron task.


nijel:

PR #3700, opened 2020-04-09
Creates basic API endpoints for users and groups

### LINKED ISSUE #3607: API for users and groups
In many cases it would be useful to have an API to manage users, their group membership and permissions.

Extend API with user management.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/90207917-api-for-users-and-groups?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SaptakS:

Currently Done:
- Adds basic endpoint for User to create, patch, list and delete user
- Adds basic endpoint for Group to create, patch, list and delete Group based on the non-relational fields.

Needs to be done:
- [x] Add relational links to assign Users to Groups

Fixes #3607

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -287,7 +287,7 @@ Groups
     :>json string name: name of a group
     :>json int project_selection: integer corresponding to group of projects
     :>json int language_selection: integer corresponding to group of languages
-    :>json array roles: array of associated roles
+    :>json array roles: link to associated groups; see :http:get:`/api/roles/(int:id)/`
 
 .. http:put:: /api/groups/(int:id)/
 
@@ -297,8 +297,7 @@ Groups
     :type id: int
     :>json string name: name of a group
     :>json int project_selection: integer corresponding to group of projects
-    :>json int language_selection: integer corresponding to group of languages
-    :>json array roles: array of associated roles
+    :>json int language_selection: integer corresponding to group of Languages
 
 .. http:patch:: /api/groups/(int:id)/
 
@@ -309,7 +308,6 @@ Groups
     :>json string name: name of a group
     :>json int project_selection: integer corresponding to group of projects
     :>json int language_selection: integer corresponding to group of languages
-    :>json array roles: array of associated roles
 
 .. http:delete:: /api/groups/(int:id)/
 
@@ -317,6 +315,41 @@ Groups
 
     :param id: Group's ID
     :type id: int
+
+
+Roles
++++++++++
+
+.. http:get:: /api/roles/
+
+    Returns a list of all roles associated with user. If user is superuser, then list of all
+    existing roles is returned.
+
+    .. seealso::
+
+        Roles object attributes are documented at :http:get:`/api/roles/(int:id)/`.
+
+.. http:get:: /api/roles/(int:id)/
+
+    Returns information about a role.
+
+    :param id: Role ID
+    :type id: int
+    :>json string name: Role code
+    :>json array permissions: list of name of permissions
+
+    **Example JSON data:**
+
+    .. code-block:: json
+
+        {
+            "name": "Access repository",
+            "permissions": [
+                "Access the internal repository",
+                "View upstream repository location"
+            ],
+            "url": "http://example.com/api/roles/1/",
+        }
 
 
 Languages
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -336,7 +336,7 @@ Roles
     :param id: Role ID
     :type id: int
     :>json string name: Role code
-    :>json array permissions: list of name of permissions
+    :>json array permissions: list of codenames of permissions
 
     **Example JSON data:**
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -193,6 +193,132 @@ API Entry Point
             "languages":"http://example.com/api/languages/"
         }
 
+
+Users
++++++++++++++++
+
+.. versionadded:: 4.0
+
+.. http:get:: /api/users/
+
+    Returns a list of users if you have permissions to see manage users. If not, then you get to see
+    only your own details.
+
+    .. seealso::
+
+        Users object attributes are documented at :http:get:`/api/users/(str:username)/`.
+
+.. http:get:: /api/users/(str:username)/
+
+    Returns information about users.
+
+    :param username: User's username
+    :type username: string
+    :>json string username: username of a user
+    :>json string full_name: full name of a user
+    :>json string email: email of a user
+    :>json boolean is_superuser: whether the user is a super user
+    :>json boolean is_active: whether the user is active
+    :>json string date_joined: date the user is created
+    :>json array groups: link to associated groups; see :http:get:`/api/groups/(str:name)/`
+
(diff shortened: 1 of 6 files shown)


weblate/api/views.py
```diff
@@ -250,6 +252,52 @@ def repository(self, request, **kwargs):
...
+    serializer_class = GroupSerializer
+    lookup_field = "name"
+
+    def get_queryset(self):
+        return Group.objects.all()
+
+    def partial_update(self, request, *args, **kwargs):
+        serializer_context = {
+            "request": request,
+        }
```
nijel, line 291: Any reason for not using `get_serializer_context`?

weblate/api/views.py
```diff
@@ -250,6 +252,52 @@ def repository(self, request, **kwargs):
...
+
+class GroupViewSet(WeblateViewSet, CreateModelMixin, DestroyModelMixin):
+    """Groups API."""
+
+    queryset = Group.objects.none()
+    serializer_class = GroupSerializer
+    lookup_field = "name"
+
+    def get_queryset(self):
+        return Group.objects.all()
```
nijel, line 286: Needs permission check, user should be probably able to only list own groups, all would be listed only with proper permission.

```suggestion
        if request.user.has_perm("group.edit"):
            return Group.objects.all()
        return request.user.groups.all()
```

weblate/api/views.py
```diff
@@ -250,6 +252,52 @@ def repository(self, request, **kwargs):
...
+        instance = User.objects.get(username=kwargs.get("username"))
+        serializer = self.serializer_class(
+            instance, data=request.data, partial=True, context=serializer_context
+        )
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data)
+
+
+class GroupViewSet(WeblateViewSet, CreateModelMixin, DestroyModelMixin):
```
nijel, line 278: Maybe using `UpdateModelMixin` to provide partial_update is simpler? Or directly `viewsets.ModelViewSet`, you definitely do not need the extras provided by `WeblateViewSet`, these are useful only for file uploads/downloads.

weblate/api/views.py
```diff
@@ -250,6 +252,111 @@ def repository(self, request, **kwargs):
...
+        self.perm_check(request)
+        return super().update(request, *args, **kwargs)
+
+    def create(self, request, *args, **kwargs):
+        self.perm_check(request)
+        return super().create(request, *args, **kwargs)
+
+    def destroy(self, request, *args, **kwargs):
+        self.perm_check(request)
+        return super().destroy(request, *args, **kwargs)
```
nijel, line 281: The user removal should be done by `weblate.accounts.utils.remove_user`. We do not actually delete the user, but remove all account data (to avoid need to rewrite lot of relations).

Not sure if we want self removal here,

SaptakS: Fixed that.

docs/api.rst
```diff
@@ -193,6 +193,132 @@ API Entry Point
...
+.. versionadded:: 4.0
+
+.. http:get:: /api/groups/
+
+    Returns a list of groups if you have permissions to see manage groups. If not, then you get to see
+    only the groups the user is a part of.
+
+    .. seealso::
+
+        Users object attributes are documented at :http:get:`/api/groups/(str:name)/`.
```
nijel, line 279:
```suggestion
        Group object attributes are documented at :http:get:`/api/groups/(str:name)/`.
```

weblate/api/views.py
```diff
@@ -253,6 +256,115 @@ def repository(self, request, **kwargs):
...
+        self.perm_check(request)
+        return super().create(request, *args, **kwargs)
+
+    def destroy(self, request, *args, **kwargs):
+        self.perm_check(request)
+        instance = self.get_object()
+        remove_user(instance, request)
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
+    def partial_update(self, request, *args, **kwargs):
```
nijel, line 289: Is this method needed? It should be already available in ModelViewSet. It calls update under the hood, so the permission check is covered as well.

weblate/api/serializers.py
```diff
@@ -113,6 +114,89 @@ def validate_code(self, value):
...
+        model = Group
+        fields = (
+            "name",
+            "roles",
+            "project_selection",
+            "language_selection",
+            "url",
+        )
+        extra_kwargs = {
+            "url": {"view_name": "api:group-detail", "lookup_field": "name"}
```
nijel, line 177: I'm not sure about this. The group name can contain any chars and using that in URL might be problematic...

weblate/api/views.py
```diff
@@ -253,6 +256,115 @@ def repository(self, request, **kwargs):
...
+
+    def create(self, request, *args, **kwargs):
+        self.perm_check(request)
+        return super().create(request, *args, **kwargs)
+
+    def destroy(self, request, *args, **kwargs):
+        self.perm_check(request)
+        return super().destroy(request, *args, **kwargs)
+
+    def partial_update(self, request, *args, **kwargs):
```
nijel, line 354: Same here - I don't think this method is needed.

weblate/api/serializers.py
SaptakS: That's true. I will update this to use ID instead then. And also for the association with users.

nijel: For users the username is already used in URLs, so that should be safer in this regard.

weblate/api/serializers.py
```diff
@@ -113,6 +114,87 @@ def validate_code(self, value):
...
+
+    def validate_name(self, value):
+        check_query = Role.objects.filter(name=value)
+        if not check_query.exists():
+            raise serializers.ValidationError("Role with this role name was not found.")
+        return value
+
+
+class GroupSerializer(serializers.ModelSerializer):
+    roles = RoleSerializer(many=True)
```
nijel, line 165: Going again through the patch, I'm not sure this is good approach as we want to add roles to the API in the future (to be able to manage permissions). Having something different in place right now will break the API once permission management will be added. So I'd either remove roles altogether for now, or have read only roles viewset + serializer which will be extended in future without breaking what would be introduced right now.

SaptakS: True. I was conflicted about this too. maybe can keep read-only like `source_language` was from `Projects` till recently.

nijel: Yes, having that read only for now is probably simplest approach.

SaptakS: Done!

nijel:

nijel:
I've made some further improvements on top of your changes:

* e6c4c53785 API: Expose list of projects and components
* 8b8ab2d18e Docs: Fix attribute description
* 80884ed737 Docs: Remove too long underlines
* 54c904da45 Docs: Document changes for #3700

PR #3727, opened 2020-04-13
Adds basic API endpoints for roles and assign roles to groups

### LINKED ISSUE #3647: API for managing permissions
Once the API for component lists is implemented (https://github.com/WeblateOrg/weblate/issues/3291) and it becomes possible to assign permissions at the component/component list level (https://github.com/WeblateOrg/weblate/issues/3646) it would be great if that could be done programmatically via the RESTful API.

A set of API endpoints, which allows for managing project/component list/component level permissions for given users. At the very least it should be possible to grant access to each of those items to a user if a project is private.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/90814815-api-for-managing-permissions?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SaptakS:

- Adds Role API to create roles with permissions, edit roles, add permission codenames to roles and remove roles.
- Assign roles to a group
- Assign projects, components, componentlist to a group

Fixes #3647

### DIFF
--- a/weblate/api/serializers.py
+++ b/weblate/api/serializers.py
@@ -143,9 +143,20 @@ class PermissionSerializer(serializers.RelatedField):
     def to_representation(self, value):
         return value.codename
 
+    def get_queryset(self):
+        return Permission.objects.all()
+
+    def to_internal_value(self, data):
+        check_query = Permission.objects.filter(codename=data)
+        if not check_query.exists():
+            raise serializers.ValidationError(
+                "Permission with this codename was not found."
+            )
+        return data
+
 
 class RoleSerializer(serializers.ModelSerializer):
-    permissions = PermissionSerializer(read_only=True, many=True)
+    permissions = PermissionSerializer(many=True)
 
     class Meta:
         model = Role
@@ -155,6 +166,26 @@ class RoleSerializer(serializers.ModelSerializer):
             "url",
         )
         extra_kwargs = {"url": {"view_name": "api:role-detail", "lookup_field": "id"}}
+
+    def create(self, validated_data):
+        permissions_validated = validated_data.pop("permissions", [])
+        role = Role.objects.create(**validated_data)
+        for codename in permissions_validated:
+            permission = Permission.objects.get(
+                codename=codename
+            )
+            role.permissions.add(permission)
+        return role
+
+    def update(self, instance, validated_data):
+        permissions_validated = validated_data.pop("permissions", [])
+        instance.name = validated_data.get('name', instance.name)
+        for codename in permissions_validated:
+            permission = Permission.objects.get(
+                codename=codename
+            )
+            instance.permissions.add(permission)
+        return instance
 
 
 class GroupSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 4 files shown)


weblate/api/serializers.py
```diff
@@ -156,6 +167,24 @@ class Meta:
         )
         extra_kwargs = {"url": {"view_name": "api:role-detail", "lookup_field": "id"}}
 
+    def create(self, validated_data):
+        permissions_validated = validated_data.pop("permissions", [])
+        role = Role.objects.create(**validated_data)
+        for codename in permissions_validated:
+            permission = Permission.objects.get(codename=codename)
+            role.permissions.add(permission)
```
nijel, line 175: Maybe this can be done simpler?

```suggestion
        role.permissions.add(Permission.objects.get(codename__in=permissions_validated))
```

weblate/api/serializers.py
```diff
@@ -156,6 +167,24 @@ class Meta:
...
+        return role
+
+    def update(self, instance, validated_data):
+        permissions_validated = validated_data.pop("permissions", [])
+        instance.name = validated_data.get("name", instance.name)
+        instance.save()
+        instance.permissions.clear()
+        for codename in permissions_validated:
+            permission = Permission.objects.get(codename=codename)
+            instance.permissions.add(permission)
```
nijel, line 185: This block could use similar simplification as in create, but there is additional concern - this will only *add* permisssions, but there is no way to remove them. Probably `set()` should be used here instead of `add()`, but interaction with partial updates has to be defined and tested. In case PATCH request is missing permissions, it should remain untouched.

SaptakS: @nijel I did the `instance.permissions.clear()` because of that. So as to remove them. Right now, yes, PATCH will need to have all permissions, missing permissions will get removed. So need to fix that. I was thinking of adding a separate definition for partial_update in the viewset for PATCH.

nijel: Ah, I didn't notice the `clear()` invocation. Anyway `set()` is better approach here as it avoids removing and adding back not changed permissions.

As for PATCH, you can figure out whether it is being done by looking at `self.partial` here.

weblate/api/serializers.py
SaptakS: I just realized there might be a problem with this. What if someone inputs more than one permission in the permissions array while creating a role. `Permission.objects.get` will throw an error then, right? `.get()` usually expects only one result of query.

nijel: Yes, filter was supposed to be there. No need to add all though.

weblate/api/views.py
```diff
@@ -339,9 +339,196 @@ def destroy(self, request, *args, **kwargs):
...
+        methods=["delete"],
+        url_path="componentlist/(?P<component_list_id>[^/.]+)",
+    )
+    def delete_componentlist(self, request, id, component_list_id):
+        obj = self.get_object()
+        self.perm_check(request)
+        try:
+            ComponentList.objects.get(
+                pk=int(component_list_id),
+            )
```
nijel, line 478:
```suggestion
            ComponentList.objects.get(pk=int(component_list_id))
```

I suggest you to install pre-commit, it will do the formatting automatically before git commit, see https://docs.weblate.org/en/latest/contributing.html#coding-standard

SaptakS: On it.

nijel:

Issue #3651, opened 2020-03-31
optional check/fixup for ensuring URLs in source match URLs in translation


eighthave:
**Is your feature request related to a problem? Please describe.**

On the f-droid.org and guardianproject.info websites, we are localizing the content, but the URLs are the same for all languages.  I have some hack link checkers to make sure that the sorted list of URLs in the source matches the sorted list of links in the translation.  This is obviously not applicable to websites that translate the paths in the URLs.

**Describe the solution you'd like**

An optional check called something like _markdown-untranslated-links_ with a potential fix up that uses a similarity check for the URLs in the translation versus the links in the source.

**Describe alternatives you've considered**

Custom script on our end, but we're close to upstreaming all of our custom checkers to Weblate.

**Additional context**

This is something that @uniqx and I will implement if you agree it makes sense.
If so, please add this issue to the Linguine project.

nijel:
Right now, the remote URLs are not checked because these might be localized:

https://github.com/WeblateOrg/weblate/blob/4c08ec868b5a4484003056563f74dbaa8eed86dc/weblate/checks/markup.py#L259-L265

eighthave:
So @uniqx and I just dug into the existing code.  It looks quite close, the only issue is that this check will mark a string as bad if source and translation have the same link collections, but they are in a different order.  Since the translator might need to reorder the words that are linked in order to write correctly, we propose that this comparison should compare the sorted lists, i.e.:
```python
 return sorted(tgt_anchors) != sorted(src_anchors) 
```

PR #3733, opened 2020-04-14
make md-text link checks order-independent

uniqx:

---

This PR makes markdown links for `md-link` checker order independent.

This closes https://github.com/WeblateOrg/weblate/issues/3651

### DIFF
--- a/weblate/checks/tests/test_markup_checks.py
+++ b/weblate/checks/tests/test_markup_checks.py
@@ -186,6 +186,36 @@ class MarkdownLinkCheckTest(CheckTestCase):
         )
 
 
+class MarkdownLinkCheckMultipleOrderIndependentLinksTest(CheckTestCase):
+    check = MarkdownLinkCheck()
+
+    def setUp(self):
+        super(MarkdownLinkCheckMultipleOrderIndependentLinksTest, self).setUp()
+
+        self.test_good_matching = (
+            "[Weblate](#weblate) has an [example]({{example}}) "
+            "for illustrating the useage of [Weblate](#weblate)",
+            "Ein [Beispiel]({{example}}) in [Webspät](#weblate) "
+            "illustriert die Verwendung von [Webspät](#weblate)",
+            "md-text",
+        )
+
+        self.test_failure_1 = (
+            "[Weblate](#weblate) has an [example]({{example}}) "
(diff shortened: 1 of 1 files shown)


weblate/checks/markup.py
```diff
@@ -256,6 +256,9 @@ def check_single(self, source, target, unit):
         if len(src_match) != len(tgt_match):
             return True
 
+        src_match.sort()
+        tgt_match.sort()
```
nijel, line 260: I think it's already order independent - what is being compared are sets. Are these changes really useful?

eighthave: Using sets will miss if a URL is used more than once in a paragraph.  Using sorted lists will confirm that the source and translation have the same number of links as well as matching links.

eighthave: Oops, now I see what you're talking about, those vars are set comprehensions not list comprehensions.

nijel: But you are right with missing the double usage, it's currently not detected.

uniqx: Actually double useage is not an issue either, because there's a length check before the parsed links get turned into a set.

uniqx:
Turns out our initial assumptions were wrong and the `md-link` check is already dealing very well with multiple links. So this PR only adds another test-case for verifying that.

weblate/checks/markup.py
eighthave: You're right!  Could you add a test of check failure so this is well documented for the future as a requirement?  Something like:
```python
        self.test_failure_1 = (
            "[Weblate](#weblate) has an [example]({{example}}) "
            "for illustrating the useage of [Weblate](#weblate)",
            "Ein [Beispiel]({{example}}) in [Webspät](#weblate) "
            "illustriert die Verwendung von [Webspät](#Webspät)",
            "md-text",
        )
        self.test_failure_2 = (
            "[Weblate](#weblate) has an [example]({{example}}) "
            "for illustrating the useage of [Weblate](#weblate)",
            "Ein [Beispiel]({{example}}) in [Webspät](#weblate) "
            "illustriert die Verwendung von Webspät",
            "md-text",
        )
```

uniqx: I've added the failure cases now.

nijel:

Issue #3648, opened 2020-03-31
"markdown-text" as standardized flag name for "md-text"


eighthave:
**Is your feature request related to a problem? Please describe.**

The [standard gettext flags](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html#PO-Files) all use full names without abbreviations.  For example `python-format` rather than `py-format`.  We would like to make _po4a_ automatically add a Markdown flag to the PO files, so that Weblate will automatically add the relevant checks.

Would it be possible to support the name "markdown-text" as a name for this flag?

**Describe the solution you'd like**

Anyway that makes it so that "markdown-text" can be used as a PO flag and do what Weblate's "md-text" currently does.  Even better, Weblate would also use the name "markdown-text".

**Describe alternatives you've considered**

We've considered just using "md-text", but that would make it the only standard PO flag that uses an abbreviation.  And "md" is not clear in many contexts.

**Additional context**

This is something that @uniqx and I could implement if you agree it makes sense. 
 If so, please add this issue to the Linguine project.

nijel:
I think it's hard to bring consistency into this. We also have `rst-text`, do you want to rename it to `restructuredtext-text`?

Introducing aliases might be way to address this. Such would be easy to implement in the flags parsing code:

https://github.com/WeblateOrg/weblate/blob/05b3b2f50273a12c2e2b6ba234f1fbfe504bf1b2/weblate/checks/flags.py#L96-L103

eighthave:
aliases would be fine by me.  I can understand migrating away from `md-text` would be a lot of work.

nijel:
It can be stored in translation files for anybody using Weblate, so breaking existing flags is not really an option.

eighthave:
How about a merge request that implements aliases and adds the "markdown-text" alias?

nijel:
Sounds good.

uniqx:
@nijel what would be the correct way to implement this alias?

I've read through all involved Check classes several times, but I couldn't find a clean way or any hint's on how to implement aliases for flags.

([weblate/checks/base.py](https://github.com/WeblateOrg/weblate/blob/weblate-3.11.3/weblate/checks/base.py), [weblate/checks/markup.py](https://github.com/WeblateOrg/weblate/blob/weblate-3.11.3/weblate/checks/markup.py))

nijel:
[...]

PR #3735, opened 2020-04-15
add alias markdown-text for md-text flag

uniqx:
Add `markdown-text` as alias for the `md-text` flag.

closes https://github.com/WeblateOrg/weblate/issues/3648


weblate/checks/flags.py
```diff
@@ -100,6 +104,9 @@ def parse(flags):
             value = flag.strip()
             if not value or value in ("fuzzy", "#"):
                 continue
+            elif value in FLAG_ALIASES.keys():
```
nijel, line 107:
```suggestion
            if value in FLAG_ALIASES.keys():
```

No need for else here.

nijel:

PR #3740, opened 2020-04-16
Squash migrations

nijel:
This should provide faster times for fresh installs skipping some of the
legacy code.

- [ ] Every commit has a message which describes it


nijel:
Not that beneficial right now because the most expensive things are still executed - trans and checks migrations, which depend on each other because of the unitdata relation change. To bring benefits, we would have to stop supporting migrations across 4.0 (similarly as we did with 3.0)...

PR #3766, opened 2020-04-22
Query Field: Separates query field from the pagination part

SaptakS:
Separates the query search field from the pagination part and adds filters
and sorting options in UI

### DIFF
--- a/weblate/static/loader-bootstrap.js
+++ b/weblate/static/loader-bootstrap.js
@@ -1528,7 +1528,11 @@ $(function () {
         $button.attr('data-field', $this.data('field'));
         $button.find('span.search-label').text($this.text());
         if ($this.closest('.query-field').length) {
-            $('#id_q').insertAtCaret(' ' + $this.data('field') + ' ');
+            if ($this.data('sort')) {
+                $('#id_sort_by').val($this.data('sort'));
+            } else {
+                $('#id_q').insertAtCaret(' ' + $this.data('field') + ' ');
+            }
         }
     });
     $('.search-group input').on('keydown', function (event) {
--- a/weblate/static/loader-bootstrap.js
+++ b/weblate/static/loader-bootstrap.js
@@ -1529,11 +1529,15 @@ $(function () {
         $button.find('span.search-label').not('#query-sort-toggle span').text($this.text());
         if ($this.data('sort')) {
             $('#id_sort_by').val($this.data('sort'));
-            $this.closest('form').submit();
+            if ($('.pagination').length) {
+                $this.closest('form').submit();
+            }
         }
         if ($this.closest('.query-field').length) {
             $('#id_q').val($this.data('field'));
-            $this.closest('form').submit();
+            if ($('.pagination').length) {
+                $this.closest('form').submit();
+            }
         }
     });
     $('#query-sort-toggle').click(function () {
@@ -1545,13 +1549,21 @@ $(function () {
         } else {
             $('#id_sort_direction').val("");
         }
-        $this.closest('form').submit();
+        if ($('.pagination').length) {
+            $this.closest('form').submit();
+        }
     });
     $('.search-group input').not('#id_q').on('keydown', function (event) {
         if (event.key === "Enter") {
             $(this).closest('.input-group').find('.search-add').click();
             event.preventDefault();
             return false;
+        }
+    });
+    $('#id_q').on('keydown', function (event) {
+        console.log("hello")
+        if (event.key === "Enter") {
+            $(this).closest('form').submit();
         }
     });
     $('.search-add').click(function () {
--- a/weblate/static/loader-bootstrap.js
+++ b/weblate/static/loader-bootstrap.js
@@ -1526,16 +1526,26 @@ $(function () {
         var $this = $(this);
         var $button = $this.closest('.input-group-btn').find('button');
         $button.attr('data-field', $this.data('field'));
-        $button.find('span.search-label').text($this.text());
-        if ($this.closest('.query-field').length) {
-            if ($this.data('sort')) {
-                $('#id_sort_by').val($this.data('sort'));
-                $this.closest('form').submit();
-            } else {
-                $('#id_q').val($this.data('field'));
-                $this.closest('form').submit();
-            }
+        $button.find('span.search-label').not('#query-sort-toggle span').text($this.text());
+        if ($this.data('sort')) {
+            $('#id_sort_by').val($this.data('sort'));
+            $this.closest('form').submit();
         }
+        if ($this.closest('.query-field').length) {
+            $('#id_q').val($this.data('field'));
+            $this.closest('form').submit();
+        }
+    });
+    $('#query-sort-toggle').click(function () {
+        var $this = $(this);
+        var $label = $this.find('span.search-label');
+        $label.toggle();
+        if (!$('#id_sort_direction').val()) {
+            $('#id_sort_direction').val("-");
+        } else {
+            $('#id_sort_direction').val("");
+        }
+        $this.closest('form').submit();
     });
     $('.search-group input').not('#id_q').on('keydown', function (event) {
         if (event.key === "Enter") {
--- a/weblate/static/loader-bootstrap.js
+++ b/weblate/static/loader-bootstrap.js
@@ -1510,22 +1510,24 @@ $(function () {
     });
 
     // Show the correct toggle button
-    var sort_name = $('#query-sort-dropdown span.search-label').text();
-    var sort_dropdown_value = $(".sort-field li a").filter(function() {
-        return $(this).text() == sort_name;
-    }).data('sort');
-    var sort_value = $('#id_sort_by').val();
-    console.log(sort_dropdown_value);
-    console.log(sort_value);
-    if (
-        sort_value.replace("-", "") === sort_dropdown_value.replace("-", "")
-        && sort_value !== sort_dropdown_value
-    ) {
-        $('#query-sort-toggle .asc').hide();
-        $('#query-sort-toggle .desc').show();
-    } else {
-        $('#query-sort-toggle .desc').hide();
-        $('#query-sort-toggle .asc').show();
+    if ($('.sort-field').length) {
+        var sort_name = $('#query-sort-dropdown span.search-label').text();
+        var sort_dropdown_value = $(".sort-field li a").filter(function() {
+            return $(this).text() == sort_name;
+        }).data('sort');
+        var sort_value = $('#id_sort_by').val();
+        if (sort_dropdown_value) {
+            if (
+                sort_value.replace("-", "") === sort_dropdown_value.replace("-", "")
+                && sort_value !== sort_dropdown_value
+            ) {
+                $('#query-sort-toggle .asc').hide();
+                $('#query-sort-toggle .desc').show();
+            } else {
+                $('#query-sort-toggle .desc').hide();
+                $('#query-sort-toggle .asc').show();
+            }
+        }
     }
 
     /* Branch loading */
--- a/weblate/static/loader-bootstrap.js
+++ b/weblate/static/loader-bootstrap.js
@@ -1525,8 +1525,11 @@ $(function () {
     $('.search-group li a').click(function () {
         var $this = $(this);
         var $button = $this.closest('.input-group-btn').find('button');
-        $button.data('field', $this.data('field'));
+        $button.attr('data-field', $this.data('field'));
         $button.find('span.search-label').text($this.text());
+        if ($this.closest('.query-field').length) {
+            $('#id_q').insertAtCaret(' ' + $this.data('field') + ' ');
+        }
     });
     $('.search-group input').on('keydown', function (event) {
         if (event.key === "Enter") {
--- a/weblate/static/loader-bootstrap.js
+++ b/weblate/static/loader-bootstrap.js
@@ -1509,6 +1509,25 @@ $(function () {
         $(this).parent().find('input').attr('disabled', '1');
     });
 
+    // Show the correct toggle button
+    var sort_name = $('#query-sort-dropdown span.search-label').text();
+    var sort_dropdown_value = $(".sort-field li a").filter(function() {
+        return $(this).text() == sort_name;
+    }).data('sort');
+    var sort_value = $('#id_sort_by').val();
+    console.log(sort_dropdown_value);
+    console.log(sort_value);
+    if (
+        sort_value.replace("-", "") === sort_dropdown_value.replace("-", "")
+        && sort_value !== sort_dropdown_value
+    ) {
+        $('#query-sort-toggle .asc').hide();
+        $('#query-sort-toggle .desc').show();
+    } else {
+        $('#query-sort-toggle .desc').hide();
+        $('#query-sort-toggle .asc').show();
+    }
+
     /* Branch loading */
     $('.branch-loader select[name=component]').change(function () {
         var $this = $(this);
--- a/weblate/static/loader-bootstrap.js
+++ b/weblate/static/loader-bootstrap.js
@@ -1544,11 +1544,15 @@ $(function () {
         var $this = $(this);
         var $label = $this.find('span.search-label');
         $label.toggle();
-        if (!$('#id_sort_direction').val()) {
-            $('#id_sort_direction').val("-");
-        } else {
-            $('#id_sort_direction').val("");
-        }
+        var sort_params = $('#id_sort_by').val().split(",")
+        sort_params.forEach(function(param, index) {
+            if (param.indexOf("-") !== -1) {
+                sort_params[index] = param.replace("-", "");
+            } else {
+                sort_params[index] = `-${param}`;
+            }
+        });
+        $('#id_sort_by').val(sort_params.join(","));
         if ($('.pagination').length) {
             $this.closest('form').submit();
         }
--- a/weblate/static/loader-bootstrap.js
+++ b/weblate/static/loader-bootstrap.js
@@ -1529,13 +1529,13 @@ $(function () {
         $button.find('span.search-label').not('#query-sort-toggle span').text($this.text());
         if ($this.data('sort')) {
             $('#id_sort_by').val($this.data('sort'));
-            if ($('.pagination').length) {
+            if ($this.closest('.result-page-form').length) {
                 $this.closest('form').submit();
             }
         }
         if ($this.closest('.query-field').length) {
             $('#id_q').val($this.data('field'));
-            if ($('.pagination').length) {
+            if ($this.closest('.result-page-form').length) {
                 $this.closest('form').submit();
             }
         }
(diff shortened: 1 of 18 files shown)


weblate/static/loader-bootstrap.js
```diff
@@ -1525,8 +1525,11 @@ $(function () {
     $('.search-group li a').click(function () {
         var $this = $(this);
         var $button = $this.closest('.input-group-btn').find('button');
-        $button.data('field', $this.data('field'));
+        $button.attr('data-field', $this.data('field'));
```
nijel, line 1528: Why this change? I'm not saying it is bad (I have no clue), but I'm curious for reasons.

weblate/trans/forms.py
```diff
@@ -657,7 +662,7 @@ def __init__(self, user, *args, **kwargs):
 
     def get_name(self):
         """Return verbose name for a search."""
-        return FILTERS.get_search_name(self.cleaned_data["q"])
+        return FILTERS.get_search_name(self.cleaned_data["q"].strip())
```
nijel, line 665: Any cleanups should IMHO go to `get_search_name` method.

weblate/trans/filter.py
```diff
@@ -128,3 +128,32 @@ def get_filter_choice(project=None):
             for label in project.label_set.values_list("name", flat=True)
         )
     return result
+
+
+def get_search_query_choices():
```
nijel, line 133: This should reuse FilterRegistry otherwise we're again duplicating the code (`get_filter_choice` should be rewritten at some point as well).

weblate/static/loader-bootstrap.js
SaptakS: So, it doesn't affect a lot to be fair. But we have `data-field` mentioned in the DOM of the button. And `.data()` updates the data value inside the javascript scope, but it doesn't update the DOM.
It's mentioned in the [docs](https://api.jquery.com/data/)
So even though the value in JS for `data-field` changes, in DOM, it still stays same, and I felt might cause problems in the future for cross-scopes.

weblate/trans/views/edit.py
```diff
@@ -141,7 +142,11 @@ def search(translation, request):
     name = form.get_name() if form_valid else ""
 
     # Grab unit IDs
-    unit_ids = list(allunits.order().values_list("id", flat=True))
+    unit_ids = list(
+        allunits.order_by(request.GET.get("sort_by", "position")).values_list(
```
nijel, line 146: We probably don't want to get this directly from request as we will do multi field ordering as well. The default one should be by priority and position.

weblate/trans/views/edit.py
```diff
@@ -141,7 +142,12 @@ def search(translation, request):
     name = form.get_name() if form_valid else ""
 
     # Grab unit IDs
-    unit_ids = list(allunits.order().values_list("id", flat=True))
+    sort_by_list = (
+        request.GET.get("sort_by").split(",")
+        if request.GET.get("sort_by")
+        else ["-priority", "position"]
```
nijel, line 148: I'd rather hard-code available choices here. This way user can do any ordering which might result in huge load on the database and DOS.

nijel: ...maybe also moving such helper method to UnitQuerySet and reuse it in weblate/trans/views/search.py

SaptakS: Yeah. Makes sense. I will put it into a helper function with a hardcoded list of allowed sortings.

weblate/static/loader-bootstrap.js
```diff
@@ -1525,8 +1525,15 @@ $(function () {
...
         $button.find('span.search-label').text($this.text());
+        if ($this.closest('.query-field').length) {
+            if ($this.data('sort')) {
+                $('#id_sort_by').val($this.data('sort'));
+            } else {
+                $('#id_q').insertAtCaret(' ' + $this.data('field') + ' ');
```
nijel, line 1534: The approach of inserting the query is not really user-friendly - if you keep changing the options, the resulting query will not make much sense. I think it makes sense here to directly perform the query after clicking on the option.

Auto submit on sort also might be more friendly than this.

weblate/templates/translate.html
```diff
@@ -35,65 +35,66 @@
...
+<div class="row">
+  <div class="col-sm-9">
+    <div class="panel-group">
+      <form action="{{ unit.translation.get_translate_url }}" method="GET">
+        {% include "snippets/query-field.html" with field=search_form.q  %}
+      </form>
+    </div>
+  </div>
+  <div class="col-sm-3">
+    <div class="btn-group">
```
nijel, line 59:
```suggestion
    <div class="btn-group pull-right flip">
```

How about right aligning this? But it is probably not that much better.

weblate/static/loader-bootstrap.js
SaptakS: >  I think it makes sense here to directly perform the query after clicking on the option.

I was thinking in the term like what if someone selects a query from dropdown but also wants to add their own search term to it? I mean in that case, they have to wait for the result and then enter the additional parts.

SaptakS: > Auto submit on sort also might be more friendly than this.

This I kind of do agree with. I will update this.

weblate/templates/translate.html
SaptakS: I will check how it looks UI wise. I just don't wanna make it seem un-intuitive too much to relate that the pagination is related to the translation.

weblate/static/loader-bootstrap.js
nijel: > I was thinking in the term like what if someone selects a query from dropdown but also wants to add their own search term to it? I mean in that case, they have to wait for the result and then enter the additional parts.

I consider this as a selection of predefined queries. To build custom query use the input box. This is how it also behaves on Sentry and I quite like it there as an user (the left dropdown was not shown in https://github.com/WeblateOrg/weblate/issues/3308#issuecomment-617066002).

The problem with adding strings is that you quickly end up with query that doesn't make sense.

Also, the auto submission avoids weird effect of changing width of the text input selecting the values.

SaptakS: Sounds good. Just one question. Let's say someone made a query using the query field before, pressed search, got the result. Now when I click on the dropdown list (or sorting), the query term is still taken into consideration when fetching the new results, right?

nijel: In Sentry it completely discards the current query and it was never an issue for me.

weblate/templates/translate.html
nijel: Let's skip UI twiddling for now, I will get proper UI design for this part.

SaptakS:
Ah nice. AccessLint is a good addition. I will add the fixes for them.

weblate/trans/models/unit.py
```diff
@@ -116,6 +116,17 @@ def same(self, unit, exclude=True):
             result = result.exclude(pk=unit.id)
         return result
 
+    def order_by_request(self, sort_list_request):
+        available_sort_choices = ["-priority", "position", "context", "num_words"]
+        sort_list = []
+        for choice in sort_list_request:
+            if choice in available_sort_choices:
+                sort_list.append(choice)
```
nijel, line 124: This is more readable and effective:

```suggestion
        sort_list = [choice choice in sort_list_request if choice in available_sort_choices]
```

weblate/trans/models/unit.py
```diff
@@ -116,6 +116,17 @@ def same(self, unit, exclude=True):
...
+        if not len(sort_list):
```
nijel, line 125: Calling len is not needed here:

```suggestion
        if not sort_list:
```

weblate/trans/views/edit.py
```diff
@@ -141,7 +141,11 @@ def search(translation, request):
     name = form.get_name() if form_valid else ""
 
     # Grab unit IDs
-    unit_ids = list(allunits.order().values_list("id", flat=True))
+    unit_ids = list(
+        allunits.order_by_request(
+            request.GET.get("sort_by", "").split(",")
```
nijel, line 146: The `request.GET.get("sort_by", "").split(",")` part repeats in every invocation. I'd rather place it into order_by_request and let it accept request as parameter.

weblate/trans/tests/test_selenium.py
```diff
@@ -865,6 +865,16 @@ def test_weblate(self):
         self.screenshot("format-highlight.png")
         self.click("Comments")
         self.screenshot("plurals.png")
+        # Test search dropdown
+        dropdown = self.driver.find_element_by_id("query-dropdown")
+        dropdown.click()
+        self.screenshot("query-dropdown.png")
+        self.click("Not translated strings")
```
nijel, line 872:
```suggestion
        with self.wait_for_page_load():
            self.click("Not translated strings")
```

weblate/trans/tests/test_selenium.py
```diff
@@ -865,6 +865,16 @@ def test_weblate(self):
...
+        # Test sort dropdown
+        sort = self.driver.find_element_by_id("query-sort-dropdown")
+        sort.click()
+        self.screenshot("query-sort.png")
+        self.click("Position")
```
nijel, line 877:
```suggestion
        with self.wait_for_page_load():
            self.click("Position")
```

SaptakS:
@nijel My reason for submit button is showing a button makes it clear that it needs some user action (click or enter) for the search results to come up. Without a button, users might expect the search results to show up as they type which is not what happens. So I think a submit button makes the intent clear that you need to `submit` the form by either pressing enter or clicking the button. Though I am open to removing it if you feel that makes more sense.

nijel:
The reason I ask is that I'd never use it - when typing the query on keyboard, it is more natural to just press enter. It also seems to be common pattern these days:

GitHub: 
image

GitLab:
image

Sentry:
image

Looking at it, I like way GitLab does sorting - the ordering is shown in the text and next to it there is an icon indicating and toggling direction - the icon also makes it clear that the dropdown controls ordering. Try it at https://gitlab.com/mailman/mailman/-/issues?sort=created_date

SaptakS:
> the ordering is shown in the text and next to it there is an icon indicating and toggling direction - the icon also makes it clear that the dropdown controls ordering.

@nijel we do have an icon, though it's not the best sorting icon given that it kind of indicates ordering by alphabet. I will try and replace it for something better and also add the display of the current ordering.

> The reason I ask is that I'd never use it - when typing the query on keyboard, it is more natural to just press enter. 

Agreed. Cool, I will remove it for now. Can always add it back if users find it unintuitive.

nijel:
Did some more testing:

* Please separate the sort widget and add support for changing sort direction, something like on GitLab:

image

* Using `col-sm3` for navigation doesn't work well on small screens, the navigvation overflows window (around width of 1000px)

* Overall it doesn't scale well to mobile screen size, but it should be better once sorting is separate widget

SaptakS:
@nijel on it.

nijel:
Looks good now. Some minor issues I see:

* The sorting selection doesn't have rounded corners as other ones.
* "Sort By" is shown when going to translate page instead of actual sorting order.
* Ordering by priority and position (current default) is missing from the selection.
* Sometimes empty search is shown as "Custom search", sometimes as "All strings".
* It probably doesn't make sense to autosubmit on search entry pages, only on the ones with results.

SaptakS:
> The sorting selection doesn't have rounded corners as other ones.

Will get this fixed

> "Sort By" is shown when going to translate page instead of actual sorting order.

Since when we go to the translate page in the beginning, there is no sorting selected (only the default sort is applied). So it just shows Sort By (as default) for users to select a sorting order.

> Ordering by priority and position (current default) is missing from the selection.

Should add this. Should I just name this as `Default` in the dropdown? or what should be the name?

> Sometimes empty search is shown as "Custom search", sometimes as "All strings".

I have noticed this as well. The reason why this happens is, if the url has a GET parameter like `q=` the filter_name returned is All Strings. But if there is no `q` parameter or if no `filter_name` is returned from backend, then Custom Search is shown. Thinking about it, I can totally check if `filter_name` received in template exists in the dropdown. If yes, then show. That would solve this problem.

> It probably doesn't make sense to autosubmit on search entry pages, only on the ones with results.

So you mean not to autosubmit in site wide search but only on the translate pages?

nijel:
> Since when we go to the translate page in the beginning, there is no sorting selected (only the default sort is applied). So it just shows Sort By (as default) for users to select a sorting order.

It should always show the current ordering to be consistent. Even if it is the default one.

> Should add this. Should I just name this as Default in the dropdown? or what should be the name?

Maybe "Priority and position"

> I have noticed this as well. The reason why this happens is, if the url has a GET parameter like q= the filter_name returned is All Strings. But if there is no q parameter or if no filter_name is returned from backend, then Custom Search is shown. Thinking about it, I can totally check if filter_name received in template exists in the dropdown. If yes, then show. That would solve this problem.

This should be addressed by abe6e59d6f881814c8d4220ec2da8ec0dd72800b (I didn't yet test it together with your changes).

> So you mean not to autosubmit in site wide search but only on the translate pages?

I think it makes sense whenever showing the search results (for example here: https://hosted.weblate.org/search/?q=+has%3Asuggestion+&checksum=&offset=), but not when creating the query for the first time (for example here: https://hosted.weblate.org/#search).

weblate/static/loader-bootstrap.js
```diff
@@ -1554,6 +1560,12 @@ $(function () {
             return false;
         }
     });
+    $('#id_q').on('keydown', function (event) {
+        console.log("hello")
```
nijel, line 1564: Looks like forgotten debugging code :-).

weblate/static/style-bootstrap.css
```diff
@@ -478,12 +478,16 @@ legend {
     width: 35px;
     padding-left: 0;
     padding-right: 0;
+    border-top-right-radius: 4px;
+    border-bottom-right-radius: 4px;
```
nijel, line 482: My guess is that this will go wrong in RTL. Why not using Bootrap button groups to get the proper styling?

weblate/utils/views.py
```diff
@@ -80,6 +80,23 @@ def dispatch(self, request, *args, **kwargs):
...
 
 
+def get_sort_name(request):
+    """Gets sort name."""
+    sort_dict = {
+        "position": "Position",
+        "priority": "Priority",
+        "num_words": "Word count",
+        "context": "Context",
+        "position,-priority": "Position & Priority",
```
nijel, line 90: Localization is missing on these texts.

nijel, line 90:
```suggestion
        "position,-priority": "Position and priority",
```

nijel:
It seems to render wrongly now for me:

obrazek

Also, I think that it should be "Position and priority" (lowercase priority).

weblate/static/style-bootstrap.css
```diff
@@ -447,6 +451,40 @@ legend {
...
+#query-sort-toggle {
+    width: 40px;
+    padding-left: 0;
+    padding-right: 0;
+}
+#query-sort-toggle span:not(.active) {
+    display: none;
+}
+#query-sort-dropdown {
+    width: 200px;
```
nijel, line 486: Hard-coding width will probably go wrong in some translation. Can this be dynamic?

SaptakS: It can be dynamic, but it will constantly resize based on the width of the test in the button. I checks in all different device widths and seems to work fine. I am okay with just keeping it dynamically change to fit the width of the text.

SaptakS:
@nijel Just pushed some bootstrap changes. Please check if things look better. Also made the copy change.

nijel:
Getting closer, but still not ready ;-).

* In RTL, the rounded corners are gone, not sure why:
obrazek
* When creating search, the sort by correctly doesn't submit, but it also doesn't update the dropdown text to indicate current selection. It works file for the predefined searches.

SaptakS:
> In RTL, the rounded corners are gone, not sure why

Not sure because I am not adding any custom CSS. Need to test and check. Might need to add some custom CSS by adding another class `rtl` to the HTML

> doesn't update the dropdown text to indicate current selection.

Will take a look.

weblate/utils/views.py
```diff
@@ -80,6 +80,24 @@ def dispatch(self, request, *args, **kwargs):
...
 
 
+def get_sort_name(request):
+    """Gets sort name."""
+    sort_dict = {
+        "position": _("Position"),
+        "priority": _("Priority"),
+        "num_words": _("Word count"),
+        "context": _("Context"),
+        "priority,position": _("Position and priority"),
```
nijel, line 90: Shouldn't it be `-priority,position` as elsewhere?

weblate/static/loader-bootstrap.js
```diff
@@ -1525,16 +1525,50 @@ $(function () {
     $('.search-group li a').click(function () {
         var $this = $(this);
         var $button = $this.closest('.input-group-btn').find('button');
-        $button.data('field', $this.data('field'));
-        $button.find('span.search-label').text($this.text());
+        $button.attr('data-field', $this.data('field'));
+        $button.find('span.search-label').not('#query-sort-toggle span').text($this.text());
+        if ($this.data('sort')) {
+            $('#id_sort_by').val($this.data('sort'));
+            if ($('.pagination').length) {
```
nijel, line 1532: This is probably not the best way to decide we're on search results as there can be no pagination on sitewide search if all matches fit into single page.

weblate/utils/views.py
SaptakS: No. This is a dictionary without the sign. So that it doesn't matter what is the toggle sort. I thought that makes sense even for future?

weblate/static/loader-bootstrap.js
SaptakS: Umm. Okay. Since it's only in `search.html` and `translate.html` I can probably add a class in those two.

nijel: That's probably the safest approach as the content of these pages can be changed and it's probably better not to rely on that as it will break without noticing later.

SaptakS: Should be fixed.

nijel:
Thanks, now it seems to work fine. I'd like to address some additional issues:

* There is way too much of blank space on search page now, compare new and old look:

  image

  image

  Ideally all would be on single line on large enough screens - search, sort, navigation and settings with zen toggle

* The zen mode doesn't seem to use the new search field.

nijel:
I'm now getting js errors on search entry page:

```
jQuery.Deferred exception: sort_dropdown_value is undefined @http://127.0.0.1:8000/static/loader-bootstrap.js?v=weblate-4.0.3-102-gfa1bfe639b:1521:20
mightThrow@http://127.0.0.1:8000/static/js/jquery.js?v=weblate-4.0.3-102-gfa1bfe639b:3762:29
resolve/</process<@http://127.0.0.1:8000/static/js/jquery.js?v=weblate-4.0.3-102-gfa1bfe639b:3830:12
 undefined jquery.js:4046:18
TypeError: sort_dropdown_value is undefined loader-bootstrap.js:1521:20
    <anonymous> http://127.0.0.1:8000/static/loader-bootstrap.js?v=weblate-4.0.3-102-gfa1bfe639b:1521
    jQuery 2
```

nijel:

PR #3776, opened 2020-04-23
Updated DeepL languages

Ch4t4r:
This PR adds languages [supported by DeepL](https://www.deepl.com/docs-api/translating-text/request/) not yet tracked in Weblate.
Not yet tracked:
 - Japanese (ja)
 - Chinese simplified (zh-hans, DeepL has it as zh)
 - Portugues Brazilian (pt-br, only supported as target language)

A potential problem might be that pt-br is only supported as target_lang (for source lang it's just pt) but the logic does not differentiate between supported source and target languages.

### DIFF
--- a/weblate/machinery/deepl.py
+++ b/weblate/machinery/deepl.py
@@ -42,7 +42,7 @@ class DeepLTranslation(MachineTranslation):
 
     def download_languages(self):
         """List of supported languages is currently hardcoded."""
-        return ("en", "de", "fr", "es", "it", "nl", "pl", "pt", "ru")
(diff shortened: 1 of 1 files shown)


Ch4t4r:
Fair enough, they seem to be adding a new language every now and then. The API you linked doesn't work for me at all though. I'd guess that it only has `pt` because `pt-BR` is only used as target language.

I'll close this PR.

nijel:
I've done changed the code to use API in d7ebf90eabee8b3384fa3d61276fea3a4b5d27d1

DwolfVW:
Cool, thanks!

Ch4t4r:
Cool, thanks!

PR #3792, opened 2020-04-27
Add 'approved' flag to Units API

dave-lang:
I'm currently wrestling with identifying approved translations from the API for a translation report. This PR adds an 'approved' flag to the Units API. As far as I can tell there's no test coverage changes required.


nijel:

PR #3797, opened 2020-04-28
Adds check for duplicate words.

### LINKED ISSUE #3702: Repetition check
Some platforms/tools have checks for repetition, which is the error that happens most often.

When something is very complex, it is easy to stop, think, and then carry on from the word already put down onwards. It can also happen when rewriting a sentence, ending up with two parts that share a repeated word in its midst.

For example this this is an example of a repeated word.

Example and example, fine. "This this" etc. is almost always an error.
If not an outright error, it is often best avoided, so quality would improve a lot overall with a check that (in most languages) has very few false positives.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/90869677-repetition-check?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

flosincapite:
Fixes #3702

This check mostly just looks for consecutive tokens that consist entirely of word characters (crucially excluding things like punctuation).

In order to eliminate some false positives, I've added a blacklist of tokens and patterns that are likely to be repeated. I generated this list in the simplest way possible: counting reduplicated terms in a large text corpus and selecting the top terms by frequency. We could do something much smarter (e.g. by using PMI or a classifier), but this is quick and scalable.

Happy to generate similar blacklists for other languages here or in a follow-up PR.

### DIFF
--- a/docs/user/checks.rst
+++ b/docs/user/checks.rst
@@ -214,6 +214,15 @@ This only checks for the length of translation characters.
 Unlike the other checks, the flag should be set as a ``key:value`` pair like
 ``max-length:100``.
 
+.. _check-duplicate:
+
+Consecutive duplicated tokens
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Checks that no duplicate tokens occur in translation.
+
+Check is false if any token occurs more than once consecutively.
+
 .. _check-python-format:
 .. _check-python-brace-format:
 .. _check-php-format:
--- a/weblate/checks/duplicate.py
+++ b/weblate/checks/duplicate.py
@@ -1,0 +1,39 @@
+#
+# Copyright © 2012 - 2020 Michal Čihař <michal@cihar.com>
+#
+# This file is part of Weblate <https://weblate.org/>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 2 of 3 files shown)


nijel:
Thanks for your contribution, the code looks good. I'm just unsure about the blacklist. 

It seems to contain country names, where I really don't see the use case. Rest seems to be heavily based on Latin names of animals, what is also questionable to include - I think that it's more likely that "gorilla gorilla" will be a wrong translation rather than Latin name of the animal. On the other side it doesn't cover some other use cases (for example see https://english.stackexchange.com/q/37827/383363).

Still, English is mostly source language in localization, what would be more useful is to have other languages covered.

Overall I'm not sure maintaining a per language blacklist here will bring enough benefits to outweigh it's size and maintenance of per language data.

flosincapite:
> Thanks for your contribution, the code looks good. I'm just unsure about the blacklist.
> 
> It seems to contain country names, where I really don't see the use case. Rest seems to be heavily based on Latin names of animals, what is also questionable to include - I think that it's more likely that "gorilla gorilla" will be a wrong translation rather than Latin name of the animal. On the other side it doesn't cover some other use cases (for example see https://english.stackexchange.com/q/37827/383363).
> 
> Still, English is mostly source language in localization, what would be more useful is to have other languages covered.
> 
> Overall I'm not sure maintaining a per language blacklist here will bring enough benefits to outweigh it's size and maintenance of per language data.

Makes sense! Okay, I have eliminated the blacklist.

nijel:
Thanks! Looks good now. Can you please also add shorts docs to `docs/user/checks.rst`?

flosincapite:
> Thanks! Looks good now. Can you please also add shorts docs to `docs/user/checks.rst`?

Done!

nijel:

PR #3800, opened 2020-04-29
Update the Squash addon

### LINKED ISSUE #3709: Ability to set "parent message" for commit via "Squash Git commits" addon
We are currently usinge the "Squash Git commits" addon.

We'd like to retain authorship information in the squashed commit by using `Co-authored-by: {{ author }}`, however with multiple authors this ends up producing a commit message like:

```
Translated using Weblate

Translation: pypa/warehouse
Translate-URL: https://hosted.weblate.org/projects/pypa/warehouse/es/

Co-authored-by: Contributor 1 <email@example.com>

Translated using Weblate

Translation: pypa/warehouse
Translate-URL: https://hosted.weblate.org/projects/pypa/warehouse/de/

Co-authored-by: Contributor 2 <email@example.com>
```

Ideally, we could produce a single commit with a commit message like:
```
Translated using Weblate


Co-authored-by: Contributor 1 <email@example.com>
Co-authored-by: Contributor 2 <email@example.com>
```
I think it would be possible to do this if were possible to configure the "Squash Git commits" to have a "parent commit message" which the squashed commit messages are appended to. 

Setting the commit message to just:

```
Co-authored-by: {{ author }}
```
doesn't work, as the first line is interpreted by git as the commit message title, not part of the body.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/90871894-ability-to-set-parent-message-for-commit-via-squash-git-commits-addon?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

di:
Fixes #3709. Support appending trailers and overriding the commit message.

Draft PR because I'm not sure how to approach adding tests for this, the existing tests are only counting the number of commits that are produced, not the commit messages produced. Some pointers to what you'd like to see here would be helpful.

### DIFF
--- a/weblate/addons/git.py
+++ b/weblate/addons/git.py
@@ -77,7 +77,7 @@ class GitSquashAddon(BaseAddon):
         if self.instance.configuration.get("append_trailers"):
             command = [
                 "log",
-                "--format={}".format("%(trailers:separator)"),
+                "--format=%(trailers:separator)",
                 "{}..HEAD".format(remote),
             ]
             if filenames:
--- a/weblate/addons/git.py
+++ b/weblate/addons/git.py
@@ -91,7 +91,16 @@ class GitSquashAddon(BaseAddon):
                 ]
             )
 
-            commit_message = "\n\n".join([commit_message, "\n".join(trailer_lines)])
+            commit_message_lines_with_trailers_removed = [
+                line for line in commit_message.split("\n") if line not in trailer_lines
+            ]
+
+            commit_message = "\n\n".join(
+                [
+                    "\n".join(commit_message_lines_with_trailers_removed),
+                    "\n".join(trailer_lines),
+                ]
+            )
 
         return commit_message
 
--- a/weblate/addons/git.py
+++ b/weblate/addons/git.py
@@ -77,7 +77,7 @@ class GitSquashAddon(BaseAddon):
         if self.instance.configuration.get("append_trailers"):
             command = [
                 "log",
-                "--format=%(trailers:separator)",
+                "--format=%(trailers)",
                 "{}..HEAD".format(remote),
             ]
             if filenames:
--- a/weblate/addons/git.py
+++ b/weblate/addons/git.py
@@ -19,7 +19,7 @@
 
 
 import os.path
-from collections import defaultdict
+from collections import OrderedDict, defaultdict
 
 from django.utils.translation import gettext_lazy as _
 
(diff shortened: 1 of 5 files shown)


weblate/addons/git.py
```diff
@@ -60,6 +60,34 @@ def get_filenames(self, component):
...
+        if self.instance.configuration.get("append_trailers"):
+            command = [
+                "log",
+                "--format={}".format("%(trailers:separator)"),
+                "{}..HEAD".format(remote),
+            ]
+            if filenames:
+                command += ["--"] + filenames
+
+            trailers = repository.execute(command)
```
nijel, line 86: I guess this will can include same entries several times, not sure how big deal that is.

weblate/addons/git.py
```diff
@@ -60,6 +60,34 @@ def get_filenames(self, component):
...
+            commit_message = "\n\n".join([commit_message, trailers])
```
nijel, line 87: Won't this include trailers twice in case custom commit message is not set? First time they would be included from the commit message and now for the second time.

What I think would make more sense is to avoid trailers from the squashed messages and add them here separately. Though I don't see way to do this using git log.

weblate/addons/git.py
```diff
@@ -60,6 +60,34 @@ def get_filenames(self, component):
...
+            ]
+            if filenames:
+                command += ["--"] + filenames
+
+            commit_message = repository.execute(command)
+
+        if self.instance.configuration.get("append_trailers"):
+            command = [
+                "log",
+                "--format={}".format("%(trailers:separator)"),
```
nijel, line 80: The formatting doesn't seem to be needed here.

This format was added in [Git 2.12.0](https://github.com/git/git/blob/master/Documentation/RelNotes/2.12.0.txt), so the git required version should be raised in `weblate/vcs/git.py`.

weblate/addons/git.py
di: I also couldn't find a way to do this with `git log`. I ended up needing to remove the trailers from the commit message anways because otherwise repeated squashes would result in an exponential increase in trailers.

weblate/addons/git.py
di: Removing trailers from the "commit body" made it easy to remove duplicates as well, so I went ahead and did that since I could see no semantic reason to have duplicate trailers.

weblate/addons/git.py
```diff
@@ -60,6 +60,50 @@ def get_filenames(self, component):
...
+        if self.instance.configuration.get("append_trailers"):
+            command = [
+                "log",
+                "--format=%(trailers)",
+                "{}..HEAD".format(remote),
+            ]
+            if filenames:
+                command += ["--"] + filenames
+
+            trailer_lines = OrderedDict.fromkeys(
```
nijel, line 86: It is not necessary to use OrderedDict, as Weblate requires Python 3.6+ and that has ordered dict, see https://stackoverflow.com/a/39980744/225718 (but it should do no harm).

PR #3808, opened 2020-05-01
Move contact info to CONTACT.md

comradekingu:
This prevents it from popping up needlessly as "Remember, contributions to this repository should follow its *", when it contains nothing to be followed, taking away attention from "contributing guidelines", which does.
While I don't think its inclusion has made any marked improvement to that effect, changing the title doesn't turn away contributors that leave when seeing it, expecting the worst.

This exact change increased donations in Remmina greatly. Could be coincidental, but a point was made to do so in the blog.


nijel:

PR #3834, opened 2020-05-08
tested macos installation

abtinmo:

i followed Installing on macOS and successfully installed and used weblate on my device. so no need for warning message


nijel:
Glad to hear that it worked, it was heavily based on my guesses.

PR #3835, opened 2020-05-08
Adds inline editing of position in translate page

SaptakS:

Addresses inline editing of position in #3308

### DIFF
--- a/weblate/static/loader-bootstrap.js
+++ b/weblate/static/loader-bootstrap.js
@@ -1542,6 +1542,12 @@ $(function () {
         });
     });
 
+    /* Click to edit position inline */
+    $('#position-input').on('click', function() {
+        $('#position-input').hide();
+        $('#position-input-editable').show();
+    });
+
     /* Advanced search */
     $('.search-group li a').click(function () {
         var $this = $(this);
--- a/weblate/static/loader-bootstrap.js
+++ b/weblate/static/loader-bootstrap.js
@@ -1542,11 +1542,32 @@ $(function () {
         });
     });
 
-    /* Click to edit position inline */
+    /* Click to edit position inline. Disable when clicked outside or pressed ESC */
     $('#position-input').on('click', function() {
         $('#position-input').hide();
         $('#position-input-editable').show();
+        $('#position-input-editable input').focus();
+        document.addEventListener('click', clickedOutsideEditableInput);
+        document.addEventListener('keyup', pressedEscape);
     });
+    var clickedOutsideEditableInput = function(event) {
+        if (!$.contains($('#position-input-editable')[0], event.target) &&
+            event.target != $('#position-input')[0]
+        ) {
+            $('#position-input').show();
+            $('#position-input-editable').hide();
+            document.removeEventListener('click', clickedOutsideEditableInput);
+            document.removeEventListener('keyup', pressedEscape);
+        }
+    }
+    var pressedEscape = function(event) {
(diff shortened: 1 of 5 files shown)


nijel:
Looks good! The only thing I'd like to add is that escape or clicking outside should hide the input.

SaptakS:
Good point. I was thinking the same. Will add some JS for the same.

weblate/static/loader-bootstrap.js
```diff
@@ -1546,6 +1546,22 @@ $(function () {
...
+    $(document).on('click', function(event) {
+        var $target = $(event.target);
+        if(!$target.closest('#position-input').length &&
+        !$target.closest('#position-input-editable input').length &&
+        $('#position-input-editable').is(":visible")) {
+            $('#position-input').show();
+            $('#position-input-editable').hide();
+        }
+    });
+    $(document).on('keyup', function(event) {
```
nijel, line 1560: Shouldn't this be only when the input is in the focus?

weblate/static/loader-bootstrap.js
```diff
@@ -1546,6 +1546,22 @@ $(function () {
     $('#position-input').on('click', function() {
         $('#position-input').hide();
         $('#position-input-editable').show();
+        $('#position-input-editable input').focus();
+    });
+    $(document).on('click', function(event) {
```
nijel, line 1551: Do I read it correctly it catches all clicks on the page? Isn't that performance issue?

julen: A better approach would be to attach this event handler upon entering inline position editing, and then remove the handler upon leaving inline editing. Otherwise, users will have a click handler firing on every single click for the lifetime of a loaded page. The same concept applies for the `keyup` below.

For the actual checking to leave inline editing, something like this should work (didn't test but shows the concept):
```js
var editableInput = document.querySelector('#position-input-editable');
if (!editableInput.contains(event.target)) {
    // logic to leave inline editing
}
```
(Note how a test like `$('#position-input-editable').is(":visible"))` would be irrelevant when such an event handler is attached only when editing inline)

SaptakS: @nijel @julen valid point. I will change how the action listener is attached.

SaptakS:
@nijel just pushed some changes that should addEventListeners for click outside and ESC only when pagination position part is clicked and removed both eventlisteners once clicked outside or pressed ESC.

weblate/static/loader-bootstrap.js
```diff
@@ -1542,6 +1542,33 @@ $(function () {
...
+    $('#position-input').on('click', function() {
+        $('#position-input').hide();
+        $('#position-input-editable').show();
+        $('#position-input-editable input').focus();
+        document.addEventListener('click', clickedOutsideEditableInput);
+        document.addEventListener('keyup', pressedEscape);
+    });
+    var clickedOutsideEditableInput = function(event) {
+        if (!$.contains($('#position-input-editable')[0], event.target) &&
+            event.target != $('#position-input')[0]
```
SaptakS, line 1555: To avoid using something like stopEventPropagation (which is something I was trying to avoid before, hence I didn't add on click) because of it's [dangers](https://css-tricks.com/dangers-stopping-event-propagation/), I have added an additional check in both the event handlers to check that the event causing the execution of the event handler isn't the same event as onclick on $('#position-input'). That way, we can avoid the event propagation in the first click that activates the input.

nijel:

PR #3850, opened 2020-05-12
Docs: fix reference to own checks

julen:
The reference wouldn't be linked otherwise.


nijel:

Issue #3710, opened 2020-04-09
Customizable string ordering


tjhietala:
**Describe the bug**
Something very strange is going on with string numbering in MyPaint on hosted.weblate.org

**To Reproduce**
Go to https://hosted.weblate.org/projects/mypaint/mypaint/fi/ and click All strings. You'll be directed to https://hosted.weblate.org/translate/mypaint/mypaint/fi/?q= which says it's string 1 out of 1272, yet there appear to be neighbouring strings above it. If you click on the one immediately above, Weblate says it's string 1086 out of 1272.

Likewise if you click the "last string" button, you'll be directed to https://hosted.weblate.org/translate/mypaint/mypaint/fi/?&offset=1272 where you can see neighbouring strings below it. If you click on the one immediately below, Weblate says it's string 1232 out of 1272.

**Expected behavior**
The first string should actually be the first and the last string should actually be the last, with no unexpected jumps in the numbering.

**Screenshots**

**Server configuration and status**
Whatever configuration is on hosted.weblate.org

**Additional context**

tjhietala:
I see. Is there any way for the translator to switch to location-based numbering?

nijel:
Not yet. I thought there is existing issue for this, but I cannot find it right now...

comradekingu:
@tjhietala They are sorted by context.
First Context: Menu (labels),then nothing, then "Menu→File (labels), Accel Editor (labels)"

nijel:
What sorting options do you find useful?

* Priority and position (currently hardcoded)
* Position
* Context
* Source

comradekingu:
Length of string?

nijel:
I also sometimes do translate short strings first, but I'm not sure it's a good practice... But it probably makes sense to include it.

tjhietala:
For me, position is important in keeping mental orientation on where I am in the component, because the context strings alone often don't provide enough context. If the ordering stays stable, it's possible to build a kind of mental map of the component over time. It's hard to say how useful other sorting options would be without trying them.

comradekingu:
Whatever makes the most sense should be selectable, no?

**Similarity**, same concept as length, but with additional benefit of keeping more topical
**Number of comments**, helps with keeping up with discussion
**New commends**, same as above
**Number of open issues**, helps people fix issues
**Number of failing checks**, same as above
**Age of string**, lets people get into the swing of things, and review what hasn't already been
[...]

PR #3851, opened 2020-05-12
Adds additing string ordering in the query tool

SaptakS:

Adds the following sortings:
- Labels
- Age of string
- Number of failing checks
- Number of comments

Refs #3710

### DIFF
--- a/weblate/trans/models/unit.py
+++ b/weblate/trans/models/unit.py
@@ -144,18 +144,24 @@ class UnitQuerySet(models.QuerySet):
                     ),
                     countable_sort_choices[unsigned_choice]["filter"],
                 )
-            elif unsigned_choice in available_sort_choices:
+            if unsigned_choice in available_sort_choices:
                 if unsigned_choice == "labels":
-                    self = self.annotate(max_labels_name=Max("labels__name"))
                     choice = choice.replace("labels", "max_labels_name")
                 sort_list.append(choice)
         if not sort_list:
             return self.order()
+        if "max_labels_name" in sort_list:
+            return self.annotate(max_labels_name=Max("labels__name")).order_by(
+                *sort_list
+            )
         return self.order_by(*sort_list)
 
     def order_by_count(self, choice, filter):
         model = choice.split("__")[0].replace("-", "")
-        return self.annotate(Count(model, filter=filter)).order_by(choice)
+        annotation_name = choice.replace("-", "")
+        return self.annotate(**{annotation_name: Count(model, filter=filter)}).order_by(
+            choice
+        )
 
     def get_unit(self, ttunit):
         """Find unit matching translate-toolkit unit.
--- a/weblate/trans/models/unit.py
+++ b/weblate/trans/models/unit.py
@@ -24,7 +24,7 @@ from copy import copy
 from django.conf import settings
 from django.core.cache import cache
 from django.db import models, transaction
-from django.db.models import Q
+from django.db.models import Count, Q
 from django.utils import timezone
 from django.utils.functional import cached_property
 from django.utils.translation import gettext_lazy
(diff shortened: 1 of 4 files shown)


weblate/trans/models/unit.py
```diff
@@ -118,16 +118,42 @@ def same(self, unit, exclude=True):
...
-        sort_list = [
-            choice
-            for choice in sort_list_request
-            if choice.replace("-", "") in available_sort_choices
+        available_sort_choices = [
+            "priority",
+            "position",
+            "context",
+            "num_words",
+            "labels",
```
nijel, line 126: This should probably translate to something like `labels__name`, but it needs to be tested.

SaptakS: Correct. I will test this out.

weblate/trans/tests/test_models.py
```diff
@@ -334,6 +335,37 @@ def test_flags(self):
...
+
+        # test sorting by count
+        unit4 = Unit.objects.filter(translation__language_code="cs")[2]
+        Comment.objects.create(unit=unit4, comment="Foo")
+        unit5 = Unit.objects.filter(translation__language_code="cs")
+        unit5 = unit5.order_by_request({"sort_by": "-num_comments"})
+        self.assertEqual(unit5[0].comment_set.count(), 1)
+        unit5 = Unit.objects.filter(translation__language_code="cs")
+        unit5 = unit5.order_by_request({"sort_by": "num_comments"})
+        self.assertEqual(unit5[0].comment_set.count(), 0)
```
nijel, line 367: Can you please test all ordering options? At least to the level that the generated queries can be executed, it doesn't have to cover validation of the ordering being correct.

weblate/trans/models/unit.py
```diff
@@ -118,16 +118,45 @@ def same(self, unit, exclude=True):
...
+            unsigned_choice = choice.replace("-", "")
+            if unsigned_choice in countable_sort_choices:
+                return self.order_by_count(
+                    choice.replace(
+                        unsigned_choice,
+                        countable_sort_choices[unsigned_choice]["order_by"],
+                    ),
+                    countable_sort_choices[unsigned_choice]["filter"],
+                )
+            elif unsigned_choice in available_sort_choices:
```
nijel, line 147:
```suggestion
            if unsigned_choice in available_sort_choices:
```
No need for else after return.

weblate/trans/models/unit.py
```diff
@@ -118,16 +118,45 @@ def same(self, unit, exclude=True):
...
+                if unsigned_choice == "labels":
+                    self = self.annotate(max_labels_name=Max("labels__name"))
```
nijel, line 149: Please do not assign to self.

weblate/trans/tests/test_models.py
SaptakS: Have added a simple test to loop over all the ordering options and check if the queryset count still works. I hope that kind of ensures what you were asking. Also, I have added a similar test to check that multiple ordering options also doesn't cause any error. Let me know if this seems good.

weblate/trans/tests/test_models.py
```diff
@@ -366,6 +366,28 @@ def test_order_by_request(self):
...
+            "num_words",
+            "labels",
+            "timestamp",
+            "num_failing_checks",
+        ]
+        for order_option in order_options:
+            ordered_unit = Unit.objects.filter(
+                translation__language_code="cs"
+            ).order_by_request({"sort_by": order_option})
+            self.assertEqual(ordered_unit.count(), 4)
```
nijel, line 383:
```suggestion
            self.assertEqual(len(ordered_unit), 4)
```

I'm not sure if `count` doesn't discard ordering (it has no use for it), so better let the query evaluate fully.

weblate/trans/tests/test_models.py
```diff
@@ -380,7 +380,7 @@ def test_order_by_request(self):
             ordered_unit = Unit.objects.filter(
                 translation__language_code="cs"
             ).order_by_request({"sort_by": order_option})
-            self.assertEqual(ordered_unit.count(), 4)
+            self.assertEqual(len(ordered_unit), 4)
```
SaptakS, line 383: Aah. I see what you mean. Makes sense.

nijel:

Issue #3784, opened 2020-04-24
security.W004 documentation not available


buck-E:
**Describe the bug**
security.W004 points user to https://docs.weblate.org/en/weblate-4.0.1/admin/install.html#production-ssl, which does not explain `SECURE_HSTS_SECONDS`

**To Reproduce**
1. Go to '/manage/performance'
2. See:
```
security.W004 | You have not set a value for the SECURE_HSTS_SECONDS setting.  If your entire site is served only over SSL, you may want to consider  setting a value and enabling HTTP Strict Transport Security. Be sure to  read the documentation first; enabling HSTS carelessly can cause  serious, irreversible problems.
```

3. Visit link at https://docs.weblate.org/en/weblate-4.0.1/admin/install.html#production-ssl
4. Fail to find information on `SECURE_HSTS_SECONDS`, and live in fear of these "serious, irreversible problems"

**Expected behavior**
The link should explain the item in the same line

rajat641:
Hi @nijel , I can start of on this bug.

nijel:
Go ahead, it's mostly about adding documentation and links to Django docs.

rajat641:
Thanks @nijel . I'm on it .

rajat641:
Hi @nijel , This is my first ever code contribution on any open-source software. Please let me know if the code changes are fine or I have made some improvements. It took me some time to do the code set up and run sphinx.

PR #3858, opened 2020-05-12
Added changes to document for SECURE_HSTS_SECONDS

rajat641:
Clarify  SECURE_HSTS_SECONDS setting documentation
- Explicitly mention how and why to enable it.
- Link to relevant Django docs.

Fixes [ #3784 ](https://github.com/WeblateOrg/weblate/issues/3784)

I have tested this on my local machine with Sphinx. As its a change in the documentation, please let me know if the sentences are clear and as per standards.

### DIFF
--- a/docs/admin/install.rst
+++ b/docs/admin/install.rst
@@ -654,12 +654,16 @@ After enabling it, you should set :setting:`ENABLE_HTTPS` in the settings:
 Set properly SECURE_HSTS_SECONDS
 ++++++++++++++++++++++++++++++++
 
-Set this value to 0 so as to make sure that everything works as expected,
-otherwise you could prevent the client from visiting the site.
+If your site is served over SSL, you have to consider setting a value for :setting:`SECURE_HSTS_SECONDS`
+in the settings.py and enabling HTTP Strict Transport Security.
+By default its set to 0 as shown below.
 
 .. code-block:: python
 
    SECURE_HSTS_SECONDS = 0
+
+If set to a non-zero integer value, the `weblate.middleware.SecurityMiddleware`
+sets the :ref:`django:http-strict-transport-security` header on all responses that do not already have it.
 
 .. warning::
 
--- a/docs/admin/install.rst
+++ b/docs/admin/install.rst
@@ -655,14 +655,15 @@ Set properly SECURE_HSTS_SECONDS
 ++++++++++++++++++++++++++++++++
 
 If your site is served over SSL, you have to consider setting a value for :setting:`SECURE_HSTS_SECONDS`
-in the settings.py and enabling HTTP Strict Transport Security.
+in the settings.py to enable HTTP Strict Transport Security.
 By default its set to 0 as shown below.
 
 .. code-block:: python
 
    SECURE_HSTS_SECONDS = 0
 
-If set to a non-zero integer value, the `weblate.middleware.SecurityMiddleware`
+If set to a non-zero integer value, the `SecurityMiddleware
+<https://docs.djangoproject.com/en/3.0/ref/middleware/#django.middleware.security.SecurityMiddleware>`_
 sets the :ref:`django:http-strict-transport-security` header on all responses that do not already have it.
 
 .. warning::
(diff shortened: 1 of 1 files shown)


docs/admin/install.rst
```diff
@@ -650,6 +650,23 @@ After enabling it, you should set :setting:`ENABLE_HTTPS` in the settings:
    :setting:`ENABLE_HTTPS`,
    :ref:`production-site`
 
+
+Set properly SECURE_HSTS_SECONDS
+++++++++++++++++++++++++++++++++
+
+Set this value to 0 so as to make sure that everything works as expected,
+otherwise you could prevent the client from visiting the site.
```
nijel, line 658: Sorry, but this is nonsense. In most cases it should be set to reasonably high value to enforce HSTS and bring increased security. It should be set to 0 only in case HTTP is also supposed to be used on the site (what I hope nobody seriously considers these days).

rajat641: @nijel , Yes you're right that it should be set to a high value so as to prevent a man-in-the-middle attack. Its set to 0 to allow HTTP and may be set to a lower value for testing purposes ensuring all assets are servers securely.
I apologize for not reading and understanding it completely.

rajat641: @nijel - If we set the value to 31536000 seconds, i.e. 1 year(As per Django documentation) is fine or I have to explain all scenarios in the document.

rajat641:
Hi @nijel , Good morning. I have tried to improve the previous version of the document. Please review.

docs/admin/install.rst
```diff
@@ -650,6 +650,27 @@ After enabling it, you should set :setting:`ENABLE_HTTPS` in the settings:
...
+
+If your site is served over SSL, you have to consider setting a value for :setting:`SECURE_HSTS_SECONDS`
+in the settings.py and enabling HTTP Strict Transport Security.
+By default its set to 0 as shown below.
+
+.. code-block:: python
+
+   SECURE_HSTS_SECONDS = 0
+
+If set to a non-zero integer value, the `weblate.middleware.SecurityMiddleware`
```
nijel, line 665: It's Django middleware taking care of this, not Weblate one.

docs/admin/install.rst
```diff
@@ -650,6 +650,27 @@ After enabling it, you should set :setting:`ENABLE_HTTPS` in the settings:
    :setting:`ENABLE_HTTPS`,
    :ref:`production-site`
 
+
+Set properly SECURE_HSTS_SECONDS
+++++++++++++++++++++++++++++++++
+
+If your site is served over SSL, you have to consider setting a value for :setting:`SECURE_HSTS_SECONDS`
+in the settings.py and enabling HTTP Strict Transport Security.
```
nijel, line 658: Setting SECURE_HSTS_SECONDS is way to enable HSTS, so it should probably be:

```suggestion
in the settings.py to enable HTTP Strict Transport Security.
```

docs/admin/install.rst
rajat641: Cool. I have added the Django(Security Middleware) link for it.

docs/admin/install.rst
rajat641: Fixed.

rajat641: @nijel  - I have updated the PR. Please review.

docs/admin/install.rst
```diff
@@ -650,6 +650,28 @@ After enabling it, you should set :setting:`ENABLE_HTTPS` in the settings:
...
+If your site is served over SSL, you have to consider setting a value for :setting:`SECURE_HSTS_SECONDS`
+in the settings.py to enable HTTP Strict Transport Security.
+By default its set to 0 as shown below.
+
+.. code-block:: python
+
+   SECURE_HSTS_SECONDS = 0
+
+If set to a non-zero integer value, the `SecurityMiddleware
+<https://docs.djangoproject.com/en/3.0/ref/middleware/#django.middleware.security.SecurityMiddleware>`_
```
nijel, line 666: Using Sphinx links is better - it will make the links work for future Django releases as well. I think following should work (but not tested):

```suggestion
If set to a non-zero integer value, the :class:`django:~django.middleware.security.SecurityMiddleware`
```

rajat641: It works after removing the tilde ~. I have updated the PR.

docs/admin/install.rst
rajat641: image

nijel:

PR #3862, opened 2020-05-13
Docs: fix typos and update internal modules list

julen:
Came across these while reading through docs.


nijel:

PR #3872, opened 2020-05-14
Docs: installation fixes

julen:
Fixes some minor issues found while reading through docs.

### DIFF
--- a/docs/admin/install.rst
+++ b/docs/admin/install.rst
@@ -41,8 +41,6 @@ translation-finder
     https://github.com/WeblateOrg/translation-finder
 Python Social Auth
     https://python-social-auth.readthedocs.io/
-Whoosh
-    https://bitbucket.org/mchaput/whoosh/wiki/Home
 Django REST Framework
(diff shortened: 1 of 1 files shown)


docs/admin/install.rst
```diff
@@ -41,8 +41,6 @@ translation-finder
     https://github.com/WeblateOrg/translation-finder
 Python Social Auth
     https://python-social-auth.readthedocs.io/
-Whoosh
-    https://bitbucket.org/mchaput/whoosh/wiki/Home
```
nijel, line 45: Whoosh is still used for query parsing, but its replacement in this case is planned as well, see https://github.com/WeblateOrg/weblate/issues/3229. But it's probably okay to remove it from this list now.

nijel:

docs/admin/install.rst
julen: Oh, good to know! I read in the changelog that it was replaced by DB FTS for searching, but I was definitely grepping one directory level above where I should 😅

PR #3876, opened 2020-05-14
Shortcuts: bind using `mod` cross-platform modifier

julen:
On Mac this ends up mapping to `command+s` whereas on Windows and Linux it maps to `ctrl+s`. This is available since Mousetrap 1.4.


nijel:

Issue #3233, opened 2019-11-21
provide more info in language API


Jibec:
**Is your feature request related to a problem? Please describe.**
I'd like to have more information about a language when interacting with Weblate's API.

**Describe the solution you'd like**
Provide more data, such as Local language name, alternate names, eventual aliases, plural forms and any other information you'll find relevant.

**Additional context**
I've build some statistics about the Fedora Documentation translation progress (https://jibecfed.fedorapeople.org/docs-l10n/localization/language/cs/).
As I was looking for English name of the language codes coming from . I discovered your API could be a really convenient way to get it.

In our tooling, having the local language name would help us a lot to use it in our language switcher tool for example.

Thanks for your help,

nijel:
Things which should be covered:

* List of plurals, with plan to support CRUD on these
* List of aliases, these are stored in `weblate/langdata/aliases.py`, it might be useful to pre-create reverse dict to avoid doing that on request time
* Statistics, this could use StatisticsSerializer and switch it to using `stats` object directly instead of helper `get_stats` (the serializer was written before generic stats were introduced and it stayed with the legacy interface)

Where I'm not sure:

* Local language name - we don't have it for most languages, Weblate is translated only to a few languages compared to what languages we support, so the local name is missing in most cases

See also https://github.com/WeblateOrg/weblate/issues/3889

Jibec:
CLDR already provides local language name. Why would you like to ask people to translate it?
It always is complex to translate language name. With exception of those really close to us, we don't know the correct translation (I use wikipedia most of the time...).
It may also save some time for translators.

nijel:
CLDR doesn't cover all languages Weblate supports, and we do already automatically include translations from there. So, what we ask to translate is not available there. 

Still, these are two separate topics - having localized language name matching UI language and having local language name for all supported languages. We currently only aim at first thing, that's why I said we have poor coverage here (360 supported languages vs. 41 translations active in Weblate).

Jibec:
Well, it's fine if it only provides the data weblate contains.

By local language name I meant: French/Français

Jibec:
glad to see you get external support for Weblate

and thanks a lot for all the fixes

PR #3901, opened 2020-05-19
API: Adds more information to the language API

SaptakS:
Refs: #3233 


weblate/lang/models.py
```diff
@@ -169,6 +169,9 @@ def aliases_get(self, code, expanded_code=None):
                     return ret
         return None
 
+    def get_aliases_names(self, code):
+        return [alias for alias, codename in ALIASES.items() if codename == code]
```
nijel, line 173: This probably more belongs to the `Language` class, using self.code direclty.

weblate/api/serializers.py
```diff
@@ -90,15 +90,35 @@ def __init__(self, *args, **kwargs):
...
+            "web_url",
+            "url",
+            "statistics_url",
+        )
         extra_kwargs = {
             "url": {"view_name": "api:language-detail", "lookup_field": "code"},
             "code": {"validators": []},
         }
 
+    def get_plural(self, obj):
```
nijel, line 116:
```suggestion
    def get_plurals(self, obj):
```

It returns all plural rules, not just one ;-)

weblate/api/serializers.py
```diff
@@ -90,15 +90,35 @@ def __init__(self, *args, **kwargs):
 
 class LanguageSerializer(serializers.ModelSerializer):
     web_url = AbsoluteURLField(source="get_absolute_url", read_only=True)
+    plural = serializers.SerializerMethodField()
```
nijel, line 93: This needs own serializer to support creating and updating plurals (in future, you don't have to implement that in this PR).

SaptakS: Should I make it into a HyperLinkedIdentityField? Or just call the serializer with many=True?

nijel: I think serializer with many is good here, these are probably not useful standalone to deserve own API endpoint.

SaptakS:
www.celeryproject.org seems to be unresponsive.

weblate/api/serializers.py
```diff
@@ -88,12 +88,43 @@ def __init__(self, *args, **kwargs):
...
+            "source",
+            "number",
+            "formula",
+            "type",
+        )
+
+
+class LanguageAliasesSerializer(serializers.ListField):
+    def get_attribute(self, instance):
+        return super().get_attribute(instance)
```
nijel, line 105: Is this needed?

nijel:

PR #3912, opened 2020-05-19
Cleanup: remove dead code

julen:
Unused since fd2bfce.


nijel:
Good catch. I guess there will be more such things in the JS code as it's not automatically tested.

PR #3919, opened 2020-05-20
Increase minimum password length to 10

dwaynebailey:
See e.g. https://blog.fleetsmith.com/password-security-guide/

- [X] Every commit has a message which describes it
- [X] Every commit is needed on its own, if you have just minor fixes to previous commits, you can squash them


nijel:

PR #3920, opened 2020-05-20
CREATE EXTENSION ... IF NOT EXISTS still requires superuser

minusf:
Use a local check to run CREATE EXTENSION only if extension is not installed already.

`0064_fulltext_index.py` tries to install `pg_trgm` with `IF NOT EXISTS`. Unfortunately that still requires superuser privileges. The manual mentions, that it's possible to install the extension manually but that is not supported by the migration in it's current form.

This is a bigger issue with django itself as well.  It already defines a migration operation `TrigramExtension` to help with the installation of this extension.  I guess it's not easy to use it in this case because of the Mariadb/Mysql/Pgsql 🍝  :D

But in the same way the django docs mention the possibility to install extensions manually, but would fail in the same way on these Operations if the db user is not a superuser...

### DIFF
--- a/weblate/trans/migrations/0064_fulltext_index.py
+++ b/weblate/trans/migrations/0064_fulltext_index.py
@@ -24,7 +24,10 @@ def create_index(apps, schema_editor):
     vendor = schema_editor.connection.vendor
     if vendor == "postgresql":
         # Install pg_trgm for trigram search and index
-        schema_editor.execute("CREATE EXTENSION IF NOT EXISTS pg_trgm")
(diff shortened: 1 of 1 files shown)


minusf:
let's where it goes.
https://code.djangoproject.com/ticket/31615#ticket

nijel:
Actually, it should be safe to use `TrigramExtension` as it checks if the database is postgres, but as that one is broken anyway, it is not really useful in this case :-).

nijel:

PR #3924, opened 2020-05-21
Spelling: EPUB, https://, Qt help

comradekingu:
As per
https://en.wikipedia.org/wiki/EPUB
https://doc.qt.io/archives/qt-4.8/qthelp-framework.html
https://doc.qt.io/qt-5/qthelp-index.html


nijel:
This was generated at Sphinx, so maybe it's better to submit corrections there: https://github.com/sphinx-doc/sphinx/blob/3.x/sphinx/templates/quickstart/make.bat_t

PR #3929, opened 2020-05-22
Added database(PostgreSQL/MySQL)/ cache(Redis) server versions

rajat641:

### DIFF
--- a/weblate/utils/requirements.py
+++ b/weblate/utils/requirements.py
@@ -22,6 +22,7 @@ import email.parser
 import sys
 
 import pkg_resources
+from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.db import connection
 
@@ -170,23 +171,31 @@ def get_versions():
     return result
 
 
-def get_version_number_db(database, version):
+def get_version_number_db(version):
     """Returns version of database in format as (Major Version.Minor Version.Revision).
 
     from an integer. To know more about database version,
     refer <https://www.psycopg.org/docs/connection.html>.
     """
-    if database == "POSTGRESQL":
-        major_version = version[0:2]
-        minor_version = version[2:4] if version[2:4] != "00" else ""
-        revision = version[5:]
-        return (
-            major_version + "." + minor_version + "." + revision
-            if minor_version != ""
-            else major_version + "." + revision
+    major_version = version[0:2]
+    major_version = (
+        major_version.split(major_version[1])[0]
+        if major_version[1] == "0"
+        else major_version
+    )
+    minor_version = version[2:4] if version[2:4] != "00" else ""
+    if minor_version != "":
+        minor_version = (
+            minor_version.split(minor_version[1])[0]
+            if minor_version[1] == "0"
+            else minor_version
         )
-    else:
-        return str(version[0]) + "." + str(version[1])
+    revision = version[4:]
+    return (
+        major_version + "." + minor_version + "." + revision
+        if minor_version != ""
+        else major_version + "." + revision
+    )
 
 
 def get_db_version():
@@ -196,7 +205,6 @@ def get_db_version():
                 connection.vendor,
                 "https://www.postgresql.org/",
                 get_version_number_db(
-                    connection.vendor,
                     str(connection.cursor().connection.server_version),
                 ),
             )
@@ -221,14 +229,14 @@ def get_db_version():
 
 def get_cache_version():
     try:
-        import os
-
-        cmd = "redis-server --version"
-        result = (
-            "Redis",
-            "https://redis.io/",
-            str(os.popen(cmd).read().split(" ")[2][2:]),
-        )
+        if (
+            settings.CACHES.get("default").get("BACKEND")
+            == "django_redis.cache.RedisCache"
+        ):
+            version = ""
+            result = ("Redis", "https://redis.io/", version)
+        else:
+            raise ImproperlyConfigured("please install a redis server.")
     except RuntimeError:
         raise ImproperlyConfigured("please install a redis server.")
     return result
--- a/weblate/utils/requirements.py
+++ b/weblate/utils/requirements.py
@@ -176,9 +176,9 @@ def get_db_version():
     if connection.vendor.upper() == "POSTGRESQL":
         try:
             cursor = connection.cursor()
-            cursor.execute("SELECT version()")
+            cursor.execute("SHOW server_version")
             version = cursor.fetchall()[0]
-            version = version[0].split(" ")[1]
+            version = version[0].split(" ")[0]
             result = (
                 connection.vendor,
                 "https://www.postgresql.org/",
@@ -191,11 +191,13 @@ def get_db_version():
             )
     else:
         try:
+            cursor = connection.cursor()
             result = (
                 connection.vendor,
                 "https://www.mysql.com/",
-                connection.cursor().connection.get_server_info().split("-", 1)[0],
+                cursor.connection.get_server_info().split("-", 1)[0],
             )
+            cursor.close()
         except RuntimeError:
             raise ImproperlyConfigured(
                 "Failed to get a database version. please install a database."
--- a/weblate/utils/requirements.py
+++ b/weblate/utils/requirements.py
@@ -177,7 +177,7 @@ def get_db_version():
         try:
             cursor = connection.cursor()
             cursor.execute("SHOW server_version")
-            version = cursor.fetchall()[0]
+            version = cursor.fetchone()
             version = version[0].split(" ")[0]
             result = (
                 connection.vendor,
@@ -207,6 +207,7 @@ def get_db_version():
 
 
 def get_cache_version():
+    result = ()
     try:
         if (
             settings.CACHES.get("default", {}).get("BACKEND")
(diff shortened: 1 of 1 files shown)


weblate/utils/requirements.py
```diff
@@ -169,10 +167,75 @@ def get_versions():
...
+            )
+
+    return result
+
+
+def get_cache_version():
+    import redis
+
+    try:
+        result = ("Redis", "https://redis.io/", str(redis.__version__))
```
nijel, line 222: Isn't this version of the redis module?

weblate/utils/requirements.py
```diff
@@ -83,9 +84,6 @@
     "boto3",
     "zeep",
     "aeidon",
-    "iniparse",
-    "mysqlclient",
-    "pyparsing",
```
nijel, line 88: The removal looks wrong.

weblate/utils/requirements.py
```diff
@@ -169,10 +167,75 @@ def get_versions():
     return result
 
 
+def get_version_number_db(database, version):
+    """
+        Returns version of database in format as (Major Version.Minor Version.Revision)
+        from an integer. To know more about database version , refer <https://www.psycopg.org/docs/connection.html>
+    """
+    if database is "POSTGRESQL":
```
nijel, line 175: This code seems used for postgres only, maybe it doesn't need the database arg and condition?

weblate/utils/requirements.py
```diff
@@ -169,10 +167,75 @@ def get_versions():
...
+        major_version = version[0:2]
```
nijel, line 176: This will break on postgres 9.x, the version there is for example "90209".

weblate/utils/requirements.py
```diff
@@ -169,10 +170,81 @@ def get_versions():
...
+
+def get_cache_version():
+    try:
+        import os
+
+        cmd = "redis-server --version"
+        result = (
+            "Redis",
+            "https://redis.io/",
+            str(os.popen(cmd).read().split(" ")[2][2:]),
```
nijel, line 230: There is no reason to have redis installed and running on same machine. Actually, in all Docker deployments it is not installed.

What should work instead (after checking that cache backend is redis) is: `cache.client.get_client().info()['redis_version']`

rajat641: Hi @nijel , 
I am checking cache backend as Redis through `settings.CACHES.get('default').get('BACKEND') == "django_redis.cache.RedisCache"`.
Also, I'm not sure what to import  for `cache.client.get_client().info()['redis_version']` to work??
I have tried checking with `import django_redis.cache` but it's throwing some error or may I'm importing something wrong.

weblate/utils/requirements.py
rajat641: I have fixed it.

weblate/utils/requirements.py
rajat641: I have fixed it as well.

weblate/utils/requirements.py
```diff
@@ -169,10 +171,88 @@ def get_versions():
...
+            )
+
+    return result
+
+
+def get_cache_version():
+    try:
+        if (
+            settings.CACHES.get("default").get("BACKEND")
+            == "django_redis.cache.RedisCache"
```
rajat641, line 234: I am checking the cache backend here and if backend is not redis. I am raising an exception. Please let me know it's correct.

weblate/utils/requirements.py
rajat641: I used pycharm code refactor tool. It modified it somehow. It's fixed now.

weblate/utils/requirements.py
rajat641: I'll use the flake8 and `pre-commit run --all` from now on .

weblate/utils/requirements.py
nijel: `from django.core.cache import cache` should do it.

weblate/utils/requirements.py
```diff
@@ -169,10 +171,88 @@ def get_versions():
     return result
 
 
+def get_version_number_db(version):
+    """Returns version of database in format as (Major Version.Minor Version.Revision).
+
+    from an integer. To know more about database version,
+    refer <https://www.psycopg.org/docs/connection.html>.
+    """
```
nijel, line 179: Can you please add tests for this? The code looks to complex for what it does and it's hard to review that it works properly. Having tests would make the verification it works correctly. It should verify all corner case you are handling (`90209` -> 9.2.9, `120003` -> 12.3, `120000` -> 12).

Maybe it would be better to avoid this postporcessing and execute `SHOW server_version;` SQL directly?

weblate/utils/requirements.py
rajat641: Yes, it works. Thanks

weblate/utils/requirements.py
```diff
@@ -169,10 +171,88 @@ def get_versions():
...
+                "Failed to get a database version. please install a database."
+            )
+
+    return result
+
+
+def get_cache_version():
+    try:
+        if (
+            settings.CACHES.get("default").get("BACKEND")
```
nijel, line 233: Chaining .get is probably not a good idea here - in case "default" is not present it will return None and next get will crash. So either rely on "default" being present or provide default value to get.

weblate/utils/requirements.py
rajat641: I'll try to avoid this post-processing and string manipulations. Will add try to execute `SHOW server_version` and parse its results.

weblate/utils/requirements.py
rajat641: Will fix this.

rajat641: `settings.CACHES.get("default", {}).get("BACKEND")` would work fine I guess.

weblate/utils/requirements.py
rajat641: @nijel , I think as per documentation 120000->12.0 
image

weblate/utils/requirements.py
rajat641: @nijel - the code `from django.core.cache import cache` looks magical. Can you refer a source where its available so that I can refer it for the future?

nijel: https://docs.djangoproject.com/en/3.0/topics/cache/#the-low-level-cache-api

weblate/utils/requirements.py
nijel: You're right with 12.0, these were mostly given as examples without deep investigation. Still, I think it's better to avoid such code if possible :-).

weblate/utils/requirements.py
rajat641: @nijel - This was fixed.

weblate/utils/requirements.py
rajat641: @nijel - Thanks for sharing. I read it over the weekend here.

weblate/utils/requirements.py
rajat641: @nijel - I have avoided the complex string manipulation that I did before. Now, I am using `connection.cursor().connection.get_server_info().split("-", 1)[0]` to parse MySQL version by quering the database and a similar approach in PostgreSQL

rajat641: Please let me if there is any better way to do things. The previous one was not definitely hackish.

rajat641: Also, after the code changes `weblate list_versions` command works fine. I have tested it after databases.

weblate/utils/requirements.py
```diff
@@ -169,10 +172,66 @@ def get_versions():
     return result
 
 
+def get_db_version():
+    if connection.vendor.upper() == "POSTGRESQL":
+        try:
+            cursor = connection.cursor()
+            cursor.execute("SHOW server_version")
+            version = cursor.fetchall()[0]
```
nijel, line 180: This should be better:

```suggestion
            version = cursor.fetchone()
```

weblate/utils/requirements.py
```diff
@@ -169,10 +172,66 @@ def get_versions():
...
+def get_cache_version():
+    try:
+        if (
+            settings.CACHES.get("default", {}).get("BACKEND")
+            == "django_redis.cache.RedisCache"
+        ):
+            version = cache.client.get_client().info()["redis_version"]
+            result = ("Redis", "https://redis.io/", version)
+        else:
+            raise ImproperlyConfigured("please install a redis server.")
```
nijel, line 218: I'm not sure about this exception. Weblate will probably work just fine with other cache backend, so raising an exception seems too strict here.

rajat641: I will remove this exception. But I'm not sure, how to get the cache version/name for some other backend.There can be many such services. OR  I have to show only cache version if the backend is `django_redis.cache.RedisCache`

weblate/utils/requirements.py
rajat641: I'll fix this and update it.

weblate/utils/requirements.py
nijel: There is no reason to cover all options, getting it only for redis is good enough as that will cover most use cases. But it should not break in others.

weblate/utils/requirements.py
rajat641: @nijel - This is fixed.

weblate/utils/requirements.py
rajat641: @nijel - I have removed the `ImproperlyConfigured` exception. Also, I tested the code with another cache namely Memcached and its working fine.

nijel:
Merged, thanks for your contribution! I've made some further improvements to the code in 5ce7ecb634

rajat641:
Thanks, @nijel . I have seen the code changes. It was a great learning experience.

PR #3939, opened 2020-05-25
[machine translation] fix mt, using microsoft api and mymemory, for language sub-code

zzazang:

- [o] Every commit has a message which describes it
- [o] Every commit is needed on its own, if you have just minor fixes to previous commits, you can squash them
- [] Any code changes come with test case
- [] Any new functionality is covered by user documentation

* when using microsoft cognitive translation api, previous code cannot translate to zh-Hant.
* mymemory is supporting language sub-code like zh-Hant, but existing code was not.

### DIFF
--- a/weblate/machinery/microsoft.py
+++ b/weblate/machinery/microsoft.py
@@ -91,16 +91,13 @@ class MicrosoftCognitiveTranslation(MachineTranslation):
         return self._access_token
 
     def map_language_code(self, code):
-        """Convert language to service specific code.
-
-        Remove second part of locale in most of cases.
-        """
-        return super().map_language_code(code).replace("_", "-").lower()
+        """Convert language to service specific code."""
+        return super().map_language_code(code).replace("_", "-")
 
     def download_languages(self):
         """Download list of supported languages from a service.
 
-        Example of the response:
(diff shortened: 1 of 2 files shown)


weblate/machinery/microsoft.py
```diff
@@ -95,8 +95,7 @@ def map_language_code(self, code):
 
         Remove second part of locale in most of cases.
         """
-        return super().map_language_code(code).replace("_", "-").lower()
-
+        return super().map_language_code(code.replace("_", "-").lower())
```
nijel, line 98: This change breaks handling of `en_devel`. Probably it's better to adjust aliases to deal Weblate language codes.

weblate/machinery/mymemory.py
```diff
@@ -31,7 +31,7 @@ class MyMemoryTranslation(MachineTranslation):
 
     def map_language_code(self, code):
         """Convert language to service specific code."""
-        return super().map_language_code(code).replace("_", "-").lower()
+        return super().map_language_code(code.replace("_", "-"))
```
nijel, line 34: Again, this breaks handling of `en_devel`.

PR #3945, opened 2020-05-25
Adds better API docs to create HyperLinked POST requests

### LINKED ISSUE #3208: API documentation does not describe create request
what the nice docu lacks is the info when it talks about parameter: does it mean header or body. parameter sounds like header, however in case of components it feels natural to take the response json from get-component call, adapt the name etc. and send it as a post
will find out when i am on the correct version.
Thanks again !

_Originally posted by @PawelSpoon in https://github.com/WeblateOrg/weblate/issues/1225#issuecomment-552076460_

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/83790431-api-documentation-does-not-describe-create-request?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SaptakS:
Fixes #3208 

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -685,7 +685,7 @@ Projects
         curl \
             -d operation=pull \
             -H "Authorization: Token TOKEN" \
-            http://example.com/api/components/hello/weblate/repository/
+            http://example.com/api/projects/hello/repository/
 
     **JSON request example:**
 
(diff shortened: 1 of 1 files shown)


nijel:

PR #3955, opened 2020-05-27
Listings: avoid extraneous underline on hover

julen:
Text before:
before_text
Text after:
after_text

Icons before:
before_img
Icons after:
after_img


nijel:

PR #3956, opened 2020-05-27
Editor: move client side code apart

### LINKED ISSUE #3875: Browser's native Cmd/Ctrl+<N> is blocked everywhere
**Describe the bug**
In any Weblate page, pressing Cmd/Ctrl+\<N\> (N being a tab number), will prevent browser's default behavior, which is to change switch to a specific tab.

**To Reproduce**
1. Go to any Weblate page
2. Press Cmd/Ctrl+\<N\> (where N != current tab number)

**Expected behavior**
I understand this is desired behavior for translate pages to enhance the translating experience, however it shouldn't be the case for any other page. Altering/hijacking browser native functionality is usually bad practice, and it interrupts visitors and their expectations.

julen:
This moves the code as-is from a globally-included script file to an
editor-specific script. It includes code for both regular and zen mode
editing.

Further improvements should be made on top of this to avoid leaking
global variables and/or ensuring only editor mode-specific code is
initialized.

Fixes #3875.

### DIFF
--- a/weblate/static/loader-bootstrap.js
+++ b/weblate/static/loader-bootstrap.js
@@ -1,10 +1,5 @@
 var loading = 0;
-var machineTranslationLoaded = false;
-var translationMemoryLoaded = false;
 var activityDataLoaded = false;
-var lastEditor = null;
-
-var IS_MAC = /Mac|iPod|iPhone|iPad/.test(navigator.platform);
 
 // Remove some weird things from location hash
 if (window.location.hash && (window.location.hash.indexOf('"') > -1 || window.location.hash.indexOf('=') > -1)) {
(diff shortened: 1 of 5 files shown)


julen:
@nijel I'm not 100% familiar with all Weblate features so please do check if something has been inadvertently broken. I did my best to identify editor-specific code, but if you spot any leftovers in _loader-bootstrap.js_, please let me know.

weblate/static/loader-bootstrap.js
```diff
@@ -850,71 +435,6 @@ $(function () {
         }
     }
 
-    /* Translation editor */
-    Mousetrap.bindGlobal(
-        ['alt+enter', 'mod+enter'],
-        submitForm
-    );
```
nijel, line 857: This is used to submit form anywhere, it's not specific to the editor.

weblate/static/loader-bootstrap.js
```diff
@@ -1449,21 +704,6 @@ $(function () {
     /* Copy to clipboard */
     var clipboard = new ClipboardJS('[data-clipboard-text]');
 
-    /* Auto translate source select */
```
nijel, line 1452: This block is not for the editor page, auto translate is standalone.

weblate/static/loader-bootstrap.js
```diff
@@ -980,69 +464,6 @@ $(function () {
         $window.scrollTop($(href).offset().top);
     });
```
nijel, line 465: This is IMHO for editor as well.

weblate/static/loader-bootstrap.js
julen: Do you have any pointers where? I understand this is needed in the editor because it's a textarea, but any other regular forms can/should be natively submitted by pressing <kbd>Enter</kbd> (the `submit` event can be captured to attach XHR behavior).

nijel: For example, settings or addon configuration forms have textareas and it's useful there as well to submit with Ctrl+Enter.

julen:
Ok, all comments have been addressed — please let me know if anything else is missing.

nijel:

PR #3962, opened 2020-05-28
Editor: use delegated events

julen:
* Moves initialization out of the global scope into a function
* Sets up events as delegated, i.e. no need to re-attach handlers when new contents are loaded (e.g. in Zen mode)


nijel:

PR #3964, opened 2020-05-28
Docs: fix typos

julen:
As spotted while reading.


nijel:
Thanks! I should have read it after writing...

PR #3965, opened 2020-05-28
Editor: separate full & zen modes

julen:
* Avoids polluting the global namespace
* Makes the distinction between full & zen modes more explicit (more to come)

@nijel I ignore what's the set of browsers Weblate is meant to support, so I kept on using ES5 language features.

### DIFF
--- a/weblate/templates/translate.html
+++ b/weblate/templates/translate.html
@@ -9,12 +9,8 @@
 
 {% block extra_script %}
 {% compress js %}
-<script defer data-cfasync="false" src="{% static 'editor.js' %}{{ cache_param }}"></script>
-<script type="text/javascript">
-document.addEventListener('DOMContentLoaded', function() {
-    new WLT.editor.Full();
-});
-</script>
+<script defer data-cfasync="false" src="{% static 'editor/base.js' %}{{ cache_param }}"></script>
+<script defer data-cfasync="false" src="{% static 'editor/full.js' %}{{ cache_param }}"></script>
 {% endcompress %}
 {% endblock %}
 
--- a/weblate/templates/translate.html
+++ b/weblate/templates/translate.html
@@ -10,6 +10,11 @@
 {% block extra_script %}
 {% compress js %}
 <script defer data-cfasync="false" src="{% static 'editor.js' %}{{ cache_param }}"></script>
+<script type="text/javascript">
+document.addEventListener('DOMContentLoaded', function() {
+    new FullEditor();
+});
+</script>
 {% endcompress %}
 {% endblock %}
 
--- a/weblate/templates/translate.html
+++ b/weblate/templates/translate.html
@@ -12,7 +12,7 @@
 <script defer data-cfasync="false" src="{% static 'editor.js' %}{{ cache_param }}"></script>
 <script type="text/javascript">
 document.addEventListener('DOMContentLoaded', function() {
-    new FullEditor();
+    new WLT.editor.Full();
 });
 </script>
 {% endcompress %}
--- a/weblate/static/editor.js
+++ b/weblate/static/editor.js
@@ -99,6 +99,141 @@ function Editor() {
     this.$translationArea[0].focus();
 }
 
+function FullEditor() {
+    Editor.call(this);
+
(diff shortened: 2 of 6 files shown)


weblate/templates/translate.html
```diff
@@ -10,6 +10,11 @@
 {% block extra_script %}
 {% compress js %}
 <script defer data-cfasync="false" src="{% static 'editor.js' %}{{ cache_param }}"></script>
+<script type="text/javascript">
+document.addEventListener('DOMContentLoaded', function() {
+    new WLT.editor.Full();
+});
+</script>
```
nijel, line 17: Inline script violates current CSP.

julen: We can add a nonce value to the script if the security middleware is in use (I don't think adding `unsafe-inline` would be an option). Do you have any other suggestions?

nijel: Does it have to be inline?

In case it is, the another question is what Django compressor does with this. You can try it in the debug mode by setting `COMPRESS_ENABLED = True` in settings.py (it is disabled by default in the debug mode).

julen: Interesting — it takes the contents out to a file. So is this a non-issue?

nijel: Maybe it is better to place it outside the HTML directly?

julen: The script file would need to be split in two files, one for each editor mode, and that'd work, however I very much think that's overkill given the few lines of script code.

nijel: I was thinking it might make sense to split it up more. In the end there are three things there:

* the editor handling itself (autosizing, copying source, keyboard)
* the translate page (tabs, glossary, navigation)
* the zen page (autosaving, autoloading)

But I'm in no way expert on JavaScript, so this is really just asking questions, and definitely not giving advices ;-).

julen: Those three things you mention are already conceptually split in this PR: a base class with common functionality and the specific modes in their own classes.

The cost of splitting into more files increases the amount of necessary network requests per each translate page (and the amount of simultaneous requests is limited by browsers, to generally 6 in HTTP/1.1), which mostly affect the initial load time.

Initial load time can be more or less important depending on the type of website users are loading: it's critical for content-heavy sites like newspapers/blogs, but not as much for long-running apps like GMail or GDocs. App-like sites also heavily cache scripts, so the price new users pay is compensated in subsequent visits.

On the other hand, splitting into more dedicated script files means there's less code in certain pages to parse and evaluate by the browser at load time. So while there's an extra initial network request (alleviated by caching in subsequent requests), the browser processes less.

Given the few lines of code needed for Zen mode, I was initially suggesting it would be overkill to incur in an extra network request, because the code parsing for the Zen mode is probably negligible. But there's no way to know that without measuring and understanding the context: do you have any insights on how people are generally using Weblate? Do they stick to one mode? Do they switch from one to the other frequently?

Maybe it's thinking way ahead of potential problems, and given the few lines of code the Weblate editor has, it's probably OK to have separate files for each mode and have them initialize themselves at script load time.

At least I hope this gives you some extra insights to take into account in the future.

nijel: > The cost of splitting into more files increases the amount of necessary network requests per each translate page (and the amount of simultaneous requests is limited by browsers, to generally 6 in HTTP/1.1), which mostly affect the initial load time.

Compressor can make it just single request, if that still matters these days when most browsers support HTTP/2.

> But there's no way to know that without measuring and understanding the context: do you have any insights on how people are generally using Weblate? Do they stick to one mode? Do they switch from one to the other frequently?

Most people do not use Zen mode at all.

julen: Cool, I'll split this into separate files then.

Also, and as I was wondering in the initial comment, what's the set of browsers/versions Weblate is committed to support?

nijel: I don't think we have some kind of commitment, so it's based of what Bootstrap 3.4 supports.

julen:
I have split the modes into two separate files as discussed. Some mode-specific code still lives in the common base file, but I'll leave that for a follow-up PR.

nijel:

PR #3970, opened 2020-05-29
fix: fix widget image 404 bug

howiezhao:
When starting with `localhost:8080`, the widget image is not found, this may be because the `get_site_url` function cannot get the port number, in `widgets.py`, i think should use `request.build_absolute_uri` function, so why use `get_site_url` function?
Screenshot from 2020-05-29 22-42-13


nijel:
`get_site_url` can get whatever you store in the sites, see https://docs.weblate.org/en/latest/admin/install.html#production-site

howiezhao:
Thanks for your explanation, it's my fault that I didn't read the document seriously, I will close this PR. @nijel

PR #3975, opened 2020-06-01
Adds API for monolingual languages

### LINKED ISSUE #3261: Add API for adding new monolingual strings
Currently, this is UI only feature and might be useful in the API as well.

It calls `Translation.new_unit` under the hood.

_Originally posted by @jornvl90 in https://github.com/WeblateOrg/weblate/issues/1145#issuecomment-555039651_

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/84738704-add-api-for-adding-new-monolingual-strings?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SaptakS:
Fixes #3261 

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1414,6 +1414,18 @@ Translations
     :type language: string
     :>json array results: array of component objects; see :http:get:`/api/units/(int:id)/`
 
+.. http:post:: /api/translations/(string:project)/(string:component)/(string:language)/units/
+
+    Add new monolingual unit.
+
+    :param project: Project URL slug
+    :type project: string
+    :param component: Component URL slug
+    :type component: string
+    :param language: Translation language code
+    :type language: string
+    :<json string key: Name of translation unit
+    :<json string value: The translation unit value
 
 .. http:get:: /api/translations/(string:project)/(string:component)/(string:language)/file/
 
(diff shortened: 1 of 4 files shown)


weblate/api/views.py
```diff
@@ -896,6 +897,28 @@ def units(self, request, **kwargs):
 
         return self.get_paginated_response(serializer.data)
 
+    @action(detail=True, methods=["post"])
+    def add_monolingual(self, request, **kwargs):
+        obj = self.get_object()
```
nijel, line 902: Should also check for permissions here: request.user.has_perm("unit.add", translation)

docs/api.rst
```diff
@@ -1491,6 +1491,19 @@ Translations
     :<json string operation: Operation to perform: one of ``push``, ``pull``, ``commit``, ``reset``, ``cleanup``
     :>json boolean result: result of the operation
 
+.. http:post:: /api/translations/(string:project)/(string:component)/(string:language)/add_monolingual/
```
nijel, line 1494: Thinking about that, I'd call the endpoint just `add`. Right now it's limited to (some) monolingual files, but that might be lifted in the future. ... Or even better, this should be an action on existing `units` endpoint - as this is what it does in the end - it adds new unit.

SaptakS: That makes sense. I think I will add it to the existing units endpoint by allowing POST if monolingual files

nijel: It just needs the permission check. It doesn't work on all monolingual files and it might work on others in future.

weblate/api/views.py
```diff
@@ -885,10 +886,32 @@ def changes(self, request, **kwargs):
...
+                    data={
+                        "result": "Unsuccessful",
+                        "detail": "Translation with this key seem to already exist!",
+                    },
+                    status=status.HTTP_400_BAD_REQUEST,
+                )
+            else:
+                obj.new_unit(request, key, value)
+                serializer = self.serializer_class(obj, context={"request": request})
+                return Response(serializer.data, status=status.HTTP_200_OK,)
```
nijel, line 913: Else is not needed:

```suggestion

            obj.new_unit(request, key, value)
            serializer = self.serializer_class(obj, context={"request": request})
            return Response(serializer.data, status=status.HTTP_200_OK,)
```

nijel:

PR #3980, opened 2020-06-02
Editor: factor out selectors

julen:
* Makes highlight initialization explicit
* Factors out its selectors


nijel:

PR #3984, opened 2020-06-03
Get stats cleanup

nijel:


nijel:
@SaptakS  This is cleanup after a1156c634742977600d7bce5020ffe6f1448970a to remove get_stats methods and use serializer only in this case.

PR #3986, opened 2020-06-03
getNumericKey: simplify implementation

julen:
Or should we rather inline it?

### DIFF
--- a/weblate/static/editor/base.js
+++ b/weblate/static/editor/base.js
@@ -11,12 +11,7 @@ WLT.Editor = (function () {
     var $document = $(document);
 
     function getNumericKey(idx) {
-        var ret = idx + 1;
-
-        if (ret === 10) {
-            return '0';
-        }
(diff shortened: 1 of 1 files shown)


weblate/static/editor/base.js
```diff
@@ -11,12 +11,7 @@ WLT.Editor = (function () {
     var $document = $(document);
 
     function getNumericKey(idx) {
-        var ret = idx + 1;
-
-        if (ret === 10) {
-            return '0';
-        }
-        return ret;
+        return (idx + 1) % 10;
```
nijel, line 14: I'm not sure if the original returning of a string had any reason. Did you look into that?

julen: It doesn't make a difference, plus it's converted to a string in the places it's used anyway (string concatenation or Django's `interpolate()` function).

nijel:

PR #3994, opened 2020-06-04
management command missing from example

minusf:
I also have a question relating to auto translation:

In the GUI i cannot find a button to auto translate a **whole project** (all components) using machine translation for a given language.

Can this management command be used for this purpose?  If yes, how?
Components seems to be a mandatory argument, even if `--mt` is specified.


nijel:

PR #3997, opened 2020-06-05
TD-630: Out of storage space

bperreault-va:
Apparently we're out of storage on the device. Hopefully that means increasing the size of the PVC.


bperreault-va:
Oof, accidentally made this PR against the WeblateOrg master branch. Closing and moving to fork.

PR #3999, opened 2020-06-07
Editor: take TM & MT code apart

julen:
Moves the code to the full editor.


nijel:

PR #4001, opened 2020-06-08
Adds code to create, update and delete languages

SaptakS:


weblate/api/views.py
```diff
@@ -934,16 +934,34 @@ def destroy(self, request, *args, **kwargs):
...
-class LanguageViewSet(viewsets.ReadOnlyModelViewSet):
+class LanguageViewSet(viewsets.ModelViewSet):
     """Languages API."""
 
     queryset = Language.objects.none()
     serializer_class = LanguageSerializer
     lookup_field = "code"
 
     def get_queryset(self):
+        if self.request.user.is_superuser:
```
nijel, line 945: I think it makes more sense to check the actual permission:

```suggestion
        if request.user.has_perm("language.edit"):
```

SaptakS:
Yes yes. I am on it. Almost fixed. Testing with all other different tests.

PR #4013, opened 2020-06-11
Editor: move remaining code to full mode

julen:
* Moves all remaining code to full mode (bug reporting, translation area persistence)
* Reuses jQuery objects as much as possible
* Binds events to `$editor` when possible
* `localStorage`-related cleanups


nijel:

PR #4018, opened 2020-06-11
Allows multiple componentlists to be associated with a group

### LINKED ISSUE #3905: The ability to add multiple component-lists to a permissions group.

Currently we can only add one component-list to a permission group and I would like to be able to add multiple. 

Be able to POST multiple component lists to `/api/groups/(int: id)/componentlists/`.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/91472860-the-ability-to-add-multiple-component-lists-to-a-permissions-group?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SaptakS:
Fixes #3905 

PS: I think might need to add some kind of data migration as well to move from foreignKey to ManyToMany. Not sure exactly what approach to take for that.


weblate/auth/migrations/0009_migrate_componentlist.py
```diff
@@ -0,0 +1,25 @@
...
+from django.db import migrations
+
+
+def migrate_componentlist(apps, schema_editor):
+    Group = apps.get_model("weblate_auth", "Group")
+    db_alias = schema_editor.connection.alias
+    groups = Group.objects.using(db_alias).all()
+    for group in groups:
+        if group.componentlist:
+            group.componentlists.add(group.componentlist)
```
nijel, line 12: Better to do the filtering on database side:

```suggestion
    groups = Group.objects.using(db_alias).filter(componentlist__isnull=False)
    for group in groups:
        group.componentlists.add(group.componentlist)
```

Issue #3911, opened 2020-05-19
Cmd+M doesn't switch to Machine Translation tab


julen:
**Describe the bug**
In Mac, pressing Cmd+M doesn't switch Weblate to the Machine Translation tab. Instead, it triggers the native Mac behavior of minimizing the current window.

Other platforms to be confirmed.

**To Reproduce**
1. Open the editor
2. Press Cmd+M

**Expected behavior**
Machine Translation tab is opened.

nijel:
It seems to sometimes work in Firefox on Linux. For most of the time the default browser action is executed.

julen:
Ok, that sounds racy. The default must be prevented the same way we did for Ctrl/Cmd+I in #3891; I verified that works fine in Mac.

Nonetheless let me get back to these shortcut issues once I split the editor code as suggested in #3875.

PR #4035, opened 2020-06-15
Editor: refactor machinery code

julen:
* Refactors machinery-related code
* Converts success/error processing logic into methods to further reuse cached objects
* Fixes machinery keyboard shortcut bugs (#3911)

@nijel please make sure to check the rendering of MT & TM results — I didn't touch the code to render machinery result rows but I wouldn't be surprised if other changes broke something related to it because it's pretty fragile in its current shape.

### DIFF
--- a/weblate/static/editor/full.js
+++ b/weblate/static/editor/full.js
@@ -111,72 +111,80 @@
     };
 
     FullEditor.prototype.initTabs = function () {
-        this.isMTLoaded = false;
-        this.isTMLoaded = false;
-
         /* Store active tab in a cookie */
         $('.translation-tabs a[data-toggle="tab"]').on('shown.bs.tab', function () {
             Cookies.remove('translate-tab', { path: '' });
             Cookies.set('translate-tab', $(this).attr('href'), { path: '/', expires: 365 });
         });
 
+        var self = this;
         /* Machine translation */
-        this.$editor.on('show.bs.tab', '[data-load="mt"]', function (e) {
-            if (this.isMTLoaded) {
+        this.isTMLoaded = false;
+        this.$editor.on('show.bs.tab', '[data-load="mt"]', function () {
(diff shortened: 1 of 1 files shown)


weblate/static/editor/full.js
```diff
@@ -111,72 +133,222 @@
...
-                    success: function (data) {
-                        processMachineTranslation(data, 'memory');
-                    },
-                    error: function (jqXHR, textStatus, errorThrown) {
-                        failedMachineTranslation(jqXHR, textStatus, errorThrown, 'memory');
-                    },
-                });
-                return false;
+    FullEditor.prototype.fetchMachinery = function (serviceName) {
+        var serviceType = serviceName === TM_SERVICE_NAME ? 'memory' : 'mt';
```
nijel, line 208: The type depends rather on source where it is triggered - the machine translation tab displays translation memory matches as well, so my guess is that this will break display of the translation memory matches on the machine translation tab.

julen: Ah I see, the MT tab also provides TM results... what's the motivation behind it if I may ask? I'd like to understand it because as a translator I find that quite confusing.

nijel: There are two reasons, but non of them is really good :-).

* There used to be just the MT tab in past and when adding TM tab, the results stayed there.
* Some third-party services are also TM (or even hybrid like MyMemory, which can provide both translation memory matches and machine translation).

julen: Ok, I see... I'll make sure not to touch existing behavior but please reconsider this in the future! None of the reasons you mentioned seem like roadblocks not to have a real split.

nijel: Issue noted: https://github.com/WeblateOrg/weblate/issues/4041

nijel:

Issue #4034, opened 2020-06-15
Evaluate usage of django-compressor


nijel:
Back then it was clearly beneficial to introduce it as number of HTTP requests was real concern. Maybe it's not the case these days with HTTP/2.

Things to consider:

* Current setup does add overhead to each request (~25 redis requests to check freshness of the cache)
* Are multiple requests still concern with HTTP/2? Maybe it would be more beneficial to utilize per script caching and serve them individually

Possible outcomes:

* Current approach is good
* Reconfigure django-compressor to better fit our use case
* Drop django-compressor usage

nijel:
Typical overhead caused by django-compressor, in total it accounts to 11ms: 

![image](https://user-images.githubusercontent.com/212189/84656064-60693080-af12-11ea-9cbe-d1bad608c763.png)

julen:
[Is HTTP/2 even an option](https://caniuse.com/#search=http2) if Weblate is meant to work in [browsers supported by Bootstrap 3.4](https://getbootstrap.com/docs/3.4/getting-started/#support-browsers)? Refs. https://github.com/WeblateOrg/weblate/pull/3965#discussion_r432321138

nijel:
The reality is that most of the users use browsers which do support it (on Hosted Weblate it is more than 99%), so I think it's okay to have slightly degraded performance for the rest. 

The browser support is something what has never been really evaluated and maybe it's time to change that :-).

julen:
> The browser support is something what has never been really evaluated and maybe it's time to change that :-).

That'd be cool for some frontend code improvements! 👍

nijel:
I'm really open to suggestions in that. I'm not really a frontend guy...

nijel:
Anyway to give some starting point, I'd like to move to Bootstrap 4 at some point (see https://github.com/WeblateOrg/weblate/issues/2916), so we might target at [Bootstrap 4 browser requirements](https://getbootstrap.com/docs/4.1/getting-started/browsers-devices/).

julen:
A common approach is to support latest 2 stable versions of browsers (e.g. [StackExchange does this](https://meta.stackexchange.com/a/71208)). Nonetheless, it'd be good (also for the web) to drop IE support as it's a browser stuck in the past. You should evaluate your userbase's browser usage data, though — it's your business after all.

nijel:
[...]

PR #4056, opened 2020-06-17
ES2015+: use arrow functions where applicable

julen:
The conversion to arrow functions for now only happens where previously
`var self = this` was necessary.

Refs. https://github.com/WeblateOrg/weblate/issues/4034#issuecomment-644618132


nijel:

PR #4058, opened 2020-06-17
Glossary: fix focusing after closing dialog

julen:
The previous code was assuming synchronous behavior, when it truly is
asynchronous. Therefore the translation textarea `focus()` was always
occuring before the modal was closed, and the modal puts the focus
elsewhere after it being closed.

### DIFF
--- a/weblate/static/editor/full.js
+++ b/weblate/static/editor/full.js
@@ -446,6 +446,15 @@
             e.preventDefault();
         });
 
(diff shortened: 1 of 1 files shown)


nijel:

julen:
@nijel please note this only fixes existing broken behavior. I don't think this is the best UX though — errors should be handled and displayed in a more graceful way, e.g. form field errors.

PR #4069, opened 2020-06-19
Adds subscriptions API associated with User

### LINKED ISSUE #2557: make user's subscription list api
I make a third party app. I can not found api that has not  users' subscription list.
please make user's subscription list.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/70179886-make-user-s-subscription-list-api?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F253393&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SaptakS:
Fixes #2557 

### DIFF
--- a/weblate/api/serializers.py
+++ b/weblate/api/serializers.py
@@ -21,6 +21,7 @@ from django.conf import settings
 from django.core.exceptions import PermissionDenied
 from rest_framework import serializers
 
+from weblate.accounts.models import Subscription
 from weblate.auth.models import Group, Permission, Role, User
 from weblate.lang import data
 from weblate.lang.models import Language, Plural
@@ -168,6 +169,9 @@ class UserSerializer(serializers.ModelSerializer):
     groups = serializers.HyperlinkedIdentityField(
         view_name="api:group-detail", lookup_field="id", many=True, read_only=True,
     )
+    subscriptions = serializers.HyperlinkedIdentityField(
+        view_name="api:user-notifications", lookup_field="username",
+    )
 
     class Meta:
         model = User
@@ -176,6 +180,7 @@ class UserSerializer(serializers.ModelSerializer):
             "full_name",
             "username",
             "groups",
+            "subscriptions",
             "is_superuser",
             "is_active",
             "date_joined",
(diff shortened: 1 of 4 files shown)


weblate/api/views.py
```diff
@@ -314,6 +316,65 @@ def groups(self, request, **kwargs):
...
+            return Response(status=status.HTTP_204_NO_CONTENT)
+
+        if request.method == "GET":
+            serializer = NotificationSerializer(
+                subscription, context={"request": request}
+            )
+        else:
+            self.perm_check(request)
+            serializer = NotificationSerializer(
+                subscription, data=request.data, context={"request": request},
```
nijel, line 371: I guess patch will not work properly as this is not handling partial data.

SaptakS: Ah yeah right. I need to do the patch check and handle partial data in serializer and write tests for that.

SaptakS: Added a commit for the same. Should handle it.

weblate/api/tests.py
```diff
@@ -203,6 +204,103 @@ def test_add_group(self):
...
+        )
+        self.assertEqual(response.data["notification"], "RepositoryNotification")
+
+    def test_patch_notifications(self):
+        user = User.objects.filter(is_active=True).first()
+        response = self.do_request(
+            "api:user-notifications-details",
+            kwargs={
+                "username": user.username,
+                "subscription_id": Subscription.objects.filter(user=user).first().id,
```
nijel, line 281: I'm affraid this is not reliable - there is no defined ordering on Subscription, so what is returned by `first()` depends on database ordering and thus is not something to rely on. It could easily happen that you already get subscription on "RepositoryNotification" and the patch doesn't work...

SaptakS: Hmm. That's true. what would be a better test? filtering by user and notification != "RepositoryNotification" ?

nijel: I'd probably just picked one of the notifications, for example "NewAnnouncementNotificaton" and use that for tests.

List of notifications subscribed by default for users is here:

https://github.com/WeblateOrg/weblate/blob/939f36c141244fc68fc73df5d6159e2792a7f129/weblate/accounts/data.py#L23-L33

weblate/api/serializers.py
```diff
@@ -168,6 +169,9 @@ class UserSerializer(serializers.ModelSerializer):
     groups = serializers.HyperlinkedIdentityField(
         view_name="api:group-detail", lookup_field="id", many=True, read_only=True,
     )
+    subscriptions = serializers.HyperlinkedIdentityField(
+        view_name="api:user-notifications", lookup_field="username",
+    )
```
nijel, line 174: I'd probably call this notifications as well to be consistent in the API. The reason I don't want to use "subscriptions" here because it's confusing - notification subscription or service subscription.

nijel: I will fix this in master branch as I've merged this accidentally when pushing other changes....

SaptakS: ok.

Issue #3248, opened 2019-11-28
Register EXPORTERS in settings.py


fdosanjuan:
I think it would be useful to be able to select which exporters to have available.

I propose to have a similar solution to the one used for machine translation services, i.e., register in settings.py the list of EXPORTERS as it is done with MT_SERVICES.

What do you think?

Thank you!!

nijel:
It makes sense to configure it for some deployments, patches are welcome :-)

55abhilash:
I would like to work on this. Shall I send a pull request with changes?

nijel:
Yes, you are welcome to do that.

nijel:
Weblate uses https://pypi.org/project/django-appconf/ to define default values for settings, it's then defined here:

https://github.com/WeblateOrg/weblate/blob/93d569739c3a3d700d57e719b55f9fb17265bd3f/weblate/formats/models.py#L57-L99

Defined with WEBLATE prefix, so the setting is in the end called WEBLATE_FORMATS, unforntunately there is not much consistence in naming settings so far.

55abhilash:
#4091

PR #4091, opened 2020-06-25
Register Exporters in Settings

55abhilash:
Added EXPORTERS under formats/models.py, which can be customized to enable/disable certain exporter.

### DIFF
--- a/weblate/formats/exporters.py
+++ b/weblate/formats/exporters.py
@@ -18,7 +18,6 @@
 #
 """Exporter using translate-toolkit."""
 
-from django.conf import settings
 from django.http import HttpResponse
 from django.utils.functional import cached_property
 from django.utils.translation import gettext_lazy as _
@@ -40,24 +39,9 @@ from weblate.formats.external import XlsxFormat
 from weblate.formats.ttkit import TTKitFormat
 from weblate.trans.util import split_plural, xliff_string_to_rich
 from weblate.utils.site import get_site_url
-from weblate.formats.models import EXPORTERS
-
 
 # Map to remove control characters except newlines and tabs
 _CHARMAP = dict.fromkeys(x for x in range(32) if x not in (9, 10, 13))
-
-
-def get_exporter(name):
-    """Return registered exporter."""
-    return EXPORTERS[name]
-
-
-def list_exporters(translation):
-    return [
-        {"name": x.name, "verbose": x.verbose}
-        for x in sorted(EXPORTERS.values(), key=lambda x: x.name)
-        if x.supports(translation)
-    ]
 
 
 class BaseExporter:
--- a/weblate/formats/exporters.py
+++ b/weblate/formats/exporters.py
@@ -47,6 +47,7 @@ _CHARMAP = dict.fromkeys(x for x in range(32) if x not in (9, 10, 13))
 EXPORTERS = {}
 WEBLATE_EXPORTERS = settings.WEBLATE_EXPORTERS
 
+
 def register_exporter(exporter):
     """Register an exporter."""
     if exporter.__name__ in WEBLATE_EXPORTERS:
--- a/weblate/formats/exporters.py
+++ b/weblate/formats/exporters.py
@@ -40,19 +40,11 @@ from weblate.formats.external import XlsxFormat
 from weblate.formats.ttkit import TTKitFormat
 from weblate.trans.util import split_plural, xliff_string_to_rich
 from weblate.utils.site import get_site_url
+from weblate.formats.models import EXPORTERS
+
 
 # Map to remove control characters except newlines and tabs
 _CHARMAP = dict.fromkeys(x for x in range(32) if x not in (9, 10, 13))
-
-EXPORTERS = {}
-WEBLATE_EXPORTERS = settings.WEBLATE_EXPORTERS
-
-
-def register_exporter(exporter):
-    """Register an exporter."""
-    if exporter.__name__ in WEBLATE_EXPORTERS:
-        EXPORTERS[exporter.name] = exporter
-    return exporter
 
 
 def get_exporter(name):
@@ -108,6 +100,10 @@ class BaseExporter:
         if len(plurals) == 1:
             return self.string_filter(plurals[0])
         return multistring([self.string_filter(plural) for plural in plurals])
+
+    @classmethod
+    def get_identifier(cls):
+        return cls.name
 
     def get_storage(self):
         return self.storage_class()
@@ -203,7 +199,6 @@ class BaseExporter:
         return
 
 
-@register_exporter
 class PoExporter(BaseExporter):
     name = "po"
     content_type = "text/x-po"
@@ -243,7 +238,6 @@ class XMLExporter(BaseExporter):
         unit.settarget(word, self.language.code)
 
 
-@register_exporter
 class PoXliffExporter(XMLExporter):
     name = "xliff"
     content_type = "application/x-xliff+xml"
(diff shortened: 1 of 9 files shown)


nijel:
I think you misunderstood #3248 - the point is to be able to fully configure what exporters are available, including adding own one. The way you implemented it allows just removing builtin ones.

What I think is needed:

* Remove register_exporter decorator
* Include fully quallified name in the setting
* Use weblate.utils.classloader.ClassLoader to register the exporters
* Merge ExportersConf into FormatsConf, there is no need to separate them when they have same prefix

55abhilash:
Ok. I understood now.
Doing the needed changes...

55abhilash:
The exporter methods (get_exporter, list_exporters) and all exporter classes are in the same file (exporters.py); it also has a variable EXPORTERS which is directly used in many views. Now, when classloader is trying to load the exporters, it is failing due to circular import, because EXPORTERS is assigned load_data() return value from ExporterLoader(child of ClassLoader) and the classloader itself tries to import exporters.py.
We can put the classes in a different file to resolve this: this new file will be imported by classloader and not the file with the variable and methods used throughout the codebase.
Sorry if I am missing something or if this is a stupid question :P The codebase is pretty new to me!

nijel:
The classloader instance is usually placed in models.py for rest of the code.

55abhilash:
I have pushed the changes master branch of the fork. Its not reflected here yet for some reason.
I have created a ExporterLoader class which inherits from class loader and have made changes in 2 places where the variable EXPORTERS is used. I have removed the register exporter decorater.
I have one question though. The class FileFormatLoader has redefinitions of methods autoload and load_data. In the redefined load_data, it is adding to self.errors the classes for which loading has failed.
`        for fileformat in list(result.values()):
            error_name = "File format: {0}".format(fileformat.format_id)
            delete_configuration_error(error_name)
            try:
                fileformat.get_class()
            except (AttributeError, ImportError) as error:
                result.pop(fileformat.format_id)
                self.errors[fileformat.format_id] = str(error)
`
I suppose this should be done for exporters as well (I haven't done this rn)?

weblate/formats/exporters.py
```diff
@@ -39,18 +40,12 @@
 from weblate.formats.ttkit import TTKitFormat
 from weblate.trans.util import split_plural, xliff_string_to_rich
 from weblate.utils.site import get_site_url
+from weblate.formats.models import EXPORTERS
```
nijel, line 43: Probably this import should not be here and rather related code moved to models.py?

weblate/formats/models.py
```diff
@@ -24,6 +24,15 @@
 from weblate.utils.classloader import ClassLoader
 
 
+class ExporterLoader(ClassLoader):
+    def __init__(self):
+        super().__init__("WEBLATE_EXPORTERS", False)
+        self.errors = {}
```
nijel, line 30: I don't think we need to collect the errors here. All exporters should import just fine. This is different with the formats because some of them need optional dependencies.

weblate/formats/exporters.py
55abhilash: Okay. EXPORTERS is used in two methods in this file. So I will move those two methods to models.py (and also change wherever the methods are called, which is like 2-3 places).

55abhilash: I have fixed this with 28ba7f1 .

55abhilash:
There is a regression after these changes in Glossaries. Glossary view has a hardcoded list of exporters; now some of them might not be registered.
Changes for this will be in next commit.

weblate/formats/models.py
```diff
@@ -24,6 +24,28 @@
...
+
+
+EXPORTERS = ExporterLoader()
+
+
+def get_exporter(name):
+    """Return registered exporter."""
+    return EXPORTERS[name]
```
nijel, line 38: Maybe this is no longer needed? I don't see reason why not to use `EXPORTERS[name]` directly.

weblate/formats/models.py
```diff
@@ -24,6 +24,28 @@
...
+
+
+def list_exporters(translation):
```
nijel, line 41: Can you please make this a method of ExporterLoader?

weblate/formats/models.py
55abhilash: Makes sense.

weblate/formats/models.py
55abhilash: Ok.

weblate/formats/models.py
```diff
@@ -24,6 +24,22 @@
 from weblate.utils.classloader import ClassLoader
 
 
+class ExporterLoader(ClassLoader):
+    def __init__(self):
+        super().__init__("WEBLATE_EXPORTERS", False)
+        self.errors = {}
```
nijel, line 30:
```suggestion
```

weblate/formats/exporters.py
```diff
@@ -83,7 +62,7 @@ def __init__(
         else:
             self.project = project
             self.language = language
-            self.plural = language.plural
+            self.plural = language.plural if language is not None else None
```
nijel, line 65: Can you please explain reasons for this? The language really should be set at this point, otherwise other things might break as well.

55abhilash: Actually this was done because when a subclass of BaseExporter (eg. CSVExporter) is accessed in template (glossary.html), it tries to initialize the class, even though all we need is attributes. So the class is called without any arguments, hence language is None and the above mentioned statement threw error.

55abhilash: We can also do it by setting do_not_call_in_templates = True in BaseExporter. But I figured since language has default value None, there **might** be other cases too where language is not passed.

nijel: Maybe the glossary page should probably use something like list_exporters instead, which serves exactly this purpose to avoid need to instantiate the class (the very same issue was on the translation page).

55abhilash: Right. Makes sense to use list_exporters. Doing the changes.

nijel: Thanks!

weblate/formats/models.py
```diff
@@ -24,6 +24,27 @@
...
+            return [
+                {"name": x.name, "verbose": x.verbose}
+                for x in sorted(self.values(), key=lambda x: x.name)
+                if x.supports(translation)
+            ]
+        else:
+            return [
+                {"name": x.name, "verbose": x.verbose}
+                for x in sorted(self.values(), key=lambda x: x.name)
+            ]
```
nijel, line 42:
```suggestion
    def list_exporters(self, translation):
        return [
            {"name": x.name, "verbose": x.verbose}
            for x in sorted(self.values(), key=lambda x: x.name)
            if x.supports(translation)
        ]

    def list_exporters_filter(self, allowed):
        return [
            {"name": x.name, "verbose": x.verbose}
            for x in sorted(self.values(), key=lambda x: x.name)
            if x.name in allowed
        ]
```

weblate/glossary/views.py
```diff
@@ -396,6 +398,8 @@ def show_glossary(request, project, lang):
         .exclude(glossary_term=None)[:10]
     )
 
+    exporters = [exp for exp in EXPORTERS.list_exporters() if exp['name'] in EXPORT_TYPES]
```
nijel, line 401:
```suggestion
    exporters = EXPORTERS.list_exporters_filter(EXPORT_TYPES)
```

nijel:
Merged, what was still missing is documentation and I've added that in d8bb691ffdd3a6a36df84d824d1dfc47675a23fa

PR #4094, opened 2020-06-25
Adds user suggestions in comment section

SaptakS:


weblate/api/views.py
```diff
@@ -263,17 +264,30 @@ def repository(self, request, **kwargs):
...
+        model = User
+        fields = ["username"]
+
+
 class UserViewSet(viewsets.ModelViewSet):
     """Users API."""
 
     queryset = User.objects.none()
-    serializer_class = UserSerializer
+    serializer_class = FullUserSerializer
```
nijel, line 279: Does it have to be defined here? I'd rather just have the get_serializer_class method

SaptakS: Yeah. Not exactly needed.

SaptakS:
> Have you looked at option of using some existing rich editor component, which would support autocompletion? We want to do more things in the editor....

I did. I was trying my best to implement it minimally. Also, most RTE don't have the @mention support. There is [CKEditor](https://github.com/ckeditor/ckeditor4) which has @mention abilities. There is also [Quill](https://github.com/quilljs/quill/) which doesn't have @mention directly but has additional npm plugins which can do it.

Seems like both the projects are active. Though Quill seems to have more adoption that ckeditor based on github "used_by" stat.

SaptakS:
I have added the code to show the user suggestion. Need to still add tests. Not sure if documentation will be needed. I would rather add the test once we are sure and happy with the design and workflow.

weblate/templates/translate.html
```diff
@@ -10,12 +10,18 @@
 {% block extra_script %}
 {% compress js %}
 <script defer data-cfasync="false" src="{% static 'editor/base.js' %}{{ cache_param }}"></script>
-<script defer data-cfasync="false" src="{% static 'editor/full.js' %}{{ cache_param }}"></script>
+<script defer data-cfasync="false" src="{% static 'editor/full.js' %}{{ cache_param }}"></script><script defer data-cfasync="false" src="{% static 'codemirror/lib/codemirror.js' %}{{ cache_param }}"></script>
```
nijel, line 13: Please keep one script per line....

weblate/templates/translate.html
```diff
@@ -10,12 +10,18 @@
...
+<script defer data-cfasync="false" src="{% static 'codemirror/addon/hint/show-hint.js' %}{{ cache_param }}"></script>
+<script defer data-cfasync="false" src="{% static 'codemirror/mode/javascript/javascript.js' %}{{ cache_param }}"></script>
```
nijel, line 15: Is JavaScript mode really needed? I don't think we will ever edit JavaScript...

weblate/static/loader-codemirror.js
```diff
@@ -0,0 +1,78 @@
...
+        return userList;
+    }
+
+    CodeMirror.registerHelper('hint', 'userSuggestions', function (editor) {
+        var cur = editor.getCursor();
+        var curLine = editor.getLine(cur.line);
+        var tokenType = editor.getTokenTypeAt(cur);
+
+        // Disable user mention inside code
+        if(!!tokenType && tokenType.indexOf('comment') !== -1)
```
nijel, line 32: I don't think there is anything in sense of code comment...

weblate/static/loader-codemirror.js
```diff
@@ -0,0 +1,78 @@
...
+            // If there is current word set, We can filter out users from the
+            // main list and display them
+            userMentionList = getUserList(curWord);
+        }
+
+        return { list: userMentionList, from: CodeMirror.Pos(cur.line, start), to: CodeMirror.Pos(cur.line, end) };
+    });
+
+
+    var commentArea = document.getElementById('id_comment')
```
nijel, line 51: Selecting by hardcoded ID is most likely a bad idea. Mentions already work in more fields than comments.

weblate/static/loader-codemirror.js
```diff
@@ -0,0 +1,76 @@
...
+    });
+
+    // Add weblate bootstrap styling on the textarea
+    $('.CodeMirror').addClass('form-control');
+    $('.CodeMirror textarea').addClass('form-control');
+
+    $('#comment-form input[type=submit').on('click', function(e) {
+        e.preventDefault();
+        $('#comment-form').submit();
+    })
```
nijel, line 73: What is reason for this? The submit button should submit without any JavaScript code...

weblate/static/loader-codemirror.js
```diff
@@ -0,0 +1,76 @@
...
+            {
+                mode: 'text/javascript',
+                theme: 'weblate',
+                lineNumbers: false,
+            }
+        );
+        console.log(codemirror)
+
+        codemirror.on("keyup", function (cm, event) {
+            if(event.key === "@" || (event.shiftKey && event.keyCode === 50)) {
```
nijel, line 58: Is the second part needed? I will behave weirdly on non-English layouts (for example on Czech keyboard, it does trigger when typing "2").

weblate/static/loader-codemirror.js
```diff
@@ -0,0 +1,76 @@
...
+    $('textarea.codemirror-markdown').each(function(idx) {
+        var codemirror = CodeMirror.fromTextArea(
+            this,
+            {
+                mode: 'text/javascript',
+                theme: 'weblate',
+                lineNumbers: false,
+            }
+        );
+        console.log(codemirror)
```
nijel, line 55: Forgotten debugging code.

weblate/static/loader-codemirror.js
```diff
@@ -0,0 +1,76 @@
...
+
+        return { list: userMentionList, from: CodeMirror.Pos(cur.line, start), to: CodeMirror.Pos(cur.line, end) };
+    });
+
+
+    $('textarea.codemirror-markdown').each(function(idx) {
+        var codemirror = CodeMirror.fromTextArea(
+            this,
+            {
+                mode: 'text/javascript',
```
nijel, line 50: Shouldn't it be markdown here?

weblate/static/loader-codemirror.js
```diff
@@ -0,0 +1,76 @@
+(function (CodeMirror) {
+
+
+    function getUserList(usernameSearch) {
+        var userList = [];
+        $.ajax({
+            type: 'GET',
+            url: `/api/users/?username=${usernameSearch}`,
+            dataType: 'json',
+            async: false,
```
nijel, line 10: Does it really have to be synchronous? It will IMHO destroy user experience when server is slow to respond.

weblate/static/loader-codemirror.js
SaptakS: I kept it as a fallback of sorts. But it does make sense for non-english layout

weblate/static/loader-codemirror.js
SaptakS: Not very sure but without this submit button doesn't work once codemirror is applied on the textarea.

weblate/static/loader-codemirror.js
SaptakS: It can be asynchronous, but then the previous response might be rendered while the next request is being performed. Will look into this.

weblate/static/loader-codemirror.js
nijel: I don't think such fallback is needed.

weblate/static/loader-codemirror.js
nijel: CodeMirror is used in more forms, so this won't address it fully. Still, I'd prefer to fix the regular submit behavior as fixing all forms where CodeMirror is used looks like something what will easily break.

weblate/static/loader-codemirror.js
```diff
@@ -0,0 +1,80 @@
...
+        var curWord = start !== end && curLine.slice(start, end);
+
+        if(curWord && curWord.length > 2) {
+            // If there is current word set, We can filter out users from the
+            // main list and display them
+            getUserList(curWord);
+        }
+
+        return { list: userMentionList, from: CodeMirror.Pos(cur.line, start), to: CodeMirror.Pos(cur.line, end) };
+    });
```
nijel, line 45: This still looks wrong. What it should do instead is to register the helper with callback and use the async mode available in the CodeMirror hints:

> It is possible to set the async property on a hinting function to true, in which case it will be called with arguments (cm, callback, ?options), and the completion interface will only be popped up when the hinting function calls the callback, passing it the object holding the completions.

weblate/static/loader-codemirror.js
```diff
@@ -0,0 +1,80 @@
...
+        codemirror.on('keyup', function (cm, event) {
+            if(event.key === '@') {
+                CodeMirror.showHint(cm, CodeMirror.hint.userSuggestions, {
+                    completeSingle: false
+                });
+            }
+        });
+    });
+
+    $('textarea.codemirror-markdown').each(function(idx) {
```
nijel, line 68: This is second time you invoke this each loop, I don't see reason to have these loops twice.

weblate/static/loader-codemirror.js
```diff
@@ -0,0 +1,80 @@
...
+        $(this).closest('form').find('input[type=submit]').on('click', function(e) {
+            e.preventDefault();
+            $(this).closest('form').submit();
+        });
```
nijel, line 72: Have you tried to debug why this is needed? It still looks weird to me.

nijel:

nijel:
Probably merged too fast - I thought tests were passing before, but apparently this broke Selenium tests. I've removed the extra test case in e7bb6e2e3928ef4bf57277bdaaaee80b15bf2975 and fixed other one in a2c0fe7998c6605ad53ccbc9b7ff2887465b60aa.

SaptakS:
Thanks for catching that.

Issue #4041, opened 2020-06-16
Remove TM matches for MT


nijel:
**Is your feature request related to a problem? Please describe.**
The machine translation UI currently contains translation memory matches as well. This is mostly a historical relict from times when there was no dedicated UI for translation memory.

**Describe the solution you'd like**
Remove translation memory matches from the machine translation tab. 

I'm not sure what is the best approach for third-party services which do provide translation memory. Especially in case they do provide translation memory, but fallback to machine translation such as MyMemory.

Additionally, there is "Weblate" service which behaves like translation memory, but scans existing translations only. Maybe it's time to remove it favoring translation memory.

**Describe alternatives you've considered**
Current setup works, but is confusing.

**Additional context**
See https://github.com/WeblateOrg/weblate/pull/4035#discussion_r440349261

julen:
I believe MyMemory offers a way to tell if a result originates from MT via the `reference` field ([example](https://api.mymemory.translated.net/get?q=This%20was%20Machine%20Translated&langpair=en|eu)).

Nonetheless, another option is to keep everything under a single tab (like e.g. Pontoon does), and changing the label of the tab, which is misleading right now. Visually telling what comes from where, emphasizing important parts translators should immediately focus on, and maybe putting TM results over MT matches should help in that case, too.

nijel:
Yes, MyMemory tells that (and I think Weblate includes that info). What I wanted to point out is that it sometimes can not be determined at service level.

Having one tab is option as well, the reason for separating it was that we don't want to allow free text entry for machine translation, but only for translation memory. But that can be probably addressed in single tab UI as well.

julen:
I can have a look at integrating both in a single tab.

I also think the initial request for `MT_BACKENDS` can be skipped either by inlining the list in the DOM or by delegating this to the backend — the endpoint should already know which services are configured and hence what to query. I ignore what was the reason to keep services in separate endpoints — worry about latency perhaps?

nijel:
We definitely want to query each service in a separate request as some might just take long, timeout, or fail.

julen:
Understood. Please feel free to assign the ticket to me.

julen:
@nijel do you have any suggestions on how to proceed with `machinery.view` and `memory.view` permissions? Should they still be kept separate?

nijel:
I think it's okay to have just `machinery.view`. I don't think anybody uses that fine grained permissions anyway.

PR #4109, opened 2020-06-29
[WIP] Editor: unify machinery

julen:
I won't have time anytime soon to finish this off but I thought it was worth sharing anyway. HTH.

Refs. #4041.


nijel:
Thanks, I've based the change on top of this. I've probably messed up some client-side concepts during this, so further review is welcome :-).

PR #4111, opened 2020-06-30
zh_Hans to zh_CN

dingpengyu:
hi team

Currently, the Chinese translation of weblate mainly uses (zh_Hans) the old (IE8N) Code,
Replace (zh_HANS) the old IE8N with (zh_CN) the new (IE8N) code.
Fix translation sorting problem (djangojs.po) in (poedit) edits, and ( Language header) error problem.
thanks


nijel:
What makes you think that zh_Hans is obsolete? It's the correct code for Simplified Chinese, using zh_CN for that is wrong. See for example https://code.djangoproject.com/ticket/18419

dingpengyu:
> What makes you think that zh_Hans is obsolete? It's the correct code for Simplified Chinese, using zh_CN for that is wrong. See for example https://code.djangoproject.com/ticket/18419
Thanks for the clarification, the main reason for submitting this pr is to fix the:
Fix translation sorting problem (djangojs.po) in (poedit) edits, and ( Language header) error .
The poedit editor does not support zh_HANS.
thanks

nijel:
Maybe poedit authors should revisit https://github.com/vslavik/poedit/issues/100...

nijel:
Sorry, we're not going to use obsolete code just because it doesn't work in Poedit.

Issue #4084, opened 2020-06-23
"This string has been translated in the past" is not helpful


okainov:
**Describe the bug**
"This string has been translated in the past"... And what? Can you show me the translation from the past so I don't need to translate it again? Without this it's completely useless warning...

**To Reproduce**
1. Do whatever needed to get this warning "This string has been translated in the past"
2. Check the UI

**Expected behavior**
I expect to see the "translation from the past"

nijel:
You can see it in the history, but it could be indeed shown more prominently.

55abhilash:
Can you assign this to me? I will send a PR shortly.

nijel:
Thanks, I've assigned you.

55abhilash:
@okainov Can you tell me steps to reproduce the warning?

okainov:
@55abhilash  that's complicated :)

What I did:
- Translated some strings, do not commit
- Go to settings, put checkmark "Allow to edit base language"
- Everything apparently got reindexed and all the translations were lost, here you see this message

Maybe also "reset changes" on the repo tab would lead to the same effect...

55abhilash:
I am unable to reproduce it with these steps.
@nijel When is this warning triggered? In my setup, the code in class TranslatedCheck : 
```
   def check_target_unit(self, sources, targets, unit):
        if unit.translated:
            return False
...
...
```
This always returns false in my case, because unit.translated is always True for me! Even though string is being translated for the first time.

@okainov : Did you reset or remove the translation from the VCS from outside weblate?

nijel:
The easiest way to trigger this should be:

* Have a component without push enabled
* Edit string which is currently not translated
* Go to repository maintenance, commit changes and once that is done, reset the repository to upstream

This way Weblate will have history that the string was translated, but the translation is gone from the repository.

nijel:
> Everything apparently got reindexed and all the translations were lost, here you see this message

This should be fixed by 75564435febc8d680143130977a51f09c7404d0a

55abhilash:
Related issue: #4105

55abhilash:
#4112

PR #4112, opened 2020-06-30
"String translated in the past" warning enhancement

55abhilash:
The string "This string has been translated in the past" changed to "This string has been translated in the past. Last translation was %s".

### DIFF
--- a/weblate/checks/consistency.py
+++ b/weblate/checks/consistency.py
@@ -103,8 +103,7 @@ class TranslatedCheck(TargetCheck):
         target = self.check_target_unit(unit.source, unit.target, unit)
         return (
             _(
-                'This string has been translated in the past. \
-                Last Translation was "%s"'
+                'This string has been translated in the past, last translation was "%s"'
             )
             % target
         )
--- a/weblate/checks/consistency.py
+++ b/weblate/checks/consistency.py
@@ -95,12 +95,19 @@ class TranslatedCheck(TargetCheck):
 
     check_id = "translated"
     name = _("Has been translated")
-    description = ""
+    description = _("This string has been translated in the past")
     ignore_untranslated = False
 
     def get_description(self, check=None):
         unit = check.unit
-        self.check_target_unit(unit.source, unit.target, unit)
+        target = self.check_target_unit(unit.source, unit.target, unit)
+        self.description = (
+            _(
+                'This string has been translated in the past. \
+                Last Translation was "%s"'
+            )
+            % target
+        )
         return self.description
 
     def check_target_unit(self, sources, targets, unit):
@@ -114,16 +121,9 @@ class TranslatedCheck(TargetCheck):
 
         changes = unit.change_set.filter(action__in=states).order()
 
-        for action in changes.values_list("action", flat=True):
+        for action, target in changes.values_list("action", "target"):
             if action in Change.ACTIONS_CONTENT:
-                self.description = (
-                    _(
-                        'This string has been translated in the past. \
-                                Last Translation was "%s"'
-                    )
-                    % changes.latest("timestamp").target
-                )
-                return True
+                return target
             if action == Change.ACTION_SOURCE_CHANGE:
                 break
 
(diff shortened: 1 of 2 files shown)


55abhilash:
For issue: #4084

weblate/checks/consistency.py
```diff
@@ -95,9 +95,14 @@ class TranslatedCheck(TargetCheck):
 
     check_id = "translated"
     name = _("Has been translated")
-    description = _("This string has been translated in the past")
+    description = ""
```
nijel, line 98: Please keep the description here as well, it is used for example when listing all the checks (/checks/ URL).

weblate/checks/consistency.py
```diff
@@ -95,9 +95,14 @@ class TranslatedCheck(TargetCheck):
...
     ignore_untranslated = False
 
+    def get_description(self, check=None):
+        unit = check.unit
+        self.check_target_unit(unit.source, unit.target, unit)
+        return self.description
```
nijel, line 104: Beware that the in database check state might be outdated, so the check_target_unit might return False when get_description is called. In this case you return random description of previously triggered check.

weblate/checks/consistency.py
```diff
@@ -111,6 +116,13 @@ def check_target_unit(self, sources, targets, unit):
 
         for action in changes.values_list("action", flat=True):
             if action in Change.ACTIONS_CONTENT:
+                self.description = (
+                    _(
+                        'This string has been translated in the past. \
+                                Last Translation was "%s"'
+                    )
+                    % changes.latest("timestamp").target
```
nijel, line 124: This hits again the database making the check twice slower than necessary. Better approach would be to include target in values_list above and use the value here directly without additional SQL query.

Also changing description class attribute doesn't sound right. Better approach might be to return the target value here and generate the verbose description in the get_description method.

weblate/locale/django.pot
```diff
@@ -2098,7 +2098,7 @@ msgid "Has been translated"
 msgstr ""
 
 #: weblate/checks/consistency.py:98
-msgid "This string has been translated in the past"
+msgid "This string has been translated in the past. Last Translation was \"%s\""
```
nijel, line 2101: This is not needed, the pot/po files are updated time to time by a dedicated script.

weblate/checks/consistency.py
55abhilash: What is the meaning of "check state might be outdated"

nijel: The get_description is called on the translate page when the check object is stored in the database. Meanwhile the string can be updated and check update in progress in the background. So the get_description should work without problems when the check is no longer failing.

weblate/checks/consistency.py
```diff
@@ -98,6 +98,18 @@ class TranslatedCheck(TargetCheck):
     description = _("This string has been translated in the past")
     ignore_untranslated = False
 
+    def get_description(self, check=None):
+        unit = check.unit
+        target = self.check_target_unit(unit.source, unit.target, unit)
+        self.description = (
```
nijel, line 104: Why updating description attribute? I see no reason for that (and it will break things).

55abhilash: Sorry, missed it before. Changed now.

weblate/checks/consistency.py
```diff
@@ -101,14 +101,14 @@ class TranslatedCheck(TargetCheck):
...
-        self.description = (
+        description = (
             _(
                 'This string has been translated in the past. \
                 Last Translation was "%s"'
             )
             % target
         )
-        return self.description
+        return description
```
nijel, line 111: ...and now the variable is useless :-). Please use return directly.

weblate/checks/consistency.py
```diff
@@ -98,6 +98,17 @@ class TranslatedCheck(TargetCheck):
     description = _("This string has been translated in the past")
     ignore_untranslated = False
 
+    def get_description(self, check=None):
+        unit = check.unit
+        target = self.check_target_unit(unit.source, unit.target, unit)
+        return (
+            _(
+                'This string has been translated in the past. \
+                Last Translation was "%s"'
```
nijel, line 107:
```suggestion
                'This string has been translated in the past, last translation was "%s"'
```

nijel:

nijel:
I've made some futher improvements in 5d7aaaf109e706fea3deecce370e3fd5bf3fb673

PR #4149, opened 2020-07-09
Spelling: Conflict handling, the option

comradekingu:
Unsure what "Remove add as translation choice." means.


nijel:

Issue #2363, opened 2018-10-30
Remove usage of Django admin interface


nijel:
**Is your feature request related to a problem? Please describe.**
Having both Django admin inteface and Weblate itself is confusing and using generic Django admin interface leads to several problems.

**Describe the solution you'd like**
All should be handled directly in Weblate.

**Describe alternatives you've considered**
More customization of Django admin could achieve better behavior, but still some things are hard to address (see https://code.djangoproject.com/ticket/10919).

**Additional context**
Things currently done *only* in the Django admin, but should be possible from Weblate:

* [x] Performance report
* [x] Repositories status
* [x] SSH keys management (replaced by https://github.com/WeblateOrg/weblate/issues/2459)
* [x] Global user and group management, automatic group assignments (related to https://github.com/WeblateOrg/weblate/issues/2246 and https://github.com/WeblateOrg/weblate/issues/2493 and https://github.com/WeblateOrg/weblate/issues/2126 and https://github.com/WeblateOrg/weblate/issues/2127)
* [x] Whiteboard
* [ ] Component lists (see https://github.com/WeblateOrg/weblate/issues/1370 and https://github.com/WeblateOrg/weblate/issues/11549 and https://github.com/WeblateOrg/weblate/issues/7181)
* [x] Deletion (see https://github.com/WeblateOrg/weblate/issues/2221)
* [x] Adding languages (see https://github.com/WeblateOrg/weblate/issues/2491)

New features in the interface:

* [x] Sending test mail (see https://github.com/WeblateOrg/weblate/issues/2517)

Things which might stay relying on Django Admin forever:

* Audit logs
* Social auth details
* Billing management

reloxx13:
i cannot see deleted/disabled users in weblate ui are not shown which i see in the admin interface.    
i would also prefer a cleaner user list like the table in djangi admin ui and not so blown up.

nijel:
As you can see, this issue is still in progress...

nijel:
The only remaining bit here are the component lists which need more than just interface changes, see https://github.com/WeblateOrg/weblate/issues/12160. Let's close this issue as all other things have been addressed and this is no longer useful.

PR #4170, opened 2020-07-12
Spelling: Site wide, matched

comradekingu:


docs/admin/access.rst
```diff
@@ -168,15 +168,15 @@ to a containing project as well.
 
 .. note::
 
-   Only first rule which is set is being used. So if you set all
+   Only the first rule will be used. So if you set all of
    :guilabel:`Component list`, :guilabel:`Components` and :guilabel:`Project`,
-   only :guilabel:`Component list` is being applied.
+   only :guilabel:`Component list` is the only one applied.
 
-Additional step is performed if checking permission for the translation:
```
comradekingu, line 175: what does "checking" mean here?

docs/admin/access.rst
```diff
@@ -168,15 +168,15 @@ to a containing project as well.
...
+An additional step is performed if checking permission for the translation:
 
 
-4. :guilabel:`Languages` are matches against scope of translations, if not set, this matches no
```
comradekingu, line 178: is matched against?
Can "if not set" be avoided?

docs/admin/access.rst
```diff
@@ -107,7 +107,7 @@ Automatic group assignments
 You can set up Weblate to automatically add users to groups based on their
 e-mail addresses. This automatic assignment happens only at the time of account creation.
 
-This can be set up in the Django admin interface for each group (in the
+This can be set up in the Weblate admin interface for each group (in the
```
nijel, line 110: Same as above, IMHO should be reverted.

docs/admin/access.rst
nijel: Checking whether user has access to a translation.

docs/admin/access.rst
nijel: Group languages are matched against the translation languages. If no group languages are set, it matches none languages (this can be still useful to grant access to a project or component, but not to actual translations.

docs/admin/access.rst
```diff
@@ -89,7 +89,7 @@ Custom
 .. image:: /images/project-access.png
 
 To allow access to this project, you have to add the privilege either
-directly to the given user, or group of users in the Django admin interface,
+directly to the given user, or group of users in the Weblate admin interface,
```
nijel, line 92: Historically, we use Django admin interface. The long-term plan is to replace it for normal management operations by Weblate one, so this is wrong - we still use Django admin interface here.

See https://github.com/WeblateOrg/weblate/issues/2363 for more details.

docs/admin/access.rst
```diff
@@ -203,7 +203,7 @@ requires explicit permission on the component (or containing component list).
 Managing users and groups
 -------------------------
 
-All users and groups can be managed using the Django admin interface,
+All users and groups can be managed using the Weblate admin interface,
```
nijel, line 206: Same as above, IMHO should be reverted.

docs/admin/access.rst
```diff
@@ -285,10 +285,10 @@ Custom access control
 ---------------------
 
 By choosing :guilabel:`Custom` as :guilabel:`Access control`, Weblate will stop
-managing access for a given project, and you can set up custom rules in the Django
+managing access for a given project, and you can set up custom rules in the Weblate
```
nijel, line 288: Same as above, IMHO should be reverted.

docs/admin/access.rst
```diff
@@ -402,30 +402,30 @@ VCS
 
     Update the internal repository [`Administration`, `Manage repository`]
 
-Global privileges
-    Use management interface (global)
+Site wide privileges
+    Use management interface
 
-    Add language definitions (global)
+    Add language definitions
```
nijel, line 408: These should match in app permission names, but it seems to be already out of sync now since https://github.com/WeblateOrg/weblate/pull/2895:

https://github.com/WeblateOrg/weblate/blob/69aa093dfd7c74e94fe2b04e181c31bd54e0bdbb/weblate/auth/data.py#L123-L142

docs/admin/access.rst
comradekingu, line 92:
```suggestion
directly to the given user, or group of users in the Django admin interface,
```

docs/admin/access.rst
comradekingu, line 206:
```suggestion
All users and groups can be managed using the Django admin interface,
```

docs/admin/access.rst
comradekingu: Changed those. I like entire, but in English /project wide/site wide is neat.

docs/admin/access.rst
comradekingu, line 288:
```suggestion
All users and groups can be managed using the Django admin interface,
```
```suggestion
managing access for a given project, and you can set up custom rules in the Django
```

docs/admin/access.rst
comradekingu, line 110:
```suggestion
This can be set up in the Django admin interface for each group (in the
```

docs/admin/access.rst
```diff
@@ -168,15 +168,15 @@ to a containing project as well.
...
-   only :guilabel:`Component list` is being applied.
+   only :guilabel:`Component list` is the only one applied.
 
-Additional step is performed if checking permission for the translation:
+An additional step is performed if checking permission for the translation:
 
 
-4. :guilabel:`Languages` are matches against scope of translations, if not set, this matches no
-   language.
+4. :guilabel:`Languages` are matches against scope of translations, if not set, this
```
comradekingu, line 178:
```suggestion
4. :guilabel:`Languages` are matched against the scope of translations if set, if not set, this
```

docs/admin/access.rst
```diff
@@ -285,10 +285,10 @@ Custom access control
 ---------------------
 
 By choosing :guilabel:`Custom` as :guilabel:`Access control`, Weblate will stop
-managing access for a given project, and you can set up custom rules in the Django
+All users and groups can be managed using the Django admin interface,
```
nijel, line 288: "Weblate will stop managing access for a given project" seems removed without any replacement (and the first part of the sentence is now unfinished).

docs/admin/install.rst
```diff
@@ -249,7 +249,7 @@ Once the key is trusted, the warning will not occur:
    gpg:                using RSA key 87E673AF83F6C3A0C344C8C3F4AA229D4D58C245
    gpg: Good signature from "Michal Čihař <michal@cihar.com>" [ultimate]
    gpg:                 aka "Michal Čihař <nijel@debian.org>" [ultimate]
-   gpg:                 aka "[jpeg image of size 8848]" [ultimate]
+   gpg:                 aka "[JPEG image of size 8848]" [ultimate]
```
nijel, line 252: This is pasted from the GnuPG output, I don't think it makes sense to modify that.

nijel, line 252:
```suggestion
   gpg:                 aka "[jpeg image of size 8848]" [ultimate]
```

docs/wlc.rst
```diff
@@ -66,7 +66,7 @@ Weblate Client is a Python library and command-line utility to manage Weblate re
 using :ref:`api`. The command-line utility can be invoked as :command:`wlc` and is
 built on :mod:`wlc`.
 
-Instance wide options
+Site wide options
 ---------------------
```
nijel, line 70:
```suggestion
-----------------
```

docs/admin/access.rst
nijel, line 288:
```suggestion
managing access for a given project, and all users and groups can be managed using the Django```

nijel:
Merged with some adjustments, thanks!

PR #4172, opened 2020-07-12
Spelling: E-mail

comradekingu:


nijel:

fitojb:
I don’t see how this change is an improvement. Modern spelling drops the hyphen.

comradekingu:
@fitojb It is consistent with the other Weblate strings. "E-mail" was picked because it is correct in the strictest sense, and it produces the fewest errors in translation. :)

fitojb:
@comradekingu Both unsubstantiated claims. Come on, I teach English. By all means keep fixing typos, but I don’t have time to retranslate strings that were correct in the first place.

comradekingu:
@fitojb So you shouldn't really have time to be on here is what you are saying? ;)
Changed strings will just be made fuzzy, no need to touch them other than for peace of mind.
Join the fight against modernity and start marking down your students for writing "email".
You made an appeal to authority, and an anecdotal one at that. https://grammarist.com/style/e-mail-email/ backs up my claim. As does any of the style guides before the spellcheckers were fired. I will change nothing to accommodate new conventions that happened in my lifetime.

https://pt.wikipedia.org/wiki/E-mail https://pt.wikipedia.org/wiki/E-mail https://ia.wikipedia.org/wiki/Currero_electronic https://eo.wikipedia.org/wiki/Retpoŝto https://sv.wikipedia.org/wiki/E-post https://nn.wikipedia.org/wiki/E-post https://no.wikipedia.org/wiki/E-post https://de.wikipedia.org/wiki/E-Mail
are some of the examples of e-*, where people generally aren't adding the hyphen by mistake, but may forget to add it.

orangesunny:
Well, I appreciate consistency, but I don’t want you to be bitter or fight about that. There is no sense in arguing what is right when both variants are right. Can we turn the flame down, please?

The **statement about this from the Weblate team** is: Both `e-mail` and `email` are possible and correct. `E-mail` is preferred due to consistency, but we are not going to yell at anyone if she/he uses the `email` version. It is even sometimes handier in small places. If @comradekingu wants to spare time making corrections like this, we are OK with that.

PR #4173, opened 2020-07-13
Spelling: Various fixes and cleanup

comradekingu:
As per https://amagama.translatehouse.org/

### DIFF
--- a/docs/admin/config.rst
+++ b/docs/admin/config.rst
@@ -61,7 +61,7 @@ Maximum number of failed authentication attempts before rate limiting is applied
 
 This is currently applied in the following locations:
 
-* Logins. Deletes the account password, preventing the user from login in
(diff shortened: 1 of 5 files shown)


docs/admin/config.rst
```diff
@@ -61,7 +61,7 @@ Maximum number of failed authentication attempts before rate limiting is applied
 
 This is currently applied in the following locations:
 
-* Logins. Deletes the account password, preventing the user from login in
+* Logins. Deletes the account password, preventing the user from logging in
```
nijel, line 64:
```suggestion
* Logins. Deletes the account password, preventing the user from signing in
```

We use "sign in" in the UI, so it probably should be consistent in the docs as well.

nijel:

PR #4178, opened 2020-07-14
Spelling: String statuses

comradekingu:


docs/api.rst
```diff
@@ -1496,22 +1496,22 @@ Translations
...
+    :>json float failing_checks_percent: percentage of strings failing checks
+    :>json int failing_checks_words: number of words with failing checks
     :>json string filename: translation filename
     :>json int fuzzy: number of strings marked for review
     :>json float fuzzy_percent: percentage of strings marked for review
     :>json int fuzzy_words: number of words marked for review
     :>json int have_comment: number of strings with comment
     :>json int have_suggestion: number of strings with suggestion
-    :>json boolean is_template: whether translation is monolingual base
+    :>json boolean is_template: whether the translation has a monolingual base
```
nijel, line 1508: This changes meaning - it really should be is.

docs/user/translating.rst
```diff
@@ -164,9 +164,9 @@ The following keyboard shortcuts can be utilized during translation:
 :kbd:`Ctrl+<NUMBER>` or :kbd:`Cmd+<NUMBER>`
     Copies placeable of given number from source string.
 :kbd:`Ctrl+M <NUMBER>` or :kbd:`Cmd+M <NUMBER>`
-    Copy machine translation of given number to current translation.
+    Copy the machine translation of given number to current translation.
 :kbd:`Ctrl+I <NUMBER>` or :kbd:`Cmd+I <NUMBER>`
-    Ignore failing check of given number.
+    Ignore failing checks beyond a given number.
```
nijel, line 169: I don't think beyond is correct here- `Ctrl+I 1` dismisses first failing check, `Ctrl+I 2` dismisses second failing check and so on.

weblate/templates/snippets/query-builder.html
```diff
@@ -34,11 +34,11 @@
...
-                  <li><a href="#" data-field="has:suggestion">{% trans "String has suggestion" %}</a></li>
-                  <li><a href="#" data-field="has:comment">{% trans "String has comment" %}</a></li>
-                  <li><a href="#" data-field="has:failing_check">{% trans "String has failing check" %}</a></li>
-                  <li><a href="#" data-field="has:plural">{% trans "String has plural" %}</a></li>
-                  <li><a href="#" data-field="has:context">{% trans "String has context" %}</a></li>
+                  <li><a href="#" data-field="has:suggestion">{% trans "String suggestion exists" %}</a></li>
+                  <li><a href="#" data-field="has:comment">{% trans "Commented string" %}</a></li>
+                  <li><a href="#" data-field="has:failing_check">{% trans "Failing checks for string" %}</a></li>
+                  <li><a href="#" data-field="has:plural">{% trans "Pluralized string" %}</a></li>
+                  <li><a href="#" data-field="has:context">{% trans "Contextualized string" %}</a></li>
```
nijel, line 41: This reminds me that this should be generated from `weblate/trans/filter.py` so it's more important to change the strings there.

docs/changes.rst
```diff
@@ -785,7 +785,7 @@ Released on August 11th 2017.
 * The max-length checks is now enforcing length of text in form.
 * Make the commit_pending age configurable per component.
 * Various user interface cleanups.
-* Fixed component/project/sitewide search for translations.
+* Fixed component/project/site wide search for translations.
```
fitojb, line 788: “site-wide” is a compound adjective and should be hyphenated.

weblate/templates/snippets/query-builder.html
comradekingu: The ones in filter.py were just plain better, the ones in here can be overwritten. "Pluralized string" isn't in there though.

docs/user/translating.rst
comradekingu, line 169:
```suggestion
    Ignore one item in the list of failing checks.
```

docs/changes.rst
comradekingu: Do you know if it should be "site-wide" when it isn't transitive? Some places it is given as a description between "local" "string only" and ~"for the entire platform/instance".
From what I can tell it is "command line", but then "command-line client"

fitojb: Sometimes, the adjective may be omitted from the phrase but still be implied (“site-wide [setting]”), so I would still hyphenate it.

nijel: This is definitely used in more places, so let's deal with "site-wide" spelling separately, I will merge this PR to get other fixes in.

weblate/templates/snippets/query-builder.html
nijel: I've switched it in c3283e2e51e7ad70473ab8d747fee0d7641ff200

PR #4194, opened 2020-07-18
add CLOUDFLARE_URL setting 

deboyblog:
add CLOUDFLARE_URL setting for CSP script domain list.

### DIFF
--- a/docs/locales/pt_BR/LC_MESSAGES/docs.po
+++ b/docs/locales/pt_BR/LC_MESSAGES/docs.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: Weblate\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2020-07-16 16:25+0200\n"
-"PO-Revision-Date: 2020-07-17 08:38+0000\n"
+"PO-Revision-Date: 2020-07-17 14:32+0000\n"
 "Last-Translator: Rafael Fontenelle <rafaelff@gnome.org>\n"
 "Language-Team: Portuguese (Brazil) <https://hosted.weblate.org/projects/"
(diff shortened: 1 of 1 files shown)


weblate/middleware.py
```diff
@@ -242,6 +242,11 @@ def __call__(self, request):
         if "://" in settings.SOCIAL_AUTH_AUTH0_IMAGE:
             domain = urlparse(settings.SOCIAL_AUTH_AUTH0_IMAGE).hostname
             image.add(domain)
+        
+        # Cloudflare use for security
+        if "://" in settings.CLOUDFLARE_URL:
```
nijel, line 247: This will crash if the setting doesn't exist.

deboyblog: This config for some domain which is use cloudflare. it prevent cloudflare's script was block by the weblate CSP limit (while DEBUG=0).
Does it make a define in `weblate/settings_docker.py` in `Third party services integration` can solve the crash problem?

nijel: The setting should either not be required (check its presence) or have default value (preferred). We use https://django-appconf.readthedocs.io/en/latest/ to provide default values for settings.

Also, I can see other issues:

* It doesn't seem to be Cloudflare specific, so I see no good reason for using CLOUDFLARE in the name.
* The URL is not really used as URL, all you use is a domain name. Why not have just this one in the setting?

How about having `CSP_SCRIPT_SRC` configuration, which would allow list of strings and those would be added to the CSP?

deboyblog: I'm agree with that, `CSP_SCRIPT_SRC` will be better.

nijel:
I've just implemented generic solution in 571f179eb1a0d18f77f966e3b218cae1ec618635

PR #4224, opened 2020-07-26
Migration from hubs and labs to GitHub and GitLab API for forks and PRs

SaptakS:

@nijel This PR is in no way ready for a full review or even close to merging. I just wanted to create a draft PR
so that you can check and tell me if the approach makes sense. I have just implemented (or migrated) GitHub
repos. It still might not tick off many of the boxes, but curious to know if this makes sense.


weblate/vcs/git.py
```diff
@@ -279,6 +280,33 @@ def configure_remote(self, pull_url, push_url, branch):
...
+        r = requests.post(
+            fork_url,
+            headers={
+                "Accept": "application/vnd.github.v3+json",
+                "Authorization": "token {}".format(settings.GITHUB_TOKEN),
+            },
+            data={},
+        )
+        response = r.json()
+        self.configure_fork_remote(response["clone_url"], self.get_username())
```
nijel, line 308: This all code is GitHub specific, it should go to the GithubRepository class.

weblate/vcs/git.py
```diff
@@ -279,6 +280,33 @@ def configure_remote(self, pull_url, push_url, branch):
         )
         self.branch = branch
 
+    def api_url(self):
+        return self.component.repo.replace(".git", "").replace(
+            "github.com", "api.github.com/repos"
+        )
```
nijel, line 286: The repo URL can have several forms and can include HTTP authentication, the code should deal with that, here is what all we support on hooks side:

https://github.com/WeblateOrg/weblate/blob/6c44517f5330ddcdc19f6934c8f30e02c90f844a/weblate/trans/views/hooks.py#L58-L65

weblate/vcs/git.py
```diff
@@ -574,7 +602,7 @@ def fork(self):
         """Create fork of original repository if one doesn't exist yet."""
         remotes = self.execute(["remote"]).splitlines()
         if self.get_username() not in remotes:
-            self.execute(["fork"])
+            super().fork()
```
nijel, line 605: This looks wrong, the super class (GitRepository) should have no knowledge on forking, it's the child classes (GithubRepository, GitLabRepository, ..) which are supposed to implement the actual logic.

weblate/vcs/git.py
```diff
@@ -654,18 +682,22 @@ def create_pull_request(self, origin_branch, fork_remote, fork_branch):
...
+            },
+            json={
+                "head": head,
+                "base": origin_branch,
+                "title": self.get_merge_message(),
+            },
         )
+        response = r.json()
+        if not response["url"]:
+            report_error(cause=response["message"])
```
nijel, line 700: What does happen here when the pull request already exists?

weblate/vcs/git.py
SaptakS: Makes sense.

weblate/vcs/git.py
SaptakS: I will move it to the child classes. Makes more sense.

weblate/vcs/git.py
SaptakS: I haven't tried that. But any time there isn't a success, they usually throw a message (i.e. wrong URL, wrong origin, if repo doesn't exist, etc.) so I am guessing it will give some message here. But I will try and test that out.

nijel: We want to gracefully handle situation when there is already a pull request opened.

weblate/vcs/base.py
```diff
@@ -50,7 +50,11 @@ def __init__(self, retcode, message):
 
     def get_message(self):
         if self.retcode != 0:
-            return "{0} ({1})".format(self.args[0], self.retcode)
+            fork_url = "https://{0}:{1}@github.com".format(
+                settings.GITHUB_USERNAME, settings.GITHUB_TOKEN
+            )
+            error_message = cleanup_repo_url(fork_url, text=self.args[0])
+            return "{0} ({1})".format(error_message, self.retcode)
```
nijel, line 57: Sorry for confusion - you don't need to introduce calls to `cleanup_repo_url`, just extend it to cleanup GITHUB credentials.

weblate/vcs/git.py
```diff
@@ -563,18 +578,20 @@ def push_to_fork(self, local_branch, fork_branch):
         """Push given local branch to branch in forked repository."""
         self.execute(
             [
+                "git",
```
nijel, line 581: I think it's better to completely remove hub usage from the GitHub class instead of rewriting it here. Right now it might be probably broken for GitLab which still relies on `lab`, but it would not be executed (not sure if that is needed though).

weblate/vcs/git.py
```diff
@@ -279,6 +280,20 @@ def configure_remote(self, pull_url, push_url, branch):
         )
         self.branch = branch
 
+    def api_url(self):
+        slug = self.component.repo.split("/")[-1].replace(".git", "")
+        owner = self.component.repo.split("/")[-2]
+        return "https://api.github.com/repos/{0}/{1}".format(owner, slug)
```
nijel, line 286: Please move this to GitHub class.

weblate/vcs/git.py
```diff
@@ -279,6 +280,20 @@ def configure_remote(self, pull_url, push_url, branch):
...
+
+    def configure_fork_remote(self, pull_url, remote_name):
+        """Configure fork remote repository."""
+        self.config_update(
+            # Pull url
+            ('remote "{}"'.format(remote_name), "url", pull_url),
+            # Push url
+            ('remote "{}"'.format(remote_name), "pushurl", pull_url),
+        )
```
nijel, line 295: This should probably go to the GitMergeRequestBase class.

weblate/vcs/git.py
```diff
@@ -654,18 +690,29 @@ def create_pull_request(self, origin_branch, fork_remote, fork_branch):
...
+        response = r.json()
+        if "url" not in response:
+            for error in response["errors"]:
+                if "message" in error:
+                    report_error(cause=error["message"])
+                else:
+                    report_error(
+                        cause="{0}: {1}".format(error["resource"], response["message"]),
+                        level="error",
+                    )
```
nijel, line 715: I'd probably do it just:

```suggestion
            report_error(cause="Pull request failed", extra_data={"errors": response["errors"]})
```

weblate/vcs/git.py
SaptakS: Aah. Makes sense. This is very specific to GitHub.

weblate/vcs/git.py
SaptakS: :+1:

weblate/vcs/git.py
```diff
@@ -621,14 +641,7 @@ def get_merge_message(self):
...
 
     name = "GitHub"
-    _cmd = "hub"
     _version = None
-    req_version = "2.7"
-
-    @classmethod
-    def _get_version(cls):
-        """Return VCS program version."""
-        return cls._popen(["--version"], merge_err=False).split()[-1]
```
nijel, line 631: This is what I meant by https://github.com/WeblateOrg/weblate/pull/4224/files#r462265471, sorry for not doing a pull request, I've messed it up somehow.

weblate/vcs/git.py
```diff
@@ -279,6 +280,20 @@ def configure_remote(self, pull_url, push_url, branch):
...
+    def api_url(self):
+        slug = self.component.repo.split("/")[-1].replace(".git", "")
+        owner = self.component.repo.split("/")[-2]
+        return "https://api.github.com/repos/{0}/{1}".format(owner, slug)
+
+    def configure_fork_remote(self, pull_url, remote_name):
+        """Configure fork remote repository."""
+        self.config_update(
+            # Pull url
+            ('remote "{}"'.format(remote_name), "url", pull_url),
```
nijel, line 292: The pull URL should be still from the original upstream, you only should chnnge push URL here.

weblate/vcs/git.py
```diff
@@ -645,6 +658,22 @@ def _getenv(cls):
...
+                "Accept": "application/vnd.github.v3+json",
+                "Authorization": "token {}".format(settings.GITHUB_TOKEN),
+            },
+            data={},
+        )
+        response = r.json()
+        pull_url = "https://{0}:{1}@github.com/{2}.git".format(
+            self.get_username(), settings.GITHUB_TOKEN, response["full_name"]
+        )
+        self.configure_fork_remote(pull_url, self.get_username())
```
nijel, line 675: Let's switch this to SSH instead to avoid need to store credentials.

weblate/vcs/git.py
SaptakS: Okay.

weblate/vcs/git.py
SaptakS: No problem

weblate/vcs/git.py
```diff
@@ -669,10 +668,7 @@ def create_fork(self):
             data={},
         )
```
nijel, line 669: Does this return name of existing fork in case one already exists? If it does it would be great to mention this in a comment, otherwise this case needs to be handled.

weblate/vcs/git.py
```diff
@@ -816,35 +804,50 @@ def _get_version(cls):
...
 
-        Used to merge branch in forked repository into branch of remote
-        repository.
+    def create_fork(self):
+        fork_url = "{}/fork".format(self.api_url())
+        r = requests.post(
+            fork_url,
+            headers={"Authorization": "Bearer {}".format(settings.GITLAB_TOKEN)},
+            data={},
+        )
```
nijel, line 826: Same remark as with GitHub - what happens in case fork exists? Either the code should contain comments describing the behavior or handle this case.

weblate/vcs/git.py
SaptakS: That's a good point. I will test this.

weblate/vcs/base.py
nijel, line 57:
```suggestion
            return "{0} ({1})".format(self.args[0], self.retcode)
```

weblate/vcs/git.py
```diff
@@ -563,18 +565,27 @@ def push_to_fork(self, local_branch, fork_branch):
         """Push given local branch to branch in forked repository."""
         self.execute(
             [
+                "git",
```
nijel, line 568:
```suggestion
```

weblate/vcs/git.py
```diff
@@ -563,18 +565,27 @@ def push_to_fork(self, local_branch, fork_branch):
...
                 "push",
                 "--force",
                 self.get_username(),
                 "{0}:{1}".format(local_branch, fork_branch),
-            ]
+            ],
+            fullcmd=True,
```
nijel, line 574:
```suggestion
            ],
```

weblate/vcs/git.py
```diff
@@ -757,9 +788,6 @@ class GitLabRepository(GitMergeRequestBase):
 
     _version = None
 
-    # docs: https://zaquestion.github.io/lab/
-    _cmd = "lab"
-
```
nijel, line 762: Please remove the _get_version method as well, it is no longer needed.

weblate/settings_example.py
```diff
@@ -200,6 +200,7 @@
 # GitHub username for sending pull requests.
 # Please see the documentation for more details.
 GITHUB_USERNAME = None
+GITHUB_TOKEN = None
```
nijel, line 203: Please add this as a default value to weblate/vcs/models.py (I've moved related settings there in 7173787ec7).

Probably same is needed for GitLab...

weblate/vcs/git.py
SaptakS: In Github API it does return the name (and all details) of the existing fork. I will add a comment to reflect the same.

weblate/vcs/git.py
SaptakS: Seems gitlab doesn't handle the fork case. It returns a message saying `"path": ["has already been taken"]`. So in GitLab side, will have to handle this case.

weblate/vcs/git.py
```diff
@@ -656,18 +683,24 @@ def create_pull_request(self, origin_branch, fork_remote, fork_branch):
...
+                "head": head,
+                "base": origin_branch,
+                "title": self.get_merge_message(),
+            },
         )
+        response = r.json()
+        if "url" not in response:
+            report_error(
+                cause="Pull request failed", extra_data={"errors": response["errors"]}
+            )
```
nijel, line 703: This probably should gracefully handle pull request already exists case, see https://github.com/WeblateOrg/weblate/issues/4266

weblate/vcs/git.py
```diff
@@ -759,54 +792,68 @@ class GitLabRepository(GitMergeRequestBase):
...
+
+        # Check if Fork already exists owned by current user. If the
+        # fork already exists, set that fork as remote.
+        # Else, create a new fork
+        r = requests.get(
+            get_fork_url,
+            headers={"Authorization": "Bearer {}".format(settings.GITLAB_TOKEN)},
+        )
+        for fork in r.json():
+            if "owner" in fork and fork["owner"]["username"] == self.get_username():
```
nijel, line 825: Shouldn't the owned=True condition return only the owned ones making this condition not needed? 

In case it doesn't filter that, we have scalability issue as projects can easily have thousands of forks...

SaptakS: It just returns the forks owned by users and any forks in a group owned by the user. So the responses will be restricted to only user's profile and all the groups, but we need that condition to identify the fork in the user's profile and not in the groups. I will add comments explaining the same.

weblate/vcs/git.py
SaptakS: So for Pull Request already exist case, I should not show any error, and report_error on other errors (and also probably raise error )?

nijel: * Pull requests already exists should be silently ignored
* Anything else should probably raise an exception here. The `report_error` only logs it, but the failure should be made obvious here and raise an exception.

SaptakS:
Updated the following:

- [x] Disable different features (apart from merge requests because needed to create merge request as well) in the Gitlab forked repo
- [x] Checks if fork already exists (added necessary comments explaining why we need to check owner logic)
- [x] If while creating fork, a repo with that name/path exist, try creating the fork with name suffixed with 4 digit (pseudo)random number
- [x] Handle PRs more gracefully in both Github and Gitlab (though for now the only form of comparison is the error message). If PR already exists, then pass, else raise a RepositoryException error
- [x] Adds API tests for both Github and Gitlab using mock and patch (Had to overwrite few of the function to make things work. Hopefully it makes sense)
- [x] Added docs.

comradekingu:
"Initial migration from Hub to API for forks and PRs"?

SaptakS:
> "Initial migration from Hub to API for forks and PRs"?

I rewrote the tittle of the PR. This PR is intended to remove the use of hubs and labs tool and instead use the APIs provided by github and gitlab to create fork and open pull/merge requests.

comradekingu:
@SaptakS It is still "GitHub and GitLab" and "forks and PRs" :)

SaptakS:
Fixed merge conflicts

weblate/vcs/tests/test_vcs.py
```diff
@@ -435,19 +434,406 @@ def add_remote_commit(self, conflict=False, rename=False):
...
 
+    def _mock_response(self, json_data, status):
+        """Mock response helper function.
+
+        This function will act as a helper function for response
+        """
+        mock_resp = mock.Mock()
+        mock_resp.status_code = status
+        mock_resp.json = mock.Mock(return_value=json_data)
+        return mock_resp
```
nijel, line 445: It would be less code to use [responses](https://pypi.org/project/responses/) for mocking the requests. We already use it in other places for testing. But it's okay to keep it this way once that was implemented...

weblate/vcs/tests/test_vcs.py
```diff
@@ -435,19 +434,406 @@ def add_remote_commit(self, conflict=False, rename=False):
...
+        mock_get_patcher = patch("requests.get")
+        mock_get = mock_get_patcher.start()
+        mock_get.side_effect = self.get_empty_side_effect
 
-class VCSGerritTest(VCSGithubTest):
+        super().test_push(branch)
+        mock_get_patcher.stop()
+        mock_put_patcher.stop()
+        mock_post_patcher.stop()
+        mock_push_to_fork.stop()
```
nijel, line 638: Can't this use contexts instead of manually stopping all patchers? The manual approach seems quite fragile and easy to miss something.

weblate/vcs/tests/test_vcs.py
```diff
@@ -435,19 +434,406 @@ def add_remote_commit(self, conflict=False, rename=False):
...
     _sets_push = True
 
+    def add_remote_commit(self, conflict=False, rename=False):
+        # Use Git to create changed upstream repo
+        backup = self._class
+        self._class = GitRepository
+        try:
+            super().add_remote_commit(conflict, rename)
+        finally:
+            self._class = backup
```
nijel, line 835: I'd prefer to reuse this code instead of having it implemented in multiple test classes.

SaptakS: I will create an intermediate class then from which I will inherit VCSGithubTest, VCSGitlabTest and VCSGerritTest.

weblate/vcs/tests/test_vcs.py
SaptakS: Okay. Will check that.

SaptakS:
I have rewrote the test using responses. That also removes the bad code of multiple manual patch stops. I have created separate mock responses focussing on different purposes in GitLab test given there were multiple API calls and bigger responses.

nijel (approved):
Looks good now. Will merge it once 4.3 development opens up.

weblate/vcs/git.py
```diff
@@ -660,18 +678,47 @@ def create_pull_request(self, origin_branch, fork_remote, fork_branch):
...
+        r = requests.post(
+            pr_url,
+            headers={
+                "Accept": "application/vnd.github.v3+json",
+                "Authorization": "token {}".format(settings.GITHUB_TOKEN),
+            },
+            json={
+                "head": head,
+                "base": origin_branch,
+                "title": self.get_merge_message(),
```
nijel, line 691: This code has just created it first pull requests on Hosted Weblate and this seems to be wrong - it injects whole message into the title, while it should be split into title and body/description, see https://github.com/WeblateOrg/weblate/pull/4427. I've fixed this in 389f1028691abdc0a16f3f9d934218a325d7baaa

PR #4242, opened 2020-08-02
Fix typo: translsate -> translate

yurchor:
Thanks in advance for fixing this typo.


nijel:

PR #4246, opened 2020-08-03
Docs: Remove duplicate quality word

rffontenelle:
'checks' link already is rendered as 'Quality checks', so having 'quality' in the paragraph results in having duplicate quality in the resulting string.

How is the documentation currently:
> All built-in quality [Quality checks](https://docs.weblate.org/pt_BR/latest/user/checks.html#checks) are turned on by default (...)


nijel:

PR #4248, opened 2020-08-03
Small doc fixes

rffontenelle:
* `admin/admin.rst`: sections below -> sections above, see commit message:
> The paragraph uses 'sections below', using as an example
    'Weblate translations'.  However, this and other sections
    are found only in one of this page's screenshots which can
    be found ABOVE, not BELOW, this paragraph.
    Therefore, use 'above' to fix the reference's direction.

* `admin/auth.rst`: spelling fixes:
> 'inbuilt' -> 'built-in'
    'Django based project' -> 'Django-based project'
    'OpenID based services' -> 'OpenID-based services'
    'tenant specific' -> tenant-specific'

### DIFF
--- a/docs/admin/admin.rst
+++ b/docs/admin/admin.rst
@@ -28,7 +28,7 @@ and other settings:
 In the :guilabel:`Reports` section, you can check the status of your site, tweak
 it for :ref:`production`, or manage SSH keys used to access :ref:`vcs-repos`.
 
-Manage database objects in any of the sections below.
+Manage database objects under any of the sections.
 The most interesting one is probably :guilabel:`Weblate translations`,
 where you can manage translatable projects, see :ref:`project` and :ref:`component`.
 
--- a/docs/admin/auth.rst
+++ b/docs/admin/auth.rst
@@ -17,10 +17,10 @@ The authentication attempts are subject to :ref:`rate-limit`.
 Authentication backends
 -----------------------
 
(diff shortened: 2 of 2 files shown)


docs/admin/admin.rst
```diff
@@ -28,7 +28,7 @@ and other settings:
 In the :guilabel:`Reports` section, you can check the status of your site, tweak
 it for :ref:`production`, or manage SSH keys used to access :ref:`vcs-repos`.
 
-Manage database objects in any of the sections below.
+Manage database objects in any of the sections above.
```
nijel, line 31: I think the original meaning here was that these sections are below "Reports" section. I see the screenshot here is broken for past few releases, in 4.0 it was still correct (I will look into this issue):

rffontenelle: @comradekingu and I discussed this, but opted to patch it differently. No problem. Considering that "below" is ambiguous (as it could refer to sections ahead in the same documentation), what do you think it would be better? `Manage database objects under any of the sections.` ? Kingu suggested "underneath".

nijel: That sounds fine to me. The key is that it is describing position on the page, not in the docs.

rffontenelle: @nijel I force-pushed the commits using "under" instead of the previous "above".

nijel:

PR #4259, opened 2020-08-05
Spelling: Back-end, except, uWSGI, Docker Machine, the, container

comradekingu:
As per
https://en.wikipedia.org/wiki/UWSGI
https://github.com/docker/machine


nijel:

Issue #4220, opened 2020-07-24
Limit the number of "Other components" displayed


tjhietala:
**Is your feature request related to a problem? Please describe.**
In projects with many components, listing them all on a component page is excessive, making the page many times longer than it needs to be. For example: https://translations.documentfoundation.org/projects/libo_ui-master/cuimessages/fi/

**Describe the solution you'd like**
Is it necessary to display "Other components" on the component page at all? If it is, the list should at least be collapsed by default, if the number of components exceeds (say) 10.

**Describe alternatives you've considered**
Including simply a link to the page that shows the components.

**Additional context**

nijel:
It was introduced to give bigger visibility to related translations, see https://github.com/WeblateOrg/weblate/issues/559, but it certainly makes sense to limit the listing to a few of them (preferring not translated ones) and include link to full list (https://translations.documentfoundation.org/languages/fi/libo_ui-master/ in your case).

comradekingu:
@tjhietala I think the current limit elsewhere is how many items you can reasonably fit on a UHD-display on its side. That is 60 or so. On https://translations.documentfoundation.org/languages/fi/libo_ui-master/ is something like 90.
I would expect to find all components in "components". What are "other components" in that regard?

nijel:
@comradekingu Just click on link in first post to see that ;-)

krishnmohan19:
Hii @nijel  i am new to open source  so can i work on this. ?

nijel:
@krishnmohan19 You are welcome.

The related code is here (note the logic for adding ghost translations):

https://github.com/WeblateOrg/weblate/blob/138d10ad14e8ef3962670fa3faea7ac2dbe1bc2c/weblate/trans/views/basic.py#L299-L316

And template is here:

https://github.com/WeblateOrg/weblate/blob/138d10ad14e8ef3962670fa3faea7ac2dbe1bc2c/weblate/templates/translation.html#L184-L191

PR #4267, opened 2020-08-05
Fix: Limit number of components displayed on translations overview page

osori:

Here is my fix to #4220, which does the following:
1. Limit the number of "Other components" displayed to 10, preferring untranslated ones
2. Template: Add a link to the full list of components

### DIFF
--- a/weblate/templates/translation.html
+++ b/weblate/templates/translation.html
@@ -186,7 +186,7 @@
   <div class="panel-heading"><h4 class="panel-title">{% trans "Other components" %}</h4></div>
 {% include "snippets/list-objects.html" with objects=other_translations name_source="component_name" label=_("Component") in_panel=True %}
   <div class="panel-footer">
-    <a href="{% url 'project-language' lang=object.language.code project=object.component.project.slug %}">See all components</a>
+    <a href="{% url 'project-language' lang=object.language.code project=object.component.project.slug %}">{% trans "Browse all components" %}</a>
   </div>
 </div>
 {% endif %}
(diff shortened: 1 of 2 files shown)


weblate/templates/translation.html
```diff
@@ -185,6 +185,9 @@ <h4 class="panel-title">
 <div class="panel panel-default">
   <div class="panel-heading"><h4 class="panel-title">{% trans "Other components" %}</h4></div>
 {% include "snippets/list-objects.html" with objects=other_translations name_source="component_name" label=_("Component") in_panel=True %}
+  <div class="panel-footer">
+    <a href="{% url 'project-language' lang=object.language.code project=object.component.project.slug %}">See all components</a>
```
nijel, line 189: Please make the text localizable. Also in similar context we use "Browse all ...", so please make the text consistent.

osori: Oops, my bad. I just changed the text to be more consistent and wrapped it with trans tag. 
Thank you for reviewing!

weblate/trans/views/basic.py
```diff
@@ -315,6 +315,12 @@ def show_translation(request, project, component, lang):
         if test_component.can_add_new_language(user):
             other_translations.append(GhostTranslation(test_component, obj.language))
 
+    # Limit the number of other components displayed to 10, preferring untranslated ones
+    other_translations = sorted(
+        other_translations,
```
osori, line 320: I'm not sure what `black` would change here... it seems fine to me. Any ideas why this happens?

weblate/trans/views/basic.py
```diff
@@ -315,6 +315,12 @@ def show_translation(request, project, component, lang):
...
+        key=lambda t : t.stats.translated_percent
```
nijel, line 321: I think it complains about this:

```suggestion
        key=lambda t : t.stats.translated_percent,
```

The best approach is install pre-commit hooks which will format the code, see https://docs.weblate.org/en/latest/contributing/code.html#coding-standard

osori: This is my first time committing using pre-commit and wow, this is great! I should have read the contributor docs thoroughly. Thanks a lot for your help!

nijel:

Issue #3251, opened 2019-11-28
[FR] User statistics per api


reloxx13:
**Is your feature request related to a problem? Please describe.**
Its not possible to export user data/statistics by api.

**Describe the solution you'd like**
Having an api action to get statistics from users
/api/users/statistics
/api/user/reloxx/statistics
/api/user/{user_id}/statistics

(maybe user&project?)

**Describe alternatives you've considered**
Grabbing the /counts but not possible cause of csrf.

**Additional context**
I would like to grab statistics with telegraf to push them in an influxdb instance and show it with grafana/chronograf
PS: telegram http plugin does not support paginations, mostly you have the oldest data on first page. having it desc sorted would be better => if no pagination is supported, atleast get the latest data.

https://github.com/influxdata/telegraf/tree/master/plugins/inputs/http

nijel:
We currently do not have much of the per user stats, quite similar state as in https://github.com/WeblateOrg/weblate/issues/2752

PR #4318, opened 2020-08-15
Adds API for user statistics

SaptakS:
Refs: #3251 

I have added the code similar to the user details view to show statistics of user owned projects and user contributed projects. Does that suffice the need? Also, not really sure about the math so can't verify if this code works as intended.

@nijel let me know your thoughts


nijel:
I can see several problems with this:

* You don't count user contributions, but you aggregate stats of projects user have contributed to. I'm not sure this is useful at all.
* The stats are never invalidated, so they will become horribly outdated (they last for 30 days in the cache).
* There are existing aggregates on user translations/comments/uploads/suggestions in the `Profile` object which could be utilized for this.

SaptakS:
@nijel Okay. I will update accordingly.

weblate/accounts/models.py
```diff
@@ -432,6 +432,7 @@ def increase_count(self, item: str, increase: int = 1):
         # Update database
         update = {item: F(item) + increase}
         Profile.objects.filter(pk=self.pk).update(**update)
+        self.user.stats.invalidate()
```
nijel, line 435: This doesn't catch language changes, so these might be outdated.

weblate/utils/stats.py
```diff
@@ -833,6 +833,31 @@ def calculate_item(self, item):
...
 
 
+class UserStats(BaseStats):
+    def invalidate(self, language: Optional[Language] = None, recurse: bool = True):
+        super().invalidate()
+
+    def _prefetch_basic(self):
+        base = self._object.profile
+        stats = {
+            "all": base.translated + base.suggested + base.uploaded + base.commented,
```
nijel, line 843: Do you see a use case for this? To me it seems like aggregating unrelated metrics...

weblate/utils/stats.py
```diff
@@ -833,6 +833,31 @@ def calculate_item(self, item):
         self.store(item, result)
 
 
+class UserStats(BaseStats):
```
nijel, line 836: I'm still not sure that introducing the stats object for user is way to go. It mostly just mirrors the fields in the database and introduces risk of inconsistencies.

This is very different from the translation stats where heavy aggregation is going on and it saves a lot of resource to calculate that.

SaptakS: Then should I just add it directly in user model as a dictionary?

weblate/utils/stats.py
SaptakS: I was thinking as "All Contributions made by user". Maybe doesn't make sense.

nijel: I think it's easy to aggregate on the client side in case somebody wants it. In case more metrics would be added in the future the meaning of "all" would change in time, what could be confusing.

weblate/utils/stats.py
nijel: My biggest concern is whether we need UserStats at all. I currently doesn't do any expensive calculations, so I'd say it's not needed for now.
You can use directly the Profile attributes in the UserStatisticsSerializer without need for intermediate object.

SaptakS:
The statistics are now fetched directly  via the serializer using Profile attributes. Removed the UserStats object given that there is no complicated aggregation of any sort as of now.

nijel (approved):
Looks good, will merge it after 4.2.1 release.

PR #4340, opened 2020-08-19
Update venv-debian.rst

stefonarch:
rabbitmq  was  missing, leading to many memory logs, celery defaulted  to 
amqp://guest:**@127.0.0.1:5672//
instead of using redis

- [ ] Every commit has a message which describes it

### DIFF
--- a/docs/admin/install/venv-debian.rst
+++ b/docs/admin/install/venv-debian.rst
@@ -28,8 +28,8 @@ Installing on Debian and Ubuntu
     # Web server option 2: Apache with ``mod_wsgi``
     apt install apache2 libapache2-mod-wsgi
 
-    # Caching backend: Redis
-    apt install redis-server
+    # Caching backend: Redis and Rabbitmq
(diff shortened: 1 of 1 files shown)


nijel:
Celery defaults to Redis in the example settings:

https://github.com/WeblateOrg/weblate/blob/2d14d76eda33f9c1a92eca148603ff5d4fa31e94/weblate/settings_example.py#L862

stefonarch:
Without rabbitqm server running here it did not respect those settings.
weblate-memory.log was plenty of
 `weblate ERROR/MainProcess] consumer: Cannot connect to amqp://guest:**@127.0.0.1:5672//: [Errno 111] Connection refused`

and a warning about celery non working properly in the GUI.
Googled tons of stuff with this error, installing rabbitqm-server solved immediately.

nijel:
Then Celery is for some reason not respecting the settings. Maybe running from different user or virtualenv? Weblate should work with both backends, but the documented setup is for redis, that's why only redis is installed.

PR #4349, opened 2020-08-20
Removes pagination from most pages

### LINKED ISSUE #3011: Sorting columns only sort following the current pagination
**Describe the bug**
When sorting the components using the available columns (Strings, words, etc), it only performs the sort on the current pagination, not on the whole list of components. I noticed it because I wasn't understanding the global translations statistics of a language (not everything at 100%), while when sorting the components on the first page, everything was at 100%.
I understood when I went to the next page and saw that some stats were not at 100%.

**To Reproduce**
1. Create enough components to have 2 pages of components
2. Translates every components from the first page
3. Order the components by "String" in desc order
4. See that the components not translated from the second page are never shown

**Expected behavior**
The sorting of columns should be applied on all components, not just the ones that are currently displayed.

SaptakS:
Fixes #3011 

Based on discussion with @nijel given that the listing of projects and other parts is much faster now, and usually the amount of things to be listed aren't that huge, we decided to remove pagination for now.

I have left pagination in `search.html` because sometime the number of units might be a little too many based on the search.
Also, the translate page still has the unit-pagination. The sorting in the search forms is handled in server side, so the sorting should still work.


weblate/templates/screenshots/screenshot_list.html
```diff
@@ -73,6 +73,4 @@
 </tbody>
 </table>
 
-{% include "paginator.html" %}
```
nijel, line 76: The paginator should stay here.

weblate/templates/trans/change_list.html
```diff
@@ -38,12 +38,8 @@
 </div>
 {% endif %}
 
-{% include "paginator.html" %}
-
 {% include "last-changes-content.html" with last_changes=object_list %}
 
-{% include "paginator.html" %}
```
nijel, line 45: The paginator should stay here.

Also, you've removed only the rendering of the paginator, but the results are still paginated, so that would not work correctly...

weblate/templates/trans/suggestion_list.html
```diff
@@ -12,6 +12,4 @@
 
 {% include "snippets/suggestions.html" with suggestions=object_list multi_unit=1 %}
 
-{% include "paginator.html" %}
```
nijel, line 15: I think the paginator should stay here. Again, the results are still paginated here, you've removed only the UI.

weblate/templates/weblate_auth/user_list.html
```diff
@@ -22,6 +22,4 @@ <h4>{{ object }}</h4>
 {% endfor %}
 </div>
 
-{% include "paginator.html" %}
```
nijel, line 25: I think the paginator should stay here. Again, the results are still paginated here, you've removed only the UI.

weblate/trans/views/basic.py
```diff
@@ -165,8 +163,7 @@ def show_project(request, project):
     )
 
     # Paginate components of project.
-    all_components = obj.component_set.filter_access(user).prefetch().order()
-    components = prefetch_tasks(prefetch_stats(get_paginator(request, all_components)))
+    components = obj.component_set.filter_access(user).prefetch().order()
```
nijel, line 166: You've removed the prefetch calls here, that will cause severe performance hit.

nijel (changes requested):
The original issue is only about components (or more generic about things which use javascript ordering). You've removed the paginator from other pages as well:

* Changes definitely need to be paginated, there are millions of them.
* Glossary needs to be paginated as well there can be thousands of them.
* Screenshots should stay paginated.
* Users should stay paginated.

PR #4380, opened 2020-08-25
skip strings with empty source for "other occurrences"

Blaimi:
this happens for translation types where source can be empty like "gettext PO file (monolingual)"

Signed-Off-By: Matthias Blümel <matthias.bluemel@krumedia.com>

- [X] Every commit has a message which describes it
- [X] Every commit is needed on its own, if you have just minor fixes to previous commits, you can squash them

It's probaly better to filter them in the lines above but I don't know how

### DIFF
--- a/weblate/trans/views/edit.py
+++ b/weblate/trans/views/edit.py
@@ -109,6 +109,7 @@ def get_other_units(unit):
             translation__component__project=translation.component.project,
             translation__language=translation.language,
         )
+        .exclude(source="")
         .filter(query)
     )
 


weblate/trans/views/edit.py
```diff
@@ -117,6 +117,8 @@ def get_other_units(unit):
         return result
 
     for item in units:
+        if item.source == "":
+            continue
```
nijel, line 121: Better to do this in the database query above to avoid fetching not needed units from the database. Appending `.exclude(source="")` should do this.

Blaimi: yep, tested, works, fixed, pushed, thanks.

I'd like to help with a unittest but I absolutely have no expierience in python.

nijel:

PR #4388, opened 2020-08-26
Various minor fixes by Weblate translators

yurchor:
Many thanks in advance for reviewing.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -471,7 +471,7 @@ not using Weblate.
 
 .. _faq-codes:
 
-Why does Weblate use language codes such sr_Latn or zh_Hans?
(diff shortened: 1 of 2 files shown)


docs/faq.rst
```diff
@@ -471,7 +471,7 @@ not using Weblate.
 
 .. _faq-codes:
 
-Why does Weblate use language codes such sr_Latn or zh_Hant?
+Why does Weblate use language codes such sr_Latn or zh_Hans?
```
nijel, line 474: Both `zh_Hant` and `zh_Hans` are valid, I'd avoid changing the string...

```suggestion
Why does Weblate use language codes such sr_Latn or zh_Hant?
```

nijel:

PR #4401, opened 2020-08-30
Adds glossary API

### LINKED ISSUE #4118: Add API for glossary
**Describe the bug**

When uploading a json file containing translations to the source language with endpoint:
POST /api/translations/(string:project)/(string:component)/(string:language)/file/

New fields saved in json file do not show as additianal translation just like when passing the same file with import glossary method in web client ui.

**To Reproduce**

1. Prepare json file with translations
2. Upload the file with curl passing the parameters:
 curl -X POST \
    -F file=@$filepath \
    -F overwrite=true \
    -H "Authorization: Token $API_KEY" \
    "https://hostname.com/api/translations/$PROJECT/$TRANSLATIONS_COMPONENT/$LANG/file/"
3. Open the glossary in the web client /dictionaries/app5/en/
4. See error - no additional fields from json file added.

If json file added manually glossary is updated with new fields in json file.

**How to extend glossary through API**

**Expected behavior**

Uploading a translation json file with additional keys extends the glossary.

SaptakS:
Fixes #4118 

### DIFF
--- a/weblate/api/serializers.py
+++ b/weblate/api/serializers.py
@@ -512,25 +512,13 @@ class TermSerializer(serializers.ModelSerializer):
             "target",
         )
 
-    def create(self, validated_data):
-        language_validated = validated_data.get("language")
-        if language_validated:
-            validated_data["language"] = Language.objects.get(
-                code=language_validated.get("code")
+    def to_internal_value(self, data):
+        result = super().to_internal_value(data)
+        if "language" in result:
+            result["language"] = Language.objects.get(
+                code=result["language"].get("code")
             )
-        term = Term.objects.create(**validated_data)
-        return term
-
-    def update(self, instance, validated_data):
-        language_validated = validated_data.get("language", None)
-        if language_validated:
-            instance.language = Language.objects.get(
-                code=language_validated.get("code")
-            )
-        instance.source = validated_data.get("source", instance.source)
-        instance.target = validated_data.get("target", instance.target)
-        instance.save()
-        return instance
+        return result
 
 
 class GlossarySerializer(serializers.ModelSerializer):
@@ -559,25 +547,13 @@ class GlossarySerializer(serializers.ModelSerializer):
             "url": {"view_name": "api:glossary-detail", "lookup_field": "id"}
         }
 
-    def create(self, validated_data):
-        source_language_validated = validated_data.get("source_language")
-        if source_language_validated:
-            validated_data["source_language"] = Language.objects.get(
-                code=source_language_validated.get("code")
+    def to_internal_value(self, data):
+        result = super().to_internal_value(data)
+        if "source_language" in result:
+            result["source_language"] = Language.objects.get(
+                code=result["source_language"].get("code")
             )
-        glossary = Glossary.objects.create(**validated_data)
-        return glossary
-
-    def update(self, instance, validated_data):
-        source_language_validated = validated_data.get("source_language", None)
-        if source_language_validated:
-            instance.source_language = Language.objects.get(
-                code=source_language_validated.get("code")
-            )
-        instance.name = validated_data.get("name", instance.name)
-        instance.color = validated_data.get("color", instance.color)
-        instance.save()
-        return instance
+        return result
 
 
 class TranslationSerializer(RemovableSerializer):
--- a/weblate/api/serializers.py
+++ b/weblate/api/serializers.py
@@ -23,6 +23,7 @@ from rest_framework import serializers
 
 from weblate.accounts.models import Subscription
 from weblate.auth.models import Group, Permission, Role, User
+from weblate.glossary.models import Glossary, Term
 from weblate.lang.models import Language, Plural
 from weblate.screenshots.models import Screenshot
 from weblate.trans.defines import LANGUAGE_NAME_LENGTH, REPO_LENGTH
@@ -177,10 +178,7 @@ class LanguageSerializer(serializers.ModelSerializer):
 
 class FullUserSerializer(serializers.ModelSerializer):
     groups = serializers.HyperlinkedIdentityField(
-        view_name="api:group-detail",
-        lookup_field="id",
-        many=True,
-        read_only=True,
+        view_name="api:group-detail", lookup_field="id", many=True, read_only=True
     )
     notifications = serializers.HyperlinkedIdentityField(
         view_name="api:user-notifications",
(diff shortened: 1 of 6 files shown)


weblate/api/views.py
```diff
@@ -752,6 +755,142 @@ def destroy(self, request, *args, **kwargs):
...
+        )
+
+    def perm_check(self, request):
+        obj = self.get_object()
+        if (
+            not request.user.has_perm("glossary.edit")
+            or obj.project_id not in request.user.allowed_project_ids
+            or not request.user.has_perm("project.edit")
+        ):
+            self.permission_denied(request, message="Can not manage glossary")
```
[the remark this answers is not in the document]
SaptakS: I have added the permission checks separately in the individual views related to the Term object managing. So I guess I will just remove the "glossary.edit" from `self.perm_check()`

weblate/api/views.py
```diff
@@ -752,6 +755,141 @@ def destroy(self, request, *args, **kwargs):
...
+        try:
+            term = obj.term_set.get(id=term_id)
+        except (Term.DoesNotExist, ValueError) as error:
+            return Response(
+                data={"result": "Unsuccessful", "detail": force_str(error)},
+                status=HTTP_400_BAD_REQUEST,
+            )
+
+        if request.method == "DELETE":
+            if not request.user.has_perm("glossary.delete"):
```
nijel, line 871: You need to pass matching project object here as well. In the views this is already scoped to a project, so we use that one, but here we should probably check all glossary projects. Something like (not tested):

```suggestion
            if not any(request.user.has_perm("glossary.delete", project) for project in obj.projects.all()):
```

Same applies to other `glossary.*` permission checks.

weblate/api/views.py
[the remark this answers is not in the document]
nijel: Yes, you can also remove the allowed_project_ids check, it is not necessary, the permission check already covers that.

weblate/api/serializers.py
```diff
@@ -507,6 +505,86 @@ class Meta:
...
+        )
+
+    def create(self, validated_data):
+        language_validated = validated_data.get("language")
+        if language_validated:
+            validated_data["language"] = Language.objects.get(
+                code=language_validated.get("code")
+            )
+        term = Term.objects.create(**validated_data)
+        return term
```
nijel, line 527: Handling of the source language attribute really should go to the  `to_internal_value` method. Handling it there makes all the validation have correct object instead of some intermediate values which fail the validation. This is not that big issue here, but I've hit the issue while creating https://github.com/WeblateOrg/weblate/pull/4458

Related code there is: https://github.com/WeblateOrg/weblate/pull/4458/files#diff-b6e5dde92daa37bd15834c3b38ed90b0R476

weblate/api/serializers.py
```diff
@@ -507,6 +505,86 @@ class Meta:
...
+    def update(self, instance, validated_data):
+        source_language_validated = validated_data.get("source_language", None)
+        if source_language_validated:
+            instance.source_language = Language.objects.get(
+                code=source_language_validated.get("code")
+            )
+        instance.name = validated_data.get("name", instance.name)
+        instance.color = validated_data.get("color", instance.color)
+        instance.save()
+        return instance
```
nijel, line 585: Similar problem as with TermSerializer, see above.

SaptakS:
Okay

SaptakS:
@nijel Updated the code to use to_internal_value

weblate/api/serializers.py
```diff
@@ -177,10 +178,7 @@ def get_value(self, dictionary):
 
 class FullUserSerializer(serializers.ModelSerializer):
     groups = serializers.HyperlinkedIdentityField(
-        view_name="api:group-detail",
-        lookup_field="id",
-        many=True,
-        read_only=True,
+        view_name="api:group-detail", lookup_field="id", many=True, read_only=True
```
nijel, line 181: Reverting change caused by mixed up black versions.

```suggestion
        view_name="api:group-detail",
        lookup_field="id",
        many=True,
        read_only=True,
```

weblate/api/serializers.py
```diff
@@ -502,6 +503,62 @@ class Meta:
...
+            "id",
+            "source",
+            "target",
+        )
+
+    def to_internal_value(self, data):
+        result = super().to_internal_value(data)
+        if "language" in result:
+            result["language"] = Language.objects.get(
+                code=result["language"].get("code")
```
nijel, line 522: The `get` invocation has no benefits here:

```suggestion
                code=result["language"]["code"]
```

weblate/api/serializers.py
```diff
@@ -502,6 +503,62 @@ class Meta:
...
+        )
+        extra_kwargs = {
+            "url": {"view_name": "api:glossary-detail", "lookup_field": "id"}
+        }
+
+    def to_internal_value(self, data):
+        result = super().to_internal_value(data)
+        if "source_language" in result:
+            result["source_language"] = Language.objects.get(
+                code=result["source_language"].get("code")
```
nijel, line 557: The `get` invocation has no benefits here:

```suggestion
                code=result["source_language"]["code"]
```

weblate/api/views.py
```diff
@@ -750,6 +753,143 @@ def destroy(self, request, *args, **kwargs):
...
+    """Translation projects API."""
+
+    queryset = Glossary.objects.none()
+    serializer_class = GlossarySerializer
+    lookup_field = "id"
+
+    def get_queryset(self):
+        return (
+            Glossary.objects.filter_access(self.request.user)
+            .prefetch_related("source_language")
```
nijel, line 766: Project objects will be needed as well:

```suggestion
            .prefetch_related("source_language", "project")
```

weblate/api/views.py
```diff
@@ -750,6 +753,143 @@ def destroy(self, request, *args, **kwargs):
...
+            .order_by("id")
+        )
+
+    def perm_check(self, request):
+        if not request.user.has_perm("project.edit"):
```
nijel, line 771: It should be permission on the object level, it should be checked on Glossary.project.

weblate/api/views.py
```diff
@@ -750,6 +753,143 @@ def destroy(self, request, *args, **kwargs):
...
+            self.permission_denied(request, message="Can not manage glossary")
+
+    def term_perm_check(self, request, permission):
+        obj = self.get_object()
```
nijel, line 775: IMHO this causes additional fetch from the database, it should reuse object which will be retrieved anyway during the request.

weblate/api/views.py
```diff
@@ -750,6 +753,143 @@ def destroy(self, request, *args, **kwargs):
...
+        self.perm_check(request)
+
+        try:
+            project = Project.objects.get(slug=project_slug)
+        except (Project.DoesNotExist, ValueError) as error:
+            return Response(
+                data={"result": "Unsuccessful", "detail": force_str(error)},
+                status=HTTP_400_BAD_REQUEST,
+            )
```
nijel, line 831: This seems better approach:

```suggestion
            raise ParseError(str(error))
```

weblate/api/views.py
```diff
@@ -750,6 +753,143 @@ def destroy(self, request, *args, **kwargs):
...
+    def terms_details(self, request, id, term_id):
+        obj = self.get_object()
+
+        try:
+            term = obj.term_set.get(id=term_id)
+        except (Term.DoesNotExist, ValueError) as error:
+            return Response(
+                data={"result": "Unsuccessful", "detail": force_str(error)},
+                status=HTTP_400_BAD_REQUEST,
+            )
```
nijel, line 870:
```suggestion
            raise ParseError(str(error))
```

weblate/api/serializers.py
```diff
@@ -502,6 +503,62 @@ class Meta:
...
+            "source",
+            "target",
+        )
+
+    def to_internal_value(self, data):
+        result = super().to_internal_value(data)
+        if "language" in result:
+            result["language"] = Language.objects.get(
+                code=result["language"]["code"]
+            )
```
nijel, line 523:
```suggestion
            result["language"] = Language.objects.get(code=result["language"]["code"])
```

weblate/api/tests.py
```diff
@@ -1217,6 +1218,295 @@ def test_create_component_local(self):
...
+            kwargs={"id": Glossary.objects.order_by("id").all()[0].id},
+            method="post",
+            superuser=True,
+            code=201,
+            format="json",
+            request={"language": {"code": "cs"}, "source": "test", "target": "test"},
+        ).data
+        self.do_request(
+            "api:glossary-terms-details",
+            kwargs={"id": Glossary.objects.order_by("id").all()[0].id, "term_id": -1},
```
nijel, line 1425:
```suggestion
            kwargs={"id": Glossary.objects.order_by("id").all()[0].id, "term_id": 1000},
```

weblate/api/tests.py
```diff
@@ -1217,6 +1218,295 @@ def test_create_component_local(self):
...
+            method="get",
+            code=400,
```
nijel, line 1427: Non existing term should fail with 404

```suggestion
            code=404,
```

weblate/api/views.py
```diff
@@ -712,6 +715,141 @@ def destroy(self, request, *args, **kwargs):
...
+        url_path="terms/(?P<term_id>[0-9]+)",
+        serializer_class=TermSerializer,
+    )
+    def terms_details(self, request, id, term_id):
+        obj = self.get_object()
+
+        try:
+            term = obj.term_set.get(id=term_id)
+        except (Term.DoesNotExist, ValueError) as error:
+            raise ParseError(str(error))
```
nijel, line 830:
```suggestion
        except Term.DoesNotExist:
            raise Http404("Term not found")
```

nijel (commented):
Correctly return 404 for not existing objects.

weblate/api/tests.py
```diff
@@ -1217,6 +1218,295 @@ def test_create_component_local(self):
...
+        response = self.client.delete(
+            reverse(
+                "api:glossary-delete-projects",
+                kwargs={
+                    "id": Glossary.objects.order_by("id").all()[0].id,
+                    "project_slug": -1,
+                },
+            ),
+        )
+        self.assertEqual(response.status_code, 400)
```
nijel, line 1369:
```suggestion
        self.assertEqual(response.status_code, 404)
```

nijel:

PR #4421, opened 2020-09-01
DOCKER: update LDAP UNION SEARCH guide

em0t:
`WEBLATE_AUTH_LDAP_USER_SEARCH_UNION` will cover `WEBLATE_AUTH_LDAP_USER_SEARCH`, so if need union search, only config union is ok, see [code](https://github.com/WeblateOrg/weblate/blob/master/weblate/settings_docker.py#L384)


nijel (approved):
Thanks, makes sense.

PR #4422, opened 2020-09-01
fix: wrong Puctuation class name to Punctuation

em0t:
A little spell bug, the right spell `Punctuation`  has occurred [here](https://github.com/em0t/weblate/blob/2b99426d55bac2387809e88fc56aaf178c5fd836/weblate/checks/chars.py#L389)


nijel (approved):
Thanks for spotting.

PR #4431, opened 2020-09-03
Sre 1254 weblate demo pvc

bstolz:
A brand spanking new pvc-demo.yml file for your repo. This is what is currently running in weblate-demo.


nijel:
I guess this should have been against https://github.com/vendasta/weblate... Maybe you want to contribute some features to Weblate separately to avoid that massive divergence?

Issue #4491, opened 2020-09-11
Project file downloads violate component visibility permissions


rageandengage:
**Describe the bug**

Our project is setup in a semi-private manner where the "Project" is publicly visible, yet not all "Components" are. It's managed this way with an expectation of secrecy: If the Component is not visible, none of its strings should be visible, only the project's statistics.

The publicly accessible file downloads option allows you to violate permissions tests and read the current state of all translations.

**To Reproduce**

Create a Project with Custom permissions. Create a Translation Component with Custom permissions. Upload a source file and create several new language files for it.

Assign your Guest role the ability to view the Project, but not the Component. As a Guest, you now have the ability to view the status of all "private" translations within the Project by downloading them through the "Download translation files as ZIP file" dropdown menu (https://instance.url/download/project-slug/?format=zip).

**Expected behavior**

It's expected that the contents of the Components should remain secret even if you have view permissions on the Project. When generating a file download response, permissions should be tested to see whether files should be included.

It would also be preferable if file download operations could be restricted by Role.

PR #4494, opened 2020-09-12
Files: Return a filtered component view for zip file download requests

rageandengage:
Fixes https://github.com/WeblateOrg/weblate/issues/4491

In addition to this, we should probably be updating weblate/templates/component-list.html to prevent showing the Download button  if no components are visible. I'm not familiar with this framework as to how to do it, though

### DIFF
--- a/weblate/trans/views/files.py
+++ b/weblate/trans/views/files.py
@@ -86,11 +86,8 @@ def download_component(request, project, component):
 
 def download_project(request, project):
     obj = get_project(request, project)
+    obj.commit_pending("download", None)
     components = obj.component_set.filter_access(request.user)
-    if not components:
-        raise PermissionDenied()
-    for component in components:
-        component.commit_pending("download", None)
     return download_multi(
         Translation.objects.filter(component__in=components), request.GET.get("format")
     )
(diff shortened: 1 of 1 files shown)


weblate/trans/views/files.py
```diff
@@ -86,18 +86,26 @@ def download_component(request, project, component):
 
 def download_project(request, project):
     obj = get_project(request, project)
-    obj.commit_pending("download", None)
+    components = obj.component_set.filter_access(request.user)
+    if not components:
+        raise PermissionDenied()
```
nijel, line 91: I think it's more appropriate to create empty file in this case. This is what happens right now when there is no component in a project.

weblate/trans/views/files.py
```diff
@@ -86,18 +86,26 @@ def download_component(request, project, component):
...
+    for component in components:
+        component.commit_pending("download", None)
```
nijel, line 93: I'd probably stay with project wide commit as that one is way more effective in case of project with many linked components. The additional commit on not accessible components should do no harm.

rageandengage:
@nijel 0cd4ba1 should resolve your suggestions. If there's anything else, please let me know

nijel:
Thanks, looks good now.

PR #4534, opened 2020-09-17
Generate 4096 bit RSA host key

tsabi:
Newer GitLab needs a host key with minimun 4096 bit lenth.


nijel:
Merged, thanks. Maybe it's time to switch to ed25519 instead of rsa keys...

Issue #4156, opened 2020-07-10
Support creating local VCS components using API


nijel:
**Is your feature request related to a problem? Please describe.**
The component creation API right now supports only creating components from remote VCS. It should also support no remote VCS setup where initial file is either created by Weblate or provided by upload.

In other words, it should support following workflows supported in the UI:

* [ ] Upload translations files - upload more files in a ZIP
* [ ] Translate document - upload main file
* [ ] Start from scratch - no file provided, starts with blank one

**Describe the solution you'd like**
Extend https://docs.weblate.org/en/latest/api.html#post--api-projects-(string-project)-components- to support local vcs.

**Describe alternatives you've considered**
Right now, this is not supported over API, but I don't think it's a viable approach.

**Additional context**
The related code is currently placed in the view, but should be shared for the API, not sure if it makes sense to move it to the models.

For example, creating component from scratch si here:

https://github.com/WeblateOrg/weblate/blob/93991891060a6eb02d8961edaf3f5aea853c7cc6/weblate/trans/views/create.py#L54-L72

nijel:
I've just implemented basics start from scratch in d9bfab2e0dfaef9fe62cd4c5ffb852558c6c8dd2

EpocDotFr:
Thank you!

PR #4553, opened 2020-09-21
Adds support to create components using docfile and zipfile via API

SaptakS:
Refs #4156 

### DIFF
--- a/weblate/api/serializers.py
+++ b/weblate/api/serializers.py
@@ -24,7 +24,6 @@ from rest_framework import serializers
 
 from weblate.accounts.models import Subscription
 from weblate.auth.models import Group, Permission, Role, User
-from weblate.formats.models import FILE_FORMATS
 from weblate.glossary.models import Glossary, Term
 from weblate.lang.models import Language, Plural
 from weblate.screenshots.models import Screenshot
@@ -42,7 +41,6 @@ from weblate.trans.util import check_upload_method_permissions, cleanup_repo_url
 from weblate.utils.site import get_site_url
 from weblate.utils.validators import validate_bitmap
 from weblate.utils.views import create_component_from_doc, create_component_from_zip
-from weblate.vcs.git import LocalRepository
 
 
 class MultiFieldHyperlinkedIdentityField(serializers.HyperlinkedIdentityField):
@@ -488,11 +486,11 @@ class ComponentSerializer(RemovableSerializer):
                 code=result["source_language"]["code"]
             )
         if "docfile" in result:
-            fake = create_component_from_doc(result)
+            create_component_from_doc(result)
             result.pop("docfile")
         if "zipfile" in result:
             try:
-                fake = create_component_from_zip(result)
+                create_component_from_zip(result)
             except BadZipfile:
                 raise serializers.ValidationError("Failed to parse uploaded ZIP file.")
             result.pop("zipfile")
--- a/weblate/api/serializers.py
+++ b/weblate/api/serializers.py
@@ -16,7 +16,6 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
-import os
 from zipfile import BadZipfile
 
 from django.conf import settings
@@ -42,6 +41,7 @@ from weblate.trans.models import (
 from weblate.trans.util import check_upload_method_permissions, cleanup_repo_url
 from weblate.utils.site import get_site_url
 from weblate.utils.validators import validate_bitmap
+from weblate.utils.views import create_component_from_doc, create_component_from_zip
 from weblate.vcs.git import LocalRepository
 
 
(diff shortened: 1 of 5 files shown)


weblate/api/serializers.py
```diff
@@ -478,6 +487,52 @@ def to_internal_value(self, data):
...
+            # Create repository
+            try:
+                LocalRepository.from_zip(fake.full_path, result["zipfile"])
+                LocalRepository.from_files(
+                    fake.full_path,
+                    {result["template"]: format_cls.get_new_file_content()},
+                )
+            except BadZipfile:
+                raise serializers.ValidationError("Failed to parse uploaded ZIP file.")
+            result.pop("zipfile")
```
nijel, line 535: This all duplicates existing code from `weblate/trans/views/create.py`. It should be extracted from there and reused in the API. It might to to the `weblate/trans/models/component.py` (where component related code is) or `weblate/utils/views.py` (where other view helpers are already stored).

weblate/api/tests.py
```diff
@@ -46,6 +46,8 @@
 
 TEST_PO = get_test_file("cs.po")
 TEST_POT = get_test_file("hello-charset.pot")
+TEST_DOC = get_test_file("short_description.txt")
```
nijel, line 49: Please use something more realistic for testing, see comment below.

```suggestion
TEST_DOC = get_test_file("cs.html")
```

weblate/api/tests.py
```diff
@@ -1235,6 +1237,54 @@ def test_patch(self):
...
+                code=201,
+                superuser=True,
+                request={
+                    "docfile": handle,
+                    "name": "Local project",
+                    "slug": "local-project",
+                    "repo": "local:",
+                    "vcs": "local",
+                    "filemask": "*.strings",
+                    "template": "en.strings",
```
nijel, line 1255: This is probably why you handle template in the create code. It should not be specified in this case - you have single document you want to translate, not set of files.

```suggestion
                    "filemask": "*.html",
                    "template": "en.html",
```

What would be more user friendly is to use separate serializer for this case and fill in all these values there, but documenting expected parameters in this case is good enough as well.

weblate/api/tests.py
```diff
@@ -1235,6 +1237,54 @@ def test_patch(self):
...
+                    "file_format": "strings-utf8",
```
nijel, line 1256: I'm surprised that this works - you're uploading plain text file and parsing it as strings, what might be a bug in the parser that it accepts it. 

```suggestion
                    "file_format": "html",
```

weblate/api/tests.py
```diff
@@ -1235,6 +1237,54 @@ def test_patch(self):
...
+                    "push": "https://username:password@github.com/example/push.git",
+                    "new_lang": "none",
```
nijel, line 1258: As there are no translations existing when uploading a single file, adding of translations should be supported.

```suggestion
                    "new_lang": "add",
```

weblate/api/tests.py
```diff
@@ -1235,6 +1237,54 @@ def test_patch(self):
...
+                superuser=True,
+                request={
+                    "zipfile": handle,
+                    "name": "Local project",
+                    "slug": "local-project",
+                    "repo": "local:",
+                    "vcs": "local",
+                    "filemask": "*.strings",
+                    "template": "en.strings",
+                    "file_format": "strings-utf8",
```
nijel, line 1280: If this passes it shows lack of validation. The zip file doesn't contain these files.

The proper configuration should be:

```suggestion
                    "filemask": "*.po",
                    "new_base": "project.pot",
                    "file_format": "po",
```

But please keep this broken test as well and make sure the validation fails in this case - the configured files are not present in the zip file and the component clean method should fail.

weblate/api/serializers.py
```diff
@@ -478,6 +487,25 @@ def to_internal_value(self, data):
             result["source_language"] = Language.objects.get(
                 code=result["source_language"]["code"]
             )
+        if "docfile" in result:
+            fake = create_component_from_doc(result)
+            format_cls = FILE_FORMATS[result["file_format"]]
+            LocalRepository.from_files(
+                fake.full_path,
+                {result["template"]: format_cls.get_new_file_content()},
+            )
```
nijel, line 496: Is this really needed? The local repository is already created in `create_component_from_doc` and creating template with default content should not be needed.

```suggestion
```

weblate/api/serializers.py
```diff
@@ -478,6 +487,25 @@ def to_internal_value(self, data):
...
+            result.pop("docfile")
+        if "zipfile" in result:
+            try:
+                fake = create_component_from_zip(result)
+                format_cls = FILE_FORMATS[result["file_format"]]
+                LocalRepository.from_files(
+                    fake.full_path,
+                    {result["template"]: format_cls.get_new_file_content()},
+                )
```
nijel, line 505: Same here.

```suggestion
```

weblate/api/serializers.py
SaptakS: Without the template file content defined, the creation from doc fails. I checked without this, but then creation would just fail.

nijel: Yes, the template has to be specified here. But either it should be specified by user or filled in automatically (see https://github.com/WeblateOrg/weblate/pull/4553/files#r496475217). Putting nonsense value will certainly not make it work (you put file content where filename is expected).

The major issue here is that validation apparently doesn't work as it should as the component creation happily accepts nonsense values.

nijel: The problem is template value you pass in the tests. In UI the creation process is two step - first the uploaded file is imported to the repo, then normal component configuration follows where user selects the file to translate and that sets all the component parameters. To make the API simple, I'd make it just single step, but that means that the params has to match the uploaded file, that's what I've tried to do in the code suggestions (but I might have missed something as I didn't test it).

nijel:
Thanks, I've fixed remaining issues in 9f4bde9a5433b49ec3f281da1632e07b4b9230b8.

SaptakS:
Thanks @nijel

PR #4586, opened 2020-09-24
Fix minor typo found by Mingun

yurchor:
Thanks in advance for reviewing.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -2262,7 +2262,7 @@ Glossary
 
 .. http:delete:: /api/glossary/(int:id)/projects/
 
-    Remove associate of a project with a glossary.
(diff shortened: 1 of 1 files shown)


docs/api.rst
```diff
@@ -2262,7 +2262,7 @@ Glossary
 
 .. http:delete:: /api/glossary/(int:id)/projects/
 
-    Remove asoociate of a project with a glossary.
+    Remove associate of a project with a glossary.
```
nijel, line 2265: I think it should actually be:

```suggestion
    Remove association of a project with a glossary.
```

yurchor: Fixed.

nijel:

PR #4626, opened 2020-09-30
Fix: Wrap comment text and make comment textarea automatically resize itself

### LINKED ISSUE #4442: Text wrapping and autosizing of CodeMirror
**Describe the bug**

The textarea in Weblate wraps text and autosizes based on length of the text. The CodeMirror editor which is recenlty used for comments (and some other inputs) doesn't do this.

**To Reproduce**

1. Go to any translation
2. Click on Comments tab
3. Write long comment

**Expected behavior**

* The long text should be wrapped for display to avoid horizontal overflow.
* The textare length should automatically size to fit the text.

osori:

This fixes #4442 and the following changes were made:
1. Wrap comment-area so that long comments without space does not overflow horizontally.
2. Wrap comment textarea using [`lineWrapping`](https://codemirror.net/doc/manual.html#option_lineWrapping) option from CodeEditor. This prevents comment text from having a horizontal scroll.
3. Make comment textarea automatically resize to fit its text

### DIFF
--- a/weblate/static/loader-codemirror.js
+++ b/weblate/static/loader-codemirror.js
@@ -62,6 +62,7 @@
       theme: "weblate",
       lineNumbers: false,
       lineWrapping: true,
+      viewportMargin: Infinity,
     });
 
     codemirror.on("change", function (cm, event) {
--- a/weblate/static/loader-codemirror.js
+++ b/weblate/static/loader-codemirror.js
@@ -61,6 +61,7 @@
       mode: "text/x-markdown",
       theme: "weblate",
       lineNumbers: false,
+      lineWrapping: true,
     });
 
     codemirror.on("change", function (cm, event) {
--- a/weblate/static/style-bootstrap.css
+++ b/weblate/static/style-bootstrap.css
@@ -225,7 +225,8 @@ textarea#id_comment {
   height: 4em;
 }
 div.CodeMirror.cm-s-weblate {
-  height: 5em;
+  min-height: 5em;
+  height: auto;
   color: #555555;
   font-family: inherit;
 }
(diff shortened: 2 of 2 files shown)


nijel:
Thanks, looks good. Unfortunately with this applied I get ugly gray bar in bottom of the editor:

image

Can you see it as well? I'm using Firefox 81.

In Chome it is shown as (long) horizontal scroll bar:

image

weblate/static/loader-codemirror.js
```diff
@@ -61,6 +61,7 @@
       mode: "text/x-markdown",
       theme: "weblate",
       lineNumbers: false,
+      lineWrapping: true,
```
nijel, line 64:
```suggestion
      lineWrapping: true,
      viewportMargin: Infinity,
```

Needed to display more than 10 rows, see https://codemirror.net/demo/resize.html and https://codemirror.net/doc/manual.html#option_viewportMargin

weblate/static/style-bootstrap.css
```diff
@@ -225,7 +225,8 @@ textarea#id_comment {
   height: 4em;
 }
 div.CodeMirror.cm-s-weblate {
-  height: 5em;
+  min-height: 5em;
```
nijel, line 228: The min-height seems to be source of displaying horizontal scroll bar as described in previous comments. 

```suggestion
```

weblate/static/style-bootstrap.css
```diff
@@ -508,6 +509,7 @@ form + .login-label {
   border-radius: 0 5px 5px 5px;
   background-color: #e5f0f2;
   padding: 8px 10px;
+  word-wrap: break-word;
```
nijel, line 512: Is this change really needed? In case it is, please submit it separately as this has nothing to do with comment editing, but affects comment rendering.

```suggestion
```

nijel: Okay, looking at the commit I understand it, but it was a bit confusing to see it in editor related pull request :-).

nijel:

PR #4646, opened 2020-10-05
Dash instead of a dot

tanolino:
I think this is a typo because it refers to the file mentioned above.


nijel (approved):
Thanks, it is indeed a typo.

PR #4648, opened 2020-10-05
Added source:contains option to unit api

adriandiazgar:
Added q (query) argument to unit api

By default the endpoint used to retrieve all the units ordered by id.
With the new implementation, the option to add **?q** like in https://docs.weblate.org/en/latest/user/search.html is possible.

- [ x] Every commit has a message which describes it
- [x ] Every commit is needed on its own, if you have just minor fixes to previous commits, you can squash them
- [] Any code changes come with test case (Can't find the testcase retrieving data from the API)
- [x ] Any new functionality is covered by user documentation


nijel:
Thanks for your contribution. I'd really prefer to introduce more flexible searching instead. There is already existing [search language](https://docs.weblate.org/en/latest/user/search.html) used in the UI and it should be reused for the API instead of adding limited filtering on single field.

Additionaly DRF has [support for filtering](https://www.django-rest-framework.org/api-guide/filtering/) and that should be used to nicely integrate it instead of manually crafting similar code.

adriandiazgar:
Hi @nijel  thanks for your feedback, I think I was able to implement the query functionality (q param) built-in in the UI.
I didn't apply it to every single API endpoint but just for the one, I need to have the functionality.
Thanks for your time

nijel:
Looks good! I think this is good to merge once you add tests.

In case you want to make it even better, look into DRF filtering and use it instead of manually processing the request params. One of the benefits of using filtering is that it is nicely rendered in the UI.

adriandiazgar:
Hi @nijel I just added the missing tests.

I am not sure why the Github actions are failing, doesn't seem to be related to my changes, I'm wondering if there is something else, could you please check it? 


Regarding DRF I've been trying to add it but I struggle to add it when the q doesn't exist in the model as a field.

nijel:
> I am not sure why the Github actions are failing, doesn't seem to be related to my changes, I'm wondering if there is something else, could you please check it?

Sorry, that's my fault, it should be fixed in master branch now.

nijel (approved):
Thanks , looks good now.

adriandiazgar:
Thanks for your time and your review, hope to contribute more in the future :)

PR #4686, opened 2020-10-13
Execute post_configure when updating an addon

cedk:


weblate/addons/management/commands/install_addon.py
```diff
@@ -67,7 +67,8 @@ def handle(self, *args, **options):
             addons = Addon.objects.filter_component(component).filter(name=addon.name)
             if addons.exists():
                 if options["update"]:
-                    addons.update(configuration=configuration)
+                    addon_component = addon.__class__(addons[0])
+                    addon_component.configure(configuration)
```
nijel, line 71: This looks wrong as it calls the `configure` method only for first addon, while the previous code did update it for all of them.

```suggestion
                    for addon_component in addons:
                        addon_component.configure(configuration)
```

weblate/addons/management/commands/install_addon.py
```diff
@@ -67,7 +67,8 @@ def handle(self, *args, **options):
             addons = Addon.objects.filter_component(component).filter(name=addon.name)
             if addons.exists():
```
nijel, line 68: Evaluate the query here as we will need all the objects in case they exist anyway:

```suggestion
            if addons:
```

weblate/addons/management/commands/install_addon.py
cedk: But there can be only one addon per component

weblate/addons/management/commands/install_addon.py
```diff
@@ -67,8 +67,8 @@ def handle(self, *args, **options):
             addons = Addon.objects.filter_component(component).filter(name=addon.name)
             if addons:
                 if options["update"]:
-                    addon_component = addon.__class__(addons[0])
```
cedk: I think the `addon.__class__` instantiation is still needed to be able to call `.configure`.

weblate/addons/management/commands/install_addon.py
```diff
@@ -65,9 +65,10 @@ def handle(self, *args, **options):
                             )
                     raise CommandError("Invalid addon configuration!")
             addons = Addon.objects.filter_component(component).filter(name=addon.name)
-            if addons.exists():
+            if addons:
                 if options["update"]:
-                    addons.update(configuration=configuration)
+                    for addon_component in addons:
+                        addon_component.configure(configuration)
```
nijel, line 71: Good point, it needs an addon not a model instance:

```suggestion
                        addon_component.addon.configure(configuration)
```

nijel:

Issue #3667, opened 2020-04-03
Add a question when removing users from the project


orangesunny:
**Is your feature request related to a problem? Please describe.**
With big number of users on the translation project, it’s quite tricky to hit the right *red bin* button.

**Describe the solution you'd like**
Ask a question *Do you want to remove `username` from this project?* With yes/no

GriffinMeltzer:
I'd like to work on this too.

nijel:
@GriffinMeltzer I've assigned you, thanks.

GriffinMeltzer:
Any ideas on where to find this in the code base?

nijel:
The HTML code is here: https://github.com/WeblateOrg/weblate/blob/master/weblate/templates/manage-access.html

Anubhav652:
I'd like to work on this.

nijel:
@Anubhav652 You're welcome, I've just assigned you the issue.

comradekingu:
@nijel That appears to be a spamtoberfest account.

Anubhav652:
> 
> 
> @nijel That appears to be a spamtoberfest account.

It's not

comradekingu:
I clicked around enough to notice, but then I had lost the thread. Sorry for implying it was.

Anubhav652:
I should implement this something along the lines of templates/trans/delete-form.html right?

nijel:
I don't thing typing the username is needed here, simple modal dialog would be okay here. Something similar is there when deleting translation memory entries:

https://github.com/WeblateOrg/weblate/blob/700594508f17e5c80474c0b71b1d343db3471f44/weblate/templates/memory/index.html#L74-L89

PR #4717, opened 2020-10-20
Added a confirmation box when removing user from a project

Anubhav652:
A confirmation box was added using modals when removing user from a project.

Issue: #3667

### DIFF
--- a/weblate/templates/manage-access.html
+++ b/weblate/templates/manage-access.html
@@ -39,13 +39,31 @@
   <td>{% if user.last_login %}{{ user.last_login|naturaltime }}{% else %}{% trans "Not yet signed in" %}{% endif %}</td>
   {% user_permissions user groups %}
   <td>
+<a href="#" data-toggle="modal" data-target="#delete_user" class="btn btn-link btn-xs red" title="{% trans "Remove" %}">
+    {% icon 'delete.svg' %}
+</a>
+ 
 <form action="{% url "delete-user" project=object.slug %}" method="post" class="inlineform">
 {% csrf_token %}
 <input type="hidden" name="user" value="{{ user.username }}" />
-<button type="submit" class="btn btn-link btn-xs red" title="{% trans "Remove" %}">
-{% icon 'delete.svg' %}
-</button>
+<div class="modal fade" tabindex="-1" role="dialog" id="delete_user">
+  <div class="modal-dialog" role="document">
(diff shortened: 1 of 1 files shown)


weblate/templates/manage-access.html
```diff
@@ -39,13 +39,31 @@
...
-</button>
+<div class="modal fade" tabindex="-1" role="dialog" id="delete_user">
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-label="{% trans "Close" %}"><span aria-hidden="true">&times;</span></button>
+        <h4 class="modal-title">{% trans "Are you absolutely sure?" %}</h4>
+      </div>
+      <div class="modal-body">
+        This will permanently delete {{ user.username }} from this project.
```
nijel, line 57: Please make this translatable using the [blocktrans](https://docs.djangoproject.com/en/3.0/topics/i18n/translation/#std:templatetag-blocktrans) tag.

weblate/templates/manage-access.html
```diff
@@ -39,13 +39,31 @@
...
+<a href="#" data-toggle="modal" data-target="#delete_user" class="btn btn-link btn-xs red" title="{% trans "Remove" %}">
+  {% icon 'delete.svg' %}
+</a>
 <form action="{% url "delete-user" project=object.slug %}" method="post" class="inlineform">
 {% csrf_token %}
 <input type="hidden" name="user" value="{{ user.username }}" />
-<button type="submit" class="btn btn-link btn-xs red" title="{% trans "Remove" %}">
-{% icon 'delete.svg' %}
-</button>
+<div class="modal fade" tabindex="-1" role="dialog" id="delete_user">
```
nijel, line 49: You should include user id in the form id, otherwise there are multiple forms with same ID in the page:
```suggestion
<div class="modal fade" tabindex="-1" role="dialog" id="delete_user_{{ user.id }}">
```

Same has to be abode in the modal toggle link.

nijel (changes requested):
Thanks for your contribution. Your code breaks when there are multiple users - it always shows and deletes the first user because the modal dialogs have same id.

Anubhav652:
Thanks for the feedback. It's all done now, I should've been more careful while testing.

nijel:

PR #4743, opened 2020-10-22
line breaks to make table smaller

orangesunny:
just to make it more convenient

Before submitting a pull request, please ensure that:

- [x] Every commit has a message which concisely describes it.
- [x] Every commit is needed on its own; if you have just minor fixes to previous commits, you can squash them.


nijel:
Sorry, but this doesn't work as intended, there doesn't seem to be any formatting possible, see https://docs.github.com/en/free-pro-team@latest/github/building-a-strong-community/configuring-issue-templates-for-your-repository

PR #4749, opened 2020-10-23
update openshift template to work with latest weblate version

SimonGolms:
Before submitting a pull request, please ensure that:

- [x] Every commit has a message which concisely describes it.
- [x] Every commit is needed on its own; if you have just minor fixes to previous commits, you can squash them.


openshift3/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+```bash
+oc delete all -l app=<APPLICATION_NAME>
+oc delete configmap -l app= <APPLICATION_NAME>
+oc delete secret -l app=<APPLICATION_NAME>
+# ATTTENTION! The following command is only optional and will permanently delete all of your data.
+oc delete pvc -l app=<APPLICATION_NAME>
+
+oc delete all -l app=weblate && oc delete secret -l app=weblate && oc delete configmap -l app=weblate && oc delete pvc -l app=weblate
+```
```
nijel, line 31: Maybe this would better fit into https://github.com/WeblateOrg/weblate/blob/master/docs/admin/install/openshift.rst so that it's rendered in the documentation - https://docs.weblate.org/en/latest/admin/install/openshift.html

nijel (commented):
Thanks for your contribution. Besides the docs placement, I'd like to hear feedback from previous contributors to the OpenShift code because I have no clue about that. Can @alesrosina or @GrahamDumpleton comment on the changes, please?

openshift3/README.md
SimonGolms: I know what you mean, personally I thought it was still too less content to list it on the official documentation page, especially regarding the further configuration. However, I wanted to at least provide some minimal information for (experience) users who click on the docs link.

nijel: I'd really prefer to have it inside the docs rather than in some README. Your instructions are way better than what we have right now. The documentation can be always better, but it's better to progress than to have it split into different locations because some parts are not perfect.

SimonGolms: Alright, based on the review and possible changes I will try to write a more detailed instruction for the docs next week. And if it's okay for you, I would do this in an extra PR to reach my hacktoberfest goals 🎃

nijel: Okay, feel free to remove the documentation changes from this PR and submit it separately as addition to existing documentation.

nijel:
Looks good, I will wait few more days in case somebody of the previous contributors is interested in a review and it case it won't arrive I will merge this.

Also, do you have any experience with testing this? It would be really great to have this covered by tests to have some confidence that the template works...

SimonGolms:
Sure, that's absolutely valid.

> Also, do you have any experience with testing this? It would be really great to have this covered by tests to have some confidence that the template works...

Unfortunately, I only have experience in running app instances under OpenShift and creating a respective template. For setting up an OpenShift Test Cluster, I'm still missing the experience. Maybe I can add the corresponding tests in the future, but unfortunately not at the moment and hope for other contributors, too.

nijel:
Okay, I will try to setup something myself using https://github.com/marketplace/actions/setup-okd-openshift-cluster, see https://github.com/WeblateOrg/weblate/pull/4766

nijel:
It seems that it does something, so at least it verifies that the template can be processed by OpenShift. I have no clue whether it actually starts the service or not, but at least we have something now :-). Further contributions are always welcome.

@SimonGolms can you please rebase this pull request on top of current master and adjust the test (as the template was changed to yaml).

alesrosina:
Late to the party - I'm already off the project, where I was doing this so sadly I don't have access to any openshift instance at the moment to test this.

nijel:

PR #4949, opened 2020-12-01
Fix typo found by Oguz Ersen

yurchor:
Thanks in advance for reviewing.


nijel:
Thanks, however this was already fixed in 533c3f447504028640c0bfeb770b1ee70ac4270b

PR #4954, opened 2020-12-01
Improve support for Azure Repositories

### LINKED ISSUE #4894: Support Commit Hooks for Azure Repositories connected via ssh
Right now it's only possible to send commit hooks from Azure Repositories that are cloned via HTTP(S) but not the ones which are cloned via SSH like for example it is supported for hooks from GitHub.

The hooks.py view should use the same mechanism to provide possible repositories that it used e.g. for GitHub (GITHUB_REPOS variable) by constructing possible URLs.

Cloning would be possible via HTTP(S) but SSH based access is more convenient e.g. in terms of expiry date since Azure DevOps PAT tokens (used to clone the git repository) can only have a max token lifetime of one year.

likemike91:
## Proposed changes

This PR should fix #4894 and also enables structured branch names (e.g. feat/localization). I tried to support as many possible URL combinations that I am aware of. What I didn't add yet is unit test but would be nice to get a first feedback on the change even without tests. I'll add them with another commit soon.

## Checklist

### DIFF
--- a/weblate/trans/tests/test_hooks.py
+++ b/weblate/trans/tests/test_hooks.py
@@ -771,7 +771,7 @@ AZURE_PAYLOAD_NEW = r"""
           "date": "2015-02-25T19:01:00Z"
         },
         "comment": "Fixed bug in web.config file",
-        "url": "https://dev.azure.com/Fabrikam-Fiber/_apis/git/repositories/278d5cd2-584d-4b63-824a-2ba458937249/commits/7d85491a4f0289f2ffcf70939b7c7160e8ce2865"
+        "url": "https://f.visualstudio.com/c/_git/ATEST/commit/33b55f7cb7e7e2453239"
       }
     ],
     "refUpdates": [
@@ -784,26 +784,26 @@ AZURE_PAYLOAD_NEW = r"""
     "repository": {
       "id": "278d5cd2-584d-4b63-824a-2ba458937249",
       "name": "ATEST",
-      "url": "https://dev.azure.com/Fabrikam-Fiber/_apis/git/repositories/278d5cd2-584d-4b63-824a-2ba458937249",
+      "url": "https://dev.azure.com/f/_apis/git/repositories/278d5cd2-584d-4b63",
       "project": {
         "id": "be9b3917-87e6-42a4-a549-2bc06a7a878f",
-        "name": "Fabrikam-Fiber-Project",
-        "url": "https://dev.azure.com/Fabrikam-Fiber/_apis/projects/be9b3917-87e6-42a4-a549-2bc06a7a878f",
+        "name": "p",
+        "url": "https://dev.azure.com/f/_apis/projects/be9b3917-87e6-42a4",
         "state": "wellFormed",
         "visibility": "unchanged",
         "lastUpdateTime": "0001-01-01T00:00:00"
       },
       "defaultBranch": "refs/heads/master",
-      "remoteUrl": "https://dev.azure.com/Fabrikam-Fiber/Fabrikam-Fiber-Project/_git/ATEST"
+      "remoteUrl": "https://dev.azure.com/f/p/_git/ATEST"
     },
     "pushedBy": {
       "displayName": "Jamal Hartnett",
       "id": "00067FFED5C7AF52@Live.com",
       "uniqueName": "fabrikamfiber4@hotmail.com"
     },
-    "pushId": 14,
+    "pushId": 1,
     "date": "2014-05-02T19:17:13.3309587Z",
-    "url": "https://dev.azure.com/Fabrikam-Fiber/_apis/git/repositories/278d5cd2-584d-4b63-824a-2ba458937249/pushes/14"
+    "url": "https://dev.azure.com/f/_apis/git/repositories/278d5cd2-584d-4b63/pushes/1"
   },
   "resourceVersion": "1.0",
   "resourceContainers": {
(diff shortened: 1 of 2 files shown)


likemike91:
@nijel I tried to fix all errors, add tests and remove warnings but the DeepSource check always fails and I have to admit that I'm not sure where to find the issues I directly introduced with my PR. May I ask you that you direct me into the right direction since I have to admit this is my first PR in the python world ;)

weblate/trans/tests/test_hooks.py
```diff
@@ -1562,15 +1643,41 @@ class AzureBackendTest(HookBackendTestCase):
...
             {
                 "branch": "master",
                 "full_name": "ATEST",
                 "repo_url": url,
-                "repos": [url],
+                "repos": [
+                    "https://dev.azure.com/f/c/_git/ATEST",
+                    "https://dev.azure.com/f/be9b3917-87e6-42a4-a549-2bc06a7a878f/"
+                    "_git/278d5cd2-584d-4b63-824a-2ba458937249"
+                    "git@ssh.dev.azure.com:v3/f/c/ATEST",
```
nijel, line 1658: This looks wrong, comma is missing before the ssh URL. To make it clear that you're relying on string concatenation, I'd suggest something like:

```suggestion
                    (
                        "https://dev.azure.com/f/be9b3917-87e6-42a4-a549-2bc06a7a878f/"
                        "_git/278d5cd2-584d-4b63-824a-2ba458937249"
                    ),
                    "git@ssh.dev.azure.com:v3/f/c/ATEST",
```

weblate/trans/views/hooks.py
```diff
@@ -71,6 +71,14 @@
     "ssh://git@{server}/{project}.git",
 )
 
+AZURE_REPOS = (
+    "https://dev.azure.com/{organization}/{project}/_git/{repository}",
+    "https://dev.azure.com/{organization}/{projectId}/_git/{repositoryId}"
```
nijel, line 76: Comma missing here:

```suggestion
    "https://dev.azure.com/{organization}/{projectId}/_git/{repositoryId}",
```

weblate/trans/tests/test_hooks.py
```diff
@@ -1562,15 +1643,45 @@ class AzureBackendTest(HookBackendTestCase):
...
+                "repos": [
+                    "https://dev.azure.com/f/c/_git/ATEST",
+                    (
+                        "https://dev.azure.com/f/be9b3917-87e6-42a4-a549-2bc06a7a878f/"
+                        "_git/278d5cd2-584d-4b63-824a-2ba458937249"
+                    ),
+                    "git@ssh.dev.azure.com:v3/f/c/ATEST",
+                    "https://f.visualstudio.com/c/_git/ATEST",
+                    "f@vs-ssh.visualstudio.com:v3/f/c/ATEST",
+                ],
```
nijel, line 1663: This still generates single URL, so either your code doesn't do what you expected or the testcase change is wrong.

weblate/trans/views/hooks.py
```diff
@@ -465,11 +473,51 @@ def pagure_hook_helper(data, request):
...
+    else:
+        repos = [http_url]
+
     return {
         "service_long_name": "Azure",
         "repo_url": http_url,
-        "repos": [http_url],
-        "branch": data["resource"]["refUpdates"][0]["name"].rsplit("/")[-1],
-        "full_name": data["resource"]["repository"]["name"],
+        "repos": [repos],
```
nijel, line 520: This creates list which contains list. You should use the list directly:

```suggestion
        "repos": repos,
```

weblate/trans/tests/test_hooks.py
nijel: Now it generates more URLs, but they do not match what is here, see the testcase failure: https://github.com/WeblateOrg/weblate/pull/4954/checks?check_run_id=1489377589

likemike91: Yeah I saw it already. I'm very sorry for creating a lot of commits that do not work but unfortunately I didn't manage to run the tests locally.

nijel: Don't worry about that, but why have you checked "Lint and unit tests pass locally with my changes." in that case? :-)

There is a guidance for that in https://docs.weblate.org/en/latest/contributing/start.html, but it might be tricky to install indeed.

likemike91: Should the tests run when I run `./rundev.sh test --failfast weblate.trans`?
Or do I need to do something differently?

nijel: Yes, that should work (though I have no experience with the dev Docker setup). You might want to run only the hook tests as the full testsuite will take quite some time: `weblate.trans.tests.test_hooks`

nijel: Looking at `rundev.sh` source, you need to first do `./rundev.sh start` and then the tests can be executed. I'm just testing the workflow whether it actually works ;-)

nijel: I've made some improvements to that in 2063025589fd14ce2661b08e018ed61f2eaefd8d, but it worked as expected.

likemike91: I finally managed to run tests locally which - obviously - helped a lot :-)

nijel:

PR #4969, opened 2020-12-03
Allow 50 characters for app store title

TheLastProject:
Google Play allows up to 50 characters for the title

## Proposed changes

Allows title.txt for Fastlane to be up to 50 characters instead of up to 30 as Google Play allows up to 50:

image

## Checklist

## Other information

I'm using this option to ping @eighthave to ensure F-Droid is also aware of the 50 character limit of Google Play.

### DIFF
--- a/weblate/formats/txt.py
+++ b/weblate/formats/txt.py
@@ -106,7 +106,7 @@ class MultiParser:
 
 class AppStoreParser(MultiParser):
     filenames = (
-        ("title.txt", "max-length:30"),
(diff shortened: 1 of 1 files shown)


eighthave:
I agree, thaks for this.  F-Droid also defaults to 50 chars for the title:
https://gitlab.com/fdroid/fdroidserver/blob/2a265cbc0b43d649a1aa0a53b77a30cd34a0681c/fdroidserver/common.py#L141

nijel:

PR #5004, opened 2020-12-08
Billing: fix typo commerial

rffontenelle:
Page shows "Commerial" but I assume the correct is "Commercial".


nijel (approved):
Thanks for spotting!

PR #5005, opened 2020-12-08
Fix wrong plural in projects

rffontenelle:
This project**s** does not -> this project does not


nijel (approved):
Thanks for fixing this.

PR #5042, opened 2020-12-15
Modify tasks.py to Support MariaDB/MySQL Backup

TheWitness:
## Proposed changes

* This change allows the creation of backup files for MariaDB and MySQL
* The password is hidden through the use of the --defaults-extra-file option
* Will look at augmenting unit testing

## Checklist

Not quite ready to merge yet.  Just saving progress.

## Other information

It'll take me a while to get the rest of the boxes ticked.


weblate/utils/tasks.py
```diff
@@ -83,20 +83,42 @@ def database_backup():
...
+            cmd += ["--file", data_dir("backups", "database.sql.gz")]
+            cmd += ["--compress", "6"]
+        else:
+            cmd += ["--file", data_dir("backups", "database.sql")]
+    elif database["ENGINE"] == "django.db.backends.mysql":
+        options = data_dir("home") + "/.my.cnf"
+
+        with open(options, "w") as handle:
+            handle.write("[mysqldump]\n")
+            handle.write("password = '" + database["PASSOWRD"] + "'\n")
```
nijel, line 104: I think using MYSQL_PWD environment variable is a better approach than creating/updating a file with configuration on every run. When creating the file, you would have to deal with non-existing directories as well (this is what fails in the CI).

TheWitness: Funny how that's not in the MySQL dump man page.  I was trying to avoid it, but at least it would be consistent with Postgresql.

Was reading this.

https://dev.mysql.com/doc/refman/8.0/en/password-security-user.html

BTW, impressed by your prechecks/CI.  Really cool stuff.

nijel: Consistency with PostgreSQL is not really needed in this, I just think it's less error prone to pass it in environment instead of creating file to hold it.

TheWitness:
I'll go ahead and change to use the deprecated password environment variable.  But in the future, where would this credential file be properly stored?

TheWitness:
I'm not sure about this latest set of errors.  Not sure how to interpret in their entirety.

weblate/utils/tasks.py
```diff
@@ -83,25 +83,45 @@ def database_backup():
...
+        env = get_clean_env({"MYSQL_PWD": database["PASSWORD"]})
     else:
-        cmd += ["--file", data_dir("backups", "database.sql")]
+        return
 
     try:
         subprocess.run(
             cmd,
-            env=get_clean_env({"PGPASSWORD": database["PASSWORD"]}),
+            env,
```
nijel, line 124:
```suggestion
            env=env,
```

Without this you are passing position based arg.

weblate/utils/tasks.py
```diff
@@ -83,25 +83,45 @@ def database_backup():
...
-    if database["HOST"]:
-        cmd += ["--host", database["HOST"]]
-    if database["PORT"]:
-        cmd += ["--port", database["PORT"]]
-    if database["USER"]:
-        cmd += ["--username", database["USER"]]
-    if settings.DATABASE_BACKUP == "compressed":
-        cmd += ["--file", data_dir("backups", "database.sql.gz")]
-        cmd += ["--compress", "6"]
+    if database["ENGINE"] == "django.db.backends.postgresql":
```
nijel, line 86:
```suggestion
    env = get_clean_env()
    if database["ENGINE"] == "django.db.backends.postgresql":
```

Call the function just once and update the env later.

weblate/utils/tasks.py
```diff
@@ -83,25 +83,45 @@ def database_backup():
...
+            cmd += ["--port", database["PORT"]]
+        if database["USER"]:
+            cmd += ["--user", database["USER"]]
+        if settings.DATABASE_BACKUP == "compressed":
+            cmd += ["--result-file", data_dir("backups", "database.sql.gz")]
+            cmd += ["--compress"]
+        else:
+            cmd += ["--result-file", data_dir("backups", "database.sql")]
+
+        env = get_clean_env({"MYSQL_PWD": database["PASSWORD"]})
```
nijel, line 117:
```suggestion
        env["MYSQL_PWD"] = database["PASSWORD"]
```

weblate/utils/tasks.py
```diff
@@ -83,25 +83,45 @@ def database_backup():
...
+            cmd += ["--port", database["PORT"]]
+        if database["USER"]:
+            cmd += ["--username", database["USER"]]
+        if settings.DATABASE_BACKUP == "compressed":
+            cmd += ["--file", data_dir("backups", "database.sql.gz")]
+            cmd += ["--compress", "6"]
+        else:
+            cmd += ["--file", data_dir("backups", "database.sql")]
+
+        env = get_clean_env({"PGPASSWORD": database["PASSWORD"]})
```
nijel, line 101:
```suggestion
        env["PGPASSWORD"] = database["PASSWORD"]
```

TheWitness:
Okay, just made another pass.  Python not first or second language, but it's not Greek either.

TheWitness:
Well, getting closer now.  What next?

nijel:
Looks the mysqldump is stuck waiting for some input. I've changed the code to not provide a valid input to the dump command in 4b7a7a9978, that should make the actual error visible.

TheWitness:
Okay, getting back to this.  Did some more research.  Worked from the command line natively, but I can tell you that the --compress option in MariaDB/MySQL does not actually compress the output.  So, switched to gzip.

TheWitness:
My guess the mysql tests are failing is likely due to a missing pip dependency.  I had to install the pip mysqlclient package to get things to work.  Is that in the pip requirements file?

nijel:
I don't think it's missing package. The tests hang, so it seems like the mysqdump is waiting for some input. Either passing the password via environment didn't work and it is waiting for it, or there is some other issue in executing that. Temporarily removing the stdout/stderr args from `subprocess.run` call might reveal the actual problem as the mysqdump output would be displayed in the CI logs.

TheWitness:
Okay, I'll do that this morning.

TheWitness:
Interesting, it's working on MariaDB 10.5.8, I wonder if Oracle quietly removed support of MYSQL_PWD.  If this is the case, where should I be storing the credential ini file?  Will the Celery process have the correct permissions?  Here is the screen shot of my production system.  Let me know.

image

TheWitness:
In any event, I commit the removal and then review the stdout from the mysql test.

TheWitness:
Nothing new in the output from my vantage point anyway.

weblate/utils/tasks.py
```diff
@@ -83,27 +83,46 @@ def database_backup():
...
+        cmd = ["mysqldump", "--databases", database["NAME"]]
+
+        if database["HOST"]:
+            cmd += ["--host", database["HOST"]]
+        if database["PORT"]:
+            cmd += ["--port", database["PORT"]]
+        if database["USER"]:
+            cmd += ["--user", database["USER"]]
+        if settings.DATABASE_BACKUP == "compressed":
+            cmd += ["|", "gzip", ">", data_dir("backups", "database.sql.gz")]
```
nijel, line 114: This doesn't work - the process is not executed in shell and these parameters are passed directly to `mysqldump`. Also removing this seems to unfreeze the test (though I'm not sure how is it related).

weblate/utils/tasks.py
```diff
@@ -83,27 +83,46 @@ def database_backup():
...
+            cmd += ["--username", database["USER"]]
+        if settings.DATABASE_BACKUP == "compressed":
+            cmd += ["--file", data_dir("backups", "database.sql.gz")]
+            cmd += ["--compress", "6"]
+        else:
+            cmd += ["--file", data_dir("backups", "database.sql")]
+
+        env["PGPASSWORD"] = database["PASSWORD"]
+    elif database["ENGINE"] == "django.db.backends.mysql":
+        cmd = ["mysqldump", "--databases", database["NAME"]]
```
nijel, line 105: It should probably include `--column-statistics=0` otherwise `mysqldump` fails with older servers (which are used in the CI environment), see https://serverfault.com/q/912162

nijel (commented):
I've played with this a bit and made the tests pass: https://github.com/nijel/weblate/runs/1625486846

My changes however completely discard the compression option, see https://github.com/nijel/weblate/commit/536d8c4a3593b620dfebdfbc7ee3f8c6bc895317 and https://github.com/nijel/weblate/commit/4e50aed9487224223ae9cf9bcbd495c4af8e4ef3

weblate/utils/tasks.py
TheWitness: Okay, so, compression off the table without a wrapper then, or just document that compression unsupported, or add shell=True?

weblate/utils/tasks.py
TheWitness: The --column-statistics is going to be a problem as that is unsupported in MariaDB.  However, so long as the client and server are from the same supplier, it won't matter.

weblate/utils/tasks.py
TheWitness: I've just added the shell option.  I don't think it's a big security issue in that all the parameters are under the admins control, and since the password, which is usually where you get into trouble, is passed through the environment, it should be pretty safe.

Either way, I defer to you as to how/if you want to support compression on MariaDB/MySQL.

TheWitness:
There is another topic, and that is backup generations.  Normally, when I back things up, I'll keep X copies of the backup just in case I only experience the problem X-2 days after it happened.  I guess the question is, do you leave that to something like logrotate, a snapshot (assuming the backup location supports such a thing), or some Weblate process to keep them around, or just leverage the backup service.  The later would be attractive in most cases.

nijel:
There is a backup service built into Weblate, the SQL dump is part of what is being backed up, see https://docs.weblate.org/en/latest/admin/backup.html

weblate/utils/tasks.py
```diff
@@ -123,8 +123,11 @@ def database_backup():
         subprocess.run(
             cmd,
             env=env,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
             stdin=subprocess.DEVNULL,
             check=True,
+            shell=True,
```
nijel, line 130: You must pass single string when executing via shell and take care of escaping. I'd rather avoid using shell here and do the compression separately if needed.

TheWitness: Lovely, too many languages, and not enough Python.  Okay, so then make it a Documentation issue "Compression not supported when using MariaDB/MySQL?

nijel: Yes, makes sense. It can be implemented later in case somebody needs it.

TheWitness: Should I take a shot at the doco?  Separate repo?

TheWitness: Found the doco.  I've taken copious notes on MariaDB and CentOS 7.  I could share.

weblate/utils/tasks.py
nijel: But this problem exists in the CI environment. I have no clue whether it's easily possible to install matching tools there, right now we use the ones shipped in the image (what seems to be MySQL 8.0.22 right now).

TheWitness: But that problem will not affect a version 8 mysqldump cli from dumping a mysql 8 database.  It would affect you attempting to dump say a MySQL 5.7-- or MariaDB 10.5-- database using a version 8 mysqldump cli.  So, there should not be a problem.

TheWitness:
I made a few documentation updates to help people using MySQL.  There are other lessons learned from my too.  Most of that is due to the fact I'm a LAMP guy and not a PIP guy, though I'm getting better at it.

TheWitness:
The output from the blacken-docs is really insufficient.  For example, I was failing a white-space test, easy enough, modify my vim configuration to trim trailing which space from lines.  But it would have been helpful to share what line in the file failed.

Now, I'm failing the blacken-docs, but again no detail on where the failure occurred, or what rule was violated.

weblate/utils/tasks.py
nijel: That probably makes sense, but still we need to make it work in the CI environment as well. Can you please try rebasing on top of 45ed219f7c86fa7d905dee2678cd26a559967384 whether it helps?

docs/admin/install.rst
```diff
@@ -413,14 +413,77 @@ Weblate requires MySQL at least 5.7.8 or MariaDB at least 10.2.7.
...
+            # Database password
+            "PASSWORD": "password",
+            # Set to empty string for localhost
+            "HOST": "127.0.0.1",
+            # Set to empty string for default
+            "PORT": "3306",
+            # In case you wish to use additional
+            # connection options
+            "OPTIONS": {
+            },
```
nijel, line 476:
```suggestion
            "OPTIONS": {},
```

I think this is what blacken-docs complains on.

TheWitness: Okay

nijel:
> The output from the blacken-docs is really insufficient. For example, I was failing a white-space test, easy enough, modify my vim configuration to trim trailing which space from lines. But it would have been helpful to share what line in the file failed.
> 
> Now, I'm failing the blacken-docs, but again no detail on where the failure occurred, or what rule was violated.

The best approach is to run it locally and will format the code correctly :-). See https://docs.weblate.org/en/latest/contributing/code.html#coding-standard

TheWitness:
As I expected Michel, a very well managed project.  Taking your lead over by us.

nijel:
Hmm, the tests on both MySQL and MariaDB still hang for some reason. Were you able to run them locally?

weblate/utils/tasks.py
```diff
@@ -83,25 +83,44 @@ def database_backup():
...
+            cmd += ["--username", database["USER"]]
+        if settings.DATABASE_BACKUP == "compressed":
+            cmd += ["--file", data_dir("backups", "database.sql.gz")]
+            cmd += ["--compress", "6"]
+        else:
+            cmd += ["--file", data_dir("backups", "database.sql")]
+
+        env["PGPASSWORD"] = database["PASSWORD"]
+    elif database["ENGINE"] == "django.db.backends.mysql":
+        cmd = ["mysqldump", "--databases", database["NAME"]]
```
nijel, line 105: Okay, the problem is here - the `--databasess` has to be the last parameter passed to mysqldump. Anything after that is considered to be a database name...

```suggestion
        cmd = ["mysqldump", "--result-file", data_dir("backups", "database.sql")]
```

weblate/utils/tasks.py
```diff
@@ -83,25 +83,44 @@ def database_backup():
...
+
+        if database["HOST"]:
+            cmd += ["--host", database["HOST"]]
+        if database["PORT"]:
+            cmd += ["--port", database["PORT"]]
+        if database["USER"]:
+            cmd += ["--user", database["USER"]]
+
+        cmd += ["--result-file", data_dir("backups", "database.sql")]
```
nijel, line 114:
```suggestion
        cmd += ["--databases", database["NAME"]]
```

weblate/utils/tasks.py
```diff
@@ -83,25 +83,44 @@ def database_backup():
...
+            cmd += ["--username", database["USER"]]
+        if settings.DATABASE_BACKUP == "compressed":
+            cmd += ["--file", data_dir("backups", "database.sql.gz")]
+            cmd += ["--compress", "6"]
+        else:
+            cmd += ["--file", data_dir("backups", "database.sql")]
+
+        env["PGPASSWORD"] = database["PASSWORD"]
+    elif database["ENGINE"] == "django.db.backends.mysql":
+        cmd = ["mysqldump", "--result-file", data_dir("backups", "database.sql")]
```
nijel, line 105: I have probably finally found out why the tests do hang. It's not the single backup test, but rather the UI test which triggers it as well (that's why the tests did pass before, I was looking at wrong ones).

The problem is that that test is in the middle of the transaction and mysqldump waits for the transaction to complete and the tests waits for mysqldump, causing nice deadlock.

Disabling the locking on mysqldump seems to address this:

```suggestion
        cmd = [
            "mysqldump",
            "--result-file",
            data_dir("backups", "database.sql"),
            "--single-transaction",
            "--skip-lock-tables",
        ]
```

nijel:

TheWitness:
Thanks for merging.  I was going to update, but been very busy.  Glad to contribute.

Issue #5040, opened 2020-12-14
XWiki translation format are marked as missing when adding a new language


surli:
**Describe the bug**

In Weblate 4.3.2, when using XWiki Page Properties format, translations are not properly handled when adding a new language. The files are committed with the translations marked as missing, even if they have been translated already. See for example: https://github.com/xwiki-contrib/application-discussions/pull/1

**To Reproduce the bug**

1. Create a component with XWiki Java Properties format
2. Add a new language to the component created and perform some translations
3. Commit the changes
4. Translations are not committed: keys are marked as missing 

**Expected behavior**

Translations should have been committed.

surli:
@nijel creating the issue for me: I'll investigate to fix it.

nijel:
Should be fixed by #5050.

PR #5050, opened 2020-12-15
Fix problems when saving XWiki properties.

surli:
## Proposed changes

Those changes aims at fixing #5040.
Basically the idea is to ensure that in case of a new language the unit is properly retrieved and not marked as missing. 

## Checklist

- [ ] I have described the changes in the commit messages

### DIFF
--- a/weblate/formats/tests/test_formats.py
+++ b/weblate/formats/tests/test_formats.py
@@ -869,7 +869,7 @@ class XWikiPropertiesFormatTest(PropertiesFormatTest):
         with open(TEST_XWIKI_PROPERTIES_NEW_LANGUAGE) as handle:
             expected = handle.read()
 
-        self.assertEqual(expected + '\n', newdata)
+        self.assertEqual(expected + "\n", newdata)
 
 
 class XWikiPagePropertiesFormatTest(PropertiesFormatTest):
--- a/weblate/formats/tests/test_formats.py
+++ b/weblate/formats/tests/test_formats.py
@@ -852,9 +852,10 @@ class XWikiPropertiesFormatTest(PropertiesFormatTest):
     def test_new_language(self):
         self.maxDiff = None
         out = os.path.join(self.tempdir, f"test_new_language.{self.EXT}")
-        self.FORMAT.add_language(out, Language.objects.get(code="cs"), self.BASE)
+        language = Language.objects.get(code="cs")
+        self.FORMAT.add_language(out, language, self.BASE)
         template_storage = self.parse_file(self.FILE)
-        new_language = self.FORMAT(out, template_storage, Language.objects.get(code="cs"))
+        new_language = self.FORMAT(out, template_storage, language.code)
         unit, add = new_language.find_unit("job.status.success")
         self.assertTrue(add)
         unit.set_target("Fait")
--- a/weblate/formats/tests/test_formats.py
+++ b/weblate/formats/tests/test_formats.py
@@ -869,7 +869,7 @@ class XWikiPropertiesFormatTest(PropertiesFormatTest):
         with open(TEST_XWIKI_PROPERTIES_NEW_LANGUAGE) as handle:
             expected = handle.read()
 
-        self.assertEqual(expected, newdata)
+        self.assertEqual(expected + '\n', newdata)
 
 
 class XWikiPagePropertiesFormatTest(PropertiesFormatTest):
--- a/weblate/formats/tests/test_formats.py
+++ b/weblate/formats/tests/test_formats.py
@@ -89,6 +89,7 @@ TEST_HE_CUSTOM = get_test_file("he-custom.po")
 TEST_HE_SIMPLE = get_test_file("he-simple.po")
 TEST_HE_THREE = get_test_file("he-three.po")
 TEST_XWIKI_PROPERTIES = get_test_file("xwiki.properties")
+TEST_XWIKI_PROPERTIES_NEW_LANGUAGE = get_test_file("xwiki_new_language.properties")
 TEST_XWIKI_PAGE_PROPERTIES = get_test_file("XWikiPageProperties.xml")
 TEST_XWIKI_PAGE_PROPERTIES_SOURCE = get_test_file("XWikiPagePropertiesSource.xml")
 TEST_XWIKI_FULL_PAGE = get_test_file("XWikiFullPage.xml")
@@ -847,6 +848,27 @@ class XWikiPropertiesFormatTest(PropertiesFormatTest):
     NEW_UNIT_MATCH = b"\nkey=Source string\n"
     EXPECTED_FLAGS = ""
     EDIT_TARGET = "[{0}] تىپتىكى خىزمەتنى باشلاش"
+
+    def test_new_language(self):
+        self.maxDiff = None
+        out = os.path.join(self.tempdir, f"test_new_language.{self.EXT}")
+        self.FORMAT.add_language(out, Language.objects.get(code="cs"), self.BASE)
+        template_storage = self.parse_file(self.FILE)
+        new_language = self.FORMAT(out, template_storage, Language.objects.get(code="cs"))
+        unit, add = new_language.find_unit("job.status.success")
+        self.assertTrue(add)
+        unit.set_target("Fait")
+        new_language.add_unit(unit.unit)
+        new_language.save()
+
+        # Read new content
+        with open(out) as handle:
+            newdata = handle.read()
+
+        with open(TEST_XWIKI_PROPERTIES_NEW_LANGUAGE) as handle:
+            expected = handle.read()
+
+        self.assertEqual(expected, newdata)
 
 
 class XWikiPagePropertiesFormatTest(PropertiesFormatTest):
--- a/weblate/formats/ttkit.py
+++ b/weblate/formats/ttkit.py
@@ -1515,18 +1515,39 @@ class XWikiPropertiesFormat(PropertiesBaseFormat):
     autoload = ("*.properties",)
     new_translation = "\n"
 
+    # Ensure that not translated units are saved too as missing properties and
+    # comments are preserved as in the original source file.
     def save_content(self, handle):
         current_units = self.all_units
+        store_units = self.store.units
+
+        # We empty the store units since we want to control what we'll serialize
         self.store.units = []
-        # Ensure that not translated units are saved too as missing properties.
+
         for unit in current_units:
-            if unit.unit is None:
-                if not unit.has_content():
-                    unit.unit = unit.mainunit
+            # If the translation unit is missing and the current unit is not
+            # only about comment.
+            if unit.unit is None and unit.has_content():
+
+                # We first check if the unit has not been translated as part of a
+                # new language: in that case the unit is not linked yet.
+                found_store_unit = None
+                for store_unit in store_units:
+                    if unit.context == store_unit.name:
+                        found_store_unit = store_unit
+
+                # If we found a unit for same context not linked, we just link it.
+                if found_store_unit is not None:
+                    unit.unit = found_store_unit
+                # else it's a missing unit: we need to mark it as missing.
                 else:
                     missingunit = self.find_unit(unit.context, unit.source)[0]
                     unit.unit = missingunit.unit
                     unit.unit.missing = True
+            # if the unit was only a comment, we take back the original source file unit
+            # to avoid any change.
+            elif not unit.has_content():
+                unit.unit = unit.mainunit
             self.add_unit(unit.unit)
 
         self.store.serialize(handle)
--- a/weblate/trans/tests/data/xwiki_new_language.properties
+++ b/weblate/trans/tests/data/xwiki_new_language.properties
@@ -67,7 +67,8 @@
 ## until 10.1
 #######################################
 
-job.status.success=Fait.
+job.status.success=Fait
 ### Missing: job.status.canceled=Canceled.
 ### Missing: job.status.error=Failed.
 #@deprecatedend
+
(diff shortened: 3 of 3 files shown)


nijel:
Can you please add tests to cover the new conditions you've introduced?

surli:
> Can you please add tests to cover the new conditions you've introduced?

Sure I pushed quickly yesterday to ensure all tests were passing. Need to write a test properly now.

surli:
@nijel I might need your help about the test I'm writing. I don't understand the stack trace it's generated right now, I do reproduce locally but even when debugging I don't understand what happens: 
```
ERROR: test_new_language (weblate.formats.tests.test_formats.XWikiPropertiesFormatTest)
  File "/home/runner/work/weblate/weblate/weblate/formats/tests/test_formats.py", line 862, in test_new_language
    new_language.save()
  File "/home/runner/work/weblate/weblate/weblate/formats/ttkit.py", line 298, in save
    self.save_atomic(self.storefile, self.save_content)
  File "/home/runner/work/weblate/weblate/weblate/formats/base.py", line 301, in save_atomic
    callback(temp)
  File "/home/runner/work/weblate/weblate/weblate/formats/ttkit.py", line 1519, in save_content
    missingunit = self.find_unit(unit.context, unit.source)[0]
  File "/home/runner/work/weblate/weblate/weblate/formats/base.py", line 284, in find_unit
    return self._find_unit_template(context)
  File "/home/runner/work/weblate/weblate/weblate/formats/base.py", line 260, in _find_unit_template
    ttkit_unit = deepcopy(template_ttkit_unit)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 270, in _reconstruct
    state = deepcopy(state, memo)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 230, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 296, in _reconstruct
    value = deepcopy(value, memo)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 270, in _reconstruct
    state = deepcopy(state, memo)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 230, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 270, in _reconstruct
    state = deepcopy(state, memo)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 230, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 205, in _deepcopy_list
    append(deepcopy(a, memo))
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 270, in _reconstruct
    state = deepcopy(state, memo)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 230, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 270, in _reconstruct
    state = deepcopy(state, memo)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 230, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 270, in _reconstruct
    state = deepcopy(state, memo)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 230, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/copy.py", line 151, in deepcopy
    copier = getattr(x, "__deepcopy__", None)
  File "/home/runner/work/weblate/weblate/weblate/utils/stats.py", line 206, in __getattr__
    self.calculate_item(name)
  File "/home/runner/work/weblate/weblate/weblate/utils/stats.py", line 619, in calculate_item
    result += getattr(translation.stats, item)
  File "/home/runner/work/weblate/weblate/weblate/utils/stats.py", line 206, in __getattr__
    self.calculate_item(name)
  File "/home/runner/work/weblate/weblate/weblate/utils/stats.py", line 513, in calculate_item
    stats = translation.unit_set.filter_type(item).aggregate(
  File "/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/home/runner/work/weblate/weblate/weblate/trans/models/unit.py", line 93, in filter_type
    raise ValueError(f"Unknown filter: {rqtype}")
ValueError: Unknown filter: __deepcopy__
```
When I debug the unit it's trying to deepcopy seems fine so not sure why it's crashing there.

nijel:
It looks like the unit adding code got reference to Weblate stats object which is not copyable...

surli:
> the unit adding code got reference to Weblate stats object

ok I don't know about those Weblate stats object, but that's weird I'm using the `AutoFormatTest#parse_file` method to obtain the template store. And if I'm correct the unit which is copied comes from this store.

weblate/formats/ttkit.py
```diff
@@ -1490,18 +1490,39 @@ class XWikiPropertiesFormat(PropertiesBaseFormat):
...
+            # If the translation unit is missing and the current unit is not
+            # only about comment.
+            if unit.unit is None and unit.has_content():
+
+                # We first check if the unit has not been translated as part of a
+                # new language: in that case the unit is not linked yet.
+                found_store_unit = None
+                for store_unit in store_units:
+                    if unit.context == store_unit.name:
+                        found_store_unit = store_unit
```
nijel, line 1512: This will perform horribly, because you are iterating N*N times for N units.

weblate/formats/ttkit.py
```diff
@@ -1490,18 +1490,39 @@ class XWikiPropertiesFormat(PropertiesBaseFormat):
...
+
+                # If we found a unit for same context not linked, we just link it.
+                if found_store_unit is not None:
+                    unit.unit = found_store_unit
+                # else it's a missing unit: we need to mark it as missing.
                 else:
                     missingunit = self.find_unit(unit.context, unit.source)[0]
```
nijel, line 1519: The problem with the tests is caused here. Because of the cleaned up units up there, this doesn't find a matching unit and tries to build a new one. The question is whether it is desired. Another question is why the component is referenced there - at this level only translate-toolkit objects should be involved and nothing from Weblate.

weblate/formats/ttkit.py
surli: Yeah I know that's not great but I didn't find any better for doing this for now.. Note that the loop only occur if there's no pounit attached and it's not a comment. 
Clearly the bug is here: apparently when adding a new language the pounit are not properly attached to the units when I loop over them. It might be related to the fact that all_units is cached, so I'm getting an old version of the units, but when I looked for another way to perform this loop I didn't find it. 

Taking any advice to better handle this :)

weblate/formats/ttkit.py
surli: > The problem with the tests is caused here. Because of the cleaned up units up there, this doesn't find a matching unit and tries to build a new one. The question is whether it is desired.

So that's actually part of the old code, and yes it's desired: whenever the translation units are not present I want them to be copied from the template store and marked as missing. 
Another way of doing it I thought about would be to generate the entire file with the missing translations when adding a language, the problem is that the method for creating a new language file is a static class method. 

> Another question is why the component is referenced there - at this level only translate-toolkit objects should be involved and nothing from Weblate.

Not sure to follow there.

nijel:
Looking at this again, I'm not convinced that this is the right place to do such transformations. You modify unit lists, but there is some caching in the class and it's super-easy to make something wrong. The similar thing is handled in resx format using an [addon](https://github.com/WeblateOrg/weblate/blob/master/weblate/addons/resx.py). It is then responsible for a single task - keeping the translation files in sync with source. The addon is triggered after upstream repo was updated, so it runs before Weblate parses updated files and can modify them. That would be more effective approach as that would be trigerred just for the updates and not on every save, but also that would be a lot of work to move the current implementation to the addon.

weblate/formats/tests/test_formats.py
```diff
@@ -848,6 +849,27 @@ class XWikiPropertiesFormatTest(PropertiesFormatTest):
     EXPECTED_FLAGS = ""
     EDIT_TARGET = "[{0}] تىپتىكى خىزمەتنى باشلاش"
 
+    def test_new_language(self):
+        self.maxDiff = None
+        out = os.path.join(self.tempdir, f"test_new_language.{self.EXT}")
+        self.FORMAT.add_language(out, Language.objects.get(code="cs"), self.BASE)
+        template_storage = self.parse_file(self.FILE)
+        new_language = self.FORMAT(out, template_storage, Language.objects.get(code="cs"))
```
nijel, line 857: This is reason of the test crash you pass Langauge object instead of language code to the translate toolkit:

```suggestion
        new_language = self.FORMAT(out, template_storage, Language.objects.get(code="cs").code)
```

This is just to show the problem, I will post improvement of whole code block the in separate comment...

nijel, line 857:
```suggestion
        language = Language.objects.get(code="cs")
        self.FORMAT.add_language(out, language, self.BASE)
        template_storage = self.parse_file(self.FILE)
        new_language = self.FORMAT(out, template_storage, language.code)
```

Complete fix - get the language just once and reuse it.

surli: thanks!

surli:
> The addon is triggered after upstream repo was updated, so it runs before Weblate parses updated files and can modify them.

I don't understand how it would help me there: on my case it's the contrary: I want the file to be saved with the missing translation units, so it needs to happen after translation have been made, not when the upstream repo is modified. 
Now I could maybe bind it to a pre-commit event, but this leads to another question: is that possible to make an addon mandatory for a translation format? Because if not that's not a proper solution for me: I want those addons to be always available for my format. 
I think for me the cleanest way to deal with my problem would be by being able to generate a new translation language file directly from the source file and not only linked to the format. So not from a static method but from an instance method. But that doesn't fit the current architecture.

surli:
@nijel apparently latest failure shows a linebreak problem: 
> AssertionError: '# --[3543 chars]\n### Missing: job.status.error=Failed.\n#@deprecatedend\n\n' != '# --[3543 chars]\n### Missing: job.status.error=Failed.\n#@deprecatedend\n\n\n'

I don't recall: do you run the CI with latest version of translation library? It might mean that my fix on it wasn't good enough.

nijel:
Did you look at the test failure? Does it depend on https://github.com/translate/translate/pull/4206?

nijel:
@surli Did you have time to look at the test failures?

surli:
> Did you look at the test failure? Does it depend on translate/translate#4206?

I did not check precisely yet, but yes it might depend on it: that was my question if the latest version of translate was using on the CI. I'll try to check it back today.

surli:
> I want to release 4.4.1 on Tuesday or Wednesday next week, if you want this to be included, it should be fixed ;-).

Thanks for the notice, sorry for being late on it...

surli:
@nijel so apparently the only check which is failing concerns changes I didn't perform here.

nijel:

PR #5065, opened 2020-12-18
Add Cloudron installation method

nebulade:
Cloudron has a 1-click installation package for Weblate and we make sure all instances are automatically kept up-to-date.
The package is at https://git.cloudron.io/cloudron/weblate-app where we also maintain selenium e2e tests, which have to succeed before we push an update. Usually updates happen withing a few days from upstream release.

### DIFF
--- a/docs/admin/install.rst
+++ b/docs/admin/install.rst
@@ -35,6 +35,7 @@ Depending on your setup and experience, choose an appropriate installation metho
 
 * :doc:`install/openshift`
 * :doc:`install/kubernetes`
+* `Installing on Cloudron <https://www.cloudron.io/store/org.weblate.cloudronapp.html>`_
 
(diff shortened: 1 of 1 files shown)


nijel:
I don't think we should list third-party packages among installation methods maintained by us. It makes more sense to have a dedicated section for these and include more providers (for example https://bitnami.com/stack/weblate). Also, it should clearly state that we have no control on these.

nijel:
Okay, it turns out we already have such section in the docs, please add Cloudron specific information to docs/admin/deployments.rst so that it's listed with other similar services (see https://docs.weblate.org/en/latest/admin/deployments.html).

nebulade:
Oh indeed! I didn't see that, thanks for your investigation. I will close this and create a new one for the other page.

PR #5080, opened 2020-12-21
Add Cloudron installation method

nebulade:
Cloudron has a 1-click installation package for Weblate and we make sure all instances are automatically kept up-to-date.
The package is at https://git.cloudron.io/cloudron/weblate-app where we also maintain selenium e2e tests, which have to succeed before we push an update. Usually updates happen withing a few days from upstream release.

Previous attempt at the wrong page https://github.com/WeblateOrg/weblate/pull/5065


nijel:

PR #5100, opened 2020-12-25
Spelling: Hamburger menu

comradekingu:


Fat-Zer:
The proposed text doesn't give a good enough hint what a "*hamburger menu*" is. IMHO it's a relatively new term especially for non-english speakers.

If I may, I'd rather suggest something like this:

* **label**: Always use hamburger menu
* **help**: Always use the hamburger menu (a button on the top right corner with a drop-down list) instead of a normal full-width version of the menu, even if there is enough space on the screen for the last one.

comradekingu:
@Fat-Zer I am not saying my suggestion is the best here, but if the user doesn't know what a hamburger menu is, repeating it in the help text isn't going to do anything.

It is nice that there is a indication of where it is, "drop down" somewhat gives away function.
"for the last one" is a conditional element that can be scrapped.

nijel:

PR #5124, opened 2021-01-01
Profile name used as profile button

comradekingu:
Not sure that works, but you get the idea.

### DIFF
--- a/weblate/templates/base.html
+++ b/weblate/templates/base.html
@@ -233,9 +233,8 @@ var _rollbarConfig = {
             <b class="caret"></b>
           </a>
           <ul class="dropdown-menu">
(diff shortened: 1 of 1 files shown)


weblate/templates/base.html
```diff
@@ -233,9 +233,8 @@
             <b class="caret"></b>
           </a>
           <ul class="dropdown-menu">
-            <li class="dropdown-header" id="profile-name"><b>{{ user }}</b><br />@{{ user.username }}</li>
+            <li><a class="button" id="profile-name"> href "{% url 'user_page' user=user.username %}"><b>{{ user }}</b><br />@{{ user.username }}</a></li>
```
nijel, line 236: This seems to mess up HTML markup and removes some used styling. But generally it makes sense, the Profile link is really confusing.

PR #5227, opened 2021-01-18
Trans: Disable key locking in select_for_update

nijel:
## Proposed changes

This reduces locking congestion on doing updates as it holds only row
level locks and not table level ones for primary key.

Needs Django 3.2 (not yet released), see https://code.djangoproject.com/ticket/30375 and https://github.com/django/django/commit/a4e6030904df63b3f10aa0729b86dc6942b0458e

## Checklist


nijel:
Can be merged once we are actually testing the changes for upcoming Django release, see https://github.com/WeblateOrg/weblate/pull/5280.

PR #5248, opened 2021-01-21
Support Lua format strings. Fixes #4289.

### LINKED ISSUE #4289: Support Lua format
xgettext supports Lua, and adds the flag `#, lua-format` appropriately. 
Unfortunately, Weblate does not support this format.

Add Lua format to Weblate.

**Additional context**
Lua format is similar to C format.
Lua string format: [documentation 1](http://lua-users.org/wiki/StringLibraryTutorial) [documentation 2](https://www.lua.org/pil/20.html).

UncombedCoconut:
## Proposed changes

Hello, I'd like to fix issue #4289 (Support Lua format) as my first contribution.
For Weblate's purposes, a Lua format string is a C format string. (The differences are that some features supported by `printf()` in C are errors in Lua, and the allowed type codes -- which Weblate does not police in any case -- include 'q'.)

## Checklist

## Other information

Because of the close relationship between Lua and C format strings, I believe we can be content with a very simple unit test (same as the corresponding feature for Perl, for instance).

As a new contributor who'd previously only used hosted Weblate, I may have missed a step in the end-to-end delivery of this feature. (Hopefully not.)

Because weblate.checks.same makes an apparent effort to maintain parity with weblate.checks.format, I'm opening a separate PR to ensure it stays caught up.

All the best,
Justin

### DIFF
--- a/docs/admin/checks.rst
+++ b/docs/admin/checks.rst
@@ -79,13 +79,13 @@ Here is a list of flags currently accepted:
     Mark this string as a variant of string with matching source. See :ref:`variants`.
 ``regex:REGEX``
     Regular expression to match translation, see :ref:`check-regex`.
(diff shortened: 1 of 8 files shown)


nijel:

PR #5249, opened 2021-01-21
Refactor checks.same to keep feature parity with checks.format.

UncombedCoconut:
## Proposed changes

Hello,
I'm [trying](https://github.com/WeblateOrg/weblate/pull/5248) to support a new programming language (Lua, flagged as `#, lua-format`), and I noticed weblate.checks.same was aware of several format string flavors in its own right.
To avoid making this worse by making weblate.checks.format better, I propose a simple refactoring so the "same" check automatically picks up enhancements like this.

## Checklist

## Other information

Either this is a good refactor (and some of the above criteria are inapplicable), or it's problematic and we can take another approach.
Honestly, it's probably not too bad to merge #5248 and leave this alone, but I felt like I had to provide an alternative.

All the best,
Justin

### DIFF
--- a/weblate/checks/same.py
+++ b/weblate/checks/same.py
@@ -25,12 +25,7 @@ from django.utils.translation import gettext_lazy as _
 
 from weblate.checks.base import TargetCheck
 from weblate.checks.data import IGNORE_WORDS
-from weblate.checks.format import (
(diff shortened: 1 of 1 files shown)


nijel:

PR #5311, opened 2021-02-01
ci: Switch back to using pre-commit integration over local

nijel:
## Proposed changes

This allows to use pre-commit.ci service which can provide fixes to pull
requests, what is really nice to new contributors.

We loose tight control over versions of flake8 plugins. In case this
will cause troubles, we might introduce automation for that, but still
that would be less work than implementing the auto fixer.

## Checklist


nijel:
Okay, this doesn't work that well - the job on pre-commit.ci timeouts after 3 minutes as that is apparently not enough time to run flake8 on our codebase.

In the end we have to wait for additional features on their side:

* https://github.com/pre-commit-ci/issues/issues/13 would allow running local tools
* https://github.com/pre-commit-ci/issues/issues/7 would allow to skip certain jobs

Until then, no autofixing via pre-commit.ci seems possible.

PR #5392, opened 2021-02-10
Glossary documentation reworked

comradekingu:
Changed one :guilabel:

Not sure if "Terms to be avoided" is better than "Forbidden", and that "forbidden" could be "avoid"?


docs/user/glossary.rst
```diff
@@ -53,12 +53,12 @@ Forbidden translations
...
-You can flag certain glossary terms as forbidden - those should not be used
-while translating. Use this to clarify translation when some words are
+You can flag certain glossary terms as forbidden, meaning ones _not_ be used
+for translations. Use this to clarify translation when some words are
 ambiguous or could have unexpected meanings.
 
 The terms can be flagged on the source language using :guilabel:`Tools` ↓
-:guilabel:`Mark as forbidden translations`. In the background this toggles the ``forbidden``
+:guilabel:`Mark as forbidden translation`. In the background this toggles the ``forbidden``
 flag on the string.
```
comradekingu, line 62:
```suggestion
flag of the string.
```

docs/user/glossary.rst
```diff
@@ -53,13 +53,13 @@ Forbidden translations
 
 .. versionadded:: 4.5
 
-You can flag certain glossary terms as forbidden - those should not be used
-while translating. Use this to clarify translation when some words are
+You can flag certain glossary terms as forbidden, meaning ones _not_ be used
```
Fat-Zer, line 56: I'm not sure but this one sounds a bit wrong... At least there is something wrong with the grammar. A bare minimum would be:

```suggestion
You can flag certain glossary terms as forbidden, meaning ones are _not_ to be used
```

But I'd suggest to change it to something like this:
```suggestion
You can flag certain glossary terms as forbidden. This will indicate that the translations of those should _not_ be used.
```

PS: Note that as for now there is an inconsistency in the logic of the Weblate's behaviour: you supposed to mark source strings as forbidden, but it's the translations who actually become forbidden. I hope this one will be changed soon because as for now this is quite unmanageable...

Fat-Zer (changes requested):

docs/user/glossary.rst
comradekingu: I think that is the "strict-same" or "read-only". "forbidden" means to avoid some word from the glossary I think? Forgot a "to" in the string above, good catch.

comradekingu, line 56:
```suggestion
You can flag certain glossary terms as forbidden, meaning ones _not_ to be used
```

docs/user/glossary.rst
```diff
@@ -53,13 +53,13 @@ Forbidden translations
 
 .. versionadded:: 4.5
 
-You can flag certain glossary terms as forbidden - those should not be used
-while translating. Use this to clarify translation when some words are
+You can flag certain glossary terms as forbidden, meaning ones _not_ to be used
+for translations. Use this to clarify translation when some words are
 ambiguous or could have unexpected meanings.
 
 The terms can be flagged on the source language using :guilabel:`Tools` ↓
```
nijel, line 60:
```suggestion
The terms can be flagged using :guilabel:`Tools` ↓
```

The source language limitation will be gone here with https://github.com/WeblateOrg/weblate/pull/5399

comradekingu, line 60:
```suggestion
Terms can be flagged using :guilabel:`Tools` ↓
```

Fat-Zer: As I can see it still requires `edit_source` permission, which most users don't have. Which should be mentioned unless it will be changed soon.

Also won't toggling the flag affect the translation of the term to all the languages? Unless it's a bug it should be mentioned too...

docs/user/glossary.rst
Fat-Zer: yep, as far as I can understand it means that the translation of the terms marked in the glossary as `forbidden` should be avoided. My note was about that even if the flag is applied to the source string, the "forbidden" becomes it's certain translation, which I've tried to highlight in the variant.

Also note that as I can see this feature is unfinished and under active development now, so I'd suggest to wait a bit with improving of the documentation until it settles down a bit as It will be changed every other week...

docs/user/glossary.rst
nijel: No, it applies it for single language only ... and I will adjust the permission check in this case as well

orangesunny (approved):

PR #5465, opened 2021-02-17
Tests: compatibility with latest Selenium

nijel:
## Proposed changes

Testing what is actually incompatibility of the Selenium 4.0 Beta 1. It don't see anything obvious in the changelog, but it breaks the tests. 

## Checklist


nijel:
This version is actaully broken for headless, the fix is https://github.com/SeleniumHQ/selenium/pull/9208

PR #5485, opened 2021-02-21
Docs: a minor reorder

Fat-Zer:
Reorder list items so it would be obvious that only one of those was added in the specified version


nijel:

PR #5501, opened 2021-02-22
Docs: Projects documentation reworked

comradekingu:


docs/admin/projects.rst
```diff
@@ -638,16 +638,16 @@ Committer name
...
 
 .. _component-committer_email:
 
 Committer e-mail
 ++++++++++++++++
 
-Email of committer used for Weblate commits, the author will always be the
-real translator. On some VCSs this might be not supported. The default value
-can be changed in :setting:`DEFAULT_COMMITER_EMAIL`.
+E-mail address of committer used for Weblate commits. The author will always
```
Fat-Zer, line 648:
```suggestion
E-mail address of the committer used for Weblate commits. The author will always
```

docs/admin/projects.rst
```diff
@@ -677,9 +677,8 @@ Default value can be changed by :setting:`COMMIT_PENDING_HOURS`.
 Lock on error
 +++++++++++++
 
-Enables locking the component on repository error (failed pull, push or merge).
-Locking in this situation avoids adding another conflict which would have to be
-resolved manually.
+Locks a component upon any repository error (failed pull, push or merge).
```
Fat-Zer, line 680:
```suggestion
Locks the component upon any repository error (failed pull, push or merge).
```

docs/admin/projects.rst
```diff
@@ -764,30 +763,29 @@ By default the component is visible to anybody who has access to the project,
 even if the person can not perform any changes in the component. This makes it
 easier to keep translation consistency within the project.
 
-Enable this in case you want to grant access to this component explicitly -
-the project level permissions will not apply and you will have to specify
-component or component list level permission in order to grant access.
+Turning this on at component-, or component-list level takes over
```
Fat-Zer, line 766:
```suggestion
Turning this on a component-, or component-list level takes over
```
or
```suggestion
Turning this at the component-, or component-list level takes over
```

docs/admin/projects.rst
```diff
@@ -764,30 +763,29 @@ By default the component is visible to anybody who has access to the project,
...
+You can choose additional projects where the component will be visible.
+Useful for shared libraries which you use in several projects.
 
 .. note::
 
-   Sharing component doesn't change its access control. It makes it only
-   visible when browsing other projects. User still need to have access to the
-   actual component in order to be able to browse or translate it.
+   Sharing a component doesn't change its access control. It only makes it
+   visible when browsing other projects. Users still need access to the
```
Fat-Zer, line 787:
```suggestion
   visible when browsing other projects. Users still need an access to the
```

Fat-Zer (changes requested):

docs/admin/projects.rst
comradekingu: Is "E-mail address for commits sent from Weblate" enough?

Fat-Zer: I'd prefer to keep words `committer` and `author`, as those are the settings name used in `git config`... It makes more clear what is meant as those are opposed to each other in the snippet...

docs/admin/projects.rst
comradekingu: turning this on for components or projects.

I am not exactly sure what "projects" are, if not just lists of projects?
Does this also mean specific component lists linked to groups that users can be put in?

docs/admin/projects.rst
comradekingu: To point out it is a privilege based system.

"to be granted access to"
or
"access permission"?

docs/admin/projects.rst
comradekingu: Yeah, it just isn't a very sleek sentence, but that is another problem.

docs/admin/projects.rst
comradekingu: I think "a" is more hypothetical since this isn't basic functionality and needs to be turned on?

comradekingu, line 680:
```suggestion
Locks any chosen component upon the first failed push or merge into its upstream repository, or pull from it.
```

docs/admin/projects.rst
Fat-Zer: Haven't noticed it before, but your snippet is much more confusing that the original... 

This describes a boolean setting in the component.

If it's turned off (default) the project-level access control settings are used.
If it's turned on the access is restricted by default and you have to define (or used one of predefined groups for `custom` or `private` access control) a group with component-level or component-list-level access control as permissions for the project-level access control settings are not used.

I'd suggested to refere the access control doc here, but as you know those are in quite badly shape yet...

docs/admin/projects.rst
Fat-Zer: I like `need to be granted` more...

docs/admin/projects.rst
comradekingu: @Fat-Zer 
"Turning it _on_" denotes both that it is normally off, and a boolean on/off setting(?).
How about "If you turn this on, at either ___ "?

docs/admin/projects.rst
```diff
@@ -764,30 +763,29 @@ By default the component is visible to anybody who has access to the project,
 even if the person can not perform any changes in the component. This makes it
 easier to keep translation consistency within the project.
 
-Enable this in case you want to grant access to this component explicitly -
-the project level permissions will not apply and you will have to specify
-component or component list level permission in order to grant access.
+Turning this on at component-, or component-list level takes over
+access permission to a component, regardless of project-level permissions.
```
comradekingu, line 767:
```suggestion
access permission to a component, regardless of project-level permissions, and you will have to grant access to it explicitly. This can be done through making a group that has access and granting it to users by putting them in it, or using the default `custom` or `private` access control groups.
```
@Fat-Zer Something like this? It could be moved to the access control docs, but good for now?

docs/admin/projects.rst
Fat-Zer: The problem with the wording here is, it just doesn't make sense to "*turn this setting on at some level*". It's just a setting for a specific component, isn't it? You just "*turn this on for a specific component*" and then you "*manage access to the component on a component- or component-list-level*" due to it overrides the project-level access control...

As for now I like the original variant more...

docs/admin/projects.rst
Fat-Zer: by the word, even for `need to be granted`, shouldn't there be an article before `access`?

docs/admin/projects.rst
Fat-Zer: Anyway, what's the problem with the original? I might try to help with a rephrasing it, but I just don't see why bother... At most I would replace a dash (`-`) with a colon (`:`) in it (thou even about that I'm not 100% sure)...

docs/admin/projects.rst
Fat-Zer: > I think "a" is more hypothetical since this isn't basic functionality and needs to be turned on?

It describes a setting which is available for a specific component. This setting locks exactly the component, in which the setting is located in... 

`a` or `any chosen component`, suggests that there is another setting to set the component. Or it's a drop-down list selection... it just sounds off-place here IMHO...

orangesunny: please close the discussion once you come to an end with it.

docs/admin/projects.rst
orangesunny: please close the discussion once you come to an end with it.

docs/admin/projects.rst
orangesunny: please don’t leave your conversations open and solution-less

docs/admin/projects.rst
```diff
@@ -663,26 +663,21 @@ well.
...
+background task or the :djadmin:`commit_pending` management command. All
+changes in a component are committed once there is at least one change
+older than this period.
 
 Default value can be changed by :setting:`COMMIT_PENDING_HOURS`.
 
-.. hint::
-
-   There are other situations where pending changes might be committed, see
-   :ref:`lazy-commit`.
```
nijel, line 676: Why it this being removed?

orangesunny: I am putting it back

orangesunny (approved):

PR #5601, opened 2021-03-07
Formats: Add scheme format.

### LINKED ISSUE #5504: Support scheme-format
I've added some translations on a self-hosted weblate instance for a scheme project. The translation uses gettext (po files) and strings have the `scheme-format` comment. While translating, I made a mistake in the format strings and broke the software. Weblate did not warn, even though msgfmt managed to report the error.

Ideally, a scheme-format check would be implemented, similar to python-format, c-format, php-format and the like that I saw in the documentation.

**Additional context**

In case it can help, here's a sample of scheme-format gettext:

```
msgid: "module ~a not found"

msgid: "  root device: ~[UUID: ~a~;label: ~s~;~a~]~%"
```

where `~a`, `~s`, `~[`, `~;` and `~%` are format specifiers like `%s` or `%d` in C (there are more, including positional arguments). Here is the documentation for the Scheme implementation I'm using: https://www.gnu.org/software/guile/manual/html_node/Formatted-Output.html.

Please tell me if there's any way I can help. If you think this is not too hard as a first contribution, can you guide me through the general steps I need to take to implement that support?

roptat:
Fixes #5504.

## Proposed changes

This adds scheme format checking.

## Checklist

## Other information

Lint passes, but I have two failing tests. I have  copied tests from php, and adapted them to scheme syntax, but I might not understand fully what they are testing. The two failing tests are:

```
2021-03-06T21:37:47.5201557Z ----------------------------------------------------------------------
2021-03-06T21:37:47.5203539Z Traceback (most recent call last):
2021-03-06T21:37:47.5206320Z   File "/home/runner/work/weblate/weblate/weblate/checks/tests/test_format_checks.py", line 306, in test_missing_tilde_format
2021-03-06T21:37:47.5208133Z     self.check.check_format("~s~~ ~~", "~s~~ tilde", False, None)
2021-03-06T21:37:47.5210472Z AssertionError: {'missing': ['~'], 'extra': []} is not false
2021-03-06T21:37:47.5211631Z 
2021-03-06T21:37:47.5212506Z ======================================================================
2021-03-06T21:37:47.5214028Z FAIL: test_space_format (weblate.checks.tests.test_format_checks.SchemeFormatCheckTest)
2021-03-06T21:37:47.5216034Z ----------------------------------------------------------------------
2021-03-06T21:37:47.5217256Z Traceback (most recent call last):
2021-03-06T21:37:47.5219118Z   File "/home/runner/work/weblate/weblate/weblate/checks/tests/test_format_checks.py", line 311, in test_space_format
2021-03-06T21:37:47.5220385Z     self.check.check_format("~d ~ string", "~d ~ other", False, None)
2021-03-06T21:37:47.5221167Z AssertionError: False is not true
2021-03-06T21:37:47.5221643Z 
2021-03-06T21:37:47.5222475Z ----------------------------------------------------------------------
```

I think I don't understand the purpose of the second value (c_format_is_position_based). I see it returns a boolean, but I'm not entirely sure what the input is, and what the boolean represents exactly. I tried to define my own `scheme_format_is_position_based`, but I'm not sure this is correct.

I don't really understand how positioning works either, like if the original source has `~a ~d` and the translated versions has `~1@*~d ~0@*~a` (btw formats are 0-based), how can weblate detect this is correct?

### DIFF
--- a/docs/user/checks.rst
+++ b/docs/user/checks.rst
@@ -443,6 +443,23 @@ Ruby format
 
     `Ruby Kernel#sprintf <https://ruby-doc.org/core/Kernel.html#method-i-sprintf>`_
 
+Scheme format
+*************
+
+*Scheme format string does not match source*
+
++------------------------+------------------------------------------------------------+
+| Simple format string   | ``There are ~d apples``                                    |
++------------------------+------------------------------------------------------------+
+| Flag to enable         | `scheme-format`                                            |
++------------------------+------------------------------------------------------------+
+
+.. seealso::
+
+    `Srfi 28 <https://srfi.schemers.org/srfi-28/srfi-28.html>`_,
+    `Chicken Scheme format <https://wiki.call-cc.org/eggref/5/format>`_,
+    `Guile Scheme formatted output <https://www.gnu.org/software/guile/manual/html_node/Formatted-Output.html>`_
+
 Vue I18n formatting
 *******************
 
--- a/weblate/checks/format.py
+++ b/weblate/checks/format.py
@@ -44,6 +44,24 @@ PYTHON_PRINTF_MATCH = re.compile(
     re.VERBOSE,
 )
 
+SCHEME_PRINTF_MATCH = re.compile(
+    r"""
+    ~(                          # initial ~
+        (?:(?P<ord>\d+)@\*~)?   # variable order, like ~1@*~d
+        (?P<fullvar>
+          (?:                   # any number of comma-separated parameters
+            #([+-]?\d+|\'.|[vV]|#)
+            ([+-]?\d+|'.|[vV]|\#)
+            (, ([+-]?\d+|'.|[vV]|\#))*
+          )?
+          :?
+          @?
+          (?P<type>[a-zA-Z%\$\?&_/|!\[\]\(\)~]) # type (~a, ~s, etc.)
+        |)                      # incomplete format string
+    )""",
+    re.VERBOSE,
+)
+
 
 PHP_PRINTF_MATCH = re.compile(
     r"""
@@ -197,6 +215,10 @@ def c_format_is_position_based(string):
     return "$" not in string and string != "%"
 
 
+def scheme_format_is_position_based(string):
+    return "@*" not in string and string != "~"
+
+
 def python_format_is_position_based(string):
     return "(" not in string and string != "%"
 
@@ -213,6 +235,7 @@ FLAG_RULES = {
     "javascript-format": (C_PRINTF_MATCH, c_format_is_position_based),
     "lua-format": (C_PRINTF_MATCH, c_format_is_position_based),
     "python-brace-format": (PYTHON_BRACE_MATCH, name_format_is_position_based),
+    "scheme-format": (SCHEME_PRINTF_MATCH, scheme_format_is_position_based),
     "c-sharp-format": (C_SHARP_MATCH, name_format_is_position_based),
     "java-format": (JAVA_MATCH, c_format_is_position_based),
 }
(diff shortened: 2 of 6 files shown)


nijel:
>  I see it returns a boolean, but I'm not entirely sure what the input is, and what the boolean represents exactly.

It should return true when the ordering can be changed in the format string. The name is indeed confusing (it's based on the C terminology, but really doesn't apply well to named format strings used mostly these days).

For C the `%s %s' should return false - the format strings have to be present in the same order. `%1$s %2$s` should return true - the format strings can appear in any order in the transaction because they include the position.

docs/user/checks.rst
```diff
@@ -443,6 +443,23 @@ Ruby format
...
+| Simple format string   | ``There are ~d apples``                                    |
++------------------------+------------------------------------------------------------+
+| Flag to enable         | `scheme-format`                                            |
++------------------------+------------------------------------------------------------+
+
+.. seealso::
+
+    `Srfi 28 <https://srfi.schemers.org/srfi-28/srfi-28.html>`
+    `Chicken Scheme format <http://wiki.call-cc.org/eggref/5/format>`
+    `Guile Scheme formatted output <https://www.gnu.org/software/guile/manual/html_node/Formatted-Output.html>`
```
nijel, line 461: The links are wrongly formatted (see https://weblate--5601.org.readthedocs.build/en/5601/user/checks.html#scheme-format). Also please use https when possible:

```suggestion
    `Srfi 28 <https://srfi.schemers.org/srfi-28/srfi-28.html>`_,
    `Chicken Scheme format <https://wiki.call-cc.org/eggref/5/format>`_,
    `Guile Scheme formatted output <https://www.gnu.org/software/guile/manual/html_node/Formatted-Output.html>`_
```

weblate/checks/format.py
```diff
@@ -197,6 +215,10 @@ def c_format_is_position_based(string):
     return "$" not in string and string != "%"
 
 
+def scheme_format_is_position_based(string):
+    return "*" not in string and string != "~"
```
nijel, line 219: Seems like this would be better fit here (guessing based on regexp above, I have no knowledge of scheme):

```suggestion
    return "@*" not in string and string != "~"
```

weblate/checks/tests/test_format_checks.py
```diff
@@ -241,6 +242,71 @@ def test_space_format(self):
...
+            self.check.check_format("~1@*~s string", "~s string", False, None)
+        )
+
+    def test_wrong_tilde_format(self):
+        self.assertTrue(
+            self.check.check_format("~s~~ (0.1~~)", "~s~~ (0.1~x)", False, None)
+        )
+
+    def test_missing_tilde_format(self):
+        self.assertFalse(self.check.check_format("~s~~ ~~", "~s~~ tilde", False, None))
```
nijel, line 302: You will probably have to override `normalize` method here - it skips % format string, but you should skip ~.

weblate/checks/format.py
```diff
@@ -439,6 +462,14 @@ class LuaFormatCheck(BasePrintfCheck):
     description = _("Lua format string does not match source")
 
 
+class SchemeFormatCheck(BasePrintfCheck):
+    """Check for Scheme format string."""
+
+    check_id = "scheme_format"
+    name = _("Scheme format")
+    description = _("Scheme format string does not match source")
```
nijel, line 470: You will probably have to override `normalize` and `format_string` methods here as well.

weblate/checks/tests/test_format_checks.py
```diff
@@ -241,6 +242,71 @@ def test_space_format(self):
...
+
+    def test_space_format(self):
+        self.assertTrue(
+            self.check.check_format("~d ~ string", "~d ~ other", False, None)
+        )
```
nijel, line 307: Your regexp doesn't capture space anywhere. In case the space can be present in the format string, please adjust the regexp, otherwise this test can be removed.

weblate/checks/format.py
```diff
@@ -439,6 +462,17 @@ class LuaFormatCheck(BasePrintfCheck):
...
+
+    def normalize(self, matches):
+        return [m for m in matches if m != "~"]
+
+
```
nijel, line 475: This will be most likely needed to properly display hints on missing format:

```suggestion

    def format_string(self, string):
        return f"~{string}"

```

roptat: oh, so that's what it does :D, thanks!

nijel:

PR #5636, opened 2021-03-11
formats: make "state" attribute literal

rffontenelle:
I believe "state" should not be translated, since it is the name of an attribute. So, marking it as ``` ``literal`` ```.

### DIFF
--- a/docs/formats.rst
+++ b/docs/formats.rst
@@ -267,7 +267,7 @@ Translation states
 
 .. versionchanged:: 3.3
 
-   Weblate ignored the state attribute prior to the 3.3 release.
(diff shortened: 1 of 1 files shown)


nijel:
Thank you for your contribution. Unfortunately, it doesn't really look useful and you've probably made it to gain points in Hacktoberfest. It will be marked as invalid, please see https://hacktoberfest.digitalocean.com/details#spam for more information.

rffontenelle:
Actually, I was just translating it and spotted. I promise to condense more changes next time ;)

nijel:
It's okay to keep changes simple - it makes it easy to review (especially compare with other doc pull requests, see https://github.com/WeblateOrg/weblate/pulls).

PR #5656, opened 2021-03-14
Discover Weblate docs update

orangesunny:
## Proposed changes

Rephrasing of the Discover docs.

@nijel 
Not sure what steps the
*2. Register your Weblate server in the discovery database*
means.

I think it needs a more detailed description, but I don’t know the process.

## Checklist


nijel:
>  2. Register your Weblate server in the discovery database

There will be a button for that on user page on weblate.org. This part is not yet implemented.

nijel:
Anyway the important point of this documentation is that it has to be activated on both sides - weblate.org and in your Weblate.

PR #5658, opened 2021-03-14
adding guidelines for docs contributing

orangesunny:
## Proposed changes

Trying to propose guidelines for contributing to the docs

@nijel 
Please give feedback and suggest other places where this should be visible.

## Checklist


nijel:
Maybe it would be useful to link this from https://github.com/WeblateOrg/weblate/blob/master/CONTRIBUTING.md

PR #5661, opened 2021-03-14
Spelling: Support answers

comradekingu:
Sounds a bit nicer

### DIFF
--- a/weblate/templates/mail/account_activity.html
+++ b/weblate/templates/mail/account_activity.html
@@ -31,7 +31,7 @@
 <p>
 {% blocktrans %}If you don’t recognize this change, you can report it by clicking the button below.{% endblocktrans %}
 <b>{% blocktrans %}Don’t use this button for support questions. Such questions won’t be answered.{% endblocktrans %}</b>
-{% blocktrans %}If you have an idea or a question, <a href="https://github.com/WeblateOrg/weblate/issues/new/choose">this</a> is a place to start.{% endblocktrans %}  
+{% blocktrans %}If you have an idea or a question, <a href="https://github.com/WeblateOrg/weblate/issues/new/choose">this</a> is a place to start.{% endblocktrans %}
 </p>
 
 <div class="line buttons">
--- a/weblate/templates/mail/account_activity.html
+++ b/weblate/templates/mail/account_activity.html
@@ -30,7 +30,8 @@
 
 <p>
 {% blocktrans %}If you don’t recognize this change, you can report it by clicking the button below.{% endblocktrans %}
-<b>{% blocktrans %}Using this button for support questions will not grant support answers.{% endblocktrans %}</b>
+<b>{% blocktrans %}Don’t use this button for support questions. Such questions won’t be answered.{% endblocktrans %}</b>
+{% blocktrans %}If you have an idea or a question, <a href="https://github.com/WeblateOrg/weblate/issues/new/choose">this</a> is a place to start.{% endblocktrans %}  
 </p>
 
 <div class="line buttons">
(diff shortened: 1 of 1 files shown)


maisoenaingtun:
Support {{#each}}-{{else}}#5661

Fat-Zer:
The former phrasing is more clear IMHO...

comradekingu:
@Fat-Zer And harsh, which is not good.

orangesunny:
Thank you for the suggestion and opinions.

1. I love being nice to all people and the Weblate community definitely. I wouldn’t say it’s harsh. It’s straightforward, plain, and informative.
2. It’s not in the app, it’s an e-mail that was misused many times before and generated a bunch of useless workload to our team. (People asking support questions the way they only should report suspicious activity, fraud, etc.) After a few weeks of having this change implemented, I am happy to say it helped a lot. So I am not going to change it this way.
3. I will add some more text to be nice and friendly.

orangesunny:
@nijel please check the formatting and merge :)

weblate/templates/mail/account_activity.html
```diff
@@ -31,6 +31,7 @@
 <p>
 {% blocktrans %}If you don’t recognize this change, you can report it by clicking the button below.{% endblocktrans %}
 <b>{% blocktrans %}Don’t use this button for support questions. Such questions won’t be answered.{% endblocktrans %}</b>
+{% blocktrans %}If you have an idea or a question, <a href="https://github.com/WeblateOrg/weblate/issues/new/choose">this</a> is a place to start.{% endblocktrans %}
```
nijel, line 34: Better to keep the link outside the translated text - it can change over the time:

```suggestion
{% blocktrans with link_start='<a href="https://github.com/WeblateOrg/weblate/issues/new/choose">' link_end='</a>'  %}If you have an idea or a question, {{ link_start }}this{{ link_end }} is a place to start.{% endblocktrans %}
```

comradekingu:
@orangesunny 
I know it has been misused, because it is evident from how the original string was put together.
"is a place to start" isn't a figure of speech. "is a good place to start" is.
If you have to add nice, you don't know what you are doing. It doesn't counteract anything, it just adds text for something that looks passive aggressive.
Comparing the amount of erroneous support questions is a false comparison.
One is the amount of questions without a message, the other is the amount of questions with a message.
Concerning the _quality_ of the message, it went from bad, to OK, to now worse than OK.
It isn't far off here, but why does it have to be so difficult I don't understand.

orangesunny:
> "is a place to start" isn't a figure of speech. "is a good place to start" is.

Thanks, you’re right, I will change that!

> If you have to add nice, you don't know what you are doing. It doesn't counteract anything, it just adds text for something that looks passive aggressive.

*...don’t do it, I won’t help you...* is not passive-aggressive. It’s informative. We provide the platform that’s widely used so we have to make sure it works not only for Hosted but for other deployments too. That’s why we are clear to users about what to do and what not.

> Comparing the amount of erroneous support questions is a false comparison.
> One is the amount of questions without a message, the other is the amount of questions with a message.

I will continue to provide support to all Weblate users in the best possible way. Without working workflows, we would end like some libre projects that only offer community forums unless you pay high support fee. We choose not to go this way so we have the rules and processes set.

> Concerning the _quality_ of the message, it went from bad, to OK, to now worse than OK.
> It isn't far off here, but why does it have to be so difficult I don't understand.

Two out of three participants in this conversation say it’s OK. You are repeating that it’s bad. Can you, please, tell me why the message has low quality?

We can discuss and try to find a better formulation of the sentence. I and @Fat-Zer just said we don’t think your one version is better. That doesn’t mean the will to change for the better is away.

PR #5662, opened 2021-03-14
Spelling: The source x2

comradekingu:
Alternative, "uneditable source string …"

### DIFF
--- a/weblate/templates/snippets/unit-readonly-badge.html
+++ b/weblate/templates/snippets/unit-readonly-badge.html
@@ -5,7 +5,7 @@
     {% elif unit.is_source and unit.translation.component.template and not unit.translation.component.edit_template %}
         <span class="badge" title="{% trans "Source string editing is disabled." %}">{% trans "Source string" %}</span>
     {% elif unit.is_source and not unit.translation.component.template and unit.translation.component.manage_units %}
-        <span class="badge" title="{% trans "You can delete and recreate the string if you want to edit it." %}">{% trans "Source string" %}</span>
+        <span class="badge" title="{% trans "Source strings can’t be edited. You can delete them and add new ones." %}">{% trans "Source string" %}</span>
     {% elif unit.is_source %}
         <span class="badge" title="{% trans "The source string cannot be edited." %}">{% trans "Source string" %}</span>
     {% else %}
--- a/weblate/templates/snippets/unit-readonly-badge.html
+++ b/weblate/templates/snippets/unit-readonly-badge.html
@@ -5,7 +5,7 @@
     {% elif unit.is_source and unit.translation.component.template and not unit.translation.component.edit_template %}
         <span class="badge" title="{% trans "Source string editing is disabled." %}">{% trans "Source string" %}</span>
     {% elif unit.is_source and not unit.translation.component.template and unit.translation.component.manage_units %}
-        <span class="badge" title="{% trans "The source string cannot be edited, but you can delete it and add a new one." %}">{% trans "Source string" %}</span>
+        <span class="badge" title="{% trans "You can delete and recreate the string if you want to edit it." %}">{% trans "Source string" %}</span>
     {% elif unit.is_source %}
         <span class="badge" title="{% trans "The source string cannot be edited." %}">{% trans "Source string" %}</span>
     {% else %}
--- a/weblate/templates/snippets/unit-readonly-badge.html
+++ b/weblate/templates/snippets/unit-readonly-badge.html
@@ -5,9 +5,9 @@
     {% elif unit.is_source and unit.translation.component.template and not unit.translation.component.edit_template %}
         <span class="badge" title="{% trans "Source string editing is disabled." %}">{% trans "Source string" %}</span>
     {% elif unit.is_source and not unit.translation.component.template and unit.translation.component.manage_units %}
-        <span class="badge" title="{% trans "Source string cannot be edited, but you can delete it and add a new one." %}">{% trans "Source string" %}</span>
+        <span class="badge" title="{% trans "The source string cannot be edited, but you can delete it and add a new one." %}">{% trans "Source string" %}</span>
     {% elif unit.is_source %}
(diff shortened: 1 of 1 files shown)


orangesunny:
> Alternative, "uneditable source string …"

This was considered. From all the talks with individual users and companies, we settle on the best version. :)

Article `the` was left out intentionally as it’s not needed here, to save space in the UI.

comradekingu:
@orangesunny It is an unmarked noun error. I'll try to shave some space out of it.

weblate/templates/snippets/unit-readonly-badge.html
```diff
@@ -5,9 +5,9 @@
     {% elif unit.is_source and unit.translation.component.template and not unit.translation.component.edit_template %}
         <span class="badge" title="{% trans "Source string editing is disabled." %}">{% trans "Source string" %}</span>
     {% elif unit.is_source and not unit.translation.component.template and unit.translation.component.manage_units %}
-        <span class="badge" title="{% trans "Source string cannot be edited, but you can delete it and add a new one." %}">{% trans "Source string" %}</span>
+        <span class="badge" title="{% trans "The source string cannot be edited, but you can delete it and add a new one." %}">{% trans "Source string" %}</span>
```
comradekingu, line 8:
```suggestion
        <span class="badge" title="{% trans "You can delete and recreate the string if you want to edit it." %}">{% trans "Source string" %}</span>
```

weblate/templates/snippets/unit-readonly-badge.html
```diff
@@ -5,9 +5,9 @@
     {% elif unit.is_source and unit.translation.component.template and not unit.translation.component.edit_template %}
         <span class="badge" title="{% trans "Source string editing is disabled." %}">{% trans "Source string" %}</span>
     {% elif unit.is_source and not unit.translation.component.template and unit.translation.component.manage_units %}
-        <span class="badge" title="{% trans "Source string cannot be edited, but you can delete it and add a new one." %}">{% trans "Source string" %}</span>
+        <span class="badge" title="{% trans "You can delete and recreate the string if you want to edit it." %}">{% trans "Source string" %}</span>
     {% elif unit.is_source %}
-        <span class="badge" title="{% trans "Source string cannot be edited." %}">{% trans "Source string" %}</span>
+        <span class="badge" title="{% trans "The source string cannot be edited." %}">{% trans "Source string" %}</span>
```
comradekingu, line 10: @orangesunny Does this fit?

Fat-Zer:
> Article `the` was left out intentionally as it’s not needed here, to save space in the UI.

@orangesunny, why do you want to save space in title text? o_O

comradekingu:
@Fat-Zer Primarily I wanted to get rid of the error. Then it was shortened to facilitate both.
The original is very staccato. It also leaves the "(that means you can then edit it)" up in the air.

Fat-Zer:
> @Fat-Zer Primarily I wanted to get rid of the error. Then it was shortened to facilitate both.
The original is very staccato. It also leaves the "(that means you can then edit it)" up in the air.

I'm genuinely don't see a reason to save space on an article in the title text: it couldn't cause any layout problems; it wraps fine by browsers; It might be less readable than it's too long, but it's less readable without an article as well... That's why I'm so confused about the notice, I suspect @orangesunny didn't notices it's a title, or was talking about the badge's main text, or there are some unspoken rules of best frontend design practices that it should be less than 75 chars...

As for your last variant, IMHO the main message of the title "It cannot be edited", and your variant looses it entirely.

By the word may be `s/Source string/It`?

PS: What's `staccato`?

comradekingu:
@Fat-Zer 
We have one style of communication, if that. It is my style of editing nijel's writing, and we don't have anything better.

Apologetic language is nowhere to be found in Weblate. As an example "just" or "simply" do this and that is avoided, along with similar cheap advertising. There is no "computer language" like debugger style errors.
The only advertising is making things sound good and interesting. Everything else is short and to the point.

There is also enough problems in other places to not get caught up in the particular way problems are solved.
If this edit is a problem, nearly everything is. Accepting fixes for other stuff is a lot better than holding up fixes for clearly broken stuff at the top of the pipeline. While we are deciding on a v 2.5, we are wasting the time of people translating the broken v1.

Staccato is a piano term. Jarred, abrupt, haphazardly put together. Can't remember what the proper term is.

Longer text is _always_ worse by that metric alone. It is harder to translate, harder to decipher, and it introduces stress.
Importantly, shorter text isn't always better overall.

"If" operative word "you want to edit it", condition "do x, y ,z." Yes it is reversed order, but it does so to actively be relaxed.

"It doesn't wrap" and ~"good enough" isn't what to aim for.

`s/Source string/It` is a good point. Since the title says "The source string", there is no reason to repeat it.
May be that it could be just "string" and then just "it", but that takes more effort to shave than I was willing to put in.
We can get there in due time.

To go through why the v1 is broken:

>Source string cannot be edited, 
why? Why is that interesting. What do I actively want to do. Give me the direct action.

>but you can delete it and add a new one
Spliced sentence, no explanation of what will happen. Yes I can figure it out, but my reasoning is just being exploited as a resource there. It annoys me to no end.

To my understanding this string would say

Title: 
>Source string cannot be edited

Body:
>Source string cannot be edited

That is a double marked noun error, in what is a 100% repetition word-for-word.
If you didn't get it the first time, you aren't about to get it the second time.
You no less have to get to the end to get a clue.

weblate/templates/snippets/unit-readonly-badge.html
orangesunny: partially yes

orangesunny:
Guys, you created a long conversation about something you are not sure about. There is no need to consume your time this way.

Anyway:
1. Thank you so much for the suggestions and opinions. Thanks to that, I was able to make it clearer.
2. You will see this in 4.5.2 once you open a source string in the editor.
3. There will be some more changes in this. Please wait with the next wave of your helpful feedback and contributions until 4.5.2 is fully released.

PR #5663, opened 2021-03-14
Spelling: occurrences

comradekingu:
Is one supposed to be actively singular here?


nijel:
This is shown only if there is more than 100 of them, so the singular in English won't be ever used.

PR #5664, opened 2021-03-14
Spelling: Suitable for offline translation

comradekingu:
Not 100% sure this is better, I think I have seen "offline translation" somewhere else.

### DIFF
--- a/weblate/templates/translation.html
+++ b/weblate/templates/translation.html
@@ -342,7 +342,7 @@
       <tr>
         <th class="number">{{ object.stats.all|intcomma }}</th>
         <td>
-          {% trans "All strings, converted files enriched with comments; suitable for translating offline" %}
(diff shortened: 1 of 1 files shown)


maisoenaingtun:
Support {{#each}}-{{else}}#5664

comradekingu:
@maisoenaingtun I am not following. The structure looks to be the way it is to make things really easy to understand.
"enriched with comments" could be "commented converted files" or "commented file conversions". or similar

orangesunny (approved):

orangesunny:
> @maisoenaingtun I am not following. The structure looks to be the way it is to make things really easy to understand.
> "enriched with comments" could be "commented converted files" or "commented file conversions". or similar

I feel these variants change the meaning so I didn’t implement them.

Issue #5716, opened 2021-03-22
Documentation API status codes


luebbe:
Just a small documentation issue:

The API docs at https://docs.weblate.org/en/latest/api.html?highlight=api#any-- mention only 200 as "OK" status code where at least "201 Created" and "204 No Content" are other 20X "OK-ish" status codes returned by Weblate.

![grafik](https://user-images.githubusercontent.com/3872576/111965439-07a87300-8af6-11eb-8fce-d19bbfff1215.png)

nijel:
Pull requests are welcome to improve the documentation :-). The source file is here: https://github.com/WeblateOrg/weblate/blob/master/docs/api.rst

PR #5721, opened 2021-03-22
Document other "success" status codes than 200 OK

luebbe:
Adresses #5716

Document other "success" status codes than 200 OK

## Checklist

## Other information

I didn't dig deeper to find more. These were the missing ones that I came across during my tests.


nijel:

PR #5730, opened 2021-03-24
Spelling: Parsing error

comradekingu:


orangesunny:
This is not an improvement. `Parse error` is the correct name of the error, widely used in many programming languages, IDEs, and systems.

comradekingu:
How do you figure it is "the correct name" if not _exclusively_ used?
If programming languages, IDEs, and systems are anything to go by, it isn't.
Which ones of those have you used? I certainly found a lot of them to be less than user-friendly, and don't really see it carrying over to a translation platform. There it is subject to translators that haven't.

Further, the unreferenced "Parse error" from "Parse error in %(component)s" is indistinguishable from the verb.
In https://hosted.weblate.org/translate/weblate/application/bg/?checksum=efa621befa14e091 it relies entirely on non-existent context.

Other than doing away with some idiosyncratic English, which is the main benefit here, it could do with a qualifier like "discovered" or "found", since it is a sentence. "Parsing error" would otherwise be followed with an ellipsis on its own. Another reason why it is better.

Not sure why it isn't in the "Could not" style of other error messages, but that is a secondary concern.
It used to be https://github.com/WeblateOrg/weblate/blob/1611b7edc3361da2a51560262a695217aa3a7f25/weblate/templates/mail/parse_error.html#L7
which is quite telling.

PR #5732, opened 2021-03-24
Contact form language reworked

comradekingu:
@orangesunny "For questions about specific translation project" which I already supplied an idiomatic suggestion for, isn't correct English.  "about ___(a/any) specific" or "projects" is the mistake. It is also a double-clause sentence. It is my name in the "reviewers".

"Bug-report" originally does away with the "bug reports to headquarters" idiosyncrasy, but is avoided here.

### DIFF
--- a/weblate/templates/accounts/contact.html
+++ b/weblate/templates/accounts/contact.html
@@ -12,9 +12,9 @@
 
 <div class="alert alert-info">
   <ul>
-    <li>{% blocktrans %}For questions about specific translation project, contact its maintainers directly, not by this form.{% endblocktrans %}</li>
-    <li>{% blocktrans %}You can only contact maintainers of the whole server here.{% endblocktrans %}</li>
-    <li>{% blocktrans %}If you seek support for Weblate, or want to file a bug-report, check its website at {{ weblate_link }}.{% endblocktrans %}</li>
+    <li>{% blocktrans %}Contact maintainers of this Weblate here.{% endblocktrans %}</li>
+    <li>{% blocktrans %}Questions about specific translation projects should instead be directed to their respective maintainers.{% endblocktrans %}</li>
+    <li>{% blocktrans %}Get support for and contribute to Weblate at {{ weblate_link }}.{% endblocktrans %}</li>
   </ul>
 </div>
 
--- a/weblate/templates/accounts/contact.html
+++ b/weblate/templates/accounts/contact.html
@@ -14,7 +14,7 @@
   <ul>
     <li>{% blocktrans %}Contact maintainers of this Weblate here.{% endblocktrans %}</li>
     <li>{% blocktrans %}Questions about specific translation projects should instead be directed to their respective maintainers.{% endblocktrans %}</li>
-    <li>{% blocktrans %}Get support for and contribute to Weblate at {{ weblate_link }}.{% endblocktrans %}</li>
+    <li>{% blocktrans %}Get support for and contribute to Weblate itself at {{ weblate_link }}.{% endblocktrans %}</li>
   </ul>
 </div>
 


weblate/templates/accounts/contact.html
```diff
@@ -12,9 +12,9 @@
 
 <div class="alert alert-info">
   <ul>
-    <li>{% blocktrans %}For questions about specific translation project, contact its maintainers directly, not by this form.{% endblocktrans %}</li>
-    <li>{% blocktrans %}You can only contact maintainers of the whole server here.{% endblocktrans %}</li>
-    <li>{% blocktrans %}If you seek support for Weblate, or want to file a bug-report, check its website at {{ weblate_link }}.{% endblocktrans %}</li>
+    <li>{% blocktrans %}Contact maintainers of this Weblate here.{% endblocktrans %}</li>
```
comradekingu, line 15: Ideally a placeholder for "contact maintainers of this _Fedora Weblate" here, but then again maybe some installations don't have "Weblate" in there.
"Placeholder" - "installation". would work

weblate/templates/accounts/contact.html
```diff
@@ -12,9 +12,9 @@
...
+    <li>{% blocktrans %}Questions about specific translation projects should instead be directed to their respective maintainers.{% endblocktrans %}</li>
+    <li>{% blocktrans %}Get support for and contribute to Weblate at {{ weblate_link }}.{% endblocktrans %}</li>
```
comradekingu, line 17:
```suggestion
    <li>{% blocktrans %}Get support for and contribute to Weblate itself at {{ weblate_link }}.{% endblocktrans %}</li>
```
possible distinction

weblate/templates/accounts/contact.html
nijel, line 15:
```suggestion
    <li>{% blocktrans %}Contact maintainers of {{ site_title }} here.{% endblocktrans %}</li>
```

nijel: The problem with this is if people don't change the site title...

orangesunny:
I cannot approve these changes as they are not solving any real issue and don’t improve the current state. They also reduce the explanatory value of the text.

- The first sentence will be right as
*For questions about a specific translation project, contact its maintainers directly, not by this form.*
Thank you for the correction.

Your name is in the previous commit's reviewers, and you don’t want it to be there?

comradekingu:
@orangesunny 
I didn't ask for your approval.

The specific issues solved are:
The first paragraph says nothing about what you actually can do.
>For questions about specific translation project, contact its maintainers directly, not by this form.

"For questions about specific translation project" is a glaring error.

Referencing "form" is just a redundant technicality for no benefit.
The "don't do this", isn't contacting maintainers of projects all, and it splits those concepts apart,
and further by what is only the slightest indication that this does anything else, right at the end, after the comma.

That is a waste of time, and it sends the wrong message.
Then it gets into why one could, but not why one would.
>You can only contact maintainers of the whole server here.

"Server" is mentioned for no good reason here. Projects as a means of contact isn't the smaller part of it btw.

>If you seek support for Weblate, or want to file a bug-report, check its website at

In the event the user is seeking support for Weblate, the concept of doing so is put before the need.
That just isn't optimal, but sub-par nonetheless.

Then there is an Oxford comma, starting at "or" for the clause of "file a bug-report", with a fully redundant "or want" to further separate it from anything to do with "Weblate".

"Get" and then "file" directly would be the elegant way, but instead there is an "if", and that is following no good premise of what to actually do. So the "if" is carried.

"Weblate" isn't even qualified or explained, so the user would already have to know what Weblate is. Preaching to the choir, as it were.

Then there is the idiosyncratic "check its website".
It is a triple-clause run-on garden path spaghetti sentence.

Only at the end do you get "it", having to do with what isn't past a semicolon, but rather two commas back "Weblate", and can then derive filing bug-reports has to do with "Weblate" as distinct from anything.

It will undoubtedly say "Weblate" right next to the site_title, so there is every reason to assume "Weblate" is of the platforms making giving no other knowledge.

That was the _entire_ text…

weblate/templates/accounts/contact.html
comradekingu: I am more worried they just call it "Weblate". Having it look wrong but understandable is sort of a corrective practice, seeing as the people in charge can be contacted right there and then. Maybe they would have had to also set up e-mailing for it to work, but alas.

TL;DR Made do without, since "itself" works nicely farther down, and the expected result always works. (Pending review.)

comradekingu:
@Fat-Zer Help for review following:

>Contact maintainers of this Weblate here.

Straight to the point, this being Weblate, as distinct from generic Weblate, if the user doesn't already know.

>Questions about specific translation projects should instead be directed to their respective maintainers.

"instead" carrying over, in carrying the first sentence, and either discarding that, or following this.
Gets the bulk of what happens to be important out of the way "specific translation projects, questions" in an easy manner, and then invokes details about doing so and what they are afterwards.
Could do "towards", could do "at", meh.

>Get support for and contribute to Weblate itself at {{ weblate_link }}.

Assumes the user has seen a website link before, "itself" being different from "this" above.
Bonus for following the exact same style here as in sentence 1.
The third sentence isn't so much a third option out of three, but rather something the user can do anyway.
It is an extra tidbit where you know what you want to do, and go "ah, OK", as it is a reassurance for some, so much as it details the separation of running Weblate platform, and Weblate the software for others.
Tried to make it sound jovial by using "contribute" and "get".

orangesunny:
> @orangesunny
> I didn't ask for your approval.

How do you want to get things merged without the approval of repository maintainers?
 
> If you actually read the commit log, as much is apparent. As you may well know I was kicked out of the reviewers team, and the organization, so now it doesn't look like I send this stuff, even if I was never actually involved in or asked to do so.

Yes, I can read the commit log, but I asked you to understand what you are thinking.
It is clear that you send this PR, it is your PR.


In this, you are reacting to the error (missing article `a`) that can be solved as I suggested above.

> > For questions about specific translation project, contact its maintainers directly, not by this form.
> 
> "For questions about specific translation project" is a glaring error.


We can cooperate if you lower the pressure and start discussing kindly and concisely.

orangesunny:
After many considerations, I close this PR as its changes are not helpful for the contact form functionality.

comradekingu:
@orangesunny What did you even consider?

PR #5895, opened 2021-04-21
Docs: Glossary spelling "Untranslatable strings"

comradekingu:
Don't know what this in-between
https://github.com/WeblateOrg/weblate/blob/c96d105cb4836991dbbeb85dab3c769fa4a9ddf2/weblate/trans/forms.py#L633
_Update only non translated strings_ (should be non-)
Could be "Update only untranslatable strings" or "Update only untranslated strings"

https://hosted.weblate.org/translate/weblate/application/nb_NO/?checksum=37b1bf47e4257ced
is "not translated strings."
I think that must have been "not translated" at some point, and then "strings" added on, because it is sub-par.

## Checklist

### DIFF
--- a/docs/user/glossary.rst
+++ b/docs/user/glossary.rst
@@ -86,7 +86,7 @@ Terminology
 .. versionadded:: 4.5
 
 Flagging certain glossary terms as ``terminology``  by bulk-editing, typing in the flag,
-or or by using :guilabel:`Tools` ↓ :guilabel:`Mark as terminology` adds entries for them
+or by using :guilabel:`Tools` ↓ :guilabel:`Mark as terminology` adds entries for them
 to all languages in the glossary. Use this for important terms that should
 be well thought out, and retain a consistent meaning across all languages.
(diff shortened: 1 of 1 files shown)


comradekingu:
Can't remember why I made this, but "not translated" has since taken the meaning of "needs work" in listings.

docs/user/glossary.rst
```diff
@@ -86,7 +86,7 @@ Terminology
 .. versionadded:: 4.5
 
 Flagging certain glossary terms as ``terminology``  by bulk-editing, typing in the flag,
-or or by using :guilabel:`Tools` ↓ :guilabel:`Mark as terminology` adds entries for them
+or by using :guilabel:`Tools` ↓ :guilabel:`Mark as terminology` adds entries for them
```
comradekingu, line 89: Spotted by Fred Maranhão at Hosted Weblate :)

comradekingu:
Neverstale

PR #5908, opened 2021-04-23
[docs] Fix --all flag

rffontenelle:
## Proposed changes

`--all` -> `\--all` to prevent Sphinx from rendering `--` as `–` (en-dash).

Current text in documentation:

all

Text after the fix:

all-fixed

Link: https://docs.weblate.org/en/latest/admin/continuous.html#updating-repositories

## Checklist

### DIFF
--- a/docs/admin/continuous.rst
+++ b/docs/admin/continuous.rst
@@ -64,7 +64,7 @@ source.
 
 * Enable :setting:`AUTO_UPDATE` to automatically update all components on your Weblate instance
 
-* Execute :djadmin:`updategit` (with selection of project or `--all` to update all)
+* Execute :djadmin:`updategit` (with selection of project or ``--all`` to update all)
 
 Whenever Weblate updates the repository, the post-update addons will be
 triggered, see :ref:`addons`.


docs/admin/continuous.rst
```diff
@@ -64,7 +64,7 @@ source.
 
 * Enable :setting:`AUTO_UPDATE` to automatically update all components on your Weblate instance
 
-* Execute :djadmin:`updategit` (with selection of project or `--all` to update all)
+* Execute :djadmin:`updategit` (with selection of project or `\--all` to update all)
```
nijel, line 67: Maybe it would be better this way? Not sure whether that renders correctly, though:

```suggestion
* Execute :djadmin:`updategit` (with selection of project or ``--all`` to update all)
```

rffontenelle: Captura de tela de 2021-04-23 16-50-33

No problem here. I'll update it later (no git right now)

rffontenelle: Force-pushed a commit with your suggestion

nijel:

PR #6086, opened 2021-05-25
Spelling: Clarify ambigous language codes in use

comradekingu:
At first I thought this meant

"Different components of the same language are using ambiguous language codes"

### DIFF
--- a/weblate/templates/trans/alert/ambiguouslanguage.html
+++ b/weblate/templates/trans/alert/ambiguouslanguage.html
@@ -2,8 +2,8 @@
 {% load translations %}
 
 <p>
-  {% trans "The component translation files are using ambiguous language code." %}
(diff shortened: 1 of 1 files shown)


weblate/templates/trans/alert/ambiguouslanguage.html
```diff
@@ -2,8 +2,8 @@
 {% load translations %}
 
 <p>
-  {% trans "The component translation files are using ambiguous language code." %}
-  {% trans "These language codes indicate a macrolanguage and it is usually better to use code of individual language instead." %}
+  {% trans "Files for translated languages are using ambiguous language codes." %}
+  {% trans "These language codes indicate a macrolanguage and it is usually better to use the code of the individual language instead." %}
```
nijel, line 6:
```suggestion
  {% trans "These language codes indicate a macrolanguage, and it is usually better to use the code of the individual language instead." %}
```

nijel:

Issue #5969, opened 2021-05-06
Using Tabs vs Spaces for the JSON addon


ahadcove:
When weblate tries to merge to our branch it merged the file with spaces causing a complete clash with our current setup.
We're already using the JSON addon and specifying the indentation, but there isn't a way to choose tabs or spaces

**Describe the solution you'd like**

Option for tabs vs spaces in the JSON indentation addon please

**Describe alternatives you've considered**

Or even an option for a custom hook before trying to merge where we could run an auto formatter. But the easier route is just adding on to the already existing json addon by selecting tabs

nijel:
It should be possible to specify a custom string for indentation instead of the integer, see https://docs.python.org/3/library/json.html#json.dump. So, it's just a matter for adding configuration in the JSON add-on.

ahadcove:
> It should be possible to specify a custom string for indentation instead of the integer, see https://docs.python.org/3/library/json.html#json.dump. So, it's just a matter for adding configuration in the JSON add-on.

This would be awesome! If you need me to help out with the PR, i can try and dig further

nijel:
The JSON add-on is here (it gets whatever is saved from the form):

https://github.com/WeblateOrg/weblate/blob/69f09794559fe83a51ce9d489725cf09658718e0/weblate/addons/json.py#L45-L48

The form to configure it is defined here:

https://github.com/WeblateOrg/weblate/blob/69f09794559fe83a51ce9d489725cf09658718e0/weblate/addons/forms.py#L184-L188

ahadcove:
Finally had time to make the PR @nijel https://github.com/WeblateOrg/weblate/pull/6088

PR #6088, opened 2021-05-26
Adding indent type to choose between integers and strings for JSON addon

ahadcove:
## Proposed changes

This is in response to the issue we was having where we need to indent based on tabs in our files vs spaces.
JSON.dump's indent can be used as integer for spaces or as a string with whatever you want inside of it. That's why in this PR I made an indent_type to parse if the type needs to be treated as an integer

Linked Issue:
https://github.com/WeblateOrg/weblate/issues/5969

## Checklist

### DIFF
--- a/docs/locales/docs.pot
+++ b/docs/locales/docs.pot
@@ -1924,11 +1924,19 @@ msgid "Sort JSON keys"
 msgstr ""
 
 #: ../../admin/addons.rst:536
+msgid "``indent_type``"
+msgstr ""
+
+#: ../../admin/addons.rst:536
+msgid "JSON indentation type"
+msgstr ""
+
+#: ../../admin/addons.rst:538
 #: ../../admin/addons.rst:620
 msgid "``indent``"
 msgstr ""
 
-#: ../../admin/addons.rst:536
+#: ../../admin/addons.rst:538
 msgid "JSON indentation"
 msgstr ""
 
--- a/docs/locales/docs.pot
+++ b/docs/locales/docs.pot
@@ -1924,19 +1924,11 @@ msgid "Sort JSON keys"
 msgstr ""
 
 #: ../../admin/addons.rst:536
-msgid "``indent_type``"
-msgstr ""
-
-#: ../../admin/addons.rst:536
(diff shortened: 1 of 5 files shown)


weblate/addons/forms.py
```diff
@@ -183,8 +183,20 @@ def __init__(self, *args, **kwargs):
...
+            ("integer", _("Treat Indentation as an integer")),
+            ("string", _("Treat Indentation as a string")),
+        ),
+        initial="integer",
+        required=True,
+    )
+    indent = forms.CharField(
+        label=_("JSON indentation"),
+        required=True,
+        initial="4",
```
nijel, line 199: I don't think this is a good idea:

* It will be hard for most user to enter tab character here. Also, even when entered it will be invisible later when displaying the configuration.
* It no longer validates the integers, so the add-on will crash in case an invalid value is provided for the `integer` selection above.

The better approach could be to make the choice to define the actual indentation style (spaces / tabs) and the number only identifies number of them.

docs/locales/docs.pot
```diff
@@ -1924,11 +1924,19 @@ msgid "Sort JSON keys"
...
+msgid "JSON indentation type"
+msgstr ""
+
+#: ../../admin/addons.rst:538
 #: ../../admin/addons.rst:620
 msgid "``indent``"
 msgstr ""
 
-#: ../../admin/addons.rst:536
+#: ../../admin/addons.rst:538
```
nijel, line 1939: Changes to the POT files are not necessary, these are updated together with PO files.

weblate/addons/forms.py
ahadcove: Well JSON.dump says to use \t for tab characters, not actual tabs
It also allows you to make indentations based on whatever you choose.

nijel: Yes, "\t" is a tab char as written in Python code. You don't want a literal `\t` there, it would end up as:

```
{
\t"foo": "bar"
}
```

Also, this clearly shows that you have not tested this patch at all...

mustafahemidaa (approved):

Issue #5513, opened 2021-02-24
A project with 1000 components is very slow


tomkolp:
In a project with a large number of components, individual segments load tragically. My project has about 1000 components, it has a lot to do with speed in all components. For comparison, I have created a new project with the same files and it loads quickly. Unfortunately, I cannot afford many projects because propagation does not work between projects.
For speed comparison, I put the same files in a test project, one is shared from a huge project.
[Link](https://weblate.eso-spolszczenie.eu/projects/101286772-test-propagacji-miedzy-projektami/)

nijel:
We definitely need to re-introduce pagination here to avoid loading all the components in such cases.

tomkolp:
I will test how the new docker image will be available.

PR #6089, opened 2021-05-26
[IMP] Default Pagination on Environment Variable

Shide:
## Proposed changes

Component pagination is great, but it's better to be able to tune in each different project (with a default option of 100, like now).

For a project with 500 components, 100 components per page it's ok (only 5 pages).
But for a project with 5000 components, 100 components is very low (50 pages are a lot of pages). So let the admin decide which size is better for his users.

## Checklist

## Other information

Improves #5513 and references #6034

### DIFF
--- a/docs/admin/config.rst
+++ b/docs/admin/config.rst
@@ -1578,7 +1578,7 @@ How many strings to show around the currently translated string. This is just a 
 DEFAULT_PAGE_LIMIT
 ------------------
 
-.. versionadded:: 4.7.1
+.. versionadded:: 4.7
 
 Default number of elements to display when pagination is active.
 
--- a/docs/admin/config.rst
+++ b/docs/admin/config.rst
@@ -1573,6 +1573,15 @@ NEARBY_MESSAGES
 
 How many strings to show around the currently translated string. This is just a default value, users can adjust this in :ref:`user-profile`.
 
+.. setting:: DEFAULT_PAGE_LIMIT
+
+DEFAULT_PAGE_LIMIT
+------------------
+
+.. versionadded:: 4.7.1
+
+Default number of elements to display when pagination is active.
+
 .. setting:: PAGURE_CREDENTIALS
 
 PAGURE_CREDENTIALS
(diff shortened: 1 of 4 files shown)


nijel:
I'm not convinced that this is something what should be user configurable:

* It does have a performance impact on the server, so it's more up to site admin to configure this instead of project admins
* Setting this to big values will also impact users - the huge list of projects will take long to render in the browser
* I don't see much value in differing this per project

Shide:
I understand. You can refuse this PR or tell me to change something.
I need this in my server and I just want to contribute with this PR.

nijel:
I see too big risk in the project-level configuration here for servers where virtually anybody can become a project admin (this pretty much applies to all big projects that run their own Weblate, for example Fedora, or openSUSE).

Would site-wide configuration be good enough for you? Changing it in settings.py (or Docker environment) is something I'd prefer for this...

Shide:
Changed as you like it! Now the only way to change this is by environment and default is 100.

Shide:
@nijel done

weblate/settings_example.py
```diff
@@ -677,6 +677,9 @@
 # Use simple language codes for default language/country combinations
 SIMPLIFY_LANGUAGES = True
 
+# Default number of elements when pagination is active
+DEFAULT_PAGE_LIMIT = 100
```
nijel, line 681: Please move this to weblate/trans/models/_conf.py which contains default values (so that it doesn't break when somebody doesn't update the settings to match new example).

nijel (commented):
Thanks! Can you please also document this new setting in docs/admin/config.rst and docs/admin/install/docker.rst?

weblate/settings_example.py
Shide: Do you want me to append 
`.. versionadded:: 4.7.1 `
to rst admin doc?

docs/admin/config.rst
```diff
@@ -1573,6 +1573,15 @@ NEARBY_MESSAGES
 
 How many strings to show around the currently translated string. This is just a default value, users can adjust this in :ref:`user-profile`.
 
+.. setting:: DEFAULT_PAGE_LIMIT
+
+DEFAULT_PAGE_LIMIT
+------------------
+
+.. versionadded:: 4.7.1
```
nijel, line 1581:
```suggestion
.. versionadded:: 4.7
```

nijel:

Shide:
Thanks! :D

PR #6097, opened 2021-05-27
[IMP] Component name available for searching

Shide:
## Proposed changes

Component name is available on search now.

Searching by component:`slug` is not useful due to the possible lack of knowledge of Users. 
Users don't need to know what `slug` is, so most of the time, they will search for **component name**, because is the only text inside the weblate page that refers to the component.

So, the new way to search for component names will be: 
`component_name`:_substring of component name_

## Checklist

### DIFF
--- a/docs/user/search.rst
+++ b/docs/user/search.rst
@@ -51,7 +51,7 @@ Fields
 ``language:TEXT``
    String target language.
 ``component:TEXT``
-   Component slug or Component name case insensitive search, see :ref:`component-slug` and :ref:`component-name`.
+   Component slug or name case insensitive search, see :ref:`component-slug` and :ref:`component-name`.
 ``project:TEXT``
    Project slug, see :ref:`project-slug`.
 ``changed_by:TEXT``
--- a/docs/user/search.rst
+++ b/docs/user/search.rst
@@ -51,7 +51,7 @@ Fields
 ``language:TEXT``
    String target language.
 ``component:TEXT``
-   Component slug, see :ref:`component-slug`.
+   Component slug or Component name case insensitive search, see :ref:`component-slug` and :ref:`component-name`.
 ``project:TEXT``
    Project slug, see :ref:`project-slug`.
 ``changed_by:TEXT``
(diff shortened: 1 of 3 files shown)


nijel:
Thanks, this is missing few bits, can you please add them?

* Add test case to `weblate/utils/tests/test_search.py`
* Add documentation to `docs/user/search.rst`

nijel:
Unfortunately, this either should not use `search` operator, or would require additional index on MySQL, the tests failed with `MySQLdb._exceptions.OperationalError: (1191, "Can't find FULLTEXT index matching the column list")`, see https://github.com/WeblateOrg/weblate/pull/6097/checks?check_run_id=2694379996#step:14:1035

Shide:
Searching components with the search operators (`:=`, `:`, etc) it's the only aceptable way to search for component name.
As always, users go fast searching and never will write the component name correctly. Almost >90% of the time will search with `:` operator.

Do you see ok to add the column **component.name** to indexes?

nijel:
I can see following options:

* Move the `component_name` lookup to EXACT_FIELD_MAP, this would do a case-insensitive lookup on a whole field.
* Add MySQL fulltext index in a migration (I don't think PostgreSQL index is needed here as it will perform good enough without it and there are usually not that many components to benefit from it). Similar thing (but for both backends) was done in weblate/trans/migrations/0111_index_explanation.py.
* Add CONTAINS_FIELD_MAP which will behave similarly as EXACT_FIELD_MAP, it will just use `icontains` lookup instead of `iexact`.
* Change the `component` lookup in `field_extra` method in a way that it would look for both slug and name. 

The second option sounds like the worst one to me - this will be a rarely used lookup and adding index for that seems merely just add overhead. Which of the others is best depends on the use-case...

Shide:
> * Change the `component` lookup in `field_extra` method in a way that it would look for both slug and name.

I like this option. The new commit reflect this

docs/user/search.rst
```diff
@@ -51,7 +51,7 @@ Fields
 ``language:TEXT``
    String target language.
 ``component:TEXT``
-   Component slug, see :ref:`component-slug`.
+   Component slug or Component name case insensitive search, see :ref:`component-slug` and :ref:`component-name`.
```
nijel, line 54:
```suggestion
   Component slug or name case insensitive search, see :ref:`component-slug` and :ref:`component-name`.
```

PR #6281, opened 2021-07-12
checks/format.py strings reworked

comradekingu:


weblate/checks/format.py
```diff
@@ -370,7 +370,7 @@ def is_position_based(self, string):
         return False
 
     def check_single(self, source, target, unit):
-        """We don't check target strings here."""
+        """Target strings are not checked at this stage."""
```
comradekingu, line 373: Or?

weblate/checks/format.py
```diff
@@ -389,15 +389,15 @@ def format_result(self, result):
...
             ) % ", ".join(self.format_string(x) for x in sorted(set(result["missing"])))
         else:
             if result["missing"]:
-                yield gettext("Following format strings are missing: %s") % ", ".join(
+                yield gettext("The following format strings are missing: %s") % ", ".join(
                     self.format_string(x) for x in sorted(set(result["missing"]))
                 )
             if result["extra"]:
-                yield gettext("Following format strings are extra: %s") % ", ".join(
+                yield gettext("The following format strings are superfluous: %s") % ", ".join(
```
comradekingu, line 400: Or "extra" in some other capacity?

weblate/checks/format.py
```diff
@@ -449,64 +453,64 @@ class PythonFormatCheck(BasePrintfCheck):
 
     check_id = "python_format"
     name = _("Python format")
-    description = _("Python format string does not match source")
+    description = _("Python-format string does not match source")
```
comradekingu, line 456: Alternatively "Mismatching PHP-format string" ? (for this and others)

weblate/checks/format.py
```diff
@@ -389,15 +389,19 @@ def format_result(self, result):
             and set(result["missing"]) == set(result["extra"])
         ):
             yield gettext(
-                "Following format strings are wrongly ordered: %s"
+                "The following format strings in the wrong order: %s"
```
nijel, line 392: are is missing here?

weblate/checks/format.py
nijel: Why "Python-format" instead of "Python format"?

weblate/checks/format.py
comradekingu:
```suggestion
                "The following format strings are in the wrong order: %s"
```

nijel:
Why "Python-format string" instead of "Python format string"?

weblate/checks/format.py
comradekingu: Because it isn't a "Python format-string".

weblate/checks/format.py
nijel:
```suggestion
        """Target strings are not checked in this class."""
```

weblate/checks/format.py
nijel: Everybody seems to call that "Python format string". See for example https://www.gnu.org/software/gettext/manual/html_node/python_002dformat.html or https://podalirius.net/en/articles/python-format-string-vulnerabilities/.

orangesunny:
The code has changed a lot since this PR’s creation. We implemented what’s useful in https://github.com/WeblateOrg/weblate/pull/9568. Possible – Hyphen(ated) skirmishes – can be performed in community discussions.

This PR is a great example why merging of some PRs take such a long time: **there is too much changed within one PR**. Every PR should focus on one thing and should be decently short.
In this case, for future references:
- Submit PR that ads dashes/hyphens in multiple places. This can be merged quickly without dense control of every single word caused by scare of complete meaning change.
- Submit another PR that changes the topic or whole sentences. This will be thoroughly checked, but merged quicker thanks to a sane separation.
- Keep all PRs sanely short and single-topic

weblate/checks/format.py
comradekingu: http://www.grammar-monster.com/lessons/hyphens_in_compound_adjectives.htm
Word-salad like "Python brace format strings" and "Python % format strings" can be avoided.
The passage doesn't use em-dash or correct quoting either. Italics would be an option.

I didn't know about it either till some guy came and complained about it here, and we had the discussion.

orangesunny: Thanks for this comment!

Yes, there are many places where hyphenated compound adjectives should be used in Weblate docs and are not. Yet. Repairing this will take a long time, and we are open to doing it. 
It just does not have the highest priority because:
- It has to be done very cautiously; in this case it is not as easy as putting `-` in front of every `format` word. Some formats’ names are a single word, some multiple words. The editor needs to know the topic and be careful; checking the documentation of that particular format can also be useful.
- I like grammar, I love rules! In Weblate docs, we do not always put grammar first, though. The majority of docs’ readers and writers do not know some grammar rules, even if they are native English speakers. Hyphenated-compound adjectives simply don’t bother many people; on the other hand, some readers are confused by it.
- Many projects (like file formats here) do not use correct grammar in their communities and docs. We simply reused bad practices, and it is hard to overcome it.

Next steps: Let’s do it. Systematically, together, without rush and spiking emotions. What do you think, @comradekingu?

PR #6304, opened 2021-07-19
Spelling: -our, the

comradekingu:


weblate/templates/snippets/login-info.html
```diff
@@ -6,7 +6,7 @@ <h4>{% trans "Web-based <strong>continuous</strong> localization" %}</h4>
 {% include "snippets/about.html" %}
 
 {% if has_legal %}
-<p>{% trans "By registering and signing in to this service you accept our:" %}</p>
+<p>{% trans "By registering and signing in to this service you accept:" %}</p>
```
comradekingu, line 9: This avoids plural or singular "our" with what could be a mix of singular/plural "terms" and "privacy".

fitojb: This is not an issue, linguistically.

comradekingu: https://hosted.weblate.org/translate/weblate/application/en/?checksum=f9ebd5ef7ce14d6b
Hoping people see that comment and that they make amends accordingly is wishful thinking.

fitojb: No, it’s not a problem. https://www.google.com/search?q=%22you+accept+our+terms+and+privacy%22

comradekingu: Correct, _in English_. The problem to avoid is baiting translators into making what is a mistake in some _other languages_.

fitojb: The solution to that is not to dumb down the original string, though.

nijel:
I've added explanation to https://hosted.weblate.org/translate/weblate/application/en/?checksum=f9ebd5ef7ce14d6b, that is IMHO better approach.

PR #6305, opened 2021-07-19
Spelling: address, e-mail, users, could not

comradekingu:


comradekingu:
Build complains about no big "W" in a URL…?
This was however changed (part-way) in https://github.com/comradekingu/Weblate/commit/68db7349161d2a5b995f2382941b861440f6a050

weblate/templates/addons/discovery_preview.html
```diff
@@ -38,7 +38,7 @@
 {% endfor %}
 {% endif %}
 {% if not matches_deleted and not matches_created and not matches_matched %}
-<tr><th colspan="3" class="danger">{% trans "No components were matched!" %}</th></tr>
+<tr><th colspan="3" class="danger">{% trans "No such components were found!" %}</th></tr>
```
nijel, line 41:
```suggestion
<tr><th colspan="3" class="danger">{% trans "No matching components were found!" %}</th></tr>
```

orangesunny (approved):
Great, thanks for the contribution!

weblate/trans/tests/test_acl.py
```diff
@@ -122,7 +122,7 @@ def test_invite_invalid(self):
             follow=True,
         )
         # This error comes from Django validation
-        self.assertContains(response, "Enter a valid email addres")
+        self.assertContains(response, "Enter a valid e-mail address")
```
nijel, line 125: Changing assertion without changing the actual message breaks testsuite. As described by the comment, the error comes from Django here.

weblate/trans/tests/test_acl.py
```diff
@@ -136,7 +136,7 @@ def test_invite_existing(self):
             },
             follow=True,
         )
-        self.assertContains(response, "User with this E-mail already exists")
+        self.assertContains(response, "A user with this e-mail already exists")
```
nijel, line 139: ...and this one as well.

weblate/trans/tests/test_acl.py
comradekingu: Found the error when I loaded githubusercontent.
Would make sense to me if weblate/weblate/trans/views/acl.py not had any such message to return.
https://hosted.weblate.org/translate/weblate/website/en/?checksum=f495c8cb6bf18bc9 is the changed statement, albeit in different files, with an ending dot (which I think is fine).
Is it in a different repo? Will I find the location in the file? Do I have to read https://docs.djangoproject.com/en/2.2/topics/testing/tools/#django.test.SimpleTestCase.assertContains

nijel: This one is at https://github.com/django/django/blob/fd999318ad78613227cdb7c5656345d9e216802b/django/core/validators.py#L161

nijel:
I've made changes to use custom validation messages in 0089a024dc9a7e8a49ac450f86292e3ad1cfe641

Issue #5962, opened 2021-05-04
Add "msgfmt --check-accelerators" to quality checks


luebbe:
Gettext msgfmt offers the --check-accelerators option to check if both, msgid and msgstr have an accelerator key.

It would be nice to have this available as an additional quality check in Weblate.

luebbe:
If I wanted to tackle this task, where should it belong? An additional class in `checks/consistency.py` would come to mind.

nijel:
You can create new file for that, that's probably best approach.

Issue #6142, opened 2021-06-05
Keyboard accelerator mark check


comradekingu:
If the source has underlines for strings, the translation should too

msgfmt -c --check-accelerators="_" --statistics po/<languagecode>.po
gives a list of missing entries, but in Weblate this goes unnoticed.

nijel:
Duplicate of #5962

PR #6451, opened 2021-08-20
Add "msgfmt --check-accelerators" to quality checks

luebbe:
## Proposed changes

Implements #5962 - Add "msgfmt --check-accelerators" to quality checks

## Checklist

- [ ] Lint and ...
- [x] ...unit tests pass locally with my changes.

### DIFF
--- a/docs/user/checks.rst
+++ b/docs/user/checks.rst
@@ -60,7 +60,7 @@ Checks that source and translation both contain zero or one accelerator key.
 
 .. note::
 
-    A string like 'Walter & Sons' may be translated without the ampersand and 
+    A string like 'Walter & Sons' may be translated without the ampersand and
     may trigger a false positive.
 
 .. _check-bbcode:
--- a/docs/user/checks.rst
+++ b/docs/user/checks.rst
@@ -42,6 +42,27 @@ Translation checks
 Executed upon every translation change, helping translators maintain
 good quality translations.
 
+.. _check-accelerators:
+
+Accelerator keys
+~~~~~~~~~~~~~~~~
+
+.. versionadded:: 4.8
+
+:Summary: Source and translation do not both contain an accelerator key
+:Scope: translated strings
+:Check class: ``weblate.checks.duplicate.AcceleratorKeyCheck``
+:Flag to ignore: ``ignore-accelerators``
+
+Accelerator keys are used to speed up UI interactions. The standard keys are
+'&' (Windows) and '_' (Linux).
+Checks that source and translation both contain zero or one accelerator key.
+
+.. note::
+
+    A string like 'Walter & Sons' may be translated without the ampersand and 
+    may trigger a false positive.
+
 .. _check-bbcode:
 
 BBcode markup
--- a/weblate/checks/chars.py
+++ b/weblate/checks/chars.py
@@ -41,7 +41,7 @@ FRENCH_PUNCTUATION = {";", ":", "?", "!"}
 class AcceleratorKeyCheck(TargetCheck):
     """Check for inconsistent accelerator keys."""
 
-    check_id = "accelerator key"
+    check_id = "accelerators"
     name = _("Accelerator key")
     description = _("Source and translation do not both contain an accelerator key")
     
--- a/weblate/checks/chars.py
+++ b/weblate/checks/chars.py
@@ -38,18 +38,22 @@ KASHIDA_CHARS = (
 )
 FRENCH_PUNCTUATION = {";", ":", "?", "!"}
 
+
 class AcceleratorKeyCheck(TargetCheck):
     """Check for inconsistent accelerator keys."""
 
     check_id = "accelerators"
     name = _("Accelerator key")
     description = _("Source and translation do not both contain an accelerator key")
-    
+
     def _check_accelerator(self, source, target, key):
         return (target.count(key) > 1) or (source.count(key) != target.count(key))
 
     def check_single(self, source, target, unit):
-        return self._check_accelerator(source, target, "&") or self._check_accelerator(source, target, "_")
+        return self._check_accelerator(source, target, "&") or self._check_accelerator(
+            source, target, "_"
+        )
+
 
 class BeginNewlineCheck(TargetCheck):
     """Check for newlines at beginning."""
--- a/weblate/checks/chars.py
+++ b/weblate/checks/chars.py
@@ -38,6 +38,15 @@ KASHIDA_CHARS = (
 )
 FRENCH_PUNCTUATION = {";", ":", "?", "!"}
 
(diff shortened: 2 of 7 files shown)


nijel (commented):
I think this will cause many false positives. It should be either fully opt-in or only enabled for formats which support accelerators.

luebbe:
> 
> 
> I think this will cause many false positives. It should be either fully opt-in or only enabled for formats which support accelerators.

Can you give me a hint how to do this? I'm still trying to find my way around.

nijel:
Just add `default_disabled = True` to the check class for opt-in behavior.

weblate/checks/chars.py
```diff
@@ -39,6 +39,22 @@
 FRENCH_PUNCTUATION = {";", ":", "?", "!"}
 
 
+class AcceleratorKeyCheck(TargetCheck):
+    """Check for inconsistent accelerator keys."""
+
+    check_id = "accelerators"
+    name = _("Accelerator key")
+    description = _("Source and translation do not both contain an accelerator key")
```
nijel, line 47: For opt-in behavior:

```suggestion
    description = _("Source and translation do not both contain an accelerator key")
    default_disabled = True
```

luebbe:
So easy, thank you ...

luebbe:
A future improvement would be to make the hotkeys configurable as a list or a string over which the check iterates.

docs/user/checks.rst
```diff
@@ -42,6 +42,27 @@ Translation checks
 Executed upon every translation change, helping translators maintain
 good quality translations.
 
+.. _check-accelerators:
+
+Accelerator keys
+~~~~~~~~~~~~~~~~
+
+.. versionadded:: 4.8
```
nijel, line 50:
```suggestion
.. versionadded:: 4.8.1
```

4.8 is already out :-)

weblate/checks/chars.py
```diff
@@ -39,6 +39,23 @@
...
+    default_disabled = True
+
+    def _check_accelerator(self, source, target, key):
+        return (target.count(key) > 1) or (source.count(key) != target.count(key))
```
nijel, line 51: Gettext check behaves slighly differently:

* We are only interested in msgids that contain exactly one '&'. 
* Count the number of '&' in msgstr, but ignore '&&'. 

See https://git.savannah.gnu.org/gitweb/?p=gettext.git;a=blob;f=gettext-tools/src/msgl-check.c;h=32c21adf546cae496b0dcccf6cc235773b8a30e8;hb=HEAD#l708

weblate/checks/chars.py
```diff
@@ -39,6 +39,23 @@
...
+
+    def check_single(self, source, target, unit):
+        return self._check_accelerator(source, target, "&") or self._check_accelerator(
+            source, target, "_"
+        )
```
nijel, line 56: Gettext checks for `&` only (unless something else is specified on the command line).

luebbe: I was checking the interwebs for different accelerators and came across this: http://docs.translatehouse.org/projects/localization-guide/en/latest/guide/translation/accelerators.html#accelerators. I knew only `&` and `_` until now, where `&`is the most common. We could add `~` too and always check for all three or we could add an `accelerator_keys` setting, make `&` the default and let the user override it. But would that setting be global, per project or even per component?
How big would the risk for false positives be, if we always check for all three?

weblate/checks/chars.py
luebbe: I was trying to catch the case that more than one accelerator is in a translation, but this may trigger a false positive on some markup like _underscores around this text_.

weblate/checks/chars.py
nijel: The flag can take argument as well. So `accelerator` would check for `&` and `accelerator:_` for `_`.

weblate/checks/chars.py
nijel: Is there really something which uses two accelerators in the string? The Gettext implementation clearly looks only for one.

luebbe: Never, this is why I wanted to flag it as a possible error. If the source or translation has more than one accelerator, there's probably something wrong. This check will pass if there's exactly zero or one accelerator in source and translation, ignoring escapes, because we can't know them all.

weblate/checks/chars.py
luebbe: is there an example somewhere how flags with arguments work?

weblate/checks/chars.py
nijel: The gettext logic is more complex:

- it doesn't check strings where source doesn't have single accelerator
- it supports escaping (I guess, it ignores double accelerator)

weblate/checks/chars.py
nijel: For example the PlaceholderCheck:

- you subclass from TargetCheckParametrized
- param_type property defines parsing (as this needs a string, `return single_value_flag(lambda x: x)` should do it)
- you define check_target_params instead of check_single, it get's parameter value as last arg
- I'm not sure if this can gracefully handle defaults when no param is specified though...

weblate/checks/chars.py
luebbe: The translationhouse docs that I linked below, mention more than one way of escaping the accelerator. All these depend on the client renderer and are independent of the translation file format. If there are still other ways of escaping, we don't know them. This is why I suggested to ignore escaping and silence false positives using the flag.
OTOH if we pass the accelerator as parameter, as discussed below, we could also pass the escaped accelerator.
The default could be `&` and `&&`, which probably covers most of the cases.
But I could be wrong, because https://github.com/WeblateOrg/weblate/issues/6142 and https://github.com/Gnucash/gnucash/pull/1039 both explicitely mention the `_` as accelerator.

weblate/checks/chars.py
luebbe: Thanks, I'll take a look

weblate/checks/chars.py
nijel: I've no real-world experience with this - I'm just describing what Gettext does here, as the original issue was to mimic its behavior. So, it doesn't have to work the way I describe, it just should be considered as somebody took the effort to implement it this way there.

luebbe: Yes, to mimic the gettext behaviour was the original intent. Since then we learned that there's more to it and gettext is just one approach, so I'd like to implement an "as good as possible" solution without overcomplicating the implementation/configuration.

luebbe:
I think I should close this PR for the time being. I don't have time to work on it right now and it probably would be better to start over again taking the input provided here into account.

PR #6459, opened 2021-08-23
Update utils.py

Meithal:
This makes it work on windows.

## Proposed changes

git-http-backend on windows has a .exe suffix, this is a simple fix.

## Checklist

What kind of test do you think would be needed?

- [x ] Lint and unit tests pass locally with my changes.
- [x ] I have squashed my commits into logic units.
- [x ] I have described the changes in the commit messages.

### DIFF
--- a/weblate/gitexport/utils.py
+++ b/weblate/gitexport/utils.py
@@ -47,9 +47,9 @@ def find_git_http_backend():
         pass
 
     for path in GIT_PATHS:
-        names = ("git-http-backend", "git-http-backend.exe")
-        for name in names:
-            name = os.path.join(path, name)
+        exes = ("git-http-backend.exe", "git-http-backend")
+        for exe in exes:
+            name = os.path.join(path, exe)
             if os.path.exists(name):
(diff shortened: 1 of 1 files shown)


weblate/gitexport/utils.py
```diff
@@ -47,8 +47,10 @@ def find_git_http_backend():
         pass
 
     for path in GIT_PATHS:
-        name = os.path.join(path, "git-http-backend")
-        if os.path.exists(name):
-            find_git_http_backend.result = name
-            return name
+        names = ("git-http-backend", "git-http-backend.exe")
+        for name in names:
+            name = os.path.join(path, "git-http-backend")
```
nijel, line 52: The name variable from the loop doesn't seem to be used, you probably had something like this in mind:

```suggestion
            name = os.path.join(path, name)
```

Meithal: Yes, good catch, my bad, I will fix once I can access a computer

nijel:

Issue #4974, opened 2020-12-04
Add project API tokens


nijel:
**Is your feature request related to a problem? If so, please describe.**

Add support for generating API tokens at project level and granting them access to certain features.

**Describe the solution you'd like**

* API tokens management at project level
* In background it generates bot users and adds them to the project
* These bot users will have NULL email and will be blocked from authenticating
* They will be shown in a user management to adjust permissions
* The username would be generated, something like project_{id}_bot{optional_unique_suffix}

**Describe alternatives you've considered**
Creating real users is an option as well, but that sucks in team management.

**Additional context**
See https://docs.gitlab.com/ee/user/project/settings/project_access_tokens.html for documentation how GitLab does it.

Also see #1066

JochenDiekenbrock:
I'm currently evaluating weblate for working with our existing private Gitlab.

For security reasons, we do not want to have *one* weblate user that has access to all repositories. Ideally the weblate user would exist in gitlab as well and weblate would push the translations with the gitlab user. If I didn't miss something, that's not possible in weblate.

Is my understanding correct, that this issue implements managing Gitlab access tokens on the project or even component level? And the generated username would be the user pushing changes to Gitlab?

Any idea, when this feature would be available?

nijel:
No, this issue is about having project wide access tokens for Weblate API. Right now the API token is for users only.

For accessing remote repos the proper solution is to have native connection to the service, ask for repository access and store these tokens. For GitHub, this is requested in https://github.com/WeblateOrg/weblate/issues/1688.

JochenDiekenbrock:
Ah ok, thank you Michal!

derkoe:
I have created a draft implementation that works so far: https://github.com/WeblateOrg/weblate/compare/main...derkoe:project-tokens

What is still open is the UI to create/delete the tokens.

nijel:
@derkoe Looks good so far. Please submit a pull request once you are ready for feedback (it doesn't have to be finished).

PR #6469, opened 2021-08-24
feat: add projects tokens

derkoe:
## Proposed changes

Add support for generating API tokens at project level. The tokens have a expiry date and can be deleted individually. With the current implementation the tokens have full access to the project.

See also #4974

## Checklist

## Other information

Possible extensions which are not implemented:
- It may be a good idea to separate read and write permissions for tokens
- Rate limiting on API token authentications for failed auths


nijel:
Thanks for the updates, it now looks better. Can you please add tests for this feature?

derkoe:
Yes, I'll try to write some tests. 

I also wanted to set the permissions on the token (Read translations, write translations and admin). Do you have an idea how this can be implemented?

nijel:
The `has_perm` method should handle this. But maybe it's better to merge the tokens first and then extend it to support this.

nijel:
Let's keep this open as the code is definitely something we want to merge. @derkoe any chance you will get to writing the tests?

derkoe:
I've added a test case (`test_projecttoken.py`) but when creating the token the test does not get the message "Token has been created: ..." in the response. Do I have to do something special for Django messages to work?

weblate/trans/tests/test_projecttoken.py
```diff
@@ -0,0 +1,31 @@
...
+        self.project.access_control = Project.ACCESS_PRIVATE
+        self.project.save()
+        self.access_url = reverse("manage-access", kwargs=self.kw_project) + "#api"
+        self.token_one = "This is token 1"
+
+    def create_token(self):
+        self.make_manager()
+        response = self.client.post(
+            reverse("create-project-token", kwargs=self.kw_project),
+            {"name": self.token_one, "expires": "2999-12-99"},
```
nijel, line 21: The date does not look valid.

nijel (commented):
Nothing special is needed, these should work without any additional changes. Most likely it fails on the form validation because of invalid expiry date...

weblate/trans/tests/test_projecttoken.py
derkoe: Thx fixed that

weblate/templates/manage-access.html
```diff
@@ -202,15 +202,83 @@ <h4 class="panel-title">
...
         <td>
-            {{ user.auth_token.key }}
-            <button type="button" class="btn btn-link btn-xs" data-clipboard-text="{{ user.auth_token.key }}" data-clipboard-message="{% trans "API key copied to clipboard." %}" title="{% trans "Copy to clipboard" %}">{% icon "copy.svg" %}</button>
+          <a href="#" data-toggle="modal" data-target="#delete_token_{{ token.id }}" class="btn btn-link btn-xs red" title="{% trans "Delete" %}">
+            {% icon 'delete.svg' %}
+          </a>
+          <form action="{% url "delete-token" project=object.slug %}" method="post" class="inlineform">
+            {% csrf_token %}
+            <input type="hidden" name="token" value="{{ token.id }}" />
+            <div class="modal fade" tabindex="-1" role="dialog" id="delete_token_{{ token.id }}">
```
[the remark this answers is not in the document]
derkoe: I'm not sure why a hidden input should have a label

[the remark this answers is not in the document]
nijel: It's about dialog, not about the hidden input, see https://udn.realityripple.com/docs/Web/Accessibility/ARIA/Roles/dialog_role

This is nice to have, but not really necessary as we miss this on nearly every dialog...

weblate/trans/models/projecttoken.py
```diff
@@ -0,0 +1,43 @@
...
+    name = models.CharField(verbose_name=gettext_lazy("Name"), max_length=100)
+    token = models.CharField(
+        verbose_name=gettext_lazy("Token value"),
+        max_length=100,
+        help_text=gettext_lazy("Token value to use in API"),
+    )
+    expires = models.DateTimeField(verbose_name=gettext_lazy("Expires"))
+
+    class Meta:
+        unique_together = [("project", "token")]
```
nijel, line 37: Maybe project, name should be unique as well? Having more tokens with same name will be confusing.

weblate/trans/forms.py
```diff
@@ -2380,3 +2388,23 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.helper = FormHelper(self)
         self.helper.form_tag = False
+
+
+class ProjectTokenDeleteForm(forms.Form):
+    token = forms.CharField(required=True, widget=forms.HiddenInput)
```
nijel, line 2394: Perhaps use `forms.IntegerField` instead? Even better would be `forms.ModelChoiceField`, the form then returns directly the object, and you can skip additional validation in the view.

weblate/trans/models/projecttoken.py
```diff
@@ -0,0 +1,43 @@
...
+
+class ProjectToken(models.Model):
+    """Project token for API access."""
+
+    project = models.ForeignKey("Project", on_delete=models.deletion.CASCADE)
+    name = models.CharField(verbose_name=gettext_lazy("Name"), max_length=100)
+    token = models.CharField(
+        verbose_name=gettext_lazy("Token value"),
+        max_length=100,
+        help_text=gettext_lazy("Token value to use in API"),
```
nijel, line 32: Perhaps provide a default value so that token is always initialized?

```suggestion
        help_text=gettext_lazy("Token value to use in API"),
        default=token_hex,
```

weblate/trans/views/acl.py
```diff
@@ -252,14 +259,60 @@ def manage_access(request, project):
...
+
+    if not request.user.has_perm("project.edit", obj):
+        raise PermissionDenied()
+
+    form = ProjectTokenCreateForm(request.POST)
+
+    if form.is_valid():
+        token_value = token_hex()
+        ProjectToken.objects.create(project=obj, token=token_value, **form.cleaned_data)
+        messages.info(request, _("Token has been created: %s") % token_value)
```
nijel, line 314: I'd move the token generating to the model via the default value, see above.

```suggestion
        token = obj.projecttoken_set.create(**form.cleaned_data)
        messages.info(request, _("Token has been created: %s") % token.token)
```

nijel (commented):
Great! I can still see some improvements, it would be great if you could look at my comments.

weblate/trans/models/projecttoken.py
derkoe: Added that

weblate/trans/forms.py
derkoe: Changed to IntegerField - I had ModelChoiceField but this did not seem to work.

weblate/trans/models/projecttoken.py
derkoe: Done that.

weblate/trans/views/acl.py
derkoe: Done this as well.

derkoe:
@nijel I have added all changes now - Flake8 is not happy with a line I haven't changed:
```
Error: weblate/trans/models/change.py:54:1: D200 One-line docstring should fit on one line with quotes
```

nijel:
@derkoe I've noticed that as well, should be fixed in 258ec08e360342febdfafdba57ba3462d2f3b85f.

nijel:

I've made the following changes on top of this:

* a0b12e42a0 docs: Document project API tokens
* 5bb03e64bb projecttoken: Use specific prefix for project tokens
* 708cea8663 projecttoken: Use ModelChoiceField instead of doing the validation manually
* 25b208356e projecttoken: Add copy button when token is created
* 7702111bad projecttoken: Properly validate uniqueness of the token name

PR #6573, opened 2021-09-21
Fix Docker paths (python3.7 -> 3.9)

SirStendec:
## Proposed changes

The Docker image for weblate was recently updated to use a newer version of Debian, and a newer version of Python (https://github.com/WeblateOrg/docker/commit/a86df5b527e20cd86df4596d6d4a59ce79da62bb). There are still docker files in the main repository assuming the use of Python 3.7, however.

This PR updates the paths directly related to docker to use `/python3.9/` rather than `/python3.7/`. Without these changes, trying to run a local development Docker image with `rundev.sh` will not actually use your local version of weblate.

## Checklist

## Other information

There are also some strings in the documentation that refer people to Python 3.7, but as those aren't directly related to Docker I've left them as they are.

### DIFF
--- a/dev-docker/docker-compose.yml
+++ b/dev-docker/docker-compose.yml
@@ -25,7 +25,7 @@ services:
       # Local data directory
       - $PWD/data:/app/data
       # Override Weblate with local dir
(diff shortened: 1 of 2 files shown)


nijel:

PR #6591, opened 2021-09-25
Spelling: The component

comradekingu:


weblate/trans/tests/test_component.py
```diff
@@ -631,7 +631,7 @@ def test_autolock(self):
         self.assertEqual(change.get_action_display(), "Component locked")
         self.assertEqual(
             change.get_details_display(),
-            "Component was automatically locked because of an alert.",
+            "The component was automatically locked because of an alert.",
```
nijel, line 634: This changes only the test. Please change the actual string as well in weblate/trans/models/check.py.

comradekingu: That file isn't there…

nijel:

PR #6605, opened 2021-09-29
Add WEBLATE_PRIVACY_URL to docker settings

MrGeneration:
## Proposed changes

As per documentation: 
https://docs.weblate.org/en/weblate-4.8.1/admin/install/docker.html#envvar-WEBLATE_PRIVACY_URL

should be available to docker installations but is not working (nor available) as per 4.8.1.

## Checklist

## Other information

`.py` files throw `EXE002` errors on flake8 check, by the sheer number I'm guessing this is supposed to be correct.


nijel:

PR #6622, opened 2021-10-01
Add missing WEBLATE_AUTO_UPDATE option to docs

MrGeneration:
This section also hints to the setting type which will help administrators understanding the scope.

## Proposed changes

The docker part of the documentation is missing the `WEBLATE_AUTO_UPDATE` setting which can be quite useful.

## Checklist

- [x] I have added documentation to describe ~~my~~a feature.

## Other information

Things to consider in terms of documentation quality:
I had to learn the hard and manual fiddling fiddling way that a lot of settings work different or not to their full extend as the source installation versions. As example: `full` is not a valid setting in docker context because the setting is evaluated by a function that looks for boolean kind of strings.

Right now this forces administrators to dig through the code which can slow down deployment processes. 
Possibly the documentation could point out the setting type. Don't know.

### DIFF
--- a/docs/admin/install/docker.rst
+++ b/docs/admin/install/docker.rst
@@ -1248,6 +1248,16 @@ Example SSL configuration:
         :ref:`production-email`,
         :setting:`django:EMAIL_BACKEND`
 
+.. envvar:: WEBLATE_AUTO_UPDATE
+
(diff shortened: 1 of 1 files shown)


nijel:

PR #6641, opened 2021-10-05
add reference to translate-toolkit

JohnRDOrazio:
While upgrading from weblate 4.7 to 4.8.1, I got some warnings about missing packages and formats such as fluent when issuing `weblate check --deploy`. I was able to get rid of these warnings by updating translate-toolkit.

## Proposed changes

Add information about updating translate-toolkit to the weblate update page.

## Checklist

### DIFF
--- a/docs/admin/upgrade.rst
+++ b/docs/admin/upgrade.rst
@@ -39,6 +39,12 @@ work, but is not as well tested as single version upgrades.
 
       pip install -U Weblate
 
+   And make sure **translate-toolkit** is up to date:
+
+   .. code-block:: sh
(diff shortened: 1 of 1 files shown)


JohnRDOrazio:
Thanks to @nijel for pointing me in the right direction in the related discussion #6629

nijel:
Thinking more about it, this is not what we want - this pulls in optional does from the translate-toolkit which Weblate will never use. What would make more sense is to mention `Weblate[all]` instead. I've at least linked that from the upgrade docs in c2e17bb.

Issue #5957, opened 2021-05-03
Add an API endpoint for "download translations as Zip"


luebbe:
**Is your feature request related to a problem? If so, please describe.**

right now it is possible to download a complete set of translations via;

`GET https://<weblate>/download/<project>/<component>/?format=zip`

This doesn't work anymore when `WEBLATE_REQUIRE_LOGIN` is enabled. One could possibly add a matching regex to `WEBLATE_LOGIN_REQUIRED_URLS_EXCEPTIONS`, but this would still allow anybody who can construct the above URL to download the translations without authorization (IIUC).

**Describe the solution you'd like**

There is already an API endpoint to download individual languages, but nothing for the whole component. The idea is to send only one request and get everything in one go.

We have:
`GET /api/translations/(string: project)/(string: component)/(string: language)/file/`

An API endpoint like:
`GET /api/translations/(string: project)/(string: component)/file/(string:format)`
or, if that doesn't make sense:
`GET /api/translations/(string: project)/(string: component)/zip`

where you have to pass in your authorization token and can only download translations when this is ok.

Vogel612:
The endpoint specification as outlined would require sweeping changes to the way the API is created within weblate. A much easier setup is available if the download is exposed on the components API instead.

Attaching the endpoint there sidesteps the issue that all translation API endpoints ***require*** the specification of `project`, `component` and `language` (implemented as the `lookup_fields` that are included into the router registration via the `MultipleFieldsMixin` and the `WeblateRouter`.

I am currently working on a patch that would add this endpoint, expect a PR by the end of this week (and feel free to poke me if I fail on that commitment)

PR #6744, opened 2021-11-02
Feature/api/download translations

Vogel612:
## Proposed changes

Introduces a new endpoint on components API's translations to allow downloading a component's translations as a ZIP file as requested by #5957

## Checklist

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1322,6 +1322,9 @@ Components
 
 .. http:get:: /api/components/(string:project)/(string:component)/file/
 
+
+    .. versionadded:: 4.9
+
     Downloads all available translations associated with the component as an archive file using the requested format.
 
     :param project: Project URL slug
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1320,6 +1320,17 @@ Components
     :type component: string
     :>json array results: array of component objects; see :http:get:`/api/changes/(int:id)/`
 
+.. http:get:: /api/components/(string:project)/(string:component)/file/
+
+    Downloads all available translations associated with the component as an archive file using the requested format.
+
+    :param project: Project URL slug
+    :type project: string
+    :param component: Component URL slug
+    :type component: string
+
+    :query string format: The archive format to use; If not specified, defaults to ``zip``; Supported formats: ``zip``
+
 .. http:get::  /api/components/(string:project)/(string:component)/screenshots/
 
     Returns a list of component screenshots.
--- a/weblate/api/tests.py
+++ b/weblate/api/tests.py
@@ -1839,6 +1839,17 @@ class ComponentAPITest(APIBaseTest):
         )
         self.assertEqual(response.headers["content-type"], "application/zip")
 
+    def test_download_translation_zip_prohibited(self):
+        project = self.component.project
+        project.access_control = Project.ACCESS_PROTECTED
+        project.save(update_fields=["access_control"])
+        response = self.do_request(
+            "api:component-download-archive",
+            self.component_kwargs,
+            method="get",
+            code=403,
+        )          
+
     def test_links(self):
         self.do_request(
             "api:component-links",
--- a/weblate/api/tests.py
+++ b/weblate/api/tests.py
@@ -1843,7 +1843,7 @@ class ComponentAPITest(APIBaseTest):
         project = self.component.project
         project.access_control = Project.ACCESS_PROTECTED
         project.save(update_fields=["access_control"])
-        response = self.do_request(
+        self.do_request(
             "api:component-download-archive",
             self.component_kwargs,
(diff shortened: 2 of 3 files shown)


Vogel612:
Created as a draft PR pending me figuring out how to run (and create) tests. I'd also be happy to move the endpoint elsewhere if the maintainers can give me any reason whatsoever :)

weblate/api/views.py
```diff
@@ -951,6 +952,23 @@ def delete_links(self, request, project__slug, slug, project_slug):
         instance.links.remove(project)
         return Response(status=HTTP_204_NO_CONTENT)
 
+    @action(detail=True, methods=["get"], url_path="translations/zip")
```
nijel, line 955: Maybe it would be better to use `/file/` to be consistent with other download endpoints? But I'm not convinced whether it's better...

```suggestion
    @action(detail=True, methods=["get"], url_path="file")
```

nijel:
Running tests locally is described here: https://docs.weblate.org/en/latest/contributing/tests.html

weblate/api/views.py
nijel: Okay, let's make it consistent with other downloads:

* call the endpoint file
* accept format in the URL and pass it to download_multi (it is currently not used, but will allow extending in future)

weblate/api/views.py
```diff
@@ -951,6 +952,23 @@ def delete_links(self, request, project__slug, slug, project_slug):
...
+        instance = self.get_object()
+        if not request.user.has_perm("translation.download", instance):
+            self.permission_denied(
+                request, "Can not download all translations for the component"
+            )
+
+        instance.commit_pending("download", None)
+        return download_multi(
+            instance.translation_set.all(),
+            "zip",
```
nijel, line 968: Avoid hardcoding this (even though there are no other formats supported right now):

```suggestion
            "request.query_params.get("format", zip"),
```

weblate/api/views.py
```diff
@@ -951,6 +952,23 @@ def delete_links(self, request, project__slug, slug, project_slug):
...
+        #  that here because the lookup for the component is different
+        instance = self.get_object()
+        if not request.user.has_perm("translation.download", instance):
+            self.permission_denied(
+                request, "Can not download all translations for the component"
+            )
+
+        instance.commit_pending("download", None)
+        return download_multi(
+            instance.translation_set.all(),
```
nijel, line 967: I've changed the API in a6e9883101 to cover committing as well (because the commit logic is more complex since dd5a3293a97f9cf328b0b3d62030f7011f95a8d2):

```suggestion
        return download_multi(
            instance.translation_set.all(),
            [instance],
```

weblate/api/tests.py
```diff
@@ -1829,6 +1829,16 @@ def test_create_translation_prohibited(self):
...
 
+    def test_download_translation_zip_ok(self):
+        response = self.do_request(
+            "api:component-download-archive",
+            self.component_kwargs,
+            method="get",
+            code=200,
+            superuser=True,
+        )
+        self.assertEqual(response.headers["content-type"], "application/zip")
```
nijel, line 1840: Can you please add test for the permission denied case as well?

nijel:
```suggestion
        self.assertEqual(response.headers["content-type"], "application/zip")

    def test_download_translation_zip_prohibited(self):
        project = self.component.project
        project.access_control = Project.ACCESS_PROTECTED
        project.save(update_fields=["access_control"])
        response = self.do_request(
            "api:component-download-archive",
            self.component_kwargs,
            method="get",
            code=403,
        )          
```

docs/api.rst
```diff
@@ -1320,6 +1320,17 @@ Components
     :type component: string
     :>json array results: array of component objects; see :http:get:`/api/changes/(int:id)/`
 
+.. http:get:: /api/components/(string:project)/(string:component)/file/
+
+    Downloads all available translations associated with the component as an archive file using the requested format.
```
nijel, line 1325:
```suggestion

    .. versionadded:: 4.9

    Downloads all available translations associated with the component as an archive file using the requested format.
```

weblate/api/tests.py
```diff
@@ -1829,6 +1829,27 @@ def test_create_translation_prohibited(self):
...
+
+    def test_download_translation_zip_prohibited(self):
+        project = self.component.project
+        project.access_control = Project.ACCESS_PROTECTED
+        project.save(update_fields=["access_control"])
+        response = self.do_request(
```
nijel, line 1846:
```suggestion
        self.do_request(
```

Vogel612:
Thanks for fixing this up, I was having internet troubles at home and didn't get back into the office until today.

PR #6846, opened 2021-11-22
clarifying the process for translators

orangesunny:
adjustment according to https://github.com/WeblateOrg/weblate/discussions/6790

## Checklist

- [ ] Lint and unit tests pass locally with my changes.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have added documentation to describe my feature.
- [ ] I have squashed my commits into logic units.
- [x] I have described the changes in the commit messages.

PR #6855, opened 2021-11-23
Refined libre trial message

comradekingu:


keunes:
> Save your work on this libre project till it is fully set up and accepted. Its libre license may change.

* 'Save your work' could be interpreted in two ways: not doing any work yet, or making sure work is stored (saved) properly. For new translators it might not be clear which of the two is implied.
* "Its libre license" implies that a license has been set, which might not be the case for trial projects.
* It does not signal to the user why this message is displayed (because of the 'trial' status), making it harder to understand the implications.

orangesunny:
This will be solved in https://github.com/WeblateOrg/weblate/pull/6846

comradekingu:
@orangesunny No, it won't. By a wide margin.
It is however solved here.

comradekingu:
> Save your work on this libre project till it is fully set up and accepted. Its libre license may change.

   > 'Save your work' could be interpreted in two ways: not doing any work yet, or making sure work is stored (saved) properly. For new translators it might not be clear which of the two is implied.

That would have to be "Save yourself from doing any work on this libre project." But it is a possible misinterpretation.
The other interpretation is a call to save ones work (with the preposition "on") ~in the project, till it is set up, and then I guess do something else.   

> "Its libre license" implies that a license has been set, which might not be the case for trial projects.
    It does not signal to the user why this message is displayed (because of the 'trial' status), making it harder to understand the implications.

Fair point. "Its selected libre licensing" would solve that.
However, 6.2 of the [terms and conditions](https://hosted.weblate.org/legal/terms/), says if a license is not set, it follows the project license.
It is possible that there is no license in the project either, but at that point who is going to spend their time translating the project?
I would report such projects for removal.

I check each and every project for license compliance, and quickly get issues fixed, so there is nothing to be gained over this approach.

Strictly speaking any work is licensed under the license set at the time, and relicensing must be agreed upon if it isn't part of the particular license. In reality nobody gets angry about it, and it has never happened that someone has taken issue having contributed to something with a license set in error.
It is possible to bait someone into contributing to a strong license, and then make it a weaker one, but who would do that?
It isn't worth it to explain the minutia of licensing over something that isn't a problem.

It is possible to "bait" someone into contributing under a weak license, but then make it a stronger one. That is legal.
If the user then doesn't know what weak licenses are, that is on the user. Also, the user could easily check the actual project license.

weblate/templates/snippets/project/billing-libre-trial.html
```diff
@@ -8,6 +8,6 @@
   {% if user_can_view_billing %}
     <a href="{{ object.billing.get_absolute_url }}" class="btn btn-primary pull-right flip">{% trans "Request approval for libre hosting" %}</a>
   {% endif %}
-  {% trans "This project is in the trial period, so be cautious before you contribute. Things like license or repository can still change before being approved as a libre project." %}
+  {% trans "Save your work on this libre project till it is fully set up and accepted. Its libre license may change." %}
```
comradekingu, line 11: ↑ Save your work on this libre project till the libre license and everything else is fully set up and accepted.

comradekingu:
@keunes What do you think about:

Save your work on this libre project manually till its libre license and configuration is [accepted as such](https://hosted.weblate.org/hosting/).
or
Save your work on this libre project manually till its libre license and configuration [is accepted](https://hosted.weblate.org/hosting/)

keunes:
> Save your work on this libre project manually till its libre license and configuration is accepted as such.
> or
> Save your work on this libre project manually till its libre license and configuration is accepted.

I still find any phrasing with 'save your work' not unambiguous and potentially unclear, especially for not-native speakers. So then I'd rather keep the existing text. But then again - it's not my decision :)

Issue #6853, opened 2021-11-23
Pull request support for GitHub Enterprise Server


gilfthde:
### Describe the problem

We would like to use the "[Pushing changes to GitHub as pull requests](https://docs.weblate.org/en/weblate-4.9.1/vcs.html#pushing-changes-to-github-as-pull-requests)" feature with *Github Enterprise Server (GHES)*. Currently this does not work because on GHES the API endpoint is different from GitHub cloud.

The current code derives the API endpoint from the repository URL (extracts the hostname) and only changes it to`api.github.com` if the repo URL points to a repository on GitHub cloud. But for GHES the base API endpoint is `<hostname>/api/v3`.

As a result all requests to create the pull request fail with an authentication error because Weblate tries to access an UI URL instead of the API, which in turn results in a redirect to the user login page.

### Describe the solution you'd like

Change the code to derive the API URL form the repository URL to append the `/api/v3` suffix to the hostname if the repository is not on GitHub cloud. Maybe an additional configuration property is required to control this.

nijel:
The logic would probably fit here:

https://github.com/WeblateOrg/weblate/blob/af562880a48f8a0ebdb7574e4cfe7eaf99f74b07/weblate/vcs/git.py#L821-L824

gilfthde:
Have a look at my draft PR. It implements the support as a new repository backend. I think this is a more clean solution.

PR #6857, opened 2021-11-23
Pull request support for Github Enterprise Server

gilfthde:
# Proposed changes

This PR implements pull request support for GitHub Enterprise Server. See #6853

I've implemented it as a new repository backend. 

## Checklist

### DIFF
--- a/docs/user/translating.rst
+++ b/docs/user/translating.rst
@@ -359,6 +359,8 @@ operation again.
 Default limits and fine-tuning is described in the administrative manual, see
 :ref:`rate-limit`.
 
+.. _search-replace:
+
 Search and replace
 ------------------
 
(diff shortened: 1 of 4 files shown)


nijel (commented):
I think that introducing another class for this will be confusing:

- users need to select the right VCS
- there will be additional configuration parameters

Having this inside existing GitHub integration would be more straightforward to users.

I think that returning `f"{host}/api/v3" in `format_api_host` in this case will work.

gilfthde:
Ok, you are right. I was thinking that just returning `{host}/api/v3` could break support for other GitHub offerings (Enterprise Cloud or AE), but it looks like Enterprise Cloud also uses `api.github.com` and AE the `/api/v3` variant. I will change it.

PR #6928, opened 2021-12-08
Add reference to Projects > Template markup

keunes:
Small change to add reference to commit message template options in the Squash commit add-on documentation.


nijel:
Thanks, I've moved it to other location in 265c099f4dbb742afb843a7f072ebe47a4d3a9e8.

Issue #5855, opened 2021-04-15
Header viewing when scrolling


tomkolp:
**Is your feature request related to a problem? If so, please describe.**

When viewing projects / component lists with a large number of components the header remains at the top of the page. This makes it difficult to identify the meaning of the numbers.

**Describe the solution you'd like**

Header viewing when scrolling

**Screenshots**

![image](https://user-images.githubusercontent.com/39633242/114857668-d49d8a80-9de8-11eb-9897-45f774831fc1.png)

ptse888:
Is anyone working on this? I would like to implement it!

nijel:
You are welcome to work on this

ptse888:
@nijel would it be possible to point me in the right direction of where the HTML files to modify are? I am having a bit of trouble locating them.

nijel:
The template for objects listing is here: https://github.com/WeblateOrg/weblate/blob/main/weblate/templates/snippets/list-objects.html

ptse888:
@nijel I tried modifying the https://github.com/WeblateOrg/weblate/blob/main/weblate/static/style-bootstrap.css and the https://github.com/WeblateOrg/weblate/blob/main/weblate/static/vendor/bootstrap/css/bootstrap.css files to try to change the header, but none of my changes would show up. I was able to do some inline CSS and was able to see my modifications. Am I looking at the wrong CSS files or am I missing something? Thank you again for the help and quick responses!

PR #6929, opened 2021-12-09
Sticky Header view change in Project screen

ptse888:
I was able to solve this issue: https://github.com/WeblateOrg/weblate/issues/5855 but was only able to get the javascript to work when it was in the HTML file.

## Checklist


nijel:
Implemented in b12c5d9239fc13da33f3952ded27399800ad7671

Issue #6169, opened 2021-06-11
Add translating tutorial to the documentation


nijel:
Having some Weblate tutorial as part of our documentation would be great. It could be probably added to https://docs.weblate.org/en/latest/user/basic.html.

Example: https://web.archive.org/web/20240315190325/https://lstn.wolba.ch/weblate/

_Originally posted by @allie-tatarian in https://github.com/WeblateOrg/weblate/discussions/5771#discussioncomment-769597_

_CC @orangesunny_

comradekingu:
Has anyone asked for it? Is it meant to be simple, or focus on what is important. Both?

nijel:
I've seen few projects having something like this on their wiki, so apparently it's something missing in our docs

comradekingu:
They do, yes. I have helped fix some. Not convinced they get read or do a lot of good if they are. If effort is to be put into translating this, then it ought to make sense.

I would want something that gets new users up to speed and avoids the common misconceptions. Basically what I want is useful translators, and people reading documentation is a good place to start.
Something like this with images and a lot of links:

> 
> The basic functionality is one field with a source string, and
> then an open field next to it or below it to put your translation in.
> This depends on the editor you pick
> 
> If your translation works, but isn't perfect, mark it as "Needs work".
> If you are unsure your translation works, you can make that a suggestion, for someone else to implement.
> 
> What you want to arrive at is a natural representation of what
> a string does, in your own language. The particular nature and
> conventions used in the source string are not important.
> 
> If the quality isn't there in the original string, you should _always_ fix it.
> A lot of source strings can be edited directly, and/or you can leave suggestions, or at the very least point out the error in a comment.
> 
> If direct editing is not possible, you will most often find a link to the source code repository next to the string,
> and if not, in the "Info" field of the project.
> 
> You can keep an eye out for errors in languages you know by adding them
> as secondary languages. That will often help you piece together a
> translation in your language.
> 
> Always go over what you have translated, and make sure errors are fixed, or dismissed.

comradekingu:
Would https://github.com/WeblateOrg/weblate/blob/main/docs/user/translating.rst be the place?

orangesunny:
> Would https://github.com/WeblateOrg/weblate/blob/main/docs/user/translating.rst be the place?

I feel this article is too complex to include the new part. I see two options on where to place this:
1. In the basics, as @nijel suggested in the first comment here ⬆️ 
2. A new article called "Quick start guide"

[...]

PR #7007, opened 2021-12-24
Docs: Translating.rst reworked and line-shifted

comradekingu:
Can I put https://github.com/WeblateOrg/weblate/issues/6169#issuecomment-860242734 here?


docs/user/translating.rst
```diff
@@ -36,8 +36,8 @@ Translation links
 
 Having navigated to a component, a set of links lead to its actual translation.
 The translation is further divided into individual checks, like
-:guilabel:`Not translated strings` or :guilabel:`Strings needing action`. If the whole project
-is translated, without error, :guilabel:`All strings` is still available.
+:guilabel:`Not translated strings` or :guilabel:`Strings needing action`.
+If the whole project is translated, without error, :guilabel:`All strings` is still available.
```
comradekingu, line 40:
```suggestion
If the whole project is translated without any errors, :guilabel:`All strings` is still available.
```
This actually meant the label is always available when a project is translated, as opposed to some of the time.

docs/user/translating.rst
```diff
@@ -49,9 +49,9 @@ Suggestions
 
     Actual permissions might vary depending on your Weblate configuration.
 
-Anonymous users can only (by default) forward suggestions. Doing so is still
-available to signed-in users, in cases where uncertainty about the translation
-arises, prompting other translators to review it.
+Anonymous users can only (by default) forward suggestions.
+Doing so is still available to signed-in users, in cases where uncertainty
```
comradekingu, line 53:
```suggestion
Doing so is still available for signed-in users, in cases where uncertainty
```

nijel:
> Can I put [#6169 (comment)](https://github.com/WeblateOrg/weblate/issues/6169#issuecomment-860242734) here?

Please no, the bigger the pull request is, the harder it is to review.

docs/user/translating.rst
```diff
@@ -1,21 +1,18 @@
 Translating using Weblate
 =========================
 
-Thank you for interest in translating using Weblate. Projects can either be
-set up for direct translation, or by way of accepting suggestions made by
-users without accounts.
+Projects either accept translations directly,
+or when a number of votes is reached.
```
orangesunny, line 5: Suggestions can be used with voting, but it is not the only use case, thus it can’t be described like this.

orangesunny (approved):
good to go now

orangesunny:
@nijel, please work out the failing checks

Issue #7008, opened 2021-12-24
Find untranslated strings


comradekingu:
### Describe the issue

I can't find untranslated strings easily anymore.

The documentation says finding untranslated strings is possible in multiple languages at once
https://docs.weblate.org/en/latest/user/translating.html?highlight=number%20of%20strings#matrix-view
but it isn't possible to sort status:empty there, and the box to do it moves farther down as more strings are loaded, so not a solution for many strings.

https://hosted.weblate.org/matrix/andors-trail/game-content/?lang=nb_NO&lang=sv

I realize opening a multi-language editor with editable fields for a whole project (or instance) is resource heavy, but I can't figure out how to operate the basic functionality. 

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar issues in this repository.

### Steps to reproduce the behavior

Finding missing translations is done by 
1. Clicking "[Project](https://hosted.weblate.org/projects/andors-trail/)" 
2. → [Languages](https://hosted.weblate.org/projects/andors-trail/#languages) 
3. → the number under "untranslated" to get to https://hosted.weblate.org/zen/andors-trail/-/nb_NO/?q=state:%3Ctranslated and then 
4. changing to "status:empty". https://hosted.weblate.org/zen/andors-trail/-/nb_NO/?offset=1&q=state%3Aempty&sort_by=-priority%2Cposition&checksum=

### Expected behavior

Clicking the number under what is correctly named "untranslated" in language view under https://hosted.weblate.org/projects/andors-trail/#languages like it used to be.

### Screenshots

![bilde](https://user-images.githubusercontent.com/13802408/147327525-3da6ba1a-db1b-4385-8497-17de96f2a682.png)

### How do you run Weblate?

weblate.org service

### Additional context

![bilde](https://user-images.githubusercontent.com/13802408/147327630-4f3f4a19-94a6-4cad-815b-8ce79cc1f6bf.png)

nijel:
> I can't find untranslated strings easily anymore.

Nothing has changed in this recently, what is the difference?

>The documentation says finding untranslated strings is possible in multiple languages at once
https://docs.weblate.org/en/latest/user/translating.html?highlight=number%20of%20strings#matrix-view
but it isn't possible to sort status:empty there, and the box to do it moves farther down as more strings are loaded, so not a solution for many strings.

There is currently no filtering in the matrix mode.

[...]

PR #7016, opened 2021-12-28
ui: Rename `Strings needing action` to `Unfinished strings`

### LINKED ISSUE #1163: "strings needing action" doesn't actually count all strings needing action
Notably, the number does not include strings that are translated but have failing checks.

Also, for our purposes, we need to count strings with suggestions as "for review"; this will probably change if we implement a proper review workflow.

I propose counting strings with failing checks into "strings needing action".

Suggestions (and possibly comments) are trickier: the checklist item doesn't differentiate between untranslated and translated, which makes it impossible to get an exact number and wordcount of "strings needing work" for people who need or want that. OTOH, counting them in "strings needing action" would annoy people who don't care about suggestions on already translated strings.

At this point, i'm thinking of implementing a "Clean strings" checklist item, in addition to "Translated strings", and possibly a corresponding "Dirty strings" which is `total - clean`
##

### LINKED ISSUE #5896: "Untranslated/words" isn't actually untranslated
I can't distinguish untranslated strings with the new always-full bars.
This is compounded by not being able to sort for untranslated strings, or number of untranslated words.

https://hosted.weblate.org/languages/nb_NO/
bilde

The third category here is "Not translated strings" or somesuch, but aren't actually _missing_, as that would suggest.

For example: https://hosted.weblate.org/languages/nb_NO/asteroidos/#overview

nijel:
## Proposed changes

* This is a better description of the category.
* Consolidated labels for the filter and overview columns.

Fixes #1163
Fixes #5896

## Checklist


comradekingu:
Renaming strings needing work to "unfinished" isn't helping me find untranslated strings.
The combination of the two I don't understand, but "untranslated" it is not.
Calling only the progress bar category "unfinished" would at least be correct.
Sorting by untranslated (state:empty) first is what helps me find untranslated strings.
For the sake of simplicity, I supposed that is a compromise between my use-case and bulk categories.

Going so far as to rename the checkbox "needing action/work" everywhere, means now it seems like "unfinished" could mean partly translated, like translating 75% of a string and then coming back to finish it later.
Those strings actually make it in, so that is where I see a problem.

Taking something out of "unfinished" doesn't finish it.
It is so hard to get people to understand coverage as different from completion, and this will not help.

TL;DR Can only the places where "untranslated" (state:empty) and "needs work" are lumped together be called "unfinished", instead of this mass change?

nijel:
This PR changes name for `state:<translated` strings to `Unfinished strings`, nothing else. It doesn't change anything else. Currently, these are sometimes called `Untranslated` and sometimes `Strings needing action`.

What you are looking for is `state:empty` and that will be covered at some point by https://github.com/WeblateOrg/weblate/issues/7008

comradekingu:
It can be "empty" or "not translated" (as I think it originally was).
It seems you have a plan, and I maybe that coincides with mine or not.
I thought "strings needing action" was the checkbox, maybe it is, maybe it isn't.

PR #7065, opened 2022-01-05
Add a note about deleting announcements

maxxer:
## Proposed changes

Just a minor addition to the configuration

## Checklist


nijel:
That's not true, whoever can create them can also delete them by clicking on the x icon when they are shown. Maybe that's not the best UI though...

maxxer:
Oh. Usually the *x* is used to dismiss the message for the user, I didn't try (as I didn't expect) it would *delete* the post

PR #7128, opened 2022-01-20
Update views.py

zjzh:
I refactored code with set comprehension which is more concise and efficient，how do think this change which has practical value？

## Checklist


nijel:
While this is more efficient from Python POV, the expensive operation here is the image processing, so there will be no practical impact of this change :-).

Also, the definition of `results` just before this block should be removed with this change.

PR #7194, opened 2022-02-03
Machinery configuration rewrite

nijel:
## Proposed changes

* Database backed configuration of machinery
* Engines can be configured in the web UI
* Engines can be configured at project level as well

## Checklist


weblate/machinery/models.py
```diff
@@ -22,78 +22,107 @@
 
 
 class WeblateConf(AppConf):
-    """Machine translation settings."""
+    """
+    Machine translation settings.
+
+    TODO: All MT_* are deprecated and should be dropped once migration
+    from Weblate 4.10 is not supported. Tentatively in Weblate 5.1.
```
comradekingu, line 29:
```suggestion
    from Weblate 4.10 is no longer supported. Tentatively in Weblate 5.1.
```

weblate/machinery/models.py
```diff
@@ -22,78 +22,107 @@
 
 
 class WeblateConf(AppConf):
-    """Machine translation settings."""
+    """
+    Machine translation settings.
```
comradekingu, line 26:
```suggestion
    Machine-translation settings.
```

PR #7217, opened 2022-02-07
Spelling: The, Perhaps, of repository updates

comradekingu:


weblate/templates/trans/alert/pushfailure.html
```diff
@@ -19,6 +19,6 @@
   </p>
 {% elif analysis.behind %}
   <p>
-    {% trans "Weblate repository is missing some commits from upstream, perhaps it is not receiving notifications on repository updates." %}
+    {% trans "The Weblate repository is missing some commits from upstream. Perhaps it is not receiving notifications of repository updates." %}
```
comradekingu, line 22:
```suggestion
    {% trans "The Weblate repository is missing some commits from upstream. Perhaps it is not receiving notifications upon remote repository updates." %}
```
?

comradekingu, line 22:
```suggestion
    {% trans "The Weblate repository is missing some commits from upstream. Perhaps it is not receiving notifications upon remote repository updates." %}
```
?

nijel: It should be "upstream repository" to be consistent. How about this?

```suggestion
    {% trans "The Weblate repository is missing some commits from upstream. Perhaps it is not receiving notifications upon the upstream repository updates." %}
```

comradekingu: "upstream repository" works to set it apart from "internal/Weblate". In my head "remote repository" is better, since making changes directly in Weblate sort of makes it the upstream considering where changes come from. That is a consideration for another day.

PR #7223, opened 2022-02-08
Docs: Reporting reworked

comradekingu:
## Checklist

### DIFF
--- a/docs/devel/reporting.rst
+++ b/docs/devel/reporting.rst
@@ -1,14 +1,14 @@
 Translation progress reporting
 ==============================
 
-Reporting features give insight into how a translation progresses over a given
-period. A summary of contributions to any given component over time is
-provided. The reporting tool is found in the :guilabel:`Insights` menu of any
-translation component, project or on the dashboard:
+Reporting features give insight into translation progress over a given period.
+A summary of contributions to any given component over time is provided.
+The reporting tool is found in the :guilabel:`Insights` menu of any
+translation component and project, and shown on the dashboard:
 
 .. image:: /screenshots/reporting.png
 
-Several reporting tools are available on this page and all can produce output
+Several reporting tools are available on this page, all of which can produce output
 in HTML, reStructuredText or JSON. The first two formats are suitable for
 embedding statistics into existing documentation, while JSON is useful for further
 processing of the data.
@@ -18,8 +18,8 @@ processing of the data.
 Translator credits
 ------------------
 
-Generates a document usable for crediting translators - sorted by language
-and lists all contributors to a given language:
+Generates a document usable for crediting translators — sorted by language
+and lists all contributors for a given language:
 
 .. code-block:: rst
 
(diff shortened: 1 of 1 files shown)


docs/devel/reporting.rst
```diff
@@ -1,14 +1,14 @@
...
 ==============================
 
-Reporting features give insight into how a translation progresses over a given
-period. A summary of contributions to any given component over time is
-provided. The reporting tool is found in the :guilabel:`Insights` menu of any
-translation component, project or on the dashboard:
+Reporting features give insight into translation progress over a given period.
+A summary of contributions to any given component over time is provided.
+The reporting tool is found in the :guilabel:`Insights` menu of any
+translation component and project, and shown on the dashboard:
```
comradekingu, line 7:
```suggestion
translation component and project, and possibly shown on the dashboard:
```

orangesunny: not needed as it changes the meaning of the description.

docs/devel/reporting.rst
```diff
@@ -30,12 +30,12 @@ and lists all contributors to a given language:
 
     * Dutch
 
-        * Jane Doe <jane@example.com> (42)
+       * Jane Doe <jane@example.com> (42)
```
comradekingu, line 33: @orangesunny 
Ref. https://github.com/WeblateOrg/weblate/commit/502660952d78a97ffd765eb63c271bf93cacaf39
"Roe" is the localised Dutch variant (same as German) https://nl.wikipedia.org/wiki/John_Doe

The idea is to be good at localisation, and that the same person (given by the identical e-mail) wouldn't be likely to edit both Dutch and Czech. Good examples are not lazy?

docs/devel/reporting.rst
```diff
@@ -1,14 +1,14 @@
...
 ==============================
 
-Reporting features give insight into how a translation progresses over a given
-period. A summary of contributions to any given component over time is
-provided. The reporting tool is found in the :guilabel:`Insights` menu of any
-translation component, project or on the dashboard:
+Reporting features give insight into translation progress over a given period.
+A summary of contributions to any given component over time is provided.
+The reporting tool is found in the :guilabel:`Insights` menu of the dashboard:,
+or any translation component, or project:
```
comradekingu, line 7:
```suggestion
any translation component, or project:
```
Ref. https://github.com/WeblateOrg/weblate/pull/7223/commits/7334d6625d0426e43c18910dece3ce2cfb2ef57b
Otherwise implies it goes missing elsewhere if found in other places.

docs/devel/reporting.rst
comradekingu: True. Seems it got fixed properly in the interim.

docs/devel/reporting.rst
orangesunny: Using localized _Does_ would be interesting, but only consistently throughout the whole docs, in localized version (other languages) of the docs. In this case (across the English docs version), one name is correct across all languages exampled in the reports' snippet; Jane Doe is clearly a productive translator as she contributed to multiple languages. Her name should not be translated to assure she is the same person.

PR #7291, opened 2022-02-22
update flags dynamically on dismiss for all languages

JohnRDOrazio:
## Proposed changes

Automatically update `Extra Flags` box when clicking on `Dismiss` check `for all languages`.

## Checklist


JohnRDOrazio:
This would be even easier if the `list-group-item` divs had more specific identifiers, whether through a class of their own or through an `id` attribute. Then the jQuery selectors would be simple and sure.

weblate/trans/views/js.py
```diff
@@ -151,7 +151,7 @@ def ignore_check_source(request, check_id):
         unit.save(same_content=True)
 
     # response for AJAX
-    return HttpResponse("ok")
+    return HttpResponse(unit.extra_flags)
```
nijel, line 154: In case we do client side processing, it should be a JSON response here. Also, it will need more content:

* `unit.extra_flags` should be stored into popup editor for flags
* `unit.all_flags.format()` should be used as display value

weblate/static/editor/full.js
```diff
@@ -324,6 +324,26 @@
...
+                return (
+                  this.nodeType === 3 &&
+                  (this.nextElementSibling === null ||
+                    this.nextElementSibling.nodeName.toLowerCase() !== "h5")
+                );
+              })[0];
+              flagsTextNode.textContent = data;
+            }
+          }
+        },
```
nijel, line 346: Instead of the template should be adjusted to contain single element to contain this:

https://github.com/WeblateOrg/weblate/blob/f6753a1a1c63fade6ad418fbda827c6750ab0bda/weblate/templates/translate.html#L648-L652

nijel (commented):
I find this useful, but needs more changes before it can be accepted:

* as you mentioned finding the element is fragile, there should be a html element for the content which could be addressed by id instead of this
* your change would remove inherited flags from display
* it should also adjust editable popup for flags so that it is consistent as well

weblate/static/editor/full.js
JohnRDOrazio: I wrapped this section in a div with an id that can be easily targeted by javascript

weblate/trans/views/js.py
JohnRDOrazio: This is now a JSON response. I included "unit.extra_flags", "unit.all_flags", and the single "ignore" flag to the response. I think it might be nice to also have a notification letting the user know that the single flag was added to the list of extra flags. I'm not sure if there is a popup notification system in weblate that can be hooked into?

JohnRDOrazio:
I would also like to add an info flag next to the "for all languages" label, something like this:

```django
<span style="display:inline-block;height:24px;vertical-align:super;margin-left:6px;" title="{% trans "Selecting this option will set the corresponding ignore check flag" %}">{% icon "info.svg" %}</span>
```

weblate/trans/tests/test_edit.py
```diff
@@ -817,22 +819,35 @@ def test_edit_check(self):
...
         self.user.save()
         response = self.client.post(ignore_url)
-        self.assertContains(response, "ok")
+        self.assertEqual(response.headers["Content-Type"], "application/json")
+
         # Should have one less check
         unit = self.get_unit()
+        obj = get_object_or_404(Check, pk=int(check_id))
+        ignore = obj.check_obj.ignore_string
+        self.assertJSONEqual(
```
JohnRDOrazio, line 843: my understanding is that this test failed the first time because it was before `unit = self.get_unit()`, therefore the value of `unit` was not yet updated with the new values. Placing it here, after the call to `unit = self.get_unit()`, should ensure that the value of `unit` is updated.

weblate/templates/translate.html
```diff
@@ -469,7 +469,7 @@ <h5>
             {% if user_can_ignore_check %}
               <a href="{% url 'js-ignore-check' check_id=check.id %}" data-check="{{ check.id }}" data-dismiss-all="{% url 'js-ignore-check-source' check_id=check.id %}" class="btn btn-warning check-dismiss check-dismiss-single">{% trans "Dismiss" %}</a>
               {% if user_can_edit_source %}
-                <label class="dismiss-all"><input type="checkbox" name="dismiss-all-{{ check.id }}" /> {% trans "For all languages" %}</label>
+                <label class="dismiss-all"><input type="checkbox" name="dismiss-all-{{ check.id }}" /> {% trans "For all languages" %}</label><span style="display:inline-block;height:24px;vertical-align:super;margin-left:6px;" title="{% trans "Selecting this option will set the corresponding \"ignore check\" flag" %}">{% icon "info.svg" %}</span>
```
nijel, line 472: Please avoid inline CSS.

weblate/trans/tests/test_edit.py
```diff
@@ -817,22 +819,35 @@ def test_edit_check(self):
...
         self.assertEqual(response.status_code, 403)
         self.user.is_superuser = True
         self.user.save()
         response = self.client.post(ignore_url)
-        self.assertContains(response, "ok")
+        self.assertEqual(response.headers["Content-Type"], "application/json")
+
         # Should have one less check
         unit = self.get_unit()
+        obj = get_object_or_404(Check, pk=int(check_id))
```
nijel, line 841: No need to go via get_object_or_404:

```suggestion
        obj = Check.objects.get(pk=int(check_id))
```

weblate/static/editor/full.js
```diff
@@ -324,6 +324,30 @@
...
+              if (transparency === 0) {
+                clearInterval(animation);
+              } else {
+                let transpVal = transparency.toString(16).padStart(2, "0");
+                $("#unit_all_flags").css({
+                  "background-color": "#ffa500" + transpVal,
+                });
+                transparency--;
+              }
+            }, 15);
```
nijel, line 348: Maybe it would be a better idea to use CSS transition instead of doing that manually in the JavaScript?

weblate/templates/translate.html
JohnRDOrazio: sure thing, which CSS file would this belong in? The relevant file I can find is https://github.com/WeblateOrg/weblate/blob/main/weblate/static/style-bootstrap.css, is this correct? Or is there a specific stylesheet for the editor?

JohnRDOrazio: okay I created a class and moved styling to [weblate/static/style-bootstrap.css](https://github.com/WeblateOrg/weblate/blob/main/weblate/static/style-bootstrap.css)

weblate/static/editor/full.js
JohnRDOrazio: done, created a class and added animation to CSS in [weblate/static/style-bootstrap.css](https://github.com/WeblateOrg/weblate/blob/main/weblate/static/style-bootstrap.css)

weblate/templates/translate.html
```diff
@@ -469,7 +469,7 @@ <h5>
             {% if user_can_ignore_check %}
               <a href="{% url 'js-ignore-check' check_id=check.id %}" data-check="{{ check.id }}" data-dismiss-all="{% url 'js-ignore-check-source' check_id=check.id %}" class="btn btn-warning check-dismiss check-dismiss-single">{% trans "Dismiss" %}</a>
               {% if user_can_edit_source %}
-                <label class="dismiss-all"><input type="checkbox" name="dismiss-all-{{ check.id }}" /> {% trans "For all languages" %}</label><span style="display:inline-block;height:24px;vertical-align:super;margin-left:6px;" title="{% trans "Selecting this option will set the corresponding \"ignore check\" flag" %}">{% icon "info.svg" %}</span>
+                <label class="dismiss-all"><input type="checkbox" name="dismiss-all-{{ check.id }}" /> {% trans "For all languages" %}</label><span class="info-tooltip" title="{% trans "Selecting this option will set the corresponding &#34;ignore check&#34; flag" %}">{% icon "info.svg" %}</span>
```
nijel, line 472: You can use single quotes to avoid escaping here:

```suggestion
                <label class="dismiss-all"><input type="checkbox" name="dismiss-all-{{ check.id }}" /> {% trans "For all languages" %}</label><span class="info-tooltip" title="{% trans 'Selecting this option will set the corresponding "ignore check" flag.' %}">{% icon "info.svg" %}</span>
```

nijel:
The code looks good now, I will get to testing this once 4.11.1 is out with fixes we need to release.

weblate/templates/translate.html
JohnRDOrazio: ok I was trying to figure out how to correctly escape, I wasn't having much luck with backslashes. Thanks for the input, implementation complete.

weblate/trans/views/js.py
```diff
@@ -151,7 +151,13 @@ def ignore_check_source(request, check_id):
         unit.save(same_content=True)
 
     # response for AJAX
-    return HttpResponse("ok")
+    return JsonResponse(
+        {
+            "extra_flags": unit.extra_flags,
+            "all_flags": unit.all_flags.format(),
```
nijel, line 157: Need to parse the translation unit flags here, not the source one (it might have extra ones, for example source string might be read-only):

```suggestion
            "all_flags": obj.unit.all_flags.format(),
```

weblate/trans/tests/test_edit.py
```diff
@@ -817,22 +818,38 @@ def test_edit_check(self):
...
+            obj = Check.objects.get(pk=int(check_id))
+        except Check.DoesNotExist:
+            obj = Check(name=check_id)
+        ignore = obj.check_obj.ignore_string
+        self.assertJSONEqual(
+            response.content.decode("utf-8"),
+            {
+                "extra_flags": unit.extra_flags.format(),
+                "all_flags": unit.all_flags.format(),
+                "ignore_check": ignore,
```
nijel, line 850:
```suggestion
                "ignore_check": ignore_flag,
```

weblate/trans/tests/test_edit.py
```diff
@@ -817,22 +818,38 @@ def test_edit_check(self):
...
         self.assertContains(response, "ok")
+
         # Should have one less failing check
         unit = self.get_unit()
         self.assertFalse(unit.has_failing_check)
         self.assertEqual(len(unit.all_checks), 1)
         self.assertEqual(len(unit.active_checks), 0)
         self.assertEqual(unit.translation.stats.allchecks, 0)
+
         # Ignore check for all languages
```
nijel, line 829:
```suggestion
        # Ignore check for all languages
        ignore_flag = Check.objects.get(pk=int(check_id)).ignore_string
```

weblate/trans/tests/test_edit.py
```diff
@@ -817,22 +818,38 @@ def test_edit_check(self):
...
-        self.assertContains(response, "ok")
+        self.assertEqual(response.headers["Content-Type"], "application/json")
+
         # Should have one less check
         unit = self.get_unit()
+        try:
+            obj = Check.objects.get(pk=int(check_id))
+        except Check.DoesNotExist:
+            obj = Check(name=check_id)
+        ignore = obj.check_obj.ignore_string
```
nijel, line 844:
```suggestion
```

nijel:

PR #7426, opened 2022-03-26
Fix links to Liberapay profiles

Changaco:
A while ago the configuration of the `liberapay.org` domain was changed to redirect all requests to the home page of the `liberapay.com` domain, so the links that Weblate generates no longer lead to the intended pages.

## Checklist


nijel:

PR #7460, opened 2022-04-02
Fix some typos (found by codespell)

stweil:
## Proposed changes

These commits fix typos in various parts of the code and documentation (ignoring code from vendors).

## Checklist

## Other information

The files for localization (*.po) still have to be updated.
Maybe add codespell to the continuous integration tests?


docs/admin/install.rst
```diff
@@ -142,7 +142,7 @@ The following dependencies have to be installed on the system:
 
 ``Git``
     https://git-scm.com/
-Pango, Cairo and related header files and gir introspection data
+Pango, Cairo and related header files and git introspection data
```
nijel, line 145: gir stands for [GObject Introspection](https://en.wikipedia.org/wiki/GObject#GObject_Introspection) here:

```suggestion
Pango, Cairo and related header files and GObject introspection data
```

docs/devel/html.rst
```diff
@@ -93,7 +93,7 @@ selector to process all elements.
 .. note::
 
    Right now, only text of the elements is extracted. This add-on doesn't support localization
-   of element attributes or elements with childs.
+   of element attributes or elements with children.
```
orangesunny, line 96: This does not make sense. A different word should be chosen, IMO. Something with sub- ?

weblate/formats/convert.py
```diff
@@ -240,12 +240,12 @@ def convertfile(storefile, template_store):
 
     def save_content(self, handle):
         """Store content to file."""
-        convertor = po2html()
+        converter = po2html()
```
orangesunny, line 243: @nijel, I am afraid that this could break the code?

weblate/wladmin/views.py
```diff
@@ -460,23 +460,23 @@ def billing(request):
     # We will list all billings anyway, so fetch  them at once
     billings = Billing.objects.prefetch().order_by("expiry", "removal", "id")
 
-    for currrent in billings:
-        if currrent.removal:
-            removal.append(currrent)
-        elif currrent.state == Billing.STATE_TRIAL:
+    for current in billings:
```
orangesunny, line 463: @nijel, this is correct, but I am unsure if it wouldn’t break anything.

orangesunny (changes requested):
Can go if my comments are solved.

docs/admin/install.rst
```diff
@@ -142,7 +142,7 @@ The following dependencies have to be installed on the system:
 
 ``Git``
     https://git-scm.com/
-Pango, Cairo and related header files and gir introspection data
```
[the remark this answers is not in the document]
nijel: I've made this change, it is correct.

docs/devel/html.rst
nijel: child elements is established in XML, so it looks okay to me.

weblate/formats/convert.py
nijel: Tests pass, so it should be okay :-).

nijel:

PR #7505, opened 2022-04-17
Docs: Workflows reworked

comradekingu:
## Checklist


docs/workflows.rst
```diff
@@ -241,20 +241,19 @@ translators to translate into additional languages.
...
 :guilabel:`Needs editing`.
 
-The bilingual formats do not allow direct editing of the source strings (these
-are typically extracted directly from the source code). In this case
+The bilingual formats do not allow direct editing of source strings (these
+are typically extracted directly from the source code). In this case a
 :guilabel:`Source needs review` label is attached to strings reported by
 translators. You should review such strings and either edit them in the source
-or remove the label.
+code, or remove the label.
```
comradekingu, line 256: ?

docs/workflows.rst
```diff
@@ -241,20 +241,19 @@ translators to translate into additional languages.
...
 
-For monolingual formats, the source string review behaves similarly as with
-:ref:`reviews` - once issue is reported on the source string, it is marked as
+For monolingual formats, source string review functions similarly to
+:ref:`reviews` — once an issue with a source string is reported, it is marked as
 :guilabel:`Needs editing`.
 
-The bilingual formats do not allow direct editing of the source strings (these
-are typically extracted directly from the source code). In this case
+The bilingual formats do not allow direct editing of source strings (these
```
comradekingu, line 252:
```suggestion
Bilingual formats do not allow direct editing of source strings (these
```

docs/workflows.rst
```diff
@@ -172,25 +172,25 @@ Reviews can be turned on in the project configuration, from the
...
-addressing this - there is additional quality gateway for the strings between
-developers and translators and users.
+In many cases the original source language strings come from developers,
+as they write the code and provide initial strings. However developers are
+often not native speakers of the source language, and may not be capable of
+attaining the desired source string quality. Intermediate translation can help you
+in addressing this — because it makes for an additional quality gateway for
+strings between developers and translators.
 
-By setting :ref:`component-intermediate`, this file will be used as source for
```
comradekingu, line 191: "edited to source language" has me confused.

docs/workflows.rst
```diff
@@ -172,25 +172,25 @@ Reviews can be turned on in the project configuration, from the
...
-In many cases the original source language strings are coming from developers,
-because they write the code and provide initial strings. However developers are
-often not a native speakers in the source language and do not provide desired
-quality of the source strings. The intermediate translation can help you in
-addressing this - there is additional quality gateway for the strings between
-developers and translators and users.
+In many cases the original source language strings come from developers,
+as they write the code and provide initial strings. However developers are
+often not native speakers of the source language, and may not be capable of
+attaining the desired source string quality. Intermediate translation can help you
```
comradekingu, line 187:
```suggestion
attaining the desired source-string quality. Intermediate translation can help you
```

docs/workflows.rst
```diff
@@ -3,24 +3,24 @@
 Translation workflows
 =====================
 
-Using Weblate is a process that brings your users closer to you, by bringing
-you closer to your translators.  It is up to you to decide how many of its
+Using Weblate is a process bringing your users closer to you, where you are
```
comradekingu, line 6: This sentence doesn't really explain anything.

docs/workflows.rst
Geeyun-JY3: `source` is corresponding to `target`. It may refer to “source text”.

See https://en.m.wikipedia.org/wiki/Translation#Source_and_target_texts

docs/workflows.rst
```diff
@@ -112,7 +112,7 @@ from additional member(s) before it is accepted as a translation.
 +------------------------+-------------+------------------------------------+
 | Suggestion voting      | off         |                                    |
 +------------------------+-------------+------------------------------------+
-| Autoaccept suggestions | 1           | You can set higher value to        |
+| Auto-accept suggestions| 1           | You can set higher value to        |
```
nijel, line 115: The UI should be changed as well when changing this in the documentation...

docs/workflows.rst
comradekingu, line 256:
```suggestion
material, or remove the label.
```

Geeyun-JY3: Totally no need. https://github.com/WeblateOrg/weblate/pull/7505#discussion_r852563528

yangyangdaji (approved):

docs/workflows.rst
comradekingu: `them in` makes it clear what it is that needs editing.

Geeyun-JY3: TBH I am not quite sure about what the “source” means here now. It seems ambiguous.

nijel: It is used in case the source strings are not editable in Weblate, so they need to be edited anywhere it is extracted from. This typically means source code for the applications.

docs/workflows.rst
comradekingu: I added every "autoaccept" I could find to the PR.

docs/workflows.rst
orangesunny: If you know that the sentence is not useful, feel free to overwrite it or remove it next time.

docs/workflows.rst
```diff
@@ -127,6 +127,9 @@ from additional member(s) before it is accepted as a translation.
 Dedicated reviewers
 -------------------
 
+
+.. versionadded:: 2.18
+
```
nijel, line 132:
```suggestion
```

I think it's time to remove these for 2.x versions.

weblate/trans/validators.py
```diff
@@ -16,30 +16,30 @@
...
 
 
 def validate_check_flags(val):
-    """Validate check influencing flags."""
+    """Validate check-influencing flags."""
     try:
         flags = Flags(val)
     except (ParseException, re.error) as error:
-        raise ValidationError(gettext("Failed to parse flags: %s") % error)
+        raise ValidationError(gettext("Could not parse flags: %s") % error)
```
nijel, line 42: Can you please avoid stacking random changes to a documentation pull request? This is a useful change, but should be done in a separate pull request, and cover all such strings. I've just done this at https://github.com/WeblateOrg/weblate/pull/9575

docs/admin/projects.rst
```diff
@@ -580,8 +580,8 @@ Turns on vote casting for suggestions, see :ref:`voting`.
 
 .. _component-suggestion_autoaccept:
 
-Autoaccept suggestions
-++++++++++++++++++++++
+Automatically accept suggestions
++++++++++++++++++++++++
```
nijel, line 584:
```suggestion
Automatically accept suggestions
++++++++++++++++++++++++++++++++
```

weblate/trans/models/component.py
```diff
@@ -472,7 +472,7 @@ class Component(models.Model, URLMixin, PathMixin, CacheKeyMixin):
         ),
     )
     suggestion_autoaccept = models.PositiveSmallIntegerField(
-        verbose_name=gettext_lazy("Autoaccept suggestions"),
+        verbose_name=gettext_lazy("Auto-accept suggestions"),
```
nijel, line 475:
```suggestion
        verbose_name=gettext_lazy("Automatically accept suggestions"),
```

orangesunny (approved):

docs/workflows.rst
comradekingu: Could it link to the relevant changelog instead?
The use-case is making sure things are supported.
Maybe for new changelogs it could go both ways?

weblate/trans/validators.py
comradekingu: Neither option is ideal really. I care for end-product and coverage, not percentage-wise total flaws.

Separate PRs would remove my only way of ensuring quality, which is knowing what files got reworked (in their entirety) based on the last time I touched anything.

I want to know where bugs aren't, according to my ability. Per-reviewer quality is my way, and it doesn't matter if I can't know what people I trust to do the work think.
If those people, which is me in some capacity are not part of what goes in at an intermediary stage, looks like that is as good as it gets once stuff is merged. That is as far as I am concerned. It can go in, but it can't stay there.
My process is to notice an error, and then to fix that file.

If I got write-access to the glossary, I could add incorrect or undesired words.
This in turn could trigger a bug-report/to-be-done queue of sorts, as it is a near endless list.
If I even attempted to make bug-reports for everything I find across the entire string-base I would never get done doing anything, and it would swarm the tracker. I have good memory of what should and is done at word-for-word level, but would lose critical oversight of what exists in a passable state that way.
Then whoever could discuss those glossary entries before they are accepted, and I don't even have to make the changes, or handle Git.
Maybe batch-suggestions would be a useful idea. I search for something, don't alter all of it, but instead adjust it to where it can be one-clicked by an admin.
F-Droid has a distinctive spell-checker triggered in the repo, but I much prefer it to be handled "in Weblate", and to also get input from all translators before going live.

I notice bugs, go check to see that I have a PR on the file, and disregard.

As an example "e.g." is used in the documentation (that I haven't fully translated so it is harder for me to know what strings are new), whereas "for example" is used in the software.
Add to this that suggestions are turned off for the "Legal" component, and it all gets a bit difficult for no real reason.

docs/admin/projects.rst
```diff
@@ -602,8 +602,8 @@ Turns on vote casting for suggestions, see :ref:`voting`.
 
 .. _component-suggestion_autoaccept:
 
-Autoaccept suggestions
-++++++++++++++++++++++
+Automatically accept suggestions
+++++++++++++++++++++++++++++++++
```
nijel, line 606: https://github.com/WeblateOrg/weblate/pull/10903 changes this in the code, docs can followup then.

nijel:

Issue #7530, opened 2022-04-23
Support for /.well-known/change-password


SethFalco:
### Describe the problem

There is a draft standard to have a standard URL users can navigate to change their password.

https://w3c.github.io/webappsec-change-password-url/

The goal is to make managing passwords easier, so software can help users change their password, browsers can direct them to where they can change their password, and for programmatic access by password managers.

To implement this, sites should (permanently) redirect `/.well-known/change-password` to the page users can go to change their password.

### Describe the solution you'd like

Make `/.well-known/change-password` permanently redirect to `https://hosted.weblate.org/accounts/password/`.

### Describe alternatives you've considered

N/A

### Additional context

Some sites are already doing this, such as:
* GitHub
* Slack
* Wikipedia

PR #7533, opened 2022-04-24
Implement webappsec change password url

SethFalco:
## Proposed changes

* Closes https://github.com/WeblateOrg/weblate/issues/7530

## Checklist

  * Not sure if it's required here, I don't see other tests for routing.
  * Not sure if it's required here. For example, I don't see dedicated docs for `robots.txt`.

## Other information

If the user is authenticated, redirects to:
`/accounts/password/`

If the user is not authenticated, Weblate automatically handles this and redirects to:
`/accounts/login/?next=/accounts/password/`

### DIFF
--- a/weblate/urls.py
+++ b/weblate/urls.py
@@ -1067,6 +1067,11 @@ real_patterns = [
             )
         ),
     ),
+    # Redirects for .well-known
+    path(
+        ".well-known/change-password",
+        RedirectView.as_view(url="/accounts/password/", permanent=True)
+    ),
 ]
 
 # Billing integration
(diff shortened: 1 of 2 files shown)


weblate/urls.py
```diff
@@ -1067,6 +1067,11 @@
             )
         ),
     ),
+    # Redirects for .well-known
+    path(
+        ".well-known/change-password",
+        RedirectView.as_view(url="/accounts/password/", permanent=True),
```
nijel, line 1073: This will most likely break with the URL_PREFIX being set.

nijel, line 1073:
```suggestion
        RedirectView.as_view(url=f"/{URL_PREFIX}accounts/password/", permanent=True),
```

SethFalco:
Sorry to leave that one to you, I was meant to get back to it eventually but have been really busy. 😭

Issue #7529, opened 2022-04-23
Config for web monetization 


SethFalco:
### Describe the problem

I think it'd be great if Web Monetization was more established, and if hosts of Weblate servers could receive additional support from users.

[Web Monetization](https://webmonetization.org/) is a proposed W3C standard for monetization on the internet.

TL;DR: Creators and websites who choose to create an [Interledger Payment Pointer](https://paymentpointers.org/) can embed it in the `<head>` element of a webpage, and users who opt-in to support such websites can stream micropayments to financially support it.

It's like Brave Rewards (for creators) except Web Monetization is a proposed W3 standard, so it's not bound to a particular browser or account.

### Describe the solution you'd like

There should be a new setting where users can configure a list of payment pointers.
If specified, Weblate will add this list in the `<head>` of the document.

When there is only one payment pointer, it's as simple as adding:
```html
<meta name="monetization" content="$alice.example">
```

If the config supports multiple payment points, which is a use-case that's officially documented on the Web Monetization site, it requires some extra JavaScript/PHP. Something to the effect of:

```html
<head>
  <script>
    // Interledger Payment Pointer mapped to weight
    const pointers = {
      '$alice.example': 50,
      '$bob.example': 40,
      '$connie.example': 9.5,
      '$dave.example': 0.5
    }

    function pickPointer () {
      const sum = Object.values(pointers).reduce((sum, weight) => sum + weight, 0);
      let choice = Math.random() * sum;
      
      for (const pointer in pointers) {
        const weight = pointers[pointer]
        if ((choice -= weight) <= 0) {
          return pointer
        }
      }
    }

    window.addEventListener('load', () => {
      const monetizationTag = document.createElement('meta')
      monetizationTag.name = 'monetization'
      monetizationTag.content = pickPointer()

      document.head.appendChild(monetizationTag)
    })
  </script>
</head>
```
> — https://webmonetization.org/docs/probabilistic-rev-sharing#code

### Describe alternatives you've considered

Instead of calculating the probabilistic revenue sharing yourself, it could be easier to use the API Web Monetization provide for it.

Suppose a config like, where the values are the Interledger Payment Pointer, weight, and a name:
```json
[
  [
    "$alice.example",
    1,
    "Alice"
  ],
  [
    "$bob.example",
    1,
    "Bob"
  ]
]
```

Just minify, stringify, and base64 encode (padding can be removed) it.
[...]

PR #7534, opened 2022-04-24
Add config for web monetization

SethFalco:
## Proposed changes

* Adds support for Web Monetization (proposed as a W3C standard)
  * Closes https://github.com/WeblateOrg/weblate/issues/7529

I named the setting "INTERLEDGER_PAYMENT_POINTERS" to refer to the value of the setting rather than the context/usage of it.

I wasn't 100% sure what the best way to configure weights in Docker, so I decided (for now) to skip weights for probabilistic revenue sharing.

If you have any thoughts there, I'd be happy to fix it now or in a separate PR.

## Checklist

  * Wasn't sure about tests, so thought I'd submit this for feedback already and come back to tests in a bit. I might be blind, but didn't see tests for other basic settings like avatar URL prefix. 🤔 

## Other information

If the setting is configured, it just adds a new element in the `<head>` of the document.

image

Users that are supporting web monetized sites, for example subscribers of Coil, will see that they're supporting the site:

image

### Alternate Solution

A possible alternate solution could be if there was a setting to just add custom `meta=[name]` properties to the head of the document. 🤔 

I'm not sure if that's a good idea, but if for whatever reason you're not into this, then we could scrap this PR and instead make a setting which is a map of properties. They'd just get translated to:

```html
<meta name="{{ key }}" content="{{ value }}">
```

Might cause some headache though as:
* Need to sanitize/escape so it can't break the DOM.
* Avoid issues with properties being defined for meta tags already set statically.


nijel:

PR #7536, opened 2022-04-24
chores: add editor config and refactors

SethFalco:
## Proposed changes

While doing other changes, these were just small things I did on the side.

### EditorConfig
Adds an [EditorConfig](https://editorconfig.org/) file, so it's more convenient for contributors to match the code styles with the rest of the project.
EditorConfig is a widely supported format that code editors use to override user preferences to adhere to project conventions.

The current settings are based on what I can see from exploring the repository.

* Most JSON files were 2 space indented, so I applied that to `.imgbotconfig` which was 4 before.

### JSON Schema
Updated `.readthedocs.yml` and `codecov.yml` with trivial changes to avoid warnings about them not matching the JSON Schema.

It shouldn't change anything functionally.

### Batch File CRLF
https://serverfault.com/questions/429594/is-it-safe-to-write-batch-files-with-unix-line-endings

It didn't have problems in this case, but because it was set in the `.editorconfig`, my editor (Visual Studio Code) automatically changed them anyway.

### Refactor
I just thought it'd be nice to do `datetime.utcnow()` once instead of three times. ^-^'

## Checklist

  * N/A
  * N/A

## Other information

Please just let me know if any changes are unfavorable and I can remove that commit.


SethFalco:
Hmm, looks like pre-commit CI specifically undid the change to the batch file. ^-^'

nijel:
I've fixed that in https://github.com/WeblateOrg/weblate/commit/1431349d6fe3d5c15ea409144fd648e51e234d54 and cherry-picked rest of the commits manually, thanks!

PR #7558, opened 2022-04-28
improvement: add `<link rel="(next|prev)"` to `<head>`

WofWca:
## Proposed changes

This should hint the browser to consider prefetching the next page,
which is especially beneficial on `/translate`.
https://html.spec.whatwg.org/multipage/links.html#link-type-next
https://developer.mozilla.org/en-US/docs/Web/HTTP/Link_prefetching_FAQ

## Checklist

## Other information

Possible drawback: e.g. if you edit a string and it is referenced on the next page (e.g. in "Nearby keys"), then if the next page gets prefetched, that referenced string will appear unchanged.
But then, in a perfect world, to negate this, I'd say it would be good to load "Nearby keys" separately from the page SPA-style so they get loaded after the next page loads, which would also benefit performance.

Did not test this since setting up a local copy is going quite rough for now. Gecko browsers currently should be doing prefetching based on these hints (see the links above), not sure about Chromium. Either way I think it's good practice.
Don't think it makes sense to add tests or docs for this.

I have also considered adding `prefetch` keyword for the `next` link, but I suppose let's let the browser decide. Maybe later.

Also I'm not sure if it's very good to use the `next_unit_url` and `prev_unit_url` in `base.html` as these variables MAY have different names in other files in the future, but currently it's just `edit.py` as far as I can tell.

### DIFF
--- a/weblate/templates/base.html
+++ b/weblate/templates/base.html
@@ -19,6 +19,13 @@
   <meta name="description" content="{{ description }}" />
   <meta name="twitter:card" content="summary_large_image" />
 
+  {% if next_unit_url %}
+  <link rel="next" href="{{ next_unit_url }}" />
+  {% endif %}
+  {% if prev_unit_url %}
+  <link rel="prev" href="{{ prev_unit_url }}" />
+  {% endif %}
+
   {% if object.slug and object.project.slug %}
   <meta property="og:image" content="{{ site_url }}{% url 'widget-image' project=object.project.slug widget='open' color='graph' extension='png' %}" />
   <link rel="alternate" type="application/rss+xml" title="RSS feed" href="{% url 'rss-component' project=object.project.slug component=object.slug %}" />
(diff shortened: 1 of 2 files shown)


WofWca:
Ok, I did some more testing, and it appears that neither Gecko nor Chromium do a prefetch. For Gecko it's because there are query parameters present in the URL, apparently.
Should we explicitly add `prefetch` then?

nijel:
I'm uncertain if prefetching is intended here. I definitely don't see the need for prefetching the previous page, most people won't go there.

> But then, in a perfect world, to negate this, I'd say it would be good to load "Nearby keys" separately from the page SPA-style so they get loaded after the next page loads, which would also benefit performance.

Not really, the things that are better to load on demand are already loaded that way.

Still, I find this patch valuable, so let's merge it this way and discuss prefetch separately.

PR #7567, opened 2022-04-29
fix: remove the duplicate `<meta name="robots"` element

WofWca:
## Proposed changes

It is defined above, also `index` and `follow` are the default anyway

## Checklist


nijel:

PR #7568, opened 2022-04-30
Rename `ckb` and `kmr`

roj1512:
## Proposed changes

Renamed `ckb` and `kmr` to Central Kurdish and Northern Kurdish.

## Checklist

## Other information

Reasons for the change:

- The current names are not popular, nor standard.
- The renamed entities are not often searched for by Kurm.. or Sora..
- To match with the name of the other dialect, `sdh` (Southern Kurdish).

### DIFF
--- a/weblate/trans/fixtures/simple-project.json
+++ b/weblate/trans/fixtures/simple-project.json
@@ -1919,7 +1919,7 @@
     "pk": 148,
     "fields": {
       "code": "ckb",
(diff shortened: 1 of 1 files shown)


nijel:
That file is a fixture for tests only, if you want to change the language names displayed, it is done in https://github.com/WeblateOrg/language-data/ (though this is already reflected there).

roj1512:
Thanks!

Issue #2916, opened 2019-07-29
Modernize frontend


guoyunhe:
**Is your feature request related to a problem? Please describe.**
We have a Weblate instance at https://l10n.opensuse.org/ and its UI is based on Bootstrap 3.x. However, other openSUSE project websites are using a theme based on Bootstrap 4.x. If we deploy Bootstrap 4.x, there will be incompatible components/styles that may break the UI functionality.

**Describe the solution you'd like**
I hope Weblate can support Bootstrap 4.x.

**Describe alternatives you've considered**

**Additional context**

nijel:
This has been already discussed in https://github.com/WeblateOrg/weblate/issues/1935, the biggest problem with Boostrap 4 is lack of RTL support. There were some third party mods which none of these looked complete, but I didn't investigate that for some time.

Even if that is handled, there is non trivial amount of work to migrate the UI between 3.x and 4.x.

guoyunhe:
Yes, here are a lot of class names changed. You have to check every component carefully. If one day you decide to open a `bootstrap4` branch, I can help with the migration.

nijel:
First we need working RTL support, then we can consider a migration. Making RTL work properly with Bootstrap 3 was painful as well, but we already have that running.

Bachstelze:
Bootstrap 4.4 is going to be a [LTS release](https://blog.getbootstrap.com/2019/07/24/lts-plan/) and could then support [RTL](https://github.com/twbs/bootstrap/issues/19545).

nijel:
There is also https://github.com/twbs/bootstrap/issues/28797 and https://github.com/twbs/bootstrap/issues/24807, but none really seems to be going anywhere close to having RTL supported.

Ahoba02:
I have created a tool to solve the issue.
Please visit
http://wahamira.net/tools/get-bootstrap-rtl

And kindly give me your feedback.

Ahoba02:
Did you check
http://wahamira.net/tools/get-bootstrap-rtl

nijel:
Boostrap 5 should have native RTL support (see https://github.com/twbs/bootstrap/pull/30980), then it makes more sense to switch to that (once it is released).

Bachstelze:
The beta version of Bootstrap 5 is released. While keeping jQuery as extra dependency most of our plugins should still work. Which changes will we have to figure out?

nijel:
[...]

PR #7586, opened 2022-05-03
WIP: split `.js`, load some asynchronously & on-demand, use ESM

WofWca:
## Proposed changes

This is super WIP, for now look at it as a discussion rather than a full-fledged MR.

In order to improve page loading performance, it would be cool to load some JS asynchronously and on-demand, and tree-shake some of it. Examples: shortcuts, syntax highlighting, username autocompletion.
Here's the audit I ran:

image

As you can see, on my PC it takes ~400ms to load a __prefetched__ page (i.e. it's only rendering, not including network download). You can imagine it's a little worse on mobile.

First step to this is rewriting the scripts with imports.

This also makes the code easier to maintain as there's fewer global vars.

TODO:
* [ ] Make sure nothing broke
* [ ] See TODOs in the code
* [ ] Perhaps a transpiler needs to be used as [ESM imports are not supported in old browsers (~92% now)](https://caniuse.com/es6-module-dynamic-import). Which one? Webpack?

## Checklist

### DIFF
--- a/weblate/static/editor/full.js
+++ b/weblate/static/editor/full.js
@@ -1,3 +1,10 @@
+import $ from "../vendor/jquery.js";
+import Cookies from "./vendor/js.cookie.js";
+import autosize from "./vendor/autosize.js";
+// TODO perf: import Mousetrap dynamically.
+import Mousetrap from "./vendor/mousetrap.js";
+import "./vendor/mousetrap-global-bind.js";
+
 (function () {
   var EditorBase = WLT.Editor.Base;
 
--- a/weblate/templates/base.html
+++ b/weblate/templates/base.html
@@ -48,6 +48,7 @@
   <script src="{% url 'js-catalog' %}" defer></script>
 {% compress js %}
 {% if rollbar_token %}
+<!-- TODO also separate this? But what if `!rollbar_token`? -->
 <script>
 var _rollbarConfig = {
     accessToken: "{{ rollbar_token }}",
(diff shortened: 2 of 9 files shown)


weblate/templates/base.html
```diff
@@ -62,23 +63,9 @@
...
 </script>
 {% endif %}
-  <script defer data-cfasync="false" src="{% static 'vendor/jquery.js' %}{{ cache_param }}"></script>
-  <script defer data-cfasync="false" src="{% static 'vendor/js.cookie.js' %}{{ cache_param }}"></script>
-  <script defer data-cfasync="false" src="{% static 'vendor/autosize.js' %}{{ cache_param }}"></script>
-  <script defer data-cfasync="false" src="{% static 'vendor/multi.js' %}{{ cache_param }}"></script>
-  <script defer data-cfasync="false" src="{% static 'vendor/mousetrap.js' %}{{ cache_param }}"></script>
-  <script defer data-cfasync="false" src="{% static 'vendor/mousetrap-global-bind.js' %}{{ cache_param }}"></script>
-  <script defer data-cfasync="false" src="{% static 'vendor/clipboard.js' %}{{ cache_param }}"></script>
+  <!-- TODO separate bootstrap? https://getbootstrap.com/docs/5.1/getting-started/javascript/#using-bootstrap-as-a-module -->
```
nijel, line 66: We're still at Bootstrap 3.4.1, see https://github.com/WeblateOrg/weblate/issues/2916

weblate/static/editor/full.js
```diff
@@ -1,3 +1,10 @@
+import $ from "../vendor/jquery.js";
+import Cookies from "./vendor/js.cookie.js";
+import autosize from "./vendor/autosize.js";
+// TODO perf: import Mousetrap dynamically.
+import Mousetrap from "./vendor/mousetrap.js";
```
nijel, line 5: Mousetrap is IMHO used on nearly every page - it is used to catch Ctrl+Enter for form submission.

nijel (commented):
Generally, any improvements to the JavaScript are welcome. The tricky part is that there are basically no tests for it, so it's super easy to break something. The current split out editor files are a typical example of that - it was tested by the author, me when merging and several issues were discovered months after merging in some not that frequently used features.

I think imports is something we can go for directly without a compatibility layer - we already started to use [arrow functions](https://caniuse.com/arrow-functions) two years ago and that was IMHO bigger drop in supported browsers than this.

weblate/static/editor/full.js
WofWca: Yeah, I saw that. Still, adding a bind asynchronously is possible.

WofWca:
A problem here is that the dependencies don't use ESM (native `export` and `import`), they do UMD or something else, so if you attempt to import them natively, the browser says that "the imported module doesn't provide export named ...".
So looks like we can't do without a transpiler.

PR #7675, opened 2022-05-20
improvement: `/translate`: turn translation filename into a repo link

WofWca:
## Checklist


weblate/trans/views/edit.py
```diff
@@ -666,6 +667,19 @@ def translate(request, project, component, lang):  # noqa: C901
...
+                '<a href="{}" target="_blank" rel="external noopener noreferrer">'
+                "{}</a>",
+                unit.translation.component.get_repoweb_link(
+                    # '1' as a placeholder, because `get_repoweb_link` can't currently
+                    # generate links without line specified. Although it's ok to use
+                    # '' or '0' on GitHub or GitLab, let's play it safe for now.
+                    unit.translation.filename,
+                    "1",
+                    user.profile.editor_link,
+                ),
```
nijel, line 680: `get_repoweb_link` returns `None` when there is no link template defined. The filename should be used instead of the link in this case. It might make more sense to factor out this logic from here and share it:

https://github.com/WeblateOrg/weblate/blob/0bf849faebe274238df4826b95a0958ffd5eac30/weblate/trans/templatetags/translations.py#L605-L633

nijel:

PR #7683, opened 2022-05-23
refactor: remove the majority of `mark_safe`

WofWca:
## Proposed changes

I tried to be careful, but I can't guarantee that nothing broke. Don't know how much of it is covered by tests. At the very least `/translate` works.

## Checklist

### DIFF
--- a/weblate/checks/duplicate.py
+++ b/weblate/checks/duplicate.py
@@ -19,8 +19,7 @@
 
 import re
 
-from django.utils.html import escape
-from django.utils.safestring import mark_safe
+from django.utils.html import format_html
 from django.utils.translation import gettext_lazy as _
 
 from weblate.checks.base import TargetCheck
@@ -94,8 +93,4 @@ class DuplicateCheck(TargetCheck):
         source = unit.source_string
         for target in unit.get_target_plurals():
             duplicate.update(self.check_single(source, target, unit))
-        return mark_safe(
-            "{} {}".format(
-                escape(self.description), escape(", ".join(sorted(duplicate)))
-            )
-        )
+        return format_html("{} {}", self.description, ", ".join(sorted(duplicate)))
--- a/weblate/checks/format.py
+++ b/weblate/checks/format.py
@@ -22,8 +22,7 @@ from collections import defaultdict
 from typing import Optional, Pattern
 
 from django.utils.functional import SimpleLazyObject
-from django.utils.html import conditional_escape
-from django.utils.safestring import mark_safe
+from django.utils.html import format_html, format_html_join
 from django.utils.translation import gettext
 from django.utils.translation import gettext_lazy as _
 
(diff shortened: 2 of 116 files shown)


weblate/checks/placeholders.py
```diff
@@ -137,7 +139,7 @@ def get_description(self, check_obj):
         if not self.has_value(unit):
             return super().get_description(check_obj)
         regex = self.get_value(unit)
-        return mark_safe(
-            escape(_("Does not match regular expression %s."))
```
WofWca, line 141: @nijel How can we avoid making translators have to edit this? Is there a way better than manually editing the `.po` files?

weblate/checks/duplicate.py
```diff
@@ -94,8 +93,4 @@ def get_description(self, check_obj):
         source = unit.source_string
         for target in unit.get_target_plurals():
             duplicate.update(self.check_single(source, target, unit))
-        return mark_safe(
-            "{} {}".format(
-                escape(self.description), escape(", ".join(sorted(duplicate)))
-            )
-        )
+        return format_html("{} {}", self.description, ", ".join(sorted(duplicate)))
```
WofWca, line 96: Looks like we can just return a formatted string, since it doesn't contain any HTML?

nijel: Sounds correct.

weblate/checks/format.py
```diff
@@ -415,7 +414,10 @@ def get_description(self, check_obj):
         if results:
             errors.extend(self.format_result(results))
         if errors:
-            return mark_safe("<br />".join(escape(error) for error in errors))
+            return format_html_join(
+                format_html("<br />"), "{}",
+                ((error,) for error in errors)
```
nijel, line 419: Not sure about this:

```suggestion
                (error for error in errors)
```

weblate/checks/placeholders.py
nijel: There is `scripts/locales-fixup` which might help with this (though it will have to be modified for this use-case).

weblate/checks/duplicate.py
WofWca: Actually I think let's keep it, there's practically no performance difference and in the future if someone decides to add HTML, they won't have to rewrite it.

weblate/trans/forms.py
```diff
@@ -197,80 +197,89 @@ def __init__(self, *args, **kwargs):
...
-                        value.encode("ascii", "xmlcharrefreplace").decode("ascii")
-                    ),
-                    char,
-                )
-            )
+        chars = format_html_join(
+            "\n",
+            BUTTON_TEMPLATE,
+            (
+                ("specialchar", name, format_html('data-value="{}"', value), char)
```
nijel, line 205: I think `encode("ascii", "xmlcharrefreplace").decode("ascii")` was used here to make sure that RTL direction markers do not confuse browsers (maybe it was an issue when looking at the page source, but I don't recall details).

WofWca: Not the biggest RTL expert, but looks like nothing broke. Having the page look like this, it produces the same HTML output (I've checked with `diff`):
image

WofWca: Whoops. It's because I copied it from devtools. Tried with `curl`, and indeed there is difference. Will search for a way.

nijel:

PR #7696, opened 2022-05-26
refactor: make `CDNJSAddon` a little safer from outputting unsafe code

WofWca:
## Proposed changes

Related: #7683

This makes tests fail though because it's invalid JS (in the template form, I mean)

## Checklist

### DIFF
--- a/weblate/templates/addons/js/weblate.js
+++ b/weblate/templates/addons/js/weblate.js
@@ -1,7 +1,7 @@
-var weblate_supported = ["{{ languages|safe }}"];
-var weblate_url = "{{ url }}";
-var weblate_selector = "{{ css_selector }}";
-var weblate_cookie_name = "{{ cookie_name }}";
+var weblate_supported = {{ languages }}
+var weblate_url = {{ url }};
+var weblate_selector = {{ css_selector }};
+var weblate_cookie_name = {{ cookie_name }};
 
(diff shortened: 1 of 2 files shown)


nijel (commented):
Thanks, looks good. The only issue is that linter now fails on the JS template. We can ignore the file in the `.pre-commit-config.yml`, not sure if there is a solution to make it lintable/formattable with the templating...

weblate/templates/addons/js/weblate.js
```diff
@@ -1,7 +1,7 @@
-var weblate_supported = ["{{ languages|safe }}"];
-var weblate_url = "{{ url }}";
-var weblate_selector = "{{ css_selector }}";
-var weblate_cookie_name = "{{ cookie_name }}";
+var weblate_supported = {{ languages }}
+var weblate_url = {{ url }};
+var weblate_selector = {{ css_selector }};
+var weblate_cookie_name = {{ cookie_name }};
```
nijel, line 4:
```suggestion
// prettier-ignore
var weblate_supported = {{ languages }}
// prettier-ignore
var weblate_url = {{ url }};
// prettier-ignore
var weblate_selector = {{ css_selector }};
// prettier-ignore
var weblate_cookie_name = {{ cookie_name }};
```

PR #7703, opened 2022-05-28
fix: `/credits` escaping HTML chars when outputting in `text/plain` mode

WofWca:
Also rewrite with `format_html` instead of `escape`
This also makes it not output leading and trailing newlines
as I think it was an error.

Same issue still exists for `/counts`, need to do the same

Before:

image

After:

image

## Checklist

### DIFF
--- a/weblate/trans/tests/test_reports.py
+++ b/weblate/trans/tests/test_reports.py
@@ -141,15 +141,14 @@ class ReportsComponentTest(BaseReportsTest):
     def test_credits_view_rst(self):
         response = self.get_credits("rst")
         self.assertEqual(response.status_code, 200)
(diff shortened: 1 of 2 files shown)


WofWca:
I see the escaping is in the tests? Is it just temporary, because of generalization? https://github.com/WeblateOrg/weblate/blob/22d577b1f1e88665a88b4569380148030e0f8389/weblate/trans/tests/test_reports.py#L155-L165 22d577b1f1e88665a88b4569380148030e0f8389

nijel:
You're right that it should not be escaped for rst.

WofWca:
Ok, I'll fix the test later today.

Issue #7708, opened 2022-05-29
Non-uppercase placeholders are not properly recognized


WofWca:
### Describe the issue

I use web extension JSON file format and use non-uppercase placeholders. It shows them in the placeholders section in uppercase, and doesn't highlight in the text:

![image](https://user-images.githubusercontent.com/39462442/170855164-5a9a8421-3fe0-4dcd-aa3f-5b596008979d.png)
![image](https://user-images.githubusercontent.com/39462442/170855171-e5c30023-2b09-4920-8b49-87c8b439acd6.png)

It starts highlighting if I change it to uppercase:

![image](https://user-images.githubusercontent.com/39462442/170855192-a0b6bd9f-c6b1-4e6c-a324-70d15ea1c455.png)

The string itself: https://hosted.weblate.org/translate/jump-cutter/extension-ui/en/?checksum=e358533698b9954c

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar issues in this repository.

### Expected behavior

Placeholders in the "flags" section preserve source string casing.

### How do you run Weblate?

weblate.org service

nijel:
Placeholders are currently case-sensitive in Weblate.

WofWca:
Ok, but the placeholder appears in lower-case in the source string, but in flags it became uppercase.

nijel:
It is extracted as uppercase because it is described that way in https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Internationalization#retrieving_message_strings_from_javascript.

It behaves this way since the beginning - eec96244458

WofWca:
Sounds like a bad practice, I don't think it should be followed.

I think we should look up how the placeholder is used in the source string (the `"message"` prop, not just "placeholders") and take casing from there.

WofWca:
I'll make an MR.

WofWca:
Also perhaps the same goes for ARB files? Everywhere I looked nobody really changes casing. But I don't know much about ARB.

https://github.com/WeblateOrg/weblate/blob/e21e8b22c5509c67a490eecc49b665c697a843a7/weblate/formats/ttkit.py#L929

Examples:
https://phrase.com/blog/posts/flutter-localization/
https://localizely.com/flutter-arb/

WofWca:
Should we maybe reopen this for ARB files? https://github.com/WeblateOrg/weblate/issues/7708#issuecomment-1141368868

[...]

PR #7722, opened 2022-05-30
fix: WebExtension placeholders not being recognized when typed non-uppercase

WofWca:
## Proposed changes

Fixes (partially?) #7708

## Checklist


WofWca:
Apparently this doesn't update casing for existing units, you have to re-parse the files.

Also I noticed that even with this change, when you update casing in the source string, the "flags" text will remain the same, and there's no highlighting, but checks are performed correctly (i.e. there's a "placeholders" failing check if the translation's placeholder casing doesn't match the source string.

image

nijel:

Issue #7713, opened 2022-05-29
Messages for the next page appear on the current one because of prefetching


WofWca:
### Describe the issue

In #7565 we added next page prefetching, and it appears to affect the current page in a weird way. See reproduction.

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar issues in this repository.

### Steps to reproduce the behavior

1. Use a Chromium-based browser for easier reproduction.
2. Go to `/translate`, to the last page of the search (e.g. https://hosted.weblate.org/translate/weblate/application/en/?&offset=2724)
3. Click "Previous page", or refresh the page.
4. A "The translation has come to an end." message will appear.

Instead of step 3 you can even open any hosted Weblate page in a new tab, it will show the message, lol.

### Expected behavior

Messages only appear for the pages they're supposed to be displayed on.

### Screenshots

![image](https://user-images.githubusercontent.com/39462442/170864236-65906d75-88db-44dd-bf72-e0e8f5c5f4ae.png)

### How do you run Weblate?

weblate.org service

### Additional context

https://github.com/WeblateOrg/weblate/blob/42220593bf4d6876ba25b1a98774eb4aec431aeb/weblate/templates/base.html#L23

https://github.com/WeblateOrg/weblate/blob/4b2e079b80103364292464b585186943b7172890/weblate/trans/views/edit.py#L560-L561

WofWca:
Looks like it's from a package.

https://github.com/WeblateOrg/weblate/blob/6f2fd4e36856f00e52da64b4c7f2d7472cd4bee7/weblate/utils/messages.py#L39-L42

https://github.com/WeblateOrg/weblate/blob/6f2fd4e36856f00e52da64b4c7f2d7472cd4bee7/weblate/utils/messages.py#L25

I thought we could maybe bail inside the `info` function if the request has the prefetch header.

nijel:
That would just silently discard the message... 

But the message itself is not the problem, this also deletes cached search results from the session, so the search won't be reproducible anymore, what could break translating the last string (if the search would now return less results).

https://github.com/WeblateOrg/weblate/blob/563f0ce76e831e652a72121ff5dfcc949ce27810/weblate/trans/views/edit.py#L561-L564

I think disabling prefetching is the safest way to go for now, I will revert 42220593bf4d6876ba25b1a98774eb4aec431aeb.

WofWca:
> That would just silently discard the message...

[...]

PR #7723, opened 2022-05-30
improvement: add `prefetch` for `<link rel="next`

WofWca:
Revert 8dc6e31637f8b1d1d43d695c97c99e0cda16e6da.
* [ ] Need to understand how to circumvent #7713.

## Checklist

### DIFF
--- a/weblate/templates/base.html
+++ b/weblate/templates/base.html
@@ -20,7 +20,7 @@
   <meta name="twitter:card" content="summary_large_image" />
 
   {% if next_unit_url %}
-  <link rel="next" href="{{ next_unit_url }}" />
+  <link rel="next prefetch" href="{{ next_unit_url }}" />
   {% endif %}
   {% if prev_unit_url %}
   <link rel="prev" href="{{ prev_unit_url }}" />


nijel:
Is really the prefetching worth of the effort?

- It will have outdated current translation, and possible some other info as well.
- It has issues with flushing the search list when going behind the last string (#7713).
- It might have a negative effect on bandwidth usage and server load.

WofWca:
I'd say it's huge. Currently it takes ~900-1500 ms to load a page. To go through 60 strings is an entire minute of waiting time. It's especially noticeable if you're just skimming through - verifying, checking.

1. Well, the same can be true without prefetching. If you've opened the page, and then translation changes. It's just like opening the page a little earlier than you would. And I believe prefetched pages get discarded after a couple of minutes reg
2. Yeah, need to fix it. I think GET requests should cause as little changes as possible.
3. Regarding client-side bandwidth - browsers decide for themselves if they can afford prefetching. Regarding server-side - I think in this case it's very likely for the next page to be visited, so almost no additional load here. If not - it's just one extra page per every `/translate` page. In the absolute worst case it's double load.

I personally was so happy that it became that fast.

WofWca:
> than fixing prefetching to work without side effects

Ok, I have an idea.
Browsers usually send an HTTP header that distinguishes a prefetch request from a regular one. On the backend we process the request as usual, except before doing a state mutation like [this](https://github.com/WeblateOrg/weblate/issues/7713#issuecomment-1141067899) we check if it's a prefetch request, and if it is, we, instead of mutating the state, send a dummy response with a header like `Cache-control: must-revalidate` that would tell the browser that it can't reuse this response when the actual request is performed.

PR #7740, opened 2022-06-04
Update installation instructions

Justman10000:
With this version of the documentation, I can easily set up Weblate... [Here](https://test-weblate.readthedocs.io) a preview is available!


docs/admin/install/steps/install-after.rst
```diff
@@ -13,7 +13,7 @@ Congratulations, your Weblate server is now running and you can start using it.
 Adding translation
 ++++++++++++++++++
 
-#. Open the admin interface (``http://localhost:8000/create/project/``) and create the project you
+#. Open the admin interface (``http://localhost:8000/create/project``) and create the project you
```
nijel, line 16: The URL has a trailing slash...

docs/admin/install/steps/install-configure.rst
```diff
@@ -47,7 +47,7 @@ Configuring Weblate
 
          ~/weblate-env/lib/python3.7/site-packages/weblate/examples/celery start
 
-#. Start the development server (see :ref:`server` for production setup):
+#. Start the development server:
```
nijel, line 50: What is the reason for removal here?

docs/admin/install/steps/install-python.rst
```diff
@@ -6,27 +6,32 @@ Python modules
...
+.. warning::
+
+   Run the following commands in an empty directory
 
 #. Create the virtualenv for Weblate:
 
    .. code-block:: sh
 
-        virtualenv --python=python3 ~/weblate-env
+        python3 -m venv .
```
nijel, line 17: What is wrong on using `virtualenv`?

docs/admin/install/steps/install-python.rst
```diff
@@ -6,27 +6,32 @@ Python modules
...
 
 #. Install Weblate including all optional dependencies:
 
    .. code-block:: sh
 
         # pkgconfig is needed to install borgbackup 1.2
-        pip install pkgconfig
+        pip3 install pkgconfig
+        # mysqlclient is needed if you want use mysql/mariadb
+        pip3 install mysqlclient
```
nijel, line 32: The recommended setup is using PostgreSQL. Still, if you want to use Weblate with MySQL, it is better to install it using `pip install Weblate[all,MySQL]`, that way you will get whatever is needed to run Weblate with MySQL.

weblate/examples/apache.conf
```diff
@@ -3,38 +3,53 @@
...
+    RewriteRule ^ https://%{HTTP_HOST}%{REQUEST_URI} [R=301,L]
+</VirtualHost>
+
+<VirtualHost *:443>
+    ServerAdmin admin@<your_domain>
+    ServerName <your_domain>
 
     # DATA_DIR/static/favicon.ico
-    Alias /favicon.ico /home/weblate/data/static/favicon.ico
+    Alias /favicon.ico /home/weblate/lib/python3.9/site-packages/data/static/favicon.ico
```
nijel, line 23: The DATA_DIR is better to be placed outside the package directories.

weblate/settings_example.py
```diff
@@ -29,13 +29,13 @@
...
 # Whether site uses https
-ENABLE_HTTPS = False
+ENABLE_HTTPS = True
 
 #
 # Django settings for Weblate project.
 #
 
-DEBUG = True
+DEBUG = False
```
nijel, line 38: The debug is intentionally enabled for development purposes. The same applies to HTTPS, which is not supported by Django built-in server.

docs/admin/install/steps/install-python.rst
```diff
@@ -6,27 +6,32 @@ Python modules
...
-        . ~/weblate-env/bin/activate
+        source bin/activate
 
 #. Install Weblate including all optional dependencies:
 
    .. code-block:: sh
 
         # pkgconfig is needed to install borgbackup 1.2
-        pip install pkgconfig
+        pip3 install pkgconfig
```
nijel, line 30: It should be `pip` as well in the virtualenv.

docs/admin/install/venv-debian.rst
```diff
@@ -8,38 +8,37 @@ Installing on Debian and Ubuntu
...
 .. include:: steps/install-system-optional.rst
 
 .. code-block:: sh
 
-    apt install tesseract-ocr libtesseract-dev libleptonica-dev
-    apt install libldap2-dev libldap-common libsasl2-dev
-    apt install libxmlsec1-dev
+    apt install tesseract-ocr libtesseract-dev libleptonica-dev -y \
+    libldap2-dev libldap-common libsasl2-dev -y \
+    libxmlsec1-dev -y
```
nijel, line 22: Too much of `-y`.

docs/admin/install/steps/install-configure.rst
Justman10000: It occurs below

docs/admin/install/steps/install-python.rst
Justman10000: It doesn't matter in that sense, so you could just as well replace this with mine, it's shorter...

weblate/settings_example.py
Justman10000: If HTTPS isn't supported, why does it exist in the config?

docs/admin/install/steps/install-python.rst
Justman10000: Does not matter? When using Python3, can one use both pip and pip3?

docs/admin/install/steps/install-after.rst
Justman10000: /create/project redirects to /create/project/ anyway

weblate/settings_example.py
nijel: It is not supported by Django built-in server, which is suitable for development only.

weblate/examples/apache.conf
nijel: It's path where Weblate stores data, it is configured in the settings.

docs/admin/install/steps/install-after.rst
nijel: Yes, it redirects, but why to change the working URL to one which redirects?

docs/admin/install/steps/install-configure.rst
nijel: Yes, it occurs, but it is important to point people to the production setup here.

docs/admin/install/steps/install-python.rst
nijel: You are doing many changes which "do not matter". These are useless and make review harder - the more changes are in a PR, the easier it is to miss something important.

docs/admin/install/steps/install-configure.rst
Justman10000: It's already mentioned below, also __Running server__ only explains how to start the devserver!

weblate/examples/apache.conf
Justman10000: Maybe so, I just have to set up the above so that Weblate starts at all...

Justman10000:
@nijel

updated! Look at my answers and if the update fits so far...

docs/admin/install/steps/install-configure.rst
nijel: Still, people did use this for production, so better to keep the mention here.

Justman10000: You can't use something for the productive environment that only applies to the development environment...

docs/admin/install/redis.rst
```diff
@@ -0,0 +1,43 @@
...
+
+    # Test the installation (optional)
+    make test
+
+    # Run redis
+    screen -S Redis src/redis-server
+
+    # Leave directory
+
+    cd ..
```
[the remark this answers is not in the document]
nijel: But there is no need to include it - users might want to compile anything Weblate needs from the source, and there is no point in including this in Weblate documentation. It's not only about writing it, but it also needs a maintenance.

weblate/examples/apache.conf
nijel: Okay, let's guide users towards configuring this in https://github.com/WeblateOrg/weblate/pull/7858

docs/admin/install/redis.rst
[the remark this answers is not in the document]
nijel: The documentation should cover things specific to Weblate, there is nothing Weblate specific in installing Redis from source. And because that I see no reason to include in here.

docs/admin/install/steps/install-python.rst
nijel: I've replaced it with even shorter in 4adc387772162f7d6b1322930644079f4873bd63

nijel:
Thanks for your pull request. I've included most of the changes in separate commits (they are linked above).

PR #7759, opened 2022-06-10
chore: Set permissions for GitHub actions

neilnaveen:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: neilnaveen <42328488+neilnaveen@users.noreply.github.com>

### DIFF
--- a/.github/workflows/closing.yml
+++ b/.github/workflows/closing.yml
@@ -5,8 +5,14 @@ on:
   issues:
     types: [closed]
 
+permissions:
+  contents: read
+
 jobs:
   issueClosed:
+    permissions:
+      issues: write  # for peter-evans/create-or-update-comment to create or update comment
+      pull-requests: write  # for peter-evans/create-or-update-comment to create or update comment
     runs-on: ubuntu-20.04
     steps:
     - name: Add closed question comment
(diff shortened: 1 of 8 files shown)


.github/workflows/closing.yml
```diff
@@ -5,8 +5,14 @@ on:
   issues:
     types: [closed]
 
+permissions:
+  contents: read
+
```
nijel, line 10: This file is maintained in https://github.com/WeblateOrg/meta/, any changes in it will be overwritten on next sync. Please do changes there.

.github/workflows/labels.yml
```diff
@@ -5,8 +5,14 @@ on:
   issues:
     types: [labeled]
 
+permissions:
+  contents: read
+
```
nijel, line 10: This one lives in https://github.com/WeblateOrg/meta/ as well (see comment on the top).

nijel:

PR #7796, opened 2022-06-22
feat: explicitly link `GettextCustomizeForm.width` choices to `xgettext` flags

cfm:
## Proposed changes

The original `GettextCustomizeForm.width` specified its choices in terms of gettext behavior.  In 54c519d, this specification was removed in favor of help text.  While the help text clarifies gettext's own behavior, highlighting the comparison to `xgettext` and its flags is helpful for choosing the right Weblate option.

Motivating discussion:  #7791

## Checklist

- [n/a] I have added tests that prove my fix is effective or that my feature works.
- [n/a] I have added documentation to describe my feature.

## Other information

Feel free to close if you prefer the current text!  In that case, just consider this user feedback.  :-)

### DIFF
--- a/weblate/addons/forms.py
+++ b/weblate/addons/forms.py
@@ -112,8 +112,8 @@ class GettextCustomizeForm(BaseAddonForm):
     width = forms.ChoiceField(
         label=_("Long lines wrapping"),
         choices=[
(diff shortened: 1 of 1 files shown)


nijel:

PR #7840, opened 2022-07-03
Enhancement: add dynamic components selection on auto translation form

warlof:
## Proposed changes

I discover last that our user where non longer able to pick a component while they use the "auto-translate" feature and need to know how to find the project/component slug to fill the field properly.

While digging in code, I discovered it was only a limitation - assume based on load against database - when components amount was exceeding 30.

This pull request is replacing the choice field/text field swap logic with a dynamic loading select field using AJAX calls.

Now, when user want to use auto-translate feature, he just need to fill the field with what he was looking for using either the component or project name. The system will back to him with a list of available components matching his criterias. From this list, he just need to pick an item which will be used to complete the field.

image

Each returned item are compounds of the component name (first line) and the related project (second line).

## Checklist

## Other information

In order to achieve this improvement, I add a third party package called Tom Select.
It is well designed, lightweight and stable.
See : https://tom-select.js.org

This enhancement is depending on a new API endpoint to allow system to retrieve available components.
The route is `/api/components/shared/` and it accepts one query parameter called `search` in order to limit results.

I tried to do my best, however, I'm not used to develop in Python / neither Django, so do not hesitate to point me on things I might did wrong :)

### DIFF
--- a/weblate/templates/translation.html
+++ b/weblate/templates/translation.html
@@ -6,6 +6,10 @@
 {% load humanize %}
 {% load metrics %}
 
+{% block extra_script %}
+<link rel="stylesheet" href="/static/vendor/bootstrap/css/tom-select.bootstrap4.css" />
+{% endblock %}
+
 {% block breadcrumbs %}
 <li><a href="{{ object.component.project.get_absolute_url }}">{{ object.component.project }}</a></li>
 {% include "snippets/component-breadcrumb.html" with object=object.component %}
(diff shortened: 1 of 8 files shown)


warlof:
@nijel thanks for the review. Regarding self project flow, what would be best approach ? Adding a filter on project, or using project components when field is empty ? (I think it was default when the field was not filled ?)

Regarding the second form, do you have its location ? 😁

weblate/templates/translation.html
```diff
@@ -234,6 +238,33 @@ <h4 class="panel-title">
...
+            .catch(() => {
+                callback();
+            });
+    },
+    render: {
+        option: function(item, escape) {
+             return `<div><h4 class="list-group-item-heading">${ escape(item.name) }</h4><div class="list-group-item-text text-primary">${ escape(item.project.name) }</div></div>`;
+        },
+    },
+});
```
nijel, line 266: Please avoid inline scripts, these are blockec by Content-Security-Policy.

warlof: Ok, I'll move it in a static file. Maybe there is already a file into which I could inject that with some refactor so it's only valid for this form?

> Ideally all needed parts would be rendered by the form itself.

I'm not sure I follow you. Does it mean I should find a way to make the class rendering the added part in the template file ?

nijel: You can add needed CSS and scripts to [weblate/templates/addons/addon_detail.html](https://github.com/WeblateOrg/weblate/blob/main/weblate/templates/addons/addon_detail.html). It's acceptable for this single use case, but once there will be more features like this, it will need refactoring to avoid loading too much of unused code.

Another option might be using ESM to load the module, this has been outlined in https://github.com/WeblateOrg/weblate/pull/7586, but it has its challenges as well. But maybe introducing it for new code might be easier than converting existing one.

PR #7900, opened 2022-07-18
memory: implement delete entry from automatic suggestions

### LINKED ISSUE #6440: Translation Memory Maintenance - Delete Entries
It would be good to be able to maintain the TM in one way or the other.

Perhaps 
- [ ] Search & Replace in TM
- [ ] Search & Delete entries in TM

Or perhpas, as a "simple" (?) starting point, giving the translator the option to delete single entries from the TM when they see the result list in **Automatic Suggestions**. Like perhaps:

- [ ] Delete single entry in **Automatic Suggestions** view:
image

ParthS007:
## Proposed changes

closes #6440 

- Added a Delete entry button for removing translation suggestion from Automatic Suggestions

## Checklist

## Other information

Screenshot 2022-07-18 at 11 19 45

### DIFF
--- a/weblate/machinery/views.py
+++ b/weblate/machinery/views.py
@@ -30,7 +30,7 @@ from django.shortcuts import get_object_or_404
 from django.urls import reverse
 from django.utils.functional import cached_property
 from django.utils.translation import gettext as _
-from django.views.decorators.http import require_POST
+from django.views.decorators.http import require_http_methods, require_POST
 from django.views.generic import TemplateView
 from django.views.generic.edit import FormView
 
@@ -344,7 +344,7 @@ def handle_machinery(request, service, unit, search=None):
     return JsonResponse(data=response)
 
 
-@require_POST
+@require_http_methods(["DELETE", "POST"])
 def translate(request, unit_id, service):
     """AJAX handler for translating."""
     unit = get_object_or_404(Unit, pk=int(unit_id))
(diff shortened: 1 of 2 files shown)


weblate/static/editor/full.js
```diff
@@ -535,6 +561,11 @@
             '<a class="js-copy-save-machinery btn btn-primary">' +
             gettext("Copy and save") +
             "</a>" +
+            "</td>" +
+            "<td>" +
+            '<a class="js-delete-machinery btn btn-danger">' +
+            gettext("Delete entry") +
+            "</a>" +
```
nijel, line 568: We're able to delete entries only from a single service (Weblate translation memory), so the button should be conditional here.

weblate/machinery/views.py
```diff
@@ -344,7 +344,7 @@ def handle_machinery(request, service, unit, search=None):
     return JsonResponse(data=response)
 
 
-@require_POST
+@require_http_methods(["DELETE", "POST"])
 def translate(request, unit_id, service):
     """AJAX handler for translating."""
     unit = get_object_or_404(Unit, pk=int(unit_id))
```
nijel, line 350: I'm uncertain if it's a good idea to mix this endpoint with removal. Anyway the removal lacks implementation here.

weblate/static/editor/full.js
ParthS007: okay, I will add the condition.

weblate/machinery/views.py
ParthS007: For the removal part, I asked a couple of question [here](https://github.com/WeblateOrg/weblate/issues/6440#issuecomment-1186991255)

PR #7946, opened 2022-07-30
Spelling: Deleted entries, translation memory

comradekingu:
?

### DIFF
--- a/weblate/memory/views.py
+++ b/weblate/memory/views.py
@@ -120,7 +120,9 @@ class RebuildView(MemoryFormView):
         import_memory.delay(project_id=project.id, component_id=component_id)
         messages.success(
             self.request,
-            _("Deleted entries and the translation memory will be rebuilt in the background."),
+            _(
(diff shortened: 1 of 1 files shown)


weblate/memory/views.py
```diff
@@ -120,7 +120,9 @@ def form_valid(self, form):
         import_memory.delay(project_id=project.id, component_id=component_id)
         messages.success(
             self.request,
-            _("Entries deleted and memory will be rebuilt in the background."),
+            _(
+                "Deleted entries and the translation memory will be rebuilt in the background."
```
nijel, line 124:
```suggestion
                "Deleted entries and the translation memory will be "
                "rebuilt in the background."
```

WofWca:
Isn't it supposed to be more like

> Entries have been deleted. memory will be rebuilt in the background.

?

nijel:
Yes, that's better. Changed in 704ea34b3ac57a3494f75e1aa93e0afb718daa17

PR #7967, opened 2022-08-04
add absolute links for source locations

### LINKED ISSUE #7960: Support absolute web links (at least) in PO files
### Describe the problem

Our translators are working on a documentation site that gets updated (in a staging area) almost instantly after they updated their translation. They have no intuitive, direct way to access the result of this rendering, or maybe even the situation _before_ the translation to get more context.

We are thinking about adding the absolute link as a reference (`#:`) to the PO files. They are then showing up as expected in the "Source String Location" section. The only issue is that weblate tries to interpret them as a VCS link, interpreting "https" as the file name and "//www.example.com/path/doc.html" as the line number.
If we could add a detection for web links (either checking for `https?://` or, more generically, for `[\w]+://`), which would lead to just linking to that absolute location, we could very seamlessly use that feature.

While this scenario is not mentioned in any documentation I could find about the "formal" gettext format, I do not think that it contradicts the intention of the reference field.

Screenshots: Adding screenshots to source strings is cumbersome and does not provide the best experience for the translator. 

I don't know if there are alternative approaches, short of buliding some kind of add-on, hook, or the like.

### Screenshots

image

### Additional context

_No response_

oliverrahner:
## Proposed changes

Implements the change proposed in #7960 (closes #7960)

## Checklist

- [X] I have added tests that prove my fix is effective or that my feature works.

### DIFF
--- a/weblate/trans/templatetags/translations.py
+++ b/weblate/trans/templatetags/translations.py
@@ -610,7 +610,7 @@ def try_linkify_filename(
     link will be an absolute link to the specified resource.
     """
     link = None
-    if re.search(r"https?://", text):
+    if re.search(r"^https?://", text):
         link = text
     elif profile:
         link = unit.translation.component.get_repoweb_link(
--- a/weblate/trans/templatetags/translations.py
+++ b/weblate/trans/templatetags/translations.py
@@ -606,11 +606,11 @@ def try_linkify_filename(
     """
     Attempt to convert `text` to a repo link to `filename:line`.
 
-    If the `text` is prefixed with a protocol, like https://, http://, ftp://, the
+    If the `text` is prefixed with http:// or https://, the
     link will be an absolute link to the specified resource.
     """
     link = None
-    if re.search(r"\w+://", text):
+    if re.search(r"https?://", text):
         link = text
     elif profile:
         link = unit.translation.component.get_repoweb_link(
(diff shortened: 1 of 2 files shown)


weblate/trans/templatetags/translations.py
```diff
@@ -603,9 +603,16 @@ def unit_state_title(unit) -> str:
...
-    """Attemp to convert `text` to a repo link to `filename:line`."""
+    """
+    Attempt to convert `text` to a repo link to `filename:line`.
+
+    If the `text` is prefixed with a protocol, like https://, http://, ftp://, the
+    link will be an absolute link to the specified resource.
+    """
     link = None
-    if profile:
+    if re.search(r"\w+://", text):
```
nijel, line 613: I don't think we want to whitelist all protocols here. I'd stick with `http://` and `https://` only.

oliverrahner: Agreed and changed!

weblate/trans/templatetags/translations.py
```diff
@@ -606,11 +606,11 @@ def try_linkify_filename(
     """
     Attempt to convert `text` to a repo link to `filename:line`.
 
-    If the `text` is prefixed with a protocol, like https://, http://, ftp://, the
+    If the `text` is prefixed with http:// or https://, the
     link will be an absolute link to the specified resource.
     """
     link = None
-    if re.search(r"\w+://", text):
+    if re.search(r"https?://", text):
```
nijel, line 613: Shouldn't it look only at string beginning?
```suggestion
    if re.match("^https?://", text):
```

oliverrahner: fixed it, but re-added the raw string literal

nijel:

PR #7969, opened 2022-08-04
api: add api to list and delete memory entries

ParthS007:
## Proposed changes

Ref: https://github.com/WeblateOrg/weblate/issues/6440

- Add API for listing and deleting memory entries

## Checklist

## Other information

Screenshot 2022-08-04 at 12 37 46

Screenshot 2022-08-04 at 14 27 57

Screenshot 2022-08-04 at 12 38 04

@nijel  Please let me know what else I can add in the documentation and for the API test. Thanks

### DIFF
--- a/weblate/api/tests.py
+++ b/weblate/api/tests.py
@@ -31,6 +31,7 @@ from weblate_language_data.languages import LANGUAGES
 from weblate.accounts.models import Subscription
 from weblate.auth.models import Group, Role, User
 from weblate.lang.models import Language
+from weblate.memory.models import Memory
 from weblate.screenshots.models import Screenshot
 from weblate.trans.models import (
     Change,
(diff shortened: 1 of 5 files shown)


weblate/api/views.py
```diff
@@ -982,6 +984,24 @@ def download_archive(self, request, **kwargs):
...
 
 
+class MemoryViewSet(viewsets.ModelViewSet, DestroyModelMixin):
+    """Memory API."""
+
+    queryset = Memory.objects.none()
+    serializer_class = MemorySerializer
+
+    def get_queryset(self):
+        return Memory.objects.order_by("id")
```
nijel, line 994: Either all this view needs to be guarded by permission, or the list filtered to list only allowed entries.

As the filtering will be quite complex, I'd stick with allowing listing to superusers only.

weblate/api/views.py
```diff
@@ -982,6 +984,24 @@ def download_archive(self, request, **kwargs):
...
+
+    def perm_check(self, request):
+        if not request.user.has_perm("memory.delete"):
```
nijel, line 997: This should get a memory instance to check permission in the correct scope.

weblate/api/tests.py
```diff
@@ -2183,6 +2183,35 @@ def test_patch(self):
...
+
+    def test_delete(self):
+        response = self.do_request(
+            "api:memory-detail",
+            kwargs={"pk": 1},
+            method="delete",
+            superuser=True,
+            code=204,
+        )
+        self.assertEqual(response.status_code, 204)
```
nijel, line 2212: The do_request already checks the status code, so it is not needed again.
```suggestion
        self.do_request(
            "api:memory-list",
            method="get",
            superuser=True,
            code=200,
        )

        self.do_request(
            "api:memory-list",
            method="get",
            superuser=False,
            code=403,
        )

    def test_delete(self):
        self.do_request(
            "api:memory-detail",
            kwargs={"pk": 1},
            method="delete",
            superuser=True,
            code=204,
        )
```

Issue #7899, opened 2022-07-18
Send notify/fedora messages when a source or translation unit is being deleted. 


jojurajan:
### Describe the problem

We use Fedora consumer for propagating changes from Weblate back to our legacy translation systems. Currently, we don't have a way to know when a translation/source string is being deleted from Weblate other than fetching data via API and comparing them.

### Describe the solution you'd like

To keep the translations in sync, we would like to get Fedora/notify messages when someone deletes a source/translation unit from Weblate.

sandratoo:
Hi, can I try this out?

nijel:
Yes! What needs to be done:

* Add new event to the `Change` model to track these (ACTION_STRING_REMOVE)
* Adjust delete endpoints to generate change entry

Once this is done, it will automatically end up in notifications.

PR #7975, opened 2022-08-05
Add notifications when a translation or source unit is deleted

superDross:
Add notifications when a translation or source unit is deleted

This is intended to address the following issue: https://github.com/WeblateOrg/weblate/issues/7899

## Proposed changes

As above, specifically for pushing deletion messages to Fedora messaging.

## Checklist

- [x ] I have added tests that prove my fix is effective or that my feature works.
- [x ] I have added documentation to describe my feature.
- [x ] I have squashed my commits into logic units.
- [x ] I have described the changes in the commit messages.


weblate/accounts/notifications.py
```diff
@@ -622,6 +622,15 @@ class TranslatedStringNotificaton(Notification):
     filter_languages = True
 
 
+@register_notification
+class TranslatedStringRemovedNotificaton(Notification):
+    actions = (Change.ACTION_STRING_REMOVE,)
+    # Translators: Notification name
+    verbose = _("Translated string removed")
+    template_name = "translated_string_removed"
+    filter_languages = True
```
nijel, line 631: I don't see a use case for this notification - not all change events need a notification e-mail.

weblate/trans/models/translation.py
```diff
@@ -1429,6 +1429,14 @@ def add_unit(  # noqa: C901
             self.notify_new(request)
         return result
 
+    def notify_deletion(self, unit, user):
+        change = Change(
+            unit=unit,
```
nijel, line 1434: As `Change` object is configured to cascade on unit removal, it will only exist for very short time and will be removed afterwards. In this case, it is probably better to link the event to translation and include the current string and context in `target` and/or `details`.

weblate/accounts/notifications.py
superDross: Fair point. I have removed the email notification and associated html files.

weblate/trans/models/translation.py
```diff
@@ -1429,6 +1429,19 @@ def add_unit(  # noqa: C901
...
+            translation=self,
+            action=Change.ACTION_STRING_REMOVE,
+            user=user,
+            target=unit.target,
+            details={
+                "source": unit.source,
+                "target": unit.target,
+            },
+        )
+        change.save(force_insert=True)
```
nijel, line 1443: This can be a bit more compact:

```suggestion
        self.change_set.create(
            action=Change.ACTION_STRING_REMOVE,
            user=user,
            target=unit.target,
            details={
                "source": unit.source,
                "target": unit.target,
            },
        )
```

weblate/trans/models/translation.py
```diff
@@ -1473,6 +1485,7 @@ def delete_unit(self, request, unit):
                     source_unit = translation_unit.source_unit
                     if source_unit.source_unit.unit_set.count() == 1:
                         source_unit.delete()
+                        self.notify_deletion(source_unit, user)
```
nijel, line 1488: This is the correct Translation object to trigger notification:

```suggestion
                        source_unit.translation.notify_deletion(source_unit, user)
```

superDross: I am not so sure that's correct. I have tested my code change locally and it works with weblate fedora messaging. I was also able to consume the message from a custom consumer too. Also the `notify_delete` method is part of the same class so `self.notify_delete` should be sufficient.

Unless there is some additional context I am unaware of?

nijel: You are deleting the source unit here, so the matching Translation object should get the notification, not the current translation. Also, this code path is not used in monolingual setup, so you are probably not testing this at all ;-).

docs/changes.rst
```diff
@@ -16,6 +16,7 @@ Released on August 22nd 2022.
 * Improved reliability of HTML and text files translation.
 * Added support for project level backups.
 * Improved performance and memory usage of translation memory lookups.
+* Added notifications for string removals.
```
nijel, line 19: This should be moved to 4.14.1 as 4.14 has already been released.

superDross: I have moved this to the correct place now

weblate/trans/models/translation.py
superDross: Oh yeah valid point, this has been altered accordingly.

PR #7984, opened 2022-08-09
memory: implement delete entry from automatic suggestions

### LINKED ISSUE #6440: Translation Memory Maintenance - Delete Entries

ParthS007:
## Proposed changes

closes #6440 

- Added a Delete entry button for removing translation suggestion from Automatic Suggestions

## Checklist

### DIFF
--- a/weblate/memory/machine.py
+++ b/weblate/memory/machine.py
@@ -17,6 +17,8 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
+from django.urls import reverse
+
 from weblate.machinery.base import MachineTranslation, get_machinery_language
 from weblate.memory.models import Memory
 
@@ -72,4 +74,7 @@ class WeblateMemory(MachineTranslation):
                 "origin": result.get_origin_display(),
                 "source": result.source,
                 "show_quality": True,
+                "delete_url": reverse("api:memory-detail", kwargs={"pk": result.id})
+                if user is not None and user.has_perm("memory.delete", result)
+                else None,
             }
--- a/weblate/static/editor/full.js
+++ b/weblate/static/editor/full.js
@@ -36,6 +36,41 @@
       submitForm({ target: this.$translationArea });
     });
 
+    /* Delete machinery results */
+    this.$editor.on("click", ".js-delete-machinery", (e) => {
+      var $el = $(e.target);
+
+      /* Delete Url dialog */
+      var $deleteEntriesDialog = null;
+      this.$editor.on("show.bs.modal", "#delete-url-modal", (e) => {
+        $deleteEntriesDialog = $(e.currentTarget);
+        $deleteEntriesDialog.find(".modal-body").html("");
+        var text = $el.parent().parent().data("raw").text;
+        var modalBody = this.machinery.renderDeleteUrls(text);
+        $deleteEntriesDialog.find(".modal-body").append(modalBody);
+      });
+
+      this.$editor.on("hide.bs.modal", "#delete-url-modal", (e) => {
+        $deleteEntriesDialog = null;
+      });
+
+      this.$editor.on("submit", ".delete-url-form", (e) => {
+        var $form = $(e.currentTarget);
+        var $deleteEntries = $form.find("input.form-check-input:checked");
+        if ($deleteEntriesDialog === null) {
+          return false;
+        }
+        $deleteEntriesDialog.modal("hide");
+
+        Object.entries($deleteEntries).forEach(([_, entry]) => {
+          if (typeof entry.id !== "undefined") {
+            this.removeTranslationEntry(entry.id);
+          }
+        });
+        return false;
+      });
+    });
+
     Mousetrap.bindGlobal("alt+end", function (e) {
       window.location = $("#button-end").attr("href");
       return false;
@@ -210,6 +245,20 @@
     });
   };
 
+  FullEditor.prototype.removeTranslationEntry = function (delete_url) {
+    $.ajax({
+      type: "DELETE",
+      url: delete_url,
+      headers: { "X-CSRFToken": this.csrfToken },
+      success: () => {
+        addAlert(gettext("Translation memory entry removed."));
+      },
+      error: (jqXHR, textStatus, errorThrown) => {
+        addAlert(errorThrown);
+      },
+    });
+  };
+
   FullEditor.prototype.fetchMachinery = function (serviceName) {
     $.ajax({
       type: "POST",
@@ -261,6 +310,7 @@
         ...this.machinery.state.translations,
         ...data.translations,
       ],
+      weblateTranslationMemory: new Set(),
       lang: data.lang,
       dir: data.dir,
     });
@@ -497,6 +547,7 @@
     constructor(initialState = {}) {
       this.state = {
         translations: [],
+        weblateTranslationMemory: new Set(),
         lang: null,
         dir: null,
       };
@@ -541,6 +592,20 @@
         )
       );
 
+      if (this.state.weblateTranslationMemory.has(el.text)) {
+        row.append(
+          $(
+            "<td>" +
+              '<a class="js-delete-machinery btn btn-danger" data-toggle="modal" data-target="#delete-url-modal">' +
+              gettext("Delete entry") +
+              "</a>" +
+              "</td>"
+          )
+        );
+      } else {
+        row.append($("<td></td>"));
+      }
+
       return row;
     }
 
@@ -549,6 +614,7 @@
       if (typeof el.origin !== "undefined") {
         service.append(" (");
         var origin;
+        var deleteUrl = false;
         if (typeof el.origin_detail !== "undefined") {
           origin = $("<abbr/>").text(el.origin).attr("title", el.origin_detail);
         } else if (typeof el.origin_url !== "undefined") {
@@ -556,10 +622,41 @@
         } else {
           origin = el.origin;
         }
+        if (el.delete_url) {
+          this.state.weblateTranslationMemory.add(el.text);
+        }
         service.append(origin);
         service.append(")");
       }
       return service;
+    }
+
+    renderDeleteUrls(text) {
+      var translations = this.state.translations;
+      var modalBody = $("<label>").text("");
+
+      translations.forEach((translation) => {
+        if (
+          text === translation.text &&
+          typeof translation.delete_url !== "undefined"
+        ) {
+          var inputElement = $("<input>")
+            .attr("class", "form-check-input")
+            .attr("type", "checkbox")
+            .attr("value", "")
+            .attr("id", translation.delete_url)
+            .attr("checked", true);
+          var labelElement = $("<label>")
+            .attr("class", "form-check-label")
+            .attr("for", translation.delete_url)
+            .text(translation.origin);
+          var divElement = $("<div>")
+            .attr("class", "form-check")
+            .append(inputElement, labelElement);
+          modalBody.append(divElement);
+        }
+      });
+      return modalBody;
     }
 
     render(translations) {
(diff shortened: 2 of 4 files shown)


weblate/memory/machine.py
```diff
@@ -66,10 +68,12 @@ def download_translations(
...
                 continue
             yield {
+                "id": result.id,
                 "text": result.target,
                 "quality": quality,
                 "service": self.name,
                 "origin": result.get_origin_display(),
                 "source": result.source,
                 "show_quality": True,
+                "delete_url": reverse("api:memory-detail"),
```
nijel, line 78: Perhaps this way?

```suggestion
                "delete_url": reverse("api:memory-detail", kwargs={"pk": result.id}),
```

weblate/static/editor/full.js
```diff
@@ -210,6 +219,20 @@
     });
   };
 
+  FullEditor.prototype.updateMachinery = function (entry_id, delete_url) {
```
nijel, line 222: Calling removal "updateMachinery" is confusing.

weblate/static/editor/full.js
```diff
@@ -525,21 +548,44 @@
...
       /* Translators: Verb for copy operation */
-      row.append(
-        $(
-          "<td>" +
-            '<a class="js-copy-machinery btn btn-warning">' +
-            gettext("Copy") +
-            '<span class="mt-number text-info"></span>' +
-            "</a>" +
-            "</td>" +
+      if (service === "Weblate Translation Memory") {
```
nijel, line 551: It's better to look for presence of the delete URL.

weblate/static/editor/full.js
```diff
@@ -210,6 +217,19 @@
     });
   };
 
+  FullEditor.prototype.removeTranslationEntry = function (delete_url) {
+    $.ajax({
+      type: "DELETE",
+      url: delete_url,
+      success: () => {
+        addAlert("Successfully Removed!");
```
nijel, line 225: Any user displayed messages should be localizable:

```suggestion
        addAlert(getttext("Translation memory entry removed."));
```

weblate/static/editor/full.js
```diff
@@ -525,21 +545,44 @@
...
+              "</a>" +
+              "</td>" +
+              "<td>" +
+              '<a class="js-copy-save-machinery btn btn-primary">' +
+              gettext("Copy and save") +
+              "</a>" +
+              "</td>"
+          )
+        );
+      }
```
nijel, line 585: Please avoid code duplication here.

weblate/static/editor/full.js
```diff
@@ -540,6 +562,17 @@
...
         )
       );
+      if (el.delete_url) {
+        row.append(
+          $(
+            "<td>" +
+              '<a class="js-delete-machinery btn btn-danger">' +
+              gettext("Delete entry") +
+              "</a>" +
+              "</td>"
```
nijel, line 572: You're adding another table column here, doesn't it break formatting?

ParthS007: It does not apparently.
Screenshot 2022-08-12 at 13 26 44

Looks fine on my screen and I tried other big and a bit small screens. Please let me know if it looks good.

ParthS007:
@nijel One thing that we discussed earlier for multiple origin suggestions, Should I add multiple delete buttons for them or an Icon with delete functionality would be more intuitive.

In the later case, then instead of delete button, we can have delete icon just next to origin instead of a Delete entry button.

What do you think would be good to go?

nijel:
We definitely need a way to delete it in selected origins only. I can see two approaches:

- Move the delete action to the origin column, probably making it icon only. This will pollute the UI quite a lot.
- Add confirmation dialog which would list all origins with checkboxes. That way the UI is consistent, and it is possible to uncheck origins which the user doesn't want to delete.

That reminds me cases which is not addressed in this PR yet:

- The code currently picks only first delete URL, and thus it always deletes first origin only.
- In case the matching string is first obtained from other sources (machine translation service), the delete button will not be present, even if later the suggestions from the translation memory arrives.

The preferred solution as I see it now (feedback welcome):

1. Collect the delete URLs at origin element.
2. Once there is at least one delete URL, display the Delete button in the row.
3. Make the Delete button open a popup listing all origins for current translation which have the delete URL. All selected by default.
4. The callback would then fire DELETE on all selected URLs.

ParthS007:
The solution sounds good to me however it will result in more clicks required to delete an entry, imo it's not good for UX. Incase of the delete icon, the number of clicks remain same and user can just click on the delete icon of entries. 

I am not sure what do you mean here by:

> This will pollute the UI quite a lot.

I am okay with implementing the preferred solution though. Please point me to any similar popup implementations if we have any just in case.

wdyt?

nijel:
I think we want confirmation here anyway - having a delete button next to copy without a confirmation doesn't sound good. The confirmation popups are used in several places, for example here:

https://github.com/WeblateOrg/weblate/blob/c8f507972102558803f29abb0f1cfacd048c3a96/weblate/templates/trans/project-access-row.html#L44-L59

ParthS007:
@nijel I worked on the proposed solution. Apologies for the delay.

> The preferred solution as I see it now (feedback welcome):

- [x]   Collect the delete URLs at origin element.
- [x] Once there is at least one delete URL, display the Delete button in the row.
- [x] Make the Delete button open a popup listing all origins for current translation which have the delete URL. All selected by default.

The popup looks like this Screenshot 2022-09-09 at 23 12 34


- [ ] The callback would then fire DELETE on all selected URLs.

I am able to callback the DELETE request but the request gets blocked by firefox (I am not sure what;s the reason, might be related to `csrf` token or something else) and I get the following error notification

Screenshot 2022-09-09 at 23 12 13

Screenshot 2022-09-09 at 23 27 44

weblate/static/editor/full.js
```diff
@@ -36,6 +36,32 @@
...
+        var $form = $(e.currentTarget);
+        var $deleteEntries = $form.find("input.form-check-input:checked");
+        $deleteEntriesDialog.modal("hide");
+
+        Object.entries($deleteEntries).forEach(([_, entry]) => {
+          if (typeof entry.id !== "undefined") {
+            this.removeTranslationEntry(entry.id);
+          }
+        });
+      });
```
nijel, line 62: You need to supress further event processing here, otherwise the form is submitted and the AJAX request will get terminated:

```suggestion
      return false;
      });
```

weblate/static/editor/full.js
```diff
@@ -210,6 +236,20 @@
     });
   };
 
+  FullEditor.prototype.removeTranslationEntry = function (delete_url) {
+    $.ajax({
+      type: "DELETE",
+      url: delete_url,
+      headers: { "X-CSRFToken": this.csrfToken },
```
nijel, line 243: The token should not be needed for the API.

```suggestion
```

ParthS007: I removed it and getting 403. I think it;s needed!

weblate/static/editor/full.js
```diff
@@ -549,23 +607,65 @@
...
+    renderDeleteUrls(text) {
+      var translations = this.state.translations;
+      var modalBody = $("<label>").text("");
+
+      translations.forEach((translation) => {
+        if (translation.service === "Weblate Translation Memory") {
+          if (
+            text === translation.text &&
+            typeof translation.delete_url !== "undefined"
+          ) {
```
nijel, line 639: Is the condition for service name needed? Maybe it's better to filter by the delete_url only?

```suggestion
          if (
            text === translation.text &&
            typeof translation.delete_url !== "undefined"
          ) {
```

weblate/memory/machine.py
```diff
@@ -72,4 +74,5 @@ def download_translations(
                 "origin": result.get_origin_display(),
                 "source": result.source,
                 "show_quality": True,
+                "delete_url": reverse("api:memory-detail", kwargs={"pk": result.id}),
```
nijel, line 77: The delete URL should be included only in case user has a permission to delete. Otherwise the UI shows button which does not work. Check for `user.has_per("memory.delete", result)` is thus needed here.

weblate/static/editor/full.js
ParthS007: okay, I added the condition to be more explicit. Now, I am thinking only the `"Weblate Translation Memory"` will have delete_url so the later check would be enough. I will remove the first one. 👍

nijel:
I've tested that and I can see few issues:

1. The UI looks weird when the delete button is not shown:

   image

   I suggest adding empty cell instead.

2. The delete button is rendered for some services where it should not be present:

   image

   Clicking on it yields empty dialog

3. The dialog is not cleared before display, so when you abort it, and show it again, it includes duplicate entries:

   image

ParthS007:
Thanks for review, I could reproduce 1 and 3. 

1. The empty cell screenshot  if delete button is not present.

Screenshot 2022-09-18 at 15 26 42

2. I couldn't reproduce 2, I am assuming 2 might have happened because of inconsistent state so I have simplified the logic for showing the delete button. I tried on multiple pages and cannot see unexpected delete button.


3. The modal screenshot after closing it several times, no duplicate entries present

Screenshot 2022-09-18 at 15 26 58

nijel:
Thanks, looks better. There is still some issue with the dismissed popup - the dialog shows three checkboxes, but the DELETE request is fired multiple times (it seems that each dismiss adds one repetition):

Snímek z 2022-09-19 08-54-03

weblate/static/editor/full.js
```diff
@@ -36,6 +36,34 @@
...
+        var $deleteEntries = $form.find("input.form-check-input:checked");
+        $deleteEntriesDialog.modal("hide");
+
+        Object.entries($deleteEntries).forEach(([_, entry]) => {
+          if (typeof entry.id !== "undefined") {
+            this.removeTranslationEntry(entry.id);
+          }
+        });
+        return false;
+      });
```
nijel, line 64: I believe here is the issue - the events are bind when the dialog is shown, so in case it is shown multiple times (dismissing it between), the `submit` handler is also triggered multiple times causing duplicate `DELETE` requests.

ParthS007: after investigating a bit, I came to this solution.

When we dismiss the modal. the `.modal('hide')` returns to the method before the modal has actually been hidden. Here I assign the dialog as null so that when it reaches the submit handler, It doesn't do the `DELETE` request if the dialog is `null`.

It works fine now after multiple dismissals, only one `DELETE` request is sent. Please let me know wdyt?

ParthS007:
@nijel Please let me know if there is something left for this PR. Thanks

weblate/memory/machine.py
```diff
@@ -72,4 +74,7 @@ def download_translations(
                 "origin": result.get_origin_display(),
                 "source": result.source,
                 "show_quality": True,
+                "delete_url": reverse("api:memory-detail", kwargs={"pk": result.id})
+                if user.has_perm("memory.delete", result)
```
nijel, line 78: User can be None in some cases (for example when invoked via automatic translation add-on):

```suggestion
                if user is not None and user.has_perm("memory.delete", result)
```

See https://github.com/WeblateOrg/weblate/actions/runs/3083360977/jobs/4984199844#step:14:5298:

> weblate.machinery.base.MachineTranslationError: AttributeError: 'NoneType' object has no attribute 'has_perm'

ParthS007:
Tests passing 🎉

nijel:

PR #8014, opened 2022-08-17
trans: link filenames when sending parse error mails

### LINKED ISSUE #2106: Analysis issues do not link to content
This is the e-mail I got about QR Code Generator/Translations
"
'Expecting property name enclosed in double quotes: line 2 column 3 (char 4)'

Erroneous file

src/_locales/en/messages.json"

Would expect link to source and respective Weblate section here.

ParthS007:
## Proposed changes

closes #2106 

- Link Filename if they exist on parsing error mail.

## Checklist

### DIFF
--- a/weblate/templates/mail/parse_error.html
+++ b/weblate/templates/mail/parse_error.html
@@ -13,7 +13,11 @@
 <h2>{% trans "Failing file" %}</h2>
 
 <p>
-{{ details.filename }}
+{% if details.filelink %}
+    <a href="{{ details.filelink }}">{{ details.filename }}</a>
+{% else %}
+    {{ details.filename }}
+{% endif %}
 </p>
 
 {% endblock %}
--- a/weblate/accounts/notifications.py
+++ b/weblate/accounts/notifications.py
@@ -472,6 +472,16 @@ class ParseErrorNotification(Notification):
     verbose = _("Parse error")
     template_name = "parse_error"
 
+    def get_context(
+        self, change=None, subscription=None, extracontext=None, changes=None
+    ):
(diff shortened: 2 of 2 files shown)


weblate/trans/models/component.py
```diff
@@ -2055,7 +2059,10 @@ def _create_translations(  # noqa: C901
                         "skipping update due to error in parsing template: %s",
                         error.nested,
                     )
-                    self.handle_parse_error(error.nested, filename=self.template)
+                    filelink = self.get_repoweb_link(self.template, "1")
+                    self.handle_parse_error(
+                        error.nested, filename=self.template, filelink=filelink
+                    )
```
nijel, line 2065: I don't think we need to parse the filelink all way down as it can be generated when rendering the notification. Somethink like this should do it (in `weblate/accounts/notifications.py`):


```py
class ParseErrorNotification(Notification):
....
    def get_context(
        self, change=None, subscription=None, extracontext=None, changes=None
    ):
        context = super().get_context(change, subscription, extracontext, changes)
        if change:
            context["filelink"] = change.component.get_repoweb_link(change.details.filename, "1")
        return context
```

weblate/templates/mail/parse_error.html
```diff
@@ -13,7 +13,7 @@ <h2>{% trans "Error message" %}</h2>
 <h2>{% trans "Failing file" %}</h2>
 
 <p>
-{{ details.filename }}
+{{ details.filename }} {{ details.filelink }}
```
nijel, line 16: Better to make this actual link.

nijel:

PR #8030, opened 2022-08-20
reduce memory usage of damerau levenshtein

maxbachmann:
This reduces the memory usage from O(N*M) to O(N). In addition this implementation is faster as well.


nijel:

Issue #8066, opened 2022-08-29
Clone to translation needs a change in order to be saved


jvdmeij:
### Describe the issue

When using the 'clone to translation' button (1 on screenshot 1) the text is copied to the other language.

When not changing the text, the text is not saved. The counter of the translation is not updated (see screenshot 2). The text will saved (and counter will be updated) if you place a random character and remove it.

Screenshot 1:
![1](https://user-images.githubusercontent.com/7727264/187227436-91ab933c-4632-4a01-a3f9-b486a7207f7c.png)

Screenshot 2:
![CleanShot 2022-08-29 at 16 38 38](https://user-images.githubusercontent.com/7727264/187227814-ed550654-029c-4e4a-afee-f399fa792cfe.png)

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar issues in this repository.

### Steps to reproduce the behavior

1. Go to 'an untranslated text'
2. Use the 'clone to translation' button
3. See that the text is copied, but the counter is not updated
4. Go to the next text, go back and you will see that the text is not saved.

### Expected behavior

I expect a text that is not changed, will still be saved.

### How do you run Weblate?

Docker container

nijel:
It should be saved as needs editing without an edit. The not updated counter is indeed a minor issue

WofWca:
Doesn't look like saving in zen mode got fixed, see #8095.

PR #8095, opened 2022-09-08
fix(edit): text not updating when cloning

WofWca:
Revert 0943c4436e and add `bubbles`.
And that commit did not fully fix_ #8066,
it only made it update the counter.

## Checklist


nijel:
Thanks! I've applied it to other code path as well in f6c6670055.

PR #8156, opened 2022-09-26
trans: generate repo based url template links

### LINKED ISSUE #7865: Default (infer) "Repository browser" link template from "Source code repository"
### Describe the problem

Unless you explicitly specify "Repository browser" link template, repository browser links will never be generated (they'll remain as text). But we can infer them (at least for popular services like GitHub, GitLab, Bitbucket).

Either fill the "Repository browser" field automatically, or default here: https://github.com/WeblateOrg/weblate/blob/5bfcb775baf855ab89dc373a4901d20dfbe0fb1f/weblate/trans/models/component.py#L1155

_No response_

### Screenshots

image

### Additional context

Inspired by this comment https://github.com/WeblateOrg/weblate/issues/7738#issuecomment-1146638364

ParthS007:
## Proposed changes

- Default (infer) "Repository browser" link template from "Source code repository"
- closes #7865 

## Checklist

### DIFF
--- a/weblate/trans/models/component.py
+++ b/weblate/trans/models/component.py
@@ -154,6 +154,25 @@ MERGE_CHOICES = (
 
 LOCKING_ALERTS = {"MergeFailure", "UpdateFailure", "PushFailure", "ParseError"}
 
+BITBUCKET_GIT_REPOS_REGEXP = [
+    r"(?:ssh|https):\/\/(?:(?:git@|)bitbucket.org)\/([^/]*)\/([^/]*)",
+    r"git@bitbucket.org:([^/]*)\/([^/]*)",
+]
+
+GITHUB_REPOS_REGEXP = [
+    r"(?:git|https):\/\/(?:github.com)\/([^/]*)\/([^/]*)",
+    r"git@github.com:([^/]*)\/([^/]*)",
+]
+
+PAGURE_REPOS_REGEXP = [r"(?:ssh|https):\/\/(?:(?:git@|)pagure.io)\/([^/]*)\/([^/]*)"]
+
+AZURE_REPOS_REGEXP = [
+    r"(?:https):\/\/(?:dev.azure.com)\/([^/]*)\/([^/]*)\/(?:_git)\/([^/]*)",
+    r"(?:https):\/\/(?:([^/]*).visualstudio.com)\/([^/]*)\/(?:_git)\/([^/]*)",
+    r"(?:[^/]*)\@vs-ssh.visualstudio.com:v3\/([^/]*)\/([^/]*)\/([^/]*)",
+    r"(?:git@ssh.dev.azure.com:v3)\/([^/]*)\/([^/]*)\/([^/]*)",
+]
+
 
 def perform_on_link(func):
     """Decorator to handle repository link."""
@@ -1155,7 +1174,11 @@ class Component(models.Model, URLMixin, PathMixin, CacheKeyMixin):
         return self.git_export
 
     def get_repoweb_link(
-        self, filename: str, line: str, template: Optional[str] = None
+        self,
+        filename: str,
+        line: str,
+        template: Optional[str] = None,
+        user=None,
     ):
         """Generate link to source code browser for given file and line.
 
@@ -1163,9 +1186,18 @@ class Component(models.Model, URLMixin, PathMixin, CacheKeyMixin):
         here.
         """
         if not template:
-            template = self.repoweb
+            if self.repoweb:
+                template = self.repoweb
+            elif user and user.has_perm("vcs.view", self):
+                template = getattr(
+                    self,
+                    f"get_{self.vcs}_repoweb_template",
+                    self.get_git_repoweb_template,
+                )()
         if self.is_repo_link:
-            return self.linked_component.get_repoweb_link(filename, line, template)
+            return self.linked_component.get_repoweb_link(
+                filename, line, template, user=self.acting_user
+            )
         if not template:
             if filename.startswith("https://"):
                 return filename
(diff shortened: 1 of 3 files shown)


weblate/trans/models/component.py
```diff
@@ -1152,17 +1152,24 @@ def get_export_url(self):
...
         here.
         """
         if not template:
-            template = self.repoweb
+            if self.repoweb:
+                template = self.repoweb
+            elif user and user.has_perm("vcs.view", self):
+                template = "{}/blob/{}/{}/#{}".format(
+                    self.get_repo_url, self.get_repo_branch, filename, line
+                )
```
nijel, line 1168: This should to be handled in a separate method, parse the repo URL, do the detection and return template. To avoid issues with using template syntax in the filename/branch name, it should return the actual template, which will be filled in by filename/line later via `render_template` call, so use `{{branch}} {{filename}} {{line}}` instead of the actual values.

nijel: I just noticed your questions in #7865, I will provide more detailed answer there...

weblate/trans/models/component.py
```diff
@@ -1176,6 +1207,118 @@ def get_repoweb_link(
             component=self,
         )
 
+    def get_git_repoweb_template(self):
+        if any(regex.match(self.repo) for regex in BITBUCKET_GIT_REPOS_REGEXP):
+            return self.get_bitbucket_git_repoweb_template()
+
+        if any(regex.match(self.repo) for regex in BITBUCKET_HG_REPOS_REGEXP):
+            raise self.get_bitbucket_hg_repoweb_template
```
nijel, line 1215: BitBucket no longer supports Mercurial (see https://bitbucket.org/blog/sunsetting-mercurial-support-in-bitbucket), so this is not needed.

weblate/trans/models/component.py
```diff
@@ -1176,6 +1207,118 @@ def get_repoweb_link(
             component=self,
         )
 
+    def get_git_repoweb_template(self):
+        if any(regex.match(self.repo) for regex in BITBUCKET_GIT_REPOS_REGEXP):
```
nijel, line 1211: Aren't these expressions matching nearly anything? I'd rather do guess based on the domain name only.

weblate/trans/models/component.py
```diff
@@ -154,6 +155,23 @@
 
 LOCKING_ALERTS = {"MergeFailure", "UpdateFailure", "PushFailure", "ParseError"}
 
+BITBUCKET_GIT_REPOS_REGEXP = [r"([a-z]+):\/\/([^/]*)\/([^/]*)", r"git@([^/]*):([^/]*)"]
```
nijel, line 158: This seems to be wrong, sample URLs:

* `https://weblate@bitbucket.org/sasajovanovic/adbukalng.git`
* `git@bitbucket.org:sasajovanovic/adbukalng.git`

nijel (changes requested):
Can you please add tests for all the new methods? That way it is much easier to see whether it does what it should.

weblate/trans/models/component.py
ParthS007: okay, removed

ParthS007:
Thanks for the review, I asked a couple of things here: https://github.com/WeblateOrg/weblate/issues/7865#issuecomment-1270695465

> I added the url detection for services, I was not sure on how the URL will look like for services other than github, so kept them as to do for now. Please let me know if it is the right direction and how can I test the functionality. I went through the tests, i am assuming now I have to test something like this? https://github.com/WeblateOrg/weblate/blob/main/weblate/trans/tests/test_templatetags.py#L129-L149

weblate/trans/models/component.py
ParthS007: 1. Can you please add different example cases for bitbucket, pagure and azure to make the guess on domain only.

2. Something like this `git@github.com:([^/]*)\/([^/]*)` is expected here after domain only guess, or just like `github.com` in `str` instance is enough?

nijel: Maybe something like this:

```python
parsed_url = urlparse(self.repo)
if self.repo.startswith("git@github.com") or parsed_url.hostname == "github.com":
    ...
```

nijel:
The URL examples can be found in the webhooks tests:

https://github.com/WeblateOrg/weblate/blob/70ace9c9a1f36aad24648db28966c62f61953e1f/weblate/trans/tests/test_hooks.py#L1485-L1788

ParthS007:
I have added a test for verifying the functionality.  I am checking the return of `get_git_repoweb_template` directly.

I observed that when I change the `repo ` attribute, It tries to `sync_git_repo` and gives `RepositoryException` in case of repo url starting without `https`.

weblate/trans/models/component.py
```diff
@@ -1179,6 +1211,95 @@
...
+        ):
+            return self.get_bitbucket_git_repoweb_template()
+
+        if (
+            self.repo.startswith("git@github.com")
+            or parsed_url.hostname == "github.com"
+        ):
+            return self.get_github_repoweb_template()
+
+        if "pagure.io" in self.repo:
```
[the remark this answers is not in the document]
nijel: This is wrong, we really need to look at hostname only.

weblate/trans/models/component.py
```diff
@@ -1179,6 +1211,95 @@ def get_repoweb_link(
...
+        owner, slug, matches = None, None, None
+        if re.match(GITHUB_REPOS_REGEXP[0], self.repo):
+            matches = re.search(GITHUB_REPOS_REGEXP[0], self.repo)
+        elif re.match(GITHUB_REPOS_REGEXP[1], self.repo):
+            matches = re.search(GITHUB_REPOS_REGEXP[1], self.repo)
+        if matches:
+            owner = matches.group(1)
+            slug = matches.group(2)
+        if owner and slug:
+            return f"https://github.com/{owner}/{slug}"
```
nijel, line 1266: What we really need here is:

```suggestion
            return f"https://{domain}/{owner}/{slug}/blob/{{branch}}/{{filename}}#L{{line}}"
```

weblate/trans/models/component.py
```diff
@@ -1179,6 +1211,95 @@ def get_repoweb_link(
...
+
+    def get_pagure_repoweb_template(self):
+        owner, slug = None, None
+        if re.match(PAGURE_REPOS_REGEXP[0], self.repo):
+            matches = re.search(PAGURE_REPOS_REGEXP[0], self.repo)
+            owner = matches.group(1)
+            slug = matches.group(2)
+
+        if owner and slug:
+            return f"https://pagure.io/{owner}/{slug}"
```
nijel, line 1278: Example final URL is https://pagure.io/fedora-logos/blob/master/f/COPYING#_4

weblate/trans/models/component.py
```diff
@@ -1179,6 +1211,95 @@ def get_repoweb_link(
...
+        elif re.match(AZURE_REPOS_REGEXP[3], self.repo):
+            matches = re.search(AZURE_REPOS_REGEXP[3], self.repo)
+
+        if matches:
+            organization = matches.group(1)
+            project = matches.group(2)
+            repository = matches.group(3)
+
+        if organization and project and repository:
+            return f"https://dev.azure.com/{organization}/{project}/_git/{repository}"
```
nijel, line 1299: Again, this needs to be a template to generate the URL for displaying the file in the web browser, see above.

weblate/trans/models/component.py
ParthS007: Thanks, @nijel Can You please add final URL for bitucket and azure as well.

weblate/trans/tests/test_component.py
```diff
@@ -586,6 +586,69 @@ def test_unlink(self):
...
+        result = component.get_bitbucket_git_repoweb_template()
+        self.assertEqual(
+            result,
+            "https://bitbucket.org/marcus/project-x.git/blob/{branch}/{filename}#{line}",  # noqa
+        )
+        component.repo = "git@bitbucket.org:marcus/project-x.git"
+        result = component.get_bitbucket_git_repoweb_template()
+        self.assertEqual(
+            result,
+            "https://bitbucket.org/marcus/project-x.git/blob/{branch}/{filename}#{line}",  # noqa
```
nijel, line 602: I think `.git` should not be part of the resulting URL (for both tests).

weblate/trans/tests/test_component.py
```diff
@@ -586,6 +586,69 @@ def test_unlink(self):
...
+        result = component.get_github_repoweb_template()
+        self.assertEqual(
+            result,
+            "https://github.com/marcus/project-x.git/blob/{branch}/{filename}#L{line}",
+        )
+        component.repo = "git@github.com:marcus/project-x.git"
+        result = component.get_github_repoweb_template()
+        self.assertEqual(
+            result,
+            "https://github.com/marcus/project-x.git/blob/{branch}/{filename}#L{line}",
```
nijel, line 618: ...same here.

weblate/trans/models/component.py
```diff
@@ -1179,6 +1211,106 @@ def get_repoweb_link(
...
+            or parsed_url.hostname == "dev.azure.com"
+            or parsed_url.hostname.split(".", 1)[-1] == "visualstudio.com"
+        ):
+            return self.get_azure_repoweb_template()
+
+        return None
+
+    def get_clean_slug(self, slug):
+        if slug.endswith(".git"):
+            slug = slug.split(".git")[0]
```
nijel, line 1244: This will break when somebody names repo for example `some.gitorious.git`.

Better to just strip last four chars when we already know them:

```suggestion
            slug = slug[:-4]
```

Alternatively, the suffix can be stripped in the regexp while parsing the URL.

ParthS007: I think regexp stripping is safer than stripping. I will do that. Thanks

nijel:

PR #8183, opened 2022-10-03
Update

Justman10000:


Justman10000:
@nijel

nijel:
Sorry, but I don't think this makes the documentation any better:

- Switching absolute paths to relative ones makes it easier to break things.
- The paths are just an example anyway, at least you need to adjust Python version to the system.
- Moving `-y` to the end makes it harder to read IMHO. Typically, options are passed to programs before arguments.

Justman10000:
Since I seem to be wasting my time here too, you can regulate your shit yourself! No more PRs or any other support from me!

PR #8187, opened 2022-10-04
Fix `conflicts` parameter in API docs

jorrit:
The parameter was lacking a trailing `s`.


nijel:

PR #8209, opened 2022-10-12
Ensure post save is triggered upon executing bulk_create

superDross:
Intends to fix: https://github.com/WeblateOrg/weblate/issues/8211

The following change introduced in 4.11.0 release resulted in some messages not being emitted:
https://github.com/WeblateOrg/weblate/commit/38bc49fa732378148f7c220df7487385101c3e99

This is due to the fact that bulk_create does not call the save() method nor the post_save signals when Changes are parsed into it.

These post_save signals are required for fedora_notify_change to receive said signal and emit the appropriate message.

## Proposed changes

This fixes this issue by over-riding the bulk_create method in the ChangeQuerySet class to ensure the post_save signal is always sent after creation.

## Checklist

- [ x] I have added tests that prove my fix is effective or that my feature works.
- [ x] I have added documentation to describe my feature.
- [ x] I have squashed my commits into logic units.
- [ x] I have described the changes in the commit messages.

### DIFF
--- a/weblate/trans/models/change.py
+++ b/weblate/trans/models/change.py
@@ -22,6 +22,7 @@ from datetime import datetime
 from django.conf import settings
 from django.db import models, transaction
 from django.db.models import Count, Q
+from django.db.models.base import post_save
 from django.utils import timezone
 from django.utils.html import escape, format_html
 from django.utils.translation import gettext as _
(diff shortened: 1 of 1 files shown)


nijel:

PR #8229, opened 2022-10-16
Addition of pa_PK to basic language exceptions

### LINKED ISSUE #8223: Maintainers not able to add Punjabi translations
### Describe the issue

The Punjabi language is used in both Pakistan and India, but with different writing systems. The locales available for this in Weblate are `pa_PK` and `pa` respectively. (This is the same as how the Android locales are set up for this.) I have been translating applications for both scripts, which is easier to do together. However, `pa_PK` is not on the list of languages to choose from in the menu for adding a new translation, which says only a basic list of languages is included. For context, Punjabi is the 10th most spoken language in the world, and the majority of speakers live in Pakistan where it is the most widely apoken language. The language selection menu even includes ancient languages but does not have this option, it is odd that it is omitted entirely.

I have tried requesting adding it manually to maintainers, but here the maintainer of F-Droid suggests this may be a bug with Weblate since Android supports this locale but they were not able to add it on their hostes Weblate instance: https://gitlab.com/fdroid/admin/-/issues/342

In the meantime, I have been using the "und" Undetermined language code to store these translationa, but this puts a burden on project maintainers to manually export these and include them under the `pa_PK` locale when the Weblate application is supposed to be able to do this.

### I already tried

1. Click "+" to add "Punjabi (Pakistan)"
2. Error appears saying project maintainer needs to add support for this
3. Project maintainer is not able to do this

### Expected behavior

Ability to translate projects on Weblate to Punjabi.

### Screenshots

Screenshots in the linked Gitlab issue re: F-Droid

### Exception traceback

_No response_

### How do you run Weblate?

weblate.org service

### Weblate versions

_No response_

### Weblate deploy checks

_No response_

### Additional context

_No response_

bgo-eiu:
## Proposed changes
This is intended to fix #8223 by adding `pa_PK` to the list of "underscore" codes to include in the basic language list. Like `pt_BR`, the majority of Punjabi speakers live in Pakistan, and like the situations with Chinese, Belarusian, and Norwegian, there are differences in the way the languages are written which mean `pa` and `pa_PK` cannot be readily substituted for eachother.

## Checklist

### DIFF
--- a/weblate/lang/data.py
+++ b/weblate/lang/data.py
@@ -24,7 +24,15 @@ from weblate_language_data.ambiguous import AMBIGUOUS
 
 NO_CODE_LANGUAGES = {lang[0] for lang in languages.LANGUAGES}
 
-UNDERSCORE_EXCEPTIONS = {"nb_NO", "zh_Hant", "zh_Hans", "be_Latn", "ro_MD", "pt_BR", "pa_PK"}
+UNDERSCORE_EXCEPTIONS = {
+    "nb_NO",
+    "zh_Hant",
+    "zh_Hans",
+    "be_Latn",
+    "ro_MD",
+    "pt_BR",
+    "pa_PK",
+}
 AT_EXCEPTIONS = {"ca@valencia"}
 
 
--- a/weblate/lang/data.py
+++ b/weblate/lang/data.py
@@ -24,7 +24,7 @@ from weblate_language_data.ambiguous import AMBIGUOUS
 
 NO_CODE_LANGUAGES = {lang[0] for lang in languages.LANGUAGES}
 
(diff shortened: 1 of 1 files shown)


nijel:
Are you confident this fixes #8223? `UNDERSCORE_EXCEPTIONS` should be in use only for non-maintainers.

nijel:
I've merged it as I think it makes sense to offer `pa_PK` to regular users as well. Maybe I just misunderstood the description in #8223 that you want that for the maintainers...

bgo-eiu:
I am not a maintainer of a project - I am not sure what the interface looks like from the maintainers' point of view so I am not sure what the issue could be there. If this adds it as an option for regular users, then asking maintainers to add it is not required anyway so that is simpler.

Issue #6508, opened 2021-09-02
Allow personal email addresses to be ommitted in commits


pappasadrian:
**Is your feature request related to a problem? If so, please describe.**

Email addresses used by translators are included in the commit files, with no option to opt out/turn off. This leads to leaks of personal information publically, by including addresses in public code. This is not respectful towards translators who contribute, and potentially illegal in some jurisdictions (Especially considering that there's no specific consent given. Everyone is opted-in by default, with no option to opt out). 

**Describe the solution you'd like**

Add an option allowing translators's email to not be included in commits.

**Describe alternatives you've considered**
Using a fake email is not an option, as a valid email is required for other features of weblate. 
Using an throwaway/single use email is also not very viable, as it is very cumbersome for users, and not a clear option for people who don't want their personal information leaked. 

**Screenshots**

**Additional context**
 
Relevant issue: https://github.com/WeblateOrg/weblate/issues/4988

bjorn3:
Github gives everyone a email address like `bjorn3@users.noreply.github.com` to be used as commit author. Alternatively giving the option to use one of the secondary email addresses associated with a github account rather than the primary one would be nice. For example zulip asks you which of the email addresses associated with your github account to use. Personally I have my private email address set as primary email on my github account, but a public email address as secondary email.

nijel:
The final solution should look like:

* ~Prefer public address from GitHub~, done in 90e101139ac8c89c7c931e0d7eb3edf1560c701a
* Fetch all verified addresses from GitHub and store them as `VerifiedEmail`
* Add separate selection for email address to use in Git commits

phw:
@nijel Thanks for getting in. Could this also be done independent of a users GitHub profile? Not every users has one. I imagine something similar to how Github does it:

- First have a separate "Commiter e-mail" field the user can set, so the e-mail being used for commits is separate from the main communication e-mail. The user can fill in any value they want to have a different e-mail for committing. If they omit this the primary e-mail address will be used.
- Have an option "Keep my e-mail address private". Activating this will automatically use some e-mail address in the form of "username@weblate_host", let's say for example "phw@hosted.weblate.org"

This is how it looks like in Github:

![grafik](https://user-images.githubusercontent.com/29852/172811374-7257f458-3185-42af-a403-e9edf46876ef.png)

[...]

PR #8268, opened 2022-10-25
Feature/hide mail

nijel:
## Proposed changes

Discussed in https://github.com/WeblateOrg/weblate/issues/6508, opening so that we can easily discuss the code.

## Checklist

### DIFF
--- a/weblate/accounts/pipeline.py
+++ b/weblate/accounts/pipeline.py
@@ -390,11 +390,11 @@ def store_email(strategy, backend, user, social, details, **kwargs):
             social.verifiedemail_set.create(email=add[0], is_deliverable=add[1])
     elif details.get("email"):
         verified, created = VerifiedEmail.objects.get_or_create(
-            social=social, defaults={"email": details["email"], "is_deliverable": details["is_deliverable"]}
+            social=social, defaults={"email": details["email"]}
         )
-        if not created and verified.email != details["email"] and verified.is_deliverable != details["is_deliverable"]:
+        if not created and verified.email != details["email"] or not verified.is_deliverable:
             verified.email = details["email"]
-            verified.is_deliverable = details["is_deliverable"]
+            verified.is_deliverable = True
             verified.save()
 
 


weblate/accounts/pipeline.py
```diff
@@ -376,17 +380,18 @@ def store_email(strategy, backend, user, social, details, **kwargs):
...
+            social.verifiedemail_set.filter(email=remove[0]).delete()
         for add in current - existing:
-            social.verifiedemail_set.create(email=add)
+            social.verifiedemail_set.create(email=add[0], is_deliverable=add[1])
     elif details.get("email"):
         verified, created = VerifiedEmail.objects.get_or_create(
-            social=social, defaults={"email": details["email"]}
+            social=social, defaults={"email": details["email"], "is_deliverable": details["is_deliverable"]}
         )
         if not created and verified.email != details["email"]:
```
nijel, line 392: This should check for `is_deliverable` being different as well.

weblate/utils/fields.py
```diff
@@ -110,3 +110,7 @@ def formfield(self, **kwargs):
                 **kwargs,
             }
         )
+
+class CommitEmailField(EmailField):
+    description = _("Commit e-mail")
+    pass
```
nijel, line 116: This class seems not useful, you override description in the `User` class anyway.

weblate/auth/models.py
```diff
@@ -639,7 +645,7 @@ def get_visible_name(self) -> str:
 
     def get_author_name(self) -> str:
         """Return formatted author name with e-mail."""
-        return f"{self.get_visible_name()} <{self.email}>"
+        return f"{self.get_visible_name()} <{self.commit_email}>"
```
nijel, line 648: Needs a fallback to `email` in case `commit_email` is not set.

weblate/auth/models.py
```diff
@@ -361,7 +367,7 @@ class User(AbstractBaseUser):
 
     EMAIL_FIELD = "email"
     USERNAME_FIELD = "username"
-    REQUIRED_FIELDS = ["email", "full_name"]
+    REQUIRED_FIELDS = ["email", "full_name", "commit_email"]
```
nijel, line 370: Is it really required? It can be blank according to the model.

ann0see: If there's no fallback, it is.

ann0see:
> Add UI to select e-mail from verified ones, this can be based on existing one

I think I need help with that. As I'm not familiar with the code base and python, could you please point me out where the GUI is to be changed? I get a KeyError at /accounts/profile with the exception value commit_email I assume I miss something somewhere (maybe I don't send commit_email in a form).

weblate/api/serializers.py
```diff
@@ -206,6 +206,7 @@ class Meta:
         model = User
         fields = (
             "email",
+            "commit_email",
```
ann0see, line 209: Not sure if that's needed.

weblate/accounts/forms.py
```diff
@@ -322,18 +334,27 @@ class UserForm(forms.ModelForm):
...
+        choices=(("", ""),),
+        required=True,
+    )
+
     full_name = FullNameField()
 
     class Meta:
         model = User
-        fields = ("username", "full_name", "email")
+        fields = ("username", "full_name", "email", "commit_email")
```
nijel, line 348:
```suggestion
        fields = ("username", "full_name", "email")
```

weblate/accounts/forms.py
```diff
@@ -322,18 +334,27 @@ class UserForm(forms.ModelForm):
         choices=(("", ""),),
         required=True,
     )
+    commit_email = forms.ChoiceField(
+        label=_("Commit E-mail"),
+        help_text=_("You can add another commit e-mail address below."),
+        choices=(("", ""),),
+        required=True,
+    )
+
```
nijel, line 343:
```suggestion
```

weblate/accounts/forms.py
```diff
@@ -322,18 +334,27 @@ class UserForm(forms.ModelForm):
...
     class Meta:
         model = User
-        fields = ("username", "full_name", "email")
+        fields = ("username", "full_name", "email", "commit_email")
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
         emails = get_all_user_mails(self.instance)
+        commit_emails = get_all_user_mails(self.instance, filter_deliverable=False)
```
nijel, line 354:
```suggestion
```

weblate/accounts/forms.py
```diff
@@ -322,18 +334,27 @@ class UserForm(forms.ModelForm):
...
 
         self.fields["email"].choices = [(x, x) for x in sorted(emails)]
+        self.fields["commit_email"].choices = [(x, x) for x in sorted(commit_emails)]
```
nijel, line 357:
```suggestion
```

weblate/api/serializers.py
nijel:
```suggestion
```

weblate/auth/models.py
```diff
@@ -328,6 +328,12 @@ class User(AbstractBaseUser):
         null=True,
         unique=True,
     )
+    commit_email = EmailField(
+        _("Commit E-mail"),
+        blank=False,
+        null=True,
+        unique=True,
+    )
```
nijel, line 336:
```suggestion
```

weblate/auth/models.py
```diff
@@ -639,7 +645,7 @@ def get_visible_name(self) -> str:
 
     def get_author_name(self) -> str:
```
nijel, line 646:
```suggestion
    def get_commit_email(self) -> str:
        return self.profile.commit_email or self.email
    
    def get_author_name(self) -> str:
```

weblate/auth/models.py
```diff
@@ -639,7 +645,7 @@ def get_visible_name(self) -> str:
...
         """Return formatted author name with e-mail."""
-        return f"{self.get_visible_name()} <{self.email}>"
+        return f"{self.get_visible_name()} <{self.commit_email if self.commit_email else self.email}>"
```
nijel, line 648:
```suggestion
        return f"{self.get_visible_name()} <{self.get_commit_email()}>"
```

nijel (commented):
You've added `commit_email` to both `Profile` and `User` models, what would lead to data duplication and caused issues in the form. I've proposed changes to place the field only in the `Profile` model (to be consistent with public e-mail field).

ann0see:
Unfortunately, I can't test on my machine at the moment due to the libffi.h issue: 
```
SystemError: ffi_prep_closure(): bad user_data (it seems that the version of the libffi library seen at runtime is different from the 'ffi.h' file seen at compile-time)
```

I think probably in the near future, libffi will be updated. pip install --force-reinstall --no-binary :all: cffi doesn't solve the issue.

weblate/accounts/pipeline.py
```diff
@@ -379,17 +383,18 @@ def store_email(strategy, backend, user, social, details, **kwargs):
     # The email can be empty for some services
     if details.get("verified_emails"):
         current = set(details["verified_emails"])
-        existing = set(social.verifiedemail_set.values_list("email", flat=True))
+        existing = set(social.verifiedemail_set.values_list("email", "is_deliverable"))
         for remove in existing - current:
-            social.verifiedemail_set.filter(email=remove).delete()
+            social.verifiedemail_set.filter(email=remove[0]).delete()
```
nijel, line 388: The filter should use both params (to avoid issue when e-mail deliverability changes):

```suggestion
            social.verifiedemail_set.filter(email=remove[0], is_deliverable=remove[1]).delete()
```

weblate/accounts/pipeline.py
```diff
@@ -379,17 +383,18 @@ def store_email(strategy, backend, user, social, details, **kwargs):
...
         for add in current - existing:
-            social.verifiedemail_set.create(email=add)
+            social.verifiedemail_set.create(email=add[0], is_deliverable=add[1])
     elif details.get("email"):
         verified, created = VerifiedEmail.objects.get_or_create(
-            social=social, defaults={"email": details["email"]}
+            social=social, defaults={"email": details["email"], "is_deliverable": details["is_deliverable"]}
```
nijel, line 393: The `is_deliverable` is never set in details so far (1/3):

```suggestion
            social=social, defaults={"email": details["email"]}
```

weblate/accounts/pipeline.py
```diff
@@ -379,17 +383,18 @@ def store_email(strategy, backend, user, social, details, **kwargs):
...
         )
-        if not created and verified.email != details["email"]:
+        if not created and verified.email != details["email"] and verified.is_deliverable != details["is_deliverable"]:
```
nijel, line 395: The `is_deliverable` is never set in details so far (2/3):

```suggestion
        if not created and verified.email != details["email"] or not verified.is_deliverable:
```

weblate/accounts/pipeline.py
```diff
@@ -379,17 +383,18 @@ def store_email(strategy, backend, user, social, details, **kwargs):
...
             verified.email = details["email"]
+            verified.is_deliverable = details["is_deliverable"]
```
nijel, line 397: The `is_deliverable` is never set in details so far (3/3):

```suggestion
            verified.is_deliverable = True
```

PR #8279, opened 2022-10-28
Ensure the "Replaced file by upload" fedora message is sent

superDross:
In handle_store_change we explicitly parse the change no. for the aforementioned message when the translation is not source. However, we never parse this when it is source.

This results in the aforementioned message never being sent when replacing source string via an upload.

Fixes: https://github.com/WeblateOrg/weblate/issues/8285

## Proposed changes

To resolve this by ensuring we can parse this via the create_translation method in the component which will then eventually parse it to check_sync downstream thereby ensuring the message is pushed to the exchange.

## Checklist


nijel:

PR #8336, opened 2022-11-11
Document component scoped changes API

jorrit:
## Proposed changes

I noticed the component scoped changes API was not documented. I copied the section from the project scoped changes API.


nijel:
It is documented since https://github.com/WeblateOrg/weblate/commit/77f9d70c6fda45f0f42b0f16a7b37e9a07653847:  https://docs.weblate.org/en/latest/api.html#get--api-components-(string-project)-(string-component)-changes-

PR #8352, opened 2022-11-15
feat: add xml add-on

warlof:
## Proposed changes

I'm using Weblate to maintain XLIFF translation files from my Business Central packages.
Compiler is using non self-closing tags format when no data is stored in a node - and as a result, huge diff are currently generated after a unit is translated from Weblate.

The purpose of this changes is to provide an optional way to handle the output formatting of XLIFF files.

It relays on the following pull request which add a flag to the XLIFF storage https://github.com/translate/translate/pull/4751

## Checklist

### DIFF
--- a/weblate/addons/forms.py
+++ b/weblate/addons/forms.py
@@ -204,14 +204,10 @@ class JSONCustomizeForm(BaseAddonForm):
 class XMLCustomizeForm(BaseAddonForm):
     """Class defining user Form to configure XML Formatting AddOn"""
 
-    tags_format = forms.ChoiceField(
-        label=_("XML closing tags format"),
-        choices=[
-            ("self_closing_tags", _("Self-Closing Tags")),
-            ("closing_tags", _("Closing Tags")),
-        ],
-        required=True,
-        initial="self_closing_tags",
+    closing_tags = forms.BooleanField(
+        label=_("Include closing tag for blank XML tags"),
+        required=False,
+        initial=True,
     )
 
 
--- a/weblate/addons/forms.py
+++ b/weblate/addons/forms.py
@@ -202,6 +202,8 @@ class JSONCustomizeForm(BaseAddonForm):
 
 
 class XMLCustomizeForm(BaseAddonForm):
+    """Class defining user Form to configure XML Formatting AddOn"""
+
     tags_format = forms.ChoiceField(
         label=_("XML closing tags format"),
         choices=[
--- a/weblate/addons/forms.py
+++ b/weblate/addons/forms.py
@@ -201,6 +201,18 @@ class JSONCustomizeForm(BaseAddonForm):
     )
 
 
+class XMLCustomizeForm(BaseAddonForm):
+    tags_format = forms.ChoiceField(
+        label=_("XML closing tags format"),
+        choices=[
+            ("self_closing_tags", _("Self-Closing Tags")),
+            ("closing_tags", _("Closing Tags")),
+        ],
+        required=True,
+        initial="self_closing_tags",
+    )
+
+
 class YAMLCustomizeForm(BaseAddonForm):
     indent = forms.IntegerField(
         label=_("YAML indentation"), min_value=1, max_value=10, initial=2, required=True
--- a/weblate/addons/xml.py
+++ b/weblate/addons/xml.py
@@ -31,7 +31,7 @@ class XMLCustomizeAddon(StoreBaseAddon):
     name = "weblate.xml.customize"
     verbose = _("Customize XML output")
     description = _(
-        "Allows adjusting XML output behavior, for example indentation or sorting."
+        "Allows adjusting XML output behavior, for example closing tags instead of self-closing tags for empty tags."
     )
     settings_form = XMLCustomizeForm
 
@@ -45,6 +45,4 @@ class XMLCustomizeAddon(StoreBaseAddon):
     def store_post_load(self, translation, store):
         """Hook triggered once component formatter has been loaded."""
         config = self.instance.configuration
-        store.store.XMLSelfClosingTags = (
-            config.get("tags_format", "closing_tags") == "self_closing_tags"
-        )
+        store.store.XMLSelfClosingTags = not config.get("closing_tags", True)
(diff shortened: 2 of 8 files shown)


weblate/addons/xliff.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class XLIFFCustomizeAddon(StoreBaseAddon):
+    name = "weblate.xliff.customize"
+    verbose = _("Customize XLIFF output")
+    description = _(
+        "Allows adjusting XLIFF output behavior, for example indentation or sorting."
+    )
+    settings_form = XLIFFCustomizeForm
+    multiple = True
```
nijel, line 34: There is no reason to install such add-ons multiple times.

```suggestion
```

weblate/addons/xliff.py
```diff
@@ -0,0 +1,41 @@
...
+    icon = "language.svg"
+
+    def store_post_load(self, translation, store):
+        config = self.instance.configuration
+        tags_format = config.get("tags_format", "closing_tags")
+        if tags == "closing_tags":
+            store.XMLSelfClosingTags = False
```
nijel, line 41: It should always set the attribute:

```suggestion
        store.XMLSelfClosingTags = config.get("tags_format", "closing_tags") == "self_closing_tags"
```

weblate/addons/xliff.py
```diff
@@ -0,0 +1,41 @@
...
+
+from django.utils.translation import gettext_lazy as _
+
+from weblate.addons.base import StoreBaseAddon
+from weblate.addons.forms import XLIFFCustomizeForm
+
+
+class XLIFFCustomizeAddon(StoreBaseAddon):
+    name = "weblate.xliff.customize"
+    verbose = _("Customize XLIFF output")
```
nijel, line 29: Maybe make it XML generic? With https://github.com/translate/translate/pull/4751, the feature is added to all Lisa subclasses.

nijel (changes requested):
Besides other comments, the add-on is missing compatibility check. Either add `compat` attribute with supported file formats, or implement `can_install` method to check whether component file format is subclass of Lisa.

weblate/addons/xliff.py
warlof: Make sense 👍

warlof:
@nijel I made some change localy, especially regarding the can_install override in order to check if yes or no we can use the formatter addon.

However, I'm still stuck for testing since when I go on the component addons page, I'm getting an exception stating the module does not exist.
I can't figure what I did wrong :(

```
ImproperlyConfigured at /addons/business-central/email-current-user-connector/
Error importing module weblate.addons.xml in WEBLATE_ADDONS: "No module named 'weblate.addons.xml'"
```

Also, in the `can_install` definition - does `cls.instance` refer to the Storage class (aka LISAfile) ?

warlof:
hey, I'm back and it's working :')

it took me a while, but I figure what was going wrong.
currently, the quick dev using docker is binding volume against python 3.10 - however, my container were boot using python 3.11
I don't know if it's normal or not - but at least, once volume is bound to proper target path, it's going better :D
https://github.com/WeblateOrg/weblate/blob/c1874814d7ea30020fff92bb11a47af8f288765c/dev-docker/docker-compose.yml#L30

let's back to our main topic :

I had override the `can_install` method in order to determine dynamically if yes or no the formatter addon can be installed.
Personal notice : component -> format -> store

I also had fix an issue within the `store_post_load` method which is taking store as parameter but it appears to be the FileFormat instead.

And now, I can confirm it's working as expected 👍

weblate/formats/ttkit.py
```diff
@@ -506,7 +506,7 @@ def all_store_units(self):
         )
 
     def delete_unit(self, ttkit_unit) -> Optional[str]:
-        self.store.removeunit(ttkit_unit)
+        return self.store.removeunit(ttkit_unit)
```
nijel, line 509: This is wrong, this method intentionally returns None. The problem with that is that adding `return None` is pointless :-). Anyway this is not related to topic of this PR.

weblate/addons/forms.py
```diff
@@ -201,6 +201,20 @@ class JSONCustomizeForm(BaseAddonForm):
...
+
+    tags_format = forms.ChoiceField(
+        label=_("XML closing tags format"),
+        choices=[
+            ("self_closing_tags", _("Self-Closing Tags")),
+            ("closing_tags", _("Closing Tags")),
+        ],
+        required=True,
+        initial="self_closing_tags",
+    )
```
nijel, line 215: This needs rephrasing so that it is easier to understand. Also, making it checkbox might be nicer.

How about this? (It would need changes in the add-on as well)

```suggestion
    closing_tags = forms.Boolean(
        label=_("Include closing tag for blank XML tags"),
        required=False,
        initial=True,
    )
```

weblate/addons/xml.py
```diff
@@ -0,0 +1,50 @@
...
+        # component are attached to a file format which is defined by a loader
+        # we want to provide this package only for component using LISAfile as loader
+        return issubclass(component.file_format_cls.loader, LISAfile)
+
+    def store_post_load(self, translation, store):
+        """Hook triggered once component formatter has been loaded."""
+        config = self.instance.configuration
+        store.store.XMLSelfClosingTags = (
+            config.get("tags_format", "closing_tags") == "self_closing_tags"
+        )
```
nijel, line 50: Change to work with `BooleanField`:

```suggestion
        store.store.XMLSelfClosingTags = config.get("closing_tags", True)
```

nijel (commented):
This needs tests as well. Once https://github.com/translate/translate/pull/4751 is merged and released, we can adjust the dependency to make sure the functionality is supported in the library.

weblate/addons/forms.py
warlof: Definitely

nijel:
Weblate now requires translate-toolkit 4.8.0 with needed changes. Can you please look into adding a testcase for this?

warlof:
> Weblate now requires translate-toolkit 4.8.0 with needed changes. Can you please look into adding a testcase for this?

I will have to figure why I cannot run existing ones first.

What to be test ?
- if unchecked => self-closing tags
- if checked => non self-closing tags

warlof:
I also had realign the branch against master `HEAD` :)

warlof:
Hi there,

Is there hope it's being shipped with next 4.15 upcoming version ? That would be a great Xmas present 😇

nijel:

nijel:
I've updated the docs in f940c51b02

PR #8445, opened 2022-12-07
accounts: Allow to use different e-mail for notifications and commits

### LINKED ISSUE #6508: Allow personal email addresses to be ommitted in commits

nijel:
## Proposed changes

* Add commit e-mail to the profile
* Import noreply e-mail from GitHub
* Store non-deliverable e-mails as well in the VerifiedEmails

Fixes #8268
Fixes #6508

This is squashed and polished #8268 by @ann0see

## Checklist


ann0see:
That's great! Thanks. You can close the other PR. 

Sorry for being unresponsive for quite some time – I'm not into python and quite busy. Good to see you took up the work.

p0358:
As I understand this means the commit email field is empty by default unless populated for example by GitHub login automatically. Would there be some way to setup such email automatically for all new users based on some template (similar to how github does/did `<username>@users.noreply.github.com`), regardless of their registration method? This way weblate instance owner could ensure no e-mails would get leaked by default. Seems like it should be some simple change over this changeset.

(Btw. still cool that now any user on any instance will be able to set different commit email to avoid leaking their real one, that's a very welcome change!)

nijel:
> Would there be some way to setup such email automatically for all new users based on some template (similar to how github does/did <username>@users.noreply.github.com), regardless of their registration method? 

Not right now. The problem is that it needs some setup outside Weblate - it needs at least DNS entry and a discarding mail server. You are welcome to open a separate issue for that. Also, on GitHub this is AFAIK opt-in.

PR #8488, opened 2022-12-15
improvement: don't stretch screenshot modal if image is narrow

WofWca:
## Proposed changes

| Before | After |
|--------|-------|
| image | image |

## Checklist

## Other information

I also tested it with a narrower screen - it doesn't overflow it (although touches the edge, as it is now on wide screens)


nijel:

PR #8518, opened 2022-12-22
[FIX] do not trim the target string when updateing a translation

tsabi:
## Proposed changes

While updating a Unit via REST API, the translated text (target field) was trimmed. During debugging, I found out that in django's rest_framework module there is a parameter called `trim_whitespace` that is enabled by default, and the value of the submitted target field is automatically trimmed.

I am not sure that i fixed this problem properly, but this quick patch fixed this exact problem.

I used this code to submit the update:

```python
HEADERS = {"Authorization": "Token wlu_yvCobi.........LZ3E4vlQtuRBMR2"}
r = requests.patch("https://weblate.xyz.tld/api/units/199129/", headers=HEADERS,
        json={"target": "\n    - OK", "state": 20},
    )
```

In the [`api/views.py#1253`](https://github.com/WeblateOrg/weblate/blob/main/weblate/api/views.py#L1253) line the `data`:

Before this patch:
```python
data = OrderedDict([('target', ['- OK']), ('state', 20)])
```

After this patch:
```python
data = OrderedDict([('target', ['\n    - OK']), ('state', 20)])
```

## Checklist

### DIFF
--- a/weblate/api/serializers.py
+++ b/weblate/api/serializers.py
@@ -911,7 +911,7 @@ class UserStatisticsSerializer(ReadOnlySerializer):
 
 
 class PluralField(serializers.ListField):
-    child = serializers.CharField()
+    child = serializers.CharField(trim_whitespace=False)
 
     def get_attribute(self, instance):
         return getattr(instance, f"get_{self.field_name}_plurals")()


nijel:
Thanks. Would you be able to add test case for this?

tsabi:
I am sorry, but i don't know too much of the inner structure of this project to be able to write a good test.

nijel:
Thanks for your patch, I've merged it with a testcase as b68e73a3975813b59d9c7b4c28c26066207ec9c8

Issue #8119, opened 2022-09-15
Reduce metrics disk usage


nijel:
### Describe the problem

Currently, the metrics and related indexes account to the most of the database usage. That is not really reasonable for a feature that displays trends and summaries.

### Describe the solution you'd like

The database usage should be reduced. The way to go might be to reduce the number of the rows by storing all data for a single day in a JSONField instead of having it in ~20 rows. This will also heavily reduce the index size as it will have one less column to index.

rob006:
> The way to go might be to reduce the number of the rows by storing all data for a single day in a JSONField instead of having it in ~20 rows. This will also heavily reduce the index size as it will have one less column to index.

Wouldn't that increase data size? Storing number as integer takes less space than storing it in JSON as string.

Alternative solutions:
1. Do not store stats with `0` as value (it over 70% metrics in [my case](https://github.com/WeblateOrg/weblate/issues/6611#issuecomment-1371400216)).
2. Store only monthly stats for metrics older than 60 days (AFAIK they're not used individually).

nijel:
> Wouldn't that increase data size? Storing number as integer takes less space than storing it in JSON as string.

That's true, but there are additional fields which are now duplicated because there can be nearly 30 values stored for each object. So, the choice really is between duplicating date and three int columns 30 times, or having it once and having JSON overhead. The change will definitely decrease index size, as there will be 1/30 of rows and one less column to index.

Still, this needs proper benchmarking before making the change, that's why this is still being postponed ;-).

> Do not store stats with 0 as value

We need to differentiate the case between no metrics calculated and 0 calculated. So even the 0 needs to be stored; otherwise it would have to be calculated again.

> Store only monthly stats for metrics older than 60 days

That would be an optimization as well. We really require daily metrics after 60 days only to calculate monthly ones. Presently, monthly ones are stored in the cache only, that's why we have to keep daily ones around as they might disappear from the cache.

> This table is still massive in my case (96% of database size) (in https://github.com/WeblateOrg/weblate/issues/6611#issuecomment-1371400216)

Can you please share table and index sizes? For my test environment, the table data has 16GB, and the two indexes 6GB + 18GB.

rob006:
> Can you please share table and index sizes? For my test environment, the table data has 16GB, and the two indexes 6GB + 18GB.

It is 5.3 GB for data and 4.2 GB for indexes. It is MySQL.

[...]

PR #8563, opened 2023-01-05
Rewrite metrics storage

nijel:
## Proposed changes

Rewritten metrics storage for reduced disk usage and improved performance.

## Checklist
- [x] https://github.com/WeblateOrg/weblate/issues/8119
- [x] https://github.com/WeblateOrg/weblate/issues/8825


weblate/metrics/migrations/0015_migrate_metrics.py
```diff
@@ -0,0 +1,135 @@
...
+METRIC_USERS = 20
+METRIC_CONTRIBUTORS = 21
+METRIC_PROJECTS = 22
+METRIC_SCREENSHOTS = 23
+METRIC_COMPONENTS = 24
+METRIC_TRANSLATIONS = 25
+METRIC_MACHINERY_INTERNAL = 26
+METRIC_MACHINERY_EXTERNAL = 27
+
+METRIC_IDS = {
```
rob006, line 33: How about using integer IDs as keys instead for string? `"27":0` will take less storage than `"machinery:internal":0`. And if each set have all metrics, keys could be skipped entirely while encoding to JSON: `["v1",0,0,0,...]` (0-index could store version of stats structure, so `METRIC_ALL` will still have `1` as index) instead of `{"all":0,"all_words":0,"translated":0,...}` - in many cases this should be more efficient than storing metrics as integers in separate columns (`0,` string takes 2 bytes vs 4 bytes for integer).

nijel: I've tried more approaches, and storing in a list is the most likely outcome. We can probably assume the list will be only extending, so storing version should not be needed at all, the list length will tell that.

See TODO at https://github.com/WeblateOrg/weblate/pull/8563/files#diff-30660b786ec366973fcef283c4f9531b57361cfc72a774d73b443fdfa25eecc9R119

This is really just proof of concept, the final patch will be worked on once 4.15.1 is released.

rob006: > We can probably assume the list will be only extending, so storing version should not be needed at all, the list length will tell that.

The main intention was to use shorter representation of list in JSON, since it will be indexed from 0: `["v1",0,0,0,...]` instead of `["1":0,"2":0,"3":0,...]`. Puting anything as first key will make JSON shorter without the need to change metrics kind IDs.

nijel: After few more iterations, I think I have a final solution now:

- the changes metric will be stored in a separate field
- others will be stored as a JSON list

This allows wiping others once we no longer needed them with a simple UPDATE SQL.

orangesunny (approved):

rob006:
Maybe this is MySQL specific, but migration `0015_migrate_metrics` uses a loooot of memory. It looks like it loads all rows from metrics table to memory, and then process this collection. For large tables this will most likely fail, because there is not enough free RAM on server. In my case I had to remove all metrics with zero as value and limit range to last 80 days.

nijel:
It uses iterator to load existing to avoid storing them in the memory. For creating bulk mode is used, so at most 1000 entries will be in memory.

Do you have debug disabled? With that enabled Django might keep too much in memory.

rob006:
I have debug mode disabled, But this looks like a limitation of MySQL driver, which does not support buffered queries and loads everything into memory:

> Without server-side cursors
> 
> MySQL doesn’t support streaming results, hence the Python database driver loads the entire result set into memory. The result set is then transformed into Python row objects by the database adapter using the fetchmany() method defined in [PEP 249](https://peps.python.org/pep-0249/).
>
> https://docs.djangoproject.com/en/4.2/ref/models/querysets/#without-server-side-cursors

nijel:
Thanks for figuring it out. I've updated the upgrading instructions to mention this.

PR #8600, opened 2023-01-13
docs: update gotext capabilities

SamWhited:
This PR updates the docs for the recent `gotext` addition to reflect the fact that gotext supports source code locations (https://github.com/translate/translate/blob/6b1b643cf687fc8a264bb62d36810c7626bf0afc/translate/storage/jsonl10n.py#L433) and descriptions in the form of both comments and translator comments (https://github.com/translate/translate/blob/6b1b643cf687fc8a264bb62d36810c7626bf0afc/translate/storage/jsonl10n.py#L426-L428).

Context is normally handled by setting the key to, eg. in the case of "Sun" for sunday vs. the star: "Sun(abbr)" or "Sun(star)" for example, but I don't know if this counts as specific support, it's just what the Go team has encouraged in the past so I left that one off.

It also updates the file used to the correct name (the `out.` version of the file is the one extracted from source by the tool that you might send off to a translator, not the one that's actually pulled into the application for use. In weblate I have used the `out.` file as the "Intermediate language file", but I never did really figure out what this meant or if it was right or how it differed from the base language file).


nijel:

Issue #8350, opened 2022-11-15
Show the source text of the source string as the content of the `New source string` event in the history


Geeyun-JY3:
### Describe the problem

The content of the `New source string` event in the history is the same as the name of the event.
It can be more useful.

### Describe the solution you'd like

Directly show the source text of the source string as the content of the `New source string` event in the history.

(Just like the `Removed string` event though not all `Removed string` events would show the text of the corresponding string. Another bug?)

### Screenshots

![image](https://user-images.githubusercontent.com/49396421/201832830-e5529648-56aa-4f15-b917-a6e20f1fde80.png)

nijel:
The event is already attached to a string, so it should be pretty easy to add the displaying.

d29107d:
@Geeyun-JY3 
I want to finish this issue.
I am a newbie to weblate.
Can you tell me how to produce the `New source string` event?

Geeyun-JY3:
> I want to finish this issue.
> I am a newbie to weblate.
> Can you tell me how to produce the `New source string` event?

I am not a developer :)
You should @nijel

d29107d:
@Geeyun-JY3 
Actually, I don't know how to produce the `New source string`  history via the website.
What you want to do is directly show the source text of the source string instead of show `New source string`, right?

Geeyun-JY3:
> Actually, I don't know how to produce the `New source string` history via the website. What you want to do is directly show the source text of the source string instead of show `New source string`, right?

Oh I got you.

Go to the changes page at any level. Then choose the `New source string` in the **Action** section of the bottom **Search** filter menu.
Here is an example: https://hosted.weblate.org/changes/?project=&lang=zh_Hans&action=13&user=&start_date=&end_date=

d29107d:
@nijel 
I don't have `New source string` in my local environment, do you know how to produce a `New source string` data?

nijel:
It is generated here:

https://github.com/WeblateOrg/weblate/blob/6ee22e1b19c9a0b9553affce7e41df524134bca2/weblate/trans/models/component.py#L1045-L1047

As mentioned, it already has a unit attached, so this is just a matter of displaying that.

PR #8750, opened 2023-02-06
change: Show the source text of the source string as the content of t…

d29107d:
Directly show the source text of the source string as the content of the New source string event in the history.
Issue: https://github.com/WeblateOrg/weblate/issues/8350

## Checklist

### DIFF
--- a/weblate/trans/models/change.py
+++ b/weblate/trans/models/change.py
@@ -592,6 +592,9 @@ class Change(models.Model, UserDisplayMixin):
     def get_action_display(self):
         if self.action in self.PLURAL_ACTIONS:
             return self.PLURAL_ACTIONS[self.action] % self.plural_count
+        if self.action == self.ACTION_NEW_SOURCE:
+            return self.get_source()
+
         return str(self.ACTIONS_DICT.get(self.action, self.action))
 
     def get_state_display(self):


d29107d:
@nijel 
Could you run the CI again or check why the checks failed?
I don't have permission to run CI again.

nijel:
Did that now. There are some issues with the azure apt mirror (see https://github.com/actions/runner-images/issues/675).

d29107d:
@nijel 
I contribute the code the first time and rebase the main branch, it needs a maintainer to approve running workflows.

weblate/trans/models/change.py
```diff
@@ -592,6 +592,9 @@ def auto_status(self):
     def get_action_display(self):
         if self.action in self.PLURAL_ACTIONS:
             return self.PLURAL_ACTIONS[self.action] % self.plural_count
+        if self.action == self.ACTION_NEW_SOURCE:
+            return self.get_source()
```
nijel, line 596: Display will be messed up with plural strings (that's also reason for failing tests). The value should be rendered using `format_translation` template tag in `weblate/templates/last-changes-content.html`.

Also, this is wrong place to display content, this method returns action name and is rendered in the left column while browsing changes.

PR #8811, opened 2023-02-17
Fix creating repository forks on gitea

Bubu:
## Proposed changes

https://github.com/WeblateOrg/weblate/commit/38008c51e6a12630348ed36f34d0370e1fb10c9c changed the request being send to gitea for creating a project
fork from sensing an empty json body with the correct Content-Type
header to sending just an empty body and no Content-Type header.

Gitea doesn't accept these requests and returns an "Empty Content-Type"
error instead.

Fix that by adding back the json body via requests `json` parameter
which also adds the correct content type header again.

## Checklist

### DIFF
--- a/weblate/vcs/git.py
+++ b/weblate/vcs/git.py
@@ -1099,7 +1099,9 @@ class GiteaRepository(GitMergeRequestBase):
     def create_fork(self, credentials: Dict):
         fork_url = "{}/forks".format(credentials["url"])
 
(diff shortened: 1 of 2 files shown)


Bubu:
Adapted the existing gitea vcs tests to catch this error. (Fails without the first commit, passes afterwards).

Also rename a misleadingly named gitea test function while at it.

PR #8875, opened 2023-03-03
Fix/next not working in redirect single

sh-TU:
## Proposed changes

The next parameter is not working when there's only 1 authentication backend available.

## Checklist

## Other information
none


nijel:

PR #8880, opened 2023-03-04
machinery: Remove a parameter that are no longer used

Xzonn:
https://github.com/WeblateOrg/weblate/commit/e3595b31c2600a92102b1f8a23e9e5ae14c166db This commit removed a parameter for the fuction `download_translations`, but did not remove the parameter of the test case, which will cause the failure when adding a new api.

image

## Proposed changes

Fix the problem that failed when adding a new api.

## Checklist


nijel:

PR #8883, opened 2023-03-05
Feature: Add goi18n v2 exporter

KuSh:
## Proposed changes

Add goi18n v2 exporter using new translate-toolkit capability
Also mark goi18n v1 format as supporting descriptions

## Checklist

## Other information

I wasn't able to test exporting explanation from ui to files. I'll appreciate a little bit of help as I don't know weblate nor python too much


requirements.txt
```diff
@@ -40,7 +40,7 @@ sentry_sdk>=1.6,<2
 siphashc>=1.2,<3.0
 social-auth-app-django==5.0.0
 social-auth-core==4.3.0
-translate-toolkit>=3.8.3,<3.9
+translate-toolkit>=3.8.6,<3.9
```
nijel, line 43: The dependency bump is not really needed, the format will just not be supported if using an older version.

KuSh: Ohh that's cool ! Thanks

nijel, line 43:
```suggestion
translate-toolkit>=3.8.3,<3.9
```

KuSh:
@nijel if you can give me advices on how to test "explanation" export into TranlsationUnit notes which should be exposed as "description" attribute in go-i18n formats, I'd love some help.

nijel:
It should be shown as "Source string description", see https://hosted.weblate.org/translate/weblate/application/cs/?checksum=39b3998dcfe68cbf

nijel:

nijel:
Description set in Weblate is not stored to the file, that one is kept in the database only.

PR #8921, opened 2023-03-13
Fix next param still not working when single social auth is enabled

sh-TU:
## Proposed changes

With this change, the POST login request should be set correctly as a form parameter instead of query url
image

## Checklist

## Other information

n/a


nijel:

PR #8924, opened 2023-03-14
Addons post-configure trigger changes

nijel:
## Proposed changes

- consistently handle changed repo after installing add-on
- trigger add-on execution in the background to avoid slowing down installation response

## Checklist


nijel:
@kadanna this should address the add-on installation issue you saw.

Issue #9010, opened 2023-03-29
AttributeError 'NoneType' object has no attribute 'xmlelement'


khrlzhfr:
### Describe the issue

I am translating an ODT document and have VCS enabled for this component. It can't commit to VCS and keep returning the following error:

### 'NoneType' object has no attribute 'xmlelement'

Request Method:	POST

Request URL:	https://l10n.space/commit/imap/state-of-internet-censorship-indonesia-2022/

Django Version:	4.1.7

Exception Type:	AttributeError

Exception Value:	
`'NoneType' object has no attribute 'xmlelement'`

Exception Location:	/usr/local/lib/python3.11/site-packages/weblate/formats/ttkit.py, line 612, in get_unit_node

Raised during:	weblate.trans.views.git.commit_component

Python Executable:	/usr/local/bin/python

Python Version:	3.11.2

Python Path:	
```
['/',
 '/usr/local/lib/python3.11/site-packages/git/ext/gitdb',
 '/',
 '/usr/local/bin',
 '/usr/local/lib/python311.zip',
 '/usr/local/lib/python3.11',
 '/usr/local/lib/python3.11/lib-dynload',
 '/usr/local/lib/python3.11/site-packages',
 '/app/data/python']

```
Server time:	Tue, 28 Mar 2023 19:31:28 +0000

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar issues in this repository.

### Steps to reproduce the behavior

2. Scroll down to '...'
3. Click on '...'

### Expected behavior

Able to commit to VCS and push to Gitlab/Github.

### Screenshots

![screencapture-mail-google-mail-u-1-2023-03-29-15_40_18](https://user-images.githubusercontent.com/1159232/228462202-07c76690-8b84-4d08-be25-9ba41d914ea0.jpg)

### Exception traceback

```pytb
See above.
```

### How do you run Weblate?

Docker container

### Weblate versions

4.16.4

PR #9012, opened 2023-03-29
Update ttkit.py

khrlzhfr:
To fix https://github.com/WeblateOrg/weblate/issues/9010

### DIFF
--- a/weblate/formats/ttkit.py
+++ b/weblate/formats/ttkit.py
@@ -617,7 +617,7 @@ class XliffUnit(TTKitUnit):
         if unit is not None:
             return unit.xmlelement.find(unit.namespaced(element))
         else:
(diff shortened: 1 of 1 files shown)


weblate/formats/ttkit.py
```diff
@@ -614,7 +614,10 @@ def _invalidate_target(self):
 
     @staticmethod
     def get_unit_node(unit, element: str = "target"):
-        return unit.xmlelement.find(unit.namespaced(element))
+        if unit is not None:
+            return unit.xmlelement.find(unit.namespaced(element))
+        else:
+            return None
```
nijel, line 620:
```suggestion
        return None
```

nijel:

Issue #8969, opened 2023-03-22
When Weblate tries to push changes to a Bitbucket Server repo while a pull request is still open, the translation component is locked


bebroll:
### Describe the issue

I have just been experimenting with the recent support for Bitbucket Server PRs. I was able to successfully create a translation component which uses "Bitbucket Server pull request" as its Version Control System type. When making modifications to the translation component and pushing these via the Repo Maintenance menu, a PR is successfully created.

Now assume that another modification is made in Weblate before the PR is merged. When using the Repo Maintenance menu again to push the changes, the translation component is locked with the alert message "Pull request failed: Only one pull request may be open for a given source and target branch".

Depending on whether "Repository push URL" and "Repository push branch" are specified or not (based on the table in the documentation https://docs.weblate.org/en/latest/admin/continuous.html#push-changes, this would be the difference between "request from fork" and "request from branch"), there is slightly different behavior:

* When using "request from branch", the original repo is not pushed to and the PR state is not updated.
* When using "request from fork", Weblate pushes the updated branch to the fork and the updated state of the fork's branch is correctly reflected in the PR.

From my understanding, if a PR is already open, the branch (either in the original repo or in the fork) should simply be pushed to. Weblate does not have to lock the translation component if creating the PR fails.

PS: I found only one related open issue on the matter (https://github.com/WeblateOrg/weblate/issues/2344). But this issue is the original request to add support for Bitbucket PRs which was used as the starting point for the recently implemented Bitbucket PR support.

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar issues in this repository.

### Steps to reproduce the behavior

1. Create a translation project
2. Create a translation component with "Bitbucket Server pull request" Version Control System; leave "Repository push URL" and "Push branch" empty
3. Make a change in the translation component and use Manage > Repository Maintenance > Push
4. Observe that a PR is created based on a branch in a fork of the original repo
5. Make another change in the translation component and use Manage > Repository Maintenance > Push
6. Observe that the operation fails with the message "Pull request failed: Only one pull request may be open for a given source and target branch" and that the translation component is locked
7. Observe that the PR created in step (3) now contains the changes that were made in (5)

### Expected behavior

If a PR already exists in Bitbucket, the latest changes should be pushed and the translation component should not be locked.

### How do you run Weblate?

Docker container

### Weblate versions

 * Weblate: 4.16.4
 * Django: 4.1.7
[...]

PR #9047, opened 2023-04-05
8969: Bitbucket Fix not throwing error and locking project when PR al…

afridlund85:
…ready exists

## Proposed changes

Fixed issue described in [8969](https://github.com/WeblateOrg/weblate/issues/8969). Weblate should no longer lock the repo when this happens. 
Bitbucket Server updates PR automatically when PR from ref changes and thus does not need to try and recreate/update a PR. Much like the Gitea implementation just does nothing when a PR already exists.

## Checklist


nijel:

PR #9064, opened 2023-04-08
Spelling: Screenshot assigned

comradekingu:
This is what I think it does, as https://hosted.weblate.org/projects/scrutin/app/#alerts doesn't have any screenshots not assigned to anything…


nijel (changes requested):
No, it is the other way around.

PR #9090, opened 2023-04-15
Docs: Config language reworked

comradekingu:
## Checklist


docs/admin/config.rst
```diff
@@ -964,8 +964,8 @@ which address from the header is used as client IP address here.
 
 .. warning::
 
-   Setting this affects the security of your installation, you should only
-   configure it to use trusted proxies for determining IP address.
+   Setting this affects the security of your installation. You should only
+   configure it to use trusted proxies for determining IP addresses.
```
comradekingu, line 968:
```suggestion
   configure it to use trusted proxies for determining the IP address.
```
Didn't check to see which is correct.

docs/admin/config.rst
```diff
@@ -221,16 +221,16 @@ BASIC_LANGUAGES
 
 .. versionadded:: 4.4
 
-List of languages to offer users for starting new translation. When not
-specified built-in list is used which includes all commonly used languages, but
-without country specific variants.
+List of languages to offer users for starting a new translation. When not
+specified a built-in list is used (which includes all commonly used languages, but
```
comradekingu, line 225:
```suggestion
specified, a built-in list is used (which includes all commonly used languages, but
```

docs/admin/config.rst
```diff
@@ -1463,10 +1463,10 @@ REGISTRATION_ALLOW_BACKENDS
 
 List of authentication backends to allow registration from. This only limits
```
comradekingu, line 1464:
```suggestion
List of authentication back-ends to allow registration from. This only limits
```
Seems this has changed now https://en.wikipedia.org/wiki/Frontend_and_backend
which could mean changing all "back-end" to "backend" is warranted.
Did not consult the Weblate glossary.

docs/admin/config.rst
```diff
@@ -1755,9 +1754,9 @@ UNUSED_ALERT_DAYS
 
 .. versionadded:: 4.17
 
-Configures when :guilabel:`Component seems unused` alert is trigged.
+Configures when :guilabel:`Component seems unused` alert is triggered.
```
comradekingu, line 1757:
```suggestion
Configures when the :guilabel:`Component seems unused` alert is triggered.
```

docs/admin/config.rst
nijel: It is called "backend" in https://python-social-auth.readthedocs.io/en/latest/backends/index.html and https://docs.djangoproject.com/en/4.2/topics/auth/customizing/#authentication-backends, so lets stick with the same naming here as well.

docs/admin/config.rst
```diff
@@ -45,8 +45,8 @@ AUDITLOG_EXPIRY
 
 .. versionadded:: 3.6
 
-How many days Weblate should keep audit logs, which contain info about account
-activity.
+How many days Weblate should keep audit logs, (which contain info about account
```
nijel, line 48:
```suggestion
How many days Weblate should keep audit logs (which contain info about account
```

docs/admin/config.rst
```diff
@@ -611,7 +611,7 @@ DEFAULT_MERGE_STYLE
 
 .. versionadded:: 3.4
 
-Merge style for any new components.
+Merge-style for any new components.
```
nijel, line 614: It is called "Merge style" in all other places. If you intend to change this, please change it in all places at once in a separate pull request for each such change.

nijel: How about referencing the actual setting instead?

```suggestion
:ref:`component-merge_style` for any new components.
```

docs/admin/config.rst
```diff
@@ -595,7 +595,7 @@ DEFAULT_LANGUAGE
 
 .. versionadded:: 4.3.2
 
-Default source language to use for example in :ref:`component-source_language`.
+Default source-language to use for example in :ref:`component-source_language`.
```
nijel, line 598:
```suggestion
:ref:`component-source_language` for any new components.
```

docs/admin/config.rst
```diff
@@ -1489,9 +1489,8 @@ Example:
 REGISTRATION_CAPTCHA
 --------------------
 
-A value of either ``True`` or ``False`` indicating whether registration of new
-accounts is protected by CAPTCHA. This setting is optional, and a default of
-``True`` will be assumed if it is not supplied.
+An optional value of either ``True`` or ``False`` indicating whether registration of new
+accounts is protected by a CAPTCHA. A default of ``True`` is assumed if not supplied.
```
nijel, line 1493: Nearly all settings are optional, so this should not be mentioned for each of them.

```suggestion
Whether registration of new accounts is protected by a CAPTCHA. Defaults to enabled.
```

docs/admin/config.rst
```diff
@@ -1521,7 +1520,7 @@ REGISTRATION_OPEN
 -----------------
 
 Whether registration of new accounts is currently permitted.
-This optional setting can remain the default ``True``, or changed to ``False``.
+This optional setting can remain the default ``True``, or be changed to ``False``.
```
nijel, line 1523: Defaults to enabled.

docs/admin/config.rst
```diff
@@ -1690,8 +1689,8 @@ SINGLE_PROJECT
 .. versionadded:: 3.8
 
 Redirects users directly to a project or component instead of showing
-the dashboard. You can either set it to ``True`` and in this case it only works in
-case there is actually only single project in Weblate. Alternatively set
+the dashboard. You can either set it to ``True`` so it only works in
+if there is actually only single project in Weblate. Alternatively, set
```
nijel, line 1693:
```suggestion
the dashboard. You can either set it to ``True`` so it only works
if there is actually only single project in Weblate. Alternatively, set
```

nijel (changes requested):

docs/admin/config.rst
comradekingu: Seems better to me. What I wanted to avoid was that it possibly reads like "Merge (the) style for any new components", rather than "(the) merge style for any new components."

docs/admin/config.rst
```diff
@@ -1463,10 +1463,10 @@ REGISTRATION_ALLOW_BACKENDS
 
 List of authentication backends to allow registration from. This only limits
 new registrations, users can still authenticate and add authentication using
-all configured authentication backends.
+all configured authentication back-ends.
```
comradekingu, line 1466:
```suggestion
all configured authentication backends.
```

docs/admin/config.rst
```diff
@@ -1463,10 +1463,10 @@ REGISTRATION_ALLOW_BACKENDS
...
 
-It is recommended to keep :setting:`REGISTRATION_OPEN` enabled while limiting
-registration backends, otherwise users will be able to register, but Weblate
+It is recommended to keep :setting:`REGISTRATION_OPEN` on while limiting
+registration back-ends, otherwise users will be able to register, but Weblate
```
comradekingu, line 1469:
```suggestion
registration backends, otherwise users will be able to register, but Weblate
```

docs/admin/config.rst
```diff
@@ -1477,7 +1477,7 @@ Example:
 
 .. hint::
 
-   The backend names match names used in URL for authentication.
+   The back-end names match names used in the URL for authentication.
```
comradekingu, line 1480:
```suggestion
   The backend names match names used in the URL for authentication.
```

docs/admin/config.rst
comradekingu, line 1523:
```suggestion
This optional setting can remain the default ``True``, or be changed to ``False``.
Defaults to enabled.
```

docs/admin/config.rst
```diff
@@ -1546,14 +1545,14 @@ REGISTRATION_REBIND
...
-Allow rebinding authentication backends for existing users. Turn this on when
+Allow rebinding authentication back-ends for existing users. Turn this on when
 migrating between authentication providers.
 
 .. note::
 
-   Disabled by default to not allow adding other authentication backends to
-   existing account. Rebinding can lead to account compromise when using
-   more third-party authentication backends.
+   Off by default to not allow adding other authentication back-ends to
```
comradekingu, line 1553:
```suggestion
   Off by default to not allow adding other authentication backends to
```

docs/admin/config.rst
```diff
@@ -1546,14 +1545,14 @@ REGISTRATION_REBIND
...
+   an existing account. Rebinding can lead to account compromise when using
+   more third-party authentication back-ends.
```
comradekingu, line 1555:
```suggestion
   more third-party authentication backends.
```

docs/admin/config.rst
```diff
@@ -1521,6 +1520,7 @@ REGISTRATION_OPEN
 
 Whether registration of new accounts is currently permitted.
 This optional setting can remain the default ``True``, or be changed to ``False``.
```
nijel, line 1522:
```suggestion
```

nijel:

Issue #9048, opened 2023-04-05
Automatic translation support adding translation as "approved"


KadAnna:
### Describe the problem

Following ticket https://github.com/WeblateOrg/weblate/issues/8979, I would like to have an option to set the automatically translated strings as "for review".   This will correspond with the "for review" workflow.

### Describe the solution you'd like

Add an option "Add as for review" to Automatic translation mode of the add-on set-up.

nijel:
The strings should be added as “for review” right now. I thought you wanted to have the possibility to add strings as “approved”.

KadAnna:
I just tried tool Automatic translation. There the Automatic translation mode Add as translation (together with Enabled reviews Setting) sets strings as for review. So it this is the same at the addon Automatic translation, then this ticket is unnecessary. Just that because of the bug https://github.com/WeblateOrg/weblate/issues/8979 I could not see it.

nijel:
I still think it makes sense to add an option to add string as approved, so let's keep this issue open.

nijel:
To implement this:

* `AutoForm` needs to extend `__init__` so that it adds `Add as approved` when reviews are enabled:

  https://github.com/WeblateOrg/weblate/blob/18f85422e98e2d3ea0960093c647ea59fef09371/weblate/trans/forms.py#L906

* `AutoTranslate` needs support for `mode` = `approved`, at leas here:
 
  https://github.com/WeblateOrg/weblate/blob/18f85422e98e2d3ea0960093c647ea59fef09371/weblate/trans/autotranslate.py#L34

Krishhhnaa:
is this issue closed ?

nijel:
No, it is open.

PR #9174, opened 2023-04-27
Add auto translation as "approved" (Closes: #9048)

jspricke:

### DIFF
--- a/weblate/trans/forms.py
+++ b/weblate/trans/forms.py
@@ -869,12 +869,6 @@ class AutoForm(forms.Form):
 
     mode = forms.ChoiceField(
         label=_("Automatic translation mode"),
-        choices=[
-            ("suggest", _("Add as suggestion")),
-            ("approved", _("Add as approved translation")),
-            ("translate", _("Add as translation")),
-            ("fuzzy", _('Add as "Needing edit"')),
-        ],
         initial="suggest",
     )
     filter_type = FilterField(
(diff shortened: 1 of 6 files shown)


weblate/trans/forms.py
```diff
@@ -871,6 +871,7 @@ class AutoForm(forms.Form):
         label=_("Automatic translation mode"),
         choices=[
             ("suggest", _("Add as suggestion")),
+            ("approved", _("Add as approved translation")),
```
nijel, line 874: This will need additional logic in `__init__` to show approved choice only if the user has permissions for that. `__init__` will need to get user object as a parameter for that and then check the permissions via `user.has_perm("unit.review", obj)`.

jspricke: done with the latest commit.

weblate/trans/forms.py
```diff
@@ -969,9 +979,13 @@ def __init__(self, obj, *args, **kwargs):
             x for x in self.fields["filter_type"].choices if x[0] in use_types
         ]
 
+        field_mode = Field("mode")
+        if user is not None and user.has_perm("unit.review", obj):
+            field_mode = Field("mode_approved")
```
jspricke, line 984: I tried this instead but it didn't work:
```self.fields["mode"].choices.append(("approved", _("Add as approved translation")))```

nijel:

PR #9209, opened 2023-05-07
Fix PunctuationSpacingCheck for repeated punctation

p0358:
## Proposed changes

There's a rule for French that ensures there's whitespace before certain punctuation characters: https://docs.weblate.org/en/latest/user/checks.html#punctuation-spacing

However there are some strings where you have doubled characters such as `!!`, `?!`, `::`. The lack of space between them was flagged by this check, which I think is wrong.

Some examples of what would be flagged before this PR:
- `%s1  ::  %s2`
- `État collecteur : dommages lourds !!`
- `BT, t'as abandonné l'Arche ! Pourquoi ?!`

## Checklist

- [X] I have added tests that prove my fix is effective or that my feature works.

## Other information

I didn't run tests/lint, but I hope the change is simple enough for that to be fine.

### DIFF
--- a/weblate/checks/chars.py
+++ b/weblate/checks/chars.py
@@ -415,7 +415,10 @@ class PunctuationSpacingCheck(TargetCheck):
             if char in FRENCH_PUNCTUATION:
                 if i + 1 < total and not target[i + 1].isspace():
                     continue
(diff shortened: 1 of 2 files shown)


nijel:

PR #9225, opened 2023-05-13
Update language plural rules link.

Maikuolan:
## Proposed changes

Unicode has migrated its CLDR pages to a new URL. Current version is also now version 43 (version 37 isn't available for viewing as a page directly via the browser anymore, though it remains available as a downloadable ZIP archive).

## Checklist

## Other information

This PR only changes an outdated link to its contemporary counterpart, and doesn't introduce any changes to any code, so I haven't done any linting or tests (referring to the above checklist), because I don't feel it should be necessary in this particular case.


nijel:

PR #9273, opened 2023-05-19
Fix the API URL for Microsoft Terminology

lahdekorpi:
The URL requires https instead of http.

## Checklist

### DIFF
--- a/weblate/machinery/microsoftterminology.py
+++ b/weblate/machinery/microsoftterminology.py
@@ -8,7 +8,7 @@ from zeep import Client
 
 from weblate.machinery.base import MachineTranslation
 
-MST_API_URL = "http://api.terminology.microsoft.com/Terminology.svc"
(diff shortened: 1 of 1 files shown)


nijel:
Thanks, can you please also adjust test so that it mocks same URL?

lahdekorpi:
@nijel I hope this is enough for the tests. I updated the URL in the envelope in the `machinery/tests.py`

lahdekorpi:
Oh, apparently the wsdl files in the mock need to be updated as well.
```
weblate.machinery.base.MachineTranslationError: ConnectionError("Connection refused by Responses - the call doesn't match any registered mock.\n\nRequest: \n- GET http://api.terminology.microsoft.com/Terminology.svc?wsdl=wsdl0\n\nAvailable matches:\n- GET https://api.terminology.microsoft.com/Terminology.svc URL does not match\n- POST https://api.terminology.microsoft.com/Terminology.svc Method does not match\n")
```

PR #9300, opened 2023-05-24
adding terminology flag description, changing IBM link

orangesunny:
## Checklist

### DIFF
--- a/docs/admin/checks.rst
+++ b/docs/admin/checks.rst
@@ -53,7 +53,7 @@ Here is a list of flags currently accepted:
 ``read-only``
     The string is read-only and should not be edited in Weblate, see :ref:`read-only-strings`.
 ``terminology``
-    Used in :ref:`glossary`. Copies the string into all glossary languages so it can be used consistently in all translations. Also useful in combination with ``read-only``, in for example product names.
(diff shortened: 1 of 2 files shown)


orangesunny:
@comradekingu, would you check the line, please? Thanks!

docs/admin/checks.rst
```diff
@@ -52,6 +52,8 @@ Here is a list of flags currently accepted:
     Uses DOS end-of-line markers instead of Unix ones (``\r\n`` instead of ``\n``).
 ``read-only``
     The string is read-only and should not be edited in Weblate, see :ref:`read-only-strings`.
+``terminology``
+    Used in :ref:`glossary`. It copies the string into all glossary languages so it can be used consistently in all transltions. Also useful in combination with ``read-only``, eg. for product names.
```
comradekingu, line 56:
```suggestion
    Used in :ref:`glossary`. Copies the string into all glossary languages so it can be used consistently in all translations. Also useful in combination with ``read-only``, in for example product names.
```
Should have been "e.g.", and also not "transitions"?

Alternatively (possibly better)
"Used in :ref:`glossary`. Copies the string into all glossary of all translated languages so it can be used consistently" is a bit simpler, but requires there to be glossaries for all translated languages, which I don't think is the case(?). Does adding as terminology start glossaries for languages that don't have one?

orangesunny:
Thanks, @comradekingu for checking and improving.

PR #9330, opened 2023-05-30
integrations: add model for integration and integration event

ParthS007:
## Checklist


nijel:
Can you please add new app to `INSTALLED_APPS`? Without that codecov complains about not covered code everywhere and that makes it hard to review the code.

weblate/integrations/models.py
```diff
@@ -0,0 +1,143 @@
...
+        return component.inegrations_cache[event]
+
+
+class Integration(models.Model):
+    component = models.ForeignKey(Component, on_delete=models.deletion.CASCADE)
+    name = models.CharField(max_length=100)
+    configuration = JSONField()
+    state = JSONField()
+    project_scope = models.BooleanField(default=False, db_index=True)
+    repo_scope = models.BooleanField(default=False, db_index=True)
```
nijel, line 39:
```suggestion
```

These are probably not a good idea here – in the add-ons, the scoping is defined by the add-on, here it will be user defined. Instead, add a foreign key to `Project` as well, and make both foreign keys nullable.

PR #9473, opened 2023-06-28
Improved plurals handling in machine translation

nijel:
## Proposed changes

This covers some situations that were missed in https://github.com/WeblateOrg/weblate/pull/8323 by @Changaco:

- blank translations when interpolation is supported by the check but not possible
- improved mapping when translating from language with more complex plural rules
- fixed plurals merging in UI when more plurals are translated with a same string
- added tests for `PluralMapper`

## Checklist

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -3,6 +3,8 @@ Weblate 5.0
 
 Not yet released.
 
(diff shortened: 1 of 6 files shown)


Changaco (approved):
Looks okay to me.

PR #9540, opened 2023-07-11
fluent: add checks for fluent units.

henry-torproject:
## Proposed changes

Relies on https://github.com/translate/translate/pull/4963

We add checks for:

+ Proper fluent syntax, to ensure the source can be written to a file.
+ Same fluent "parts", to ensure the translation has the same attributes and presence of a value as the original.
+ Same fluent references, to ensure the translation includes the same expected references, taking variants into account.
+ Valid as inner HTML, to ensure the source can be parsed as inner HTML without a loss of content.
+ Same HTML elements, to ensure the translation includes the same expected elements, taking variants into account.

As part of this, we disable the placeholders and XML checks since they are redundant now, and do not work well with more complex fluent entries.

Fixes https://github.com/translate/translate/issues/4448

## Checklist

## Other information

This was tested locally using the changes to translate toolkit from https://github.com/translate/translate/pull/4963


docs/user/checks.rst
```diff
@@ -269,6 +269,314 @@ ECMAScript template literals
    :ref:`check-formats`,
    `Template literals <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals>`_
 
+.. _check-fluent-source-syntax:
+
+Fluent Source Syntax
+********************
+
+.. versionadded:: ??
```
henry-torproject, line 277: Still need to add a version number.

orangesunny (changes requested):
Docs are good.
- @henry-torproject can read the docs again and validate my changes.
- version number needs to be added. Are we ready to put it to 5.0, @nijel?

henry-torproject:
> Docs are good.
> 
>     * @henry-torproject can read the docs again and validate my changes.

@orangesunny thanks for proof reading them. All the changes look good to me! I would just change:

> This check ensures that the translated Message also has a value if the source Message has it, or no value if the source has no value

to

> This check ensures that the translated Message also has a value if the source Message has one, or no value if the source has none

just to emphasis that it doesn't have to be the *same* value.

I'm not sure about your merge request process. Is it ok for me to take your commit and merge it into mine and do another (force) push?

orangesunny:
> > Docs are good.
> > ```
> > * @henry-torproject can read the docs again and validate my changes.
> > ```
> 
> @orangesunny thanks for proof reading them. All the changes look good to me! I would just change:
thank you!
> 
> > This check ensures that the translated Message also has a value if the source Message has it, or no value if the source has no value
> 
> to
> 
> > This check ensures that the translated Message also has a value if the source Message has one, or no value if the source has none
> 
> just to emphasis that it doesn't have to be the _same_ value.
> 
just included this, thank you!

> I'm not sure about your merge request process. Is it ok for me to take your commit and merge it into mine and do another (force) push?

Yes, within pull requests, force push is OK and squashing is great.

docs/user/checks.rst
nijel: It will be 5.0

weblate/checks/fluent/inner_html.py
```diff
@@ -0,0 +1,1290 @@
+# Copyright © Henry Wilkes <henry@torproject.org>
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
```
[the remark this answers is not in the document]
nijel: Yes, we follow https://reuse.software/ for license annotations.

nijel (commented):
The inner HTML check is very complex. Could it be based on the Safe HTML check that Weblate already has? It doesn't provide that detailed feedback, but that would heavily reduce the size of this patch.

Code is here:

https://github.com/WeblateOrg/weblate/blob/f46a0747355979f104a93eda8cbfbd4a2f024cd8/weblate/checks/markup.py#L344-L355

henry-torproject:
> The inner HTML check is very complex. Could it be based on the Safe HTML check that Weblate already has? It doesn't provide that detailed feedback, but that would heavily reduce the size of this patch.
> 
> Code is here:
> 
> https://github.com/WeblateOrg/weblate/blob/f46a0747355979f104a93eda8cbfbd4a2f024cd8/weblate/checks/markup.py#L344-L355

@nijel unfortunately, I looked at a lot of existing HTML parsers and I didn't find one that does what we want for fluent.

Note that this is not meant as a HTML sanitizer or linter.

Basically, translators (and devs) will most of the time not realise they are setting inner HTML, nor do we want them to worry about that when it isn't important. So we only want to prevent HTML parsing errors that lead to a loss of content, like [eof-in-tag](https://html.spec.whatwg.org/#parse-error-eof-in-tag), but not warn about HTML parsing errors that do not lead to a loss of content, like [invalid-first-character-of-tag-name for start tags](https://html.spec.whatwg.org/#parse-error-invalid-first-character-of-tag-name).

Here are examples of fluent strings in firefox to highlight this:

+ https://searchfox.org/l10n/rev/4a820d9ece0e5561bcdd4fef4e924c98b5c79590/be/toolkit/toolkit/about/aboutProcesses.ftl#160
+ https://searchfox.org/l10n/rev/4a820d9ece0e5561bcdd4fef4e924c98b5c79590/hye/browser/browser/preferences/preferences.ftl#1417
+ https://searchfox.org/l10n/rev/4a820d9ece0e5561bcdd4fef4e924c98b5c79590/ar/browser/browser/preferences/preferences.ftl#474

So a generic HTML5 parser would be too lenient (it would parse and ignore things that would cause parts of the string to be lost), whilst an XML parser would be much too strict. I did a little bit of research to see if a parser could be configured but had no luck.

I tried with the standard `html.HTMLParser` and `lxml.etree.HTMLParser` but they were too lenient at ignoring losses of content. So I ended up writing my own HTML parser which does the right amount.

But I'm open to other ideas if there is some other way to have this same behaviour.

weblate/checks/fluent/utils.py
```diff
@@ -0,0 +1,255 @@
...
+import re
+from typing import TYPE_CHECKING
+
+from django.utils.html import escape, format_html, format_html_join
+from translate.storage.fluent import (
+    FluentPart,
+    FluentReference,
+    FluentSelectorBranch,
+    FluentUnit,
+)
```
nijel, line 16: Fluent support is optional in Weblate, but this makes it a requirement. Solution is needed:

* Deal with the import failure here.
* Adjust dependencies to require fluent.syntax unconditionally.

nijel (commented):
Okay, let's leave the HTML validation as is.

weblate/checks/fluent/utils.py
henry-torproject: Which would you prefer?

I guess with this added support it could be considered unconditional.

henry-torproject:
I'm going to be away next week and the week after (17th - 30th July), so I won't able able to respond to questions in that period until I'm back. So it is just tomorrow that I can respond (within CET working time).

There's a lot to review here, so I think waiting the two weeks until I'm back would be good. You could start reviewing it in that time, but we can pick up when I'm back.

I think I'll mark this as "work in progress" during that period so it doesn't accidentally progress and land. Is that ok?

weblate/checks/fluent/utils.py
nijel: As it doesn't pull any huge third-party deps, I'd go for making it required.

henry-torproject:
ok, putting this on ice in the Draft status until I am back. Feel free to add comments whilst I'm away

weblate/checks/fluent/utils.py
nijel: After some considerations, I've decided to go this way for all formats in https://github.com/WeblateOrg/weblate/pull/9584.

weblate/checks/fluent/references.py
```diff
@@ -0,0 +1,751 @@
...
+    |    *[no] a { -term }
+    | }
+
+    Here a reference to ``-term.starts-with-vowel`` is not expected to appear in
+    translations, but a reference to ``-term`` is.
+    """
+
+    check_id = "fluent-references"
+    name = _("Fluent References")
+    descirption = _("Fluent references should match")
```
nijel, line 671: Typo

```suggestion
    description = _("Fluent references should match")
```

weblate/checks/fluent/syntax.py
```diff
@@ -0,0 +1,79 @@
...
+        return escape(
+            gettext("Fluent syntax error: {error}.").format(error=syntax_error)
+        )
+
+
+class FluentSourceSyntaxCheck(_FluentSyntaxCheck, SourceCheck):
+    """Check that the source uses valid Fluent syntax."""
+
+    check_id = "fluent-source-syntax"
+    name = _("Fluent Source Syntax")
```
nijel, line 47: None of the current checks use title-case, so I'd avoid it here as well.

docs/user/checks.rst
```diff
@@ -269,6 +269,314 @@ ECMAScript template literals
    :ref:`check-formats`,
    `Template literals <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals>`_
 
+.. _check-fluent-source-syntax:
```
nijel, line 272: The skeleton of the documentation should be generated by `weblate list_checks` to ensure consistency.

nijel (commented):
Thanks, this really looks good, besides some minor issues which I've pointed out.

henry-torproject:
ok, should be ready for review again. I'm planning on adapting the existing punctuation checks to work with fluent as well, but I think that will be a separate pull request.

nijel:
Thanks, I think it looks good. Waiting for tests to pass to see the code coverage before merging.

nijel:
File format test failed:

```
FAIL: test_flags (weblate.formats.tests.test_formats.FluentFormatTest.test_flags)
Check flags on corresponding translatable units.
  File "/home/runner/work/weblate/weblate/weblate/formats/tests/test_formats.py", line 327, in test_flags
    self.assertEqual(unit.flags, expected_flag)
AssertionError: 'fluent-type:Message' != ''
- fluent-type:Message
+ 
```

Expected flags need to be adjusted here:

https://github.com/WeblateOrg/weblate/blob/7d43b975bd9039a1d60416df0557d39fd958ba33/weblate/formats/tests/test_formats.py#L1282

henry-torproject:
thanks @nijel Do you have an idea for when these changes would reach hosted.weblate.org?

nijel:
They are already there ;-).

henry-torproject:
> They are already there ;-).

oh great! Does hosted.weblate pull from the "main" branch rather than the releases?

nijel:
It is manually deployed from main when I think it's good to do so.

PR #9557, opened 2023-07-13
Reporting templates reworked

comradekingu:
## Proposed changes

Fairly standard stuff. Nothing untowards.

## Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -10,9 +10,13 @@ body:
 - type: markdown
   attributes:
     value: |
-      Thank you for reporting an issue on Weblate! This form will guide you to create a useful issue report.
+      Thank you for requesting a change.
+      This form guides you in creating a useful feature request.
 
-      Want your answer quickly and guaranteed? Visit https://weblate.org/support/ to reach our dedicated support team. As a subscriber, you will always have priority and help Weblate growing.
+      Get in contact with our our dedicated support team
+      for quick and guaranteed answers on https://weblate.org/support/
+      As a subscriber, you always have priority and help Weblate grow.
+- type: textarea
 - type: textarea
   id: describe
   attributes:
@@ -20,21 +24,21 @@ body:
     description: >
       Is your feature request related to a problem? If so, please provide
       a clear and concise description of what the problem is.
-      Ex. I'm always frustrated when [...]
+      Ex. I'm always frustrated when […]
   validations:
     required: true
 - type: textarea
   id: solition
   attributes:
-    label: Describe the solution you'd like
+    label: Describe the solution you would like
     description: A clear and concise description of what the new feature should do.
   validations:
     required: true
 - type: textarea
   id: alternatives
   attributes:
-    label: Describe alternatives you've considered
-    description: A clear and concise description of any alternative solutions or features you've considered.
+    label: Describe alternatives you have considered
+    description: A clear and concise description of any alternative solutions or features you have considered.
 - type: textarea
   id: screenshots
   attributes:
@@ -44,4 +48,4 @@ body:
   id: additional
   attributes:
     label: Additional context
-    description: Add any other context about the problem here.
+    description: Add any other contextual info about the problem here.
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -10,9 +10,12 @@ body:
 - type: markdown
   attributes:
     value: |
-      Thank you for reporting an issue on Weblate! This form will guide you to create a useful issue report.
+      Thank you for reporting an issue.
+      This form guides you in creating a useful issue report.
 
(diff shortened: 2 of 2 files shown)


.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -10,31 +10,35 @@ body:
 - type: markdown
   attributes:
```
comradekingu, line 11: Got this, but the bug-report template has one…

>There is a problem with this template
>
>body[1]: Required attribute key label is missing. [Learn more about this error.](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/common-validation-errors-when-creating-issue-forms#bodyi-required-attribute-key-value-is-missing)

nijel:
The changes have to be implemented in https://github.com/WeblateOrg/meta/tree/main/.github/ISSUE_TEMPLATE/snippets, otherwise they will be overwritten next time we sync this. But we can use this for discussion and then do final changes there...

orangesunny:
@comradekingu, I appreciate your willingness to help, though these changes are not all needed. Unlike the docs, parts of these templates are a bit of advertising text that does not aim to be the most grammatically correct, precise and polished version of a sentence. It should show the Weblate team’s attitude and kindly motivate to either fill the form or buy services.

As this PR will be closed either way, we need to open a new one in the correct place if we want to pursue these changes. Do you want me to create that PR, use some of your ideas, and mention you as a co-author, or will you open that PR yourself and I will then edit it to a desired outcome?

comradekingu:
@orangesunny
I worked in advertisement. I studied psychology a fair bit when doing philosophy.
The issue here isn't that. The issue here is failing at English.
Even the particular style is unimportant in so doing.
Not a good look for an outfit that facilitates translations.

I can't begin to understand what kind of idea it is to reject perfectly good help in a way that makes the reception more arduous than doing all the work. It is consistent with the
bulk of the language in the entire project at that. (There is no "advertisement" anywhere,
which is a concept Weblate is built on.)
That is a consistency that matters. It is appreciated by the types of very smart people that buy Weblate services.
I couldn't outsmart these people if my life depended on it.
Can I at least sell you that much insight?
The availability of second-guessing yourself is readily available. It just takes an A/B comparison from either trying it,
or getting an opinion on it.
There are people that can do advertisment, and they do it well. You also can't buy your way into it, and especially not for this type of operation. It could plain and simple go either way.

So, what you want in a bug-reporting or feature-requesting template is _not_ advertisement,
but rather a study in interaction design.
As a precursory guideline, take note of where you have seen the current approach and consider why it doesn't happen.

It is damaging to have straight garbage in production, as an added insult to injury for people wanting to report errors.
Care about the product. Care about what the end-result is, not the end of the line of what the intention may be.
The "Weblate team" to my knowledge doesn't include anyone to either catch these templates in the making,
or correct them afterwards. You could ask me, as this is easy for me to do, and/or consult most any native speaker.

I for the most part don't have "ideas", but rather some command of the English language.
You can't take my work, because sadly you think of it as something that isn't considered, or you aren't making those considerations in changing it. That is how to end up with what was the root issue again.
You have tried this, and it didn't work.

There is an insistence in protecting flaws, and you have the power to do.
If the small and irrelevant minutia all mattered so much after the fact,
consider there are other instances of mistakes just as egregious.
You couldn't control your way into perfection if you tried, and starting with the stuff I touch isn't
where to learn. Fortunately there is a lot of very low-hanging fruit.

When you take my work I could point out every single little flaw and even detail the considerations,
but you somehow don't learn from that. It is also easy for me to do.
Readjust the approach here. Intending good and doing well aren't always aligned.
Do you want to keep failing because you can tell yourself you are doing the right thing?
Take what is good, and shed what doesn't work.
Between me an nijel we worked out a concept of all the things I do that just don't cut it. I learnt a great deal from that.

I sometimes use too advanced language still, and you catch that.
Point it out when that happens, and I will readjust those passages. Done deal.
Similarly, if you instead detailed what the actual problem was here in the sense of language rather than concepts,
you wouldn't be able to.

"either fill the form or buy services" is more-so now down to the letter what it does.
The added fluff, non-attempt at grammar, etc. is neither kind nor welcoming.
You don't get credit for attempts. It has to all work.
We can't do how I do things, simply because that would be too restrictive, it is harder to translate, etc.
Meeting at the middle is where the author just disappears.

I can make a PR where it is supposed to go, but for love or money first point out what changes are needed beforehand.

As a perfect example of why added fluff isn't anything other than infuriating,
the "these changes are not all needed" part is either a misconstrued formulation, or fails to account for an objective world.

orangesunny:
TL;DR, @comradekingu. Definitely not now, I might get back to it later.

I understand that you want to show your opinion is truly right, but writing a paragraph(s) instead of a point(s) is not a way to be successful in Weblate repo, as well as many other repositories on GitHub, as you already know. Please, earnestly, focus more on being kind and open to discussion than trying to bulletproof and cement your opinions in large paragraphs.

nijel:
I've taken most of the changes and integrated them in https://github.com/WeblateOrg/meta/commit/38991e839fc16e85beba411879c7c2f901eb42b0. That is the place for changing these texts.

Issue #9531, opened 2023-07-10
Admin password step during setup is easy to miss.


yilmazdurmaz:
### Describe the problem

I was trying to have a local instance to work on local for a project.

This is the first time I try Weblate on local.

Most steps are clear, but this one is crucial and this line is very easy to misinterpret as I did and spent the next few hours where to find the admin password:
>Create an account for the administrator user and copy its password to the clipboard, and also save it for later use:
> `weblate createadmin`

### Describe the solution you'd like

This expression should emphasize the "copy the password" part, something like this:

__Create an account for the administrator user. This command will generate an admin password for the first use. Copy it to the clipboard, and also save it for later use: __

In addition, this would make it much better if you include this:

__ If you previously missed/lost the admin password, renew it with this command__
`weblate createadmin --update`

### Describe alternatives you've considered

The alternative was to check lots of places to see if there is a default password somewhere while spending a lot of time (few hours) just for this simple solution:
`weblate createadmin --update`

swrp-j:
Could you please assign this issue to me?

PR #9558, opened 2023-07-13
subsequent line update to "install-configure.rst" following #9531

yilmazdurmaz:
The main admin account is named "admin". It is better to emphasize this.

this follows my issue https://github.com/WeblateOrg/weblate/issues/9531 and subsequent fixes

## Checklist


orangesunny (approved):
We can merge this or keep the current working state. This seems better, so thank you for showing a different point of view that helped to figure this out, @yilmazdurmaz!

Issue #9532, opened 2023-07-10
Show language menu in language selection


comradekingu:
### Describe the problem

The top menu items all focus-shift (in language view) because the second entry is "language".

### Describe the solution you'd like

Add "Language" to this menu
![image](https://github.com/WeblateOrg/weblate/assets/13802408/da5faab6-97c4-4c59-9a1c-217594594cfa)

### Describe alternatives you've considered

Also considered adding all the submenus to the top menu, but that is more of an involved approach.

### Screenshots

This is what it usually looks like
![image](https://github.com/WeblateOrg/weblate/assets/13802408/ffa22c67-e21e-4ff8-892b-e3904b48c8d6)

### Additional context

"Administrate" and "Share" could be grayed out?

nijel:
What should the “Language” tab contain when you are already at language level? There are different actions at each level and that's not going to change.

yilmazdurmaz:
Hi @nijel,
I am not sure what is the intention of the OP here, but I have a similar idea for that.

While we are already in Language, having the list of "other" languages as well would prove benefical to switch

comradekingu:
@nijel 
The problem occurs when I am in a specific language.
When I am in https://hosted.weblate.org/projects/openrefine/#languages I can click nb_NO,
but then I am in 
https://hosted.weblate.org/languages/nb_NO/openrefine/
and I can't click "Language" again.

If I was able to click "Language", I would expect it to go back to
https://hosted.weblate.org/projects/openrefine/#languages

That means added functionality, and the first 6 entries in the menu are always in the same place.
I see no downside to it.

yilmazdurmaz:
Yep, we want very similar things here.

the "language" button in a project links to `{host}/projects/{project_name}/#languages`

If I am not missing something, just simply adding the same button when under `{host}/languages/{lang_code}/{project_name}` should do the job.

for me, it is the same button but for a component pages at `{host}/projects/{project_name}/{component_name}/{lang_code}` to have a link to `{host}/projects/{project_name}/{component_name}/#translations`

need I open a new issue, or is it ok to merge these two requests?

nijel:
> That means added functionality, and the first 6 entries in the menu are always in the same place.

These entries are always context-specific. This is not a place for navigation to other contexts.

>  I see no downside to it.

I do see downsides here.

> If I am not missing something, just simply adding the same button when under {host}/languages/{lang_code}/{project_name} should do the job.

[...]

PR #9561, opened 2023-07-13
Update language-project.html to include "Language" tab

yilmazdurmaz:
Adding "Language" tab as per #9532

This is to switch the language of a project faster from tabs and also supply consistency to the tab bar.

Since it is now a natural part of the tabs, it should help speed up checking language statuses for project owners and contributors.

## Checklist


nijel:
See https://github.com/WeblateOrg/weblate/issues/9532#issuecomment-1635354832

PR #9563, opened 2023-07-14
docs: merge instructions for upgrading from older releases

nijel:
## Proposed changes

The process is always the same and will be the same for 5.0 as well, so let's document it just once.

## Checklist

### DIFF
--- a/docs/admin/upgrade.rst
+++ b/docs/admin/upgrade.rst
@@ -114,7 +114,7 @@ Version specific instructions
 -----------------------------
 
 Upgrade from an older major version
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Upgrades across major versions are not supported. Always upgrade to the latest
 patch level for the initial major release. Upgrades skipping this step are not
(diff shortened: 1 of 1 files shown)


orangesunny (approved):

docs/admin/upgrade.rst
```diff
@@ -113,18 +113,18 @@ work, but is not as well tested as single version upgrades.
 Version specific instructions
 -----------------------------
 
-Upgrade from older major version
+Upgrade from an older major version
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
nijel:
```suggestion
Upgrade from an older major version
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

docs/admin/upgrade.rst
```diff
@@ -113,26 +113,22 @@ work, but is not as well tested as single version upgrades.
 Version specific instructions
 -----------------------------
 
-Upgrade from 2.x
-~~~~~~~~~~~~~~~~
+Upgrade from an older major version
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
nijel, line 117:
```suggestion
Upgrade from an older major version
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Issue #9594, opened 2023-07-18
doc: Inconsistent Python version requirements


surli:
### Describe the issue

The requirements for Python says 3.6 or newer in the doc (here: https://docs.weblate.org/en/latest/admin/install.html#software-requirements) while when checking migration instructions I noticed that it's written that support has been dropped for Python 3.7 in Weblate 4.18 (here: https://docs.weblate.org/en/weblate-4.18.2/admin/upgrade.html#upgrade-from-4-17-to-4-18)

So I believe that now Python 3.8 or newer is required?

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar issues in this repository.

### Steps to reproduce the behavior

Just read the doc :)

### How do you run Weblate?

weblate.org service

nijel:
Should be changed to 3.9 for the upcoming release.

ghubrakesh:
Hello there @surli, As far as I have understood, I will just need to change the requirements to 3.9 or newer version of python on both the pages. Is that right?

nijel:
Yes, it's just the documentation that needs update.

PR #9600, opened 2023-07-19
Update install.rst #9594

ghubrakesh:
Edited mistake in required python version. it's changed from 3.6 to 3.9 as addressed in issue #9594


nijel:

PR #9605, opened 2023-07-20
CACHE_DIR documentation

### LINKED ISSUE #9601: doc: inconsistent path for serving static files
### Describe the issue

There is an inconsistency in the documentation regarding serving the static files.
In `settings_example.py` there is since 4.16:
`STATIC_ROOT = os.path.join(CACHE_DIR, "static")`

Before 4.16 this value was:
`STATIC_ROOT = os.path.join(DATA_DIR, "static")`

The impact of this change is that all documentation talking about serving static file should point out that those files are coming from the CACHE_DIR, e.g. in https://docs.weblate.org/en/latest/admin/install.html#serving-static-files and in particular in https://docs.weblate.org/en/latest/admin/install.html#sample-configuration-for-nginx-and-uwsgi where the paths are now wrong. 

I just had the issue when performing an upgrade: suddenly my static files were not properly served anymore. I had to change my nginx configuration to fix the problem. 

### I already tried

Upgrade weblate from 4.15 to 4.16 and apply the changes in settings_example.py

### Expected behavior

_No response_

### Screenshots

_No response_

### Exception traceback

_No response_

### How do you run Weblate?

weblate.org service

### Weblate versions

_No response_

### Weblate deploy checks

_No response_

### Additional context

_No response_

nijel:
## Proposed changes

* document content of CACHE_DIR
* adjust related documentation to cover change in 4.16, fixes #9601

## Checklist

### DIFF
--- a/docs/admin/config.rst
+++ b/docs/admin/config.rst
@@ -278,6 +278,17 @@ network filesystem.
 
 The Docker container uses a separate volume for this, see :ref:`docker-volume`.
 
+The following subdirectories usually exist:
+
+:file:`fonts`
+   :program:`font-config` cache for :ref:`fonts`.
+:file:`avatar`
+   Cached user avatars, see :ref:`avatars`.
+:file:`static`
+   Default location for static Django files, specified by :setting:`django:STATIC_ROOT`. See :ref:`static-files`.
+:file:`tesseract`
+   OCR trained data for :ref:`screenshots`.
+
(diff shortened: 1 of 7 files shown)


surli:
It's probably worth adding also a mention in upgrade from 4.15 to 4.16 to be careful with this.

nijel:
I've added mention to the upgrade instructions as well.

surli:
LGTM

Issue #9486, opened 2023-06-30
Notifications issue


ib:
### Describe the issue

@nijel This is a addition to #9464.

I still don't understand why the "Translated string" notification is nothing but a subset of "Changed string".

The usual translation workflow is: (1) "New strings to translate" → (2) "New translation" or "Suggestion accepted" → (possibly later) (3) "Translation changed".

There are options to be notified separately for (1) and (3), but not for (2).

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar issues in this repository.

### Steps to reproduce the behavior

Select "Translated string" and "Changed string" in the user settings.

### Expected behavior

I would expect there to be a notification option that informs on first-time, new translations only.

### How do you run Weblate?

weblate.org service

nijel:
"Translated string" is indeed subset of "Changed string". It covers only changes done by translator in Weblate. It excludes changes done in the repository, automatic translation, or state changes, so it is less noisy. Similarly there is "Approved string" notifications, which is subset of "Changed string", but only triggered on strings approval with the review workflow.

ib:
My point is that there is no notification setting for (2).

nijel:
“Translated string” is actually triggered on both “Translation changed” and “New translation”. There is indeed no separate notification for accepting suggestions.

ib:
(2) isn't about "Suggestion accepted".

What I miss is a notification for a previously untranslated string that got translated for the first time, which can happen either by a manual translation or an accepted suggestion. This is what I have labeled (2).

I thought that "Translated string" meant exactly this (but I was wrong).

You can close this issue if you feel that workflow steps (1), (2) and (3) do not require separate notifications.

nijel:
I think it makes sense to include suggestion accepted action in the translated string notification. I'd rather not introduce another notification choice.

Geeyun-JY3:
> I think it makes sense to include suggestion accepted action in the translated string notification. I'd rather not introduce another notification choice.

If implemented, I think a more specific name for the “Translated string” notification is needed.

ib:
@nijel Thank you, I appreciate your efforts and can understand not introducing a new notification choice.

However, I think that the current notification labels might be confusing to users.

[...]

PR #9727, opened 2023-08-11
notifications: Improve naming

nijel:
## Proposed changes

- there is no UI reason to stick with really short names here, so make them more descriptive
- use pgettext to make sure these are translated differently from change events

See also #9486

## Checklist


nijel:
Current ones:

image

Proposed changes:

image

orangesunny (approved):
I would go with my changes. Checking them rendered once again might be useful, though.

weblate/accounts/notifications.py
```diff
@@ -446,8 +447,9 @@ class RepositoryNotification(Notification):
         Change.ACTION_REBASE,
         Change.ACTION_MERGE,
     )
-    # Translators: Notification name
-    verbose = gettext_lazy("Repository operation")
+    verbose = pgettext_lazy(
+        "Notification name", "Successful operation of the repository"
+    )
```
nijel, line 452: This is any operation on the repository, so perhaps:

```suggestion
    verbose = pgettext_lazy(
        "Notification name", "Operation of the repository"
    )
```

PR #9847, opened 2023-09-01
Font fixes for bar chart widgets

JanRei:
## Proposed changes

I have noticed some problems with the bar chart widgets generated by Weblate:
- The font "Source Sans Pro" was replaced by "Source Sans 3" and moved to a subdirectory. However, the widgets still try to use the old font from the old path.
- My browser (Firefox) did not recognize the specified fallback ``Sans``. As a result, the serif font "Times New Roman" was used. This seems to be unintentional.

In this PR, I updated the SVG templates. The font CSS should now be imported from the correct path. And I changed the fallback to ``sans-serif`` which is also used in other places.

## Checklist

### DIFF
--- a/weblate/templates/svg/multi-language-badge-horizontal.svg
+++ b/weblate/templates/svg/multi-language-badge-horizontal.svg
@@ -8,7 +8,7 @@
 <a xlink:href="{{ translation.6 }}" xlink:title="{{ translation.0 }}">
 <text
    xml:space="preserve"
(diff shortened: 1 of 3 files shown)


m-aciek:
For the reference:
* `vendor` directory move commit: https://github.com/WeblateOrg/weblate/commit/e2ed5d3326380750e1d76e11444435ca4b34ec35
* https://github.com/WeblateOrg/weblate/pull/6321

m-aciek (approved):

nijel:

Issue #9854, opened 2023-08-25
Support `WEBLATE_IP_PROXY_OFFSET` Environment Variable in Container


michael-smt:
### Describe the issue

There is no `WEBLATE_IP_PROXY_OFFSET` environment variable support to customize the [IP_PROXY_OFFSET](https://docs.weblate.org/en/latest/admin/config.html#std-setting-IP_PROXY_OFFSET) configuration.

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar issues in this repository.

### Steps to reproduce the behavior

1. Launch Weblate Container with `WEBLATE_IP_PROXY_OFFSET`.
2. Observe no change in behaviour

### Additional context

When running behind a AWS Application Loadbalancer (ALB) the only options [supported](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/x-forwarded-headers.html#x-forwarded-for) are to either "preserve", "extend" or "remove" the `X-Forwarded-For` header.

So to get the IP that has connected to the ALB passed to Weblate the "extend" option has to be used. This adds the IP to the **end** of any already existing data in `X-Forwarded-For` as seen in the example table in the [AWS docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/x-forwarded-headers.html#x-forwarded-for). 

This header can already contain arbitrary (malicious) content from clients connecting to the ALB. The Weblate container currently parses the _first_ address (`IP_PROXY_OFFSET` is set to `0`). This can lead to bogus address values and also generate "Enter a valid IPv4 or IPv6 address." exceptions from the Django core validator `validate_ipv46_address()`.

Therefor in our AWS setup I'd like to set `WEBLATE_IP_PROXY_OFFSET=-1`.

If making this configurable via environment variable is not desired I'd like to propose changing the default from `1` to `-1` to always use the last and (presumably) most trusted value.

PR #9859, opened 2023-09-04
Proxy: change default offset to take final IP

michael-smt:
Relates to #9854 

## Proposed changes

The `X-Forwarded-For` header is appended to by every proxy between the client and Weblate. The header is split by `,` in `weblate/middleware.py`, then the value at the index set by `IP_PROXY_OFFSET` is passed to Django validators.

Change the default to `-1` to take the last entry in the list of addresses. This is the proxy closest to Weblate. Which is assumed to be the most trusted when `IP_BEHIND_REVERSE_PROXY` is enabled.

## Checklist

## Other information
The MDN web docs also [recommends](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#security_and_privacy_concerns) using the final value(s):

> If any trusted reverse proxies are between the client and server, the final `X-Forwarded-For IP` addresses (one for each trusted proxy) are trustworthy, as they were added by trusted proxies. 
> [...]
>  Conversely, leftmost (untrusted) values must only be used where there will be no negative impact from the possibility of using spoofed values.


nijel:
Can you please document the change as it is possibly breaking one? In docs/changes.rst and in docs/admin/config.rst.

nijel:

PR #9864, opened 2023-09-05
ui: Add approved percentage to the object listing

### LINKED ISSUE #9855: Add Approved Percentage as a Column in the Object List
### Describe the problem

Currently in the object list (for projects, components & languages) there is no percentage shown that represents the blue bar.

It is cumbersome to collect a overview on the approval percentage over multiple languages or components.

Similar to the "Translated" column there would also be a "Approved" column with the same behaviour (percentage/green checkmark).

Re-use the existing "Translated" column and rename it to "Translated/Approved". When approval is applicable, replace the values in the "Translated" column with the approval percentage. Or show two percentage values. While this saves space, readability suffers and can become confusing.

Making the columns customizeable seem like much too much effort. There is still enough horizontal space left in desktop mode that a additional column would fit.

### Screenshots

Screenshot

### Additional context

In our translation projects we use the precentage of approved strings as a quality gateway for two things:
- When using a intermediate language file, the main translation work shall only start when the monolingual base file's approval is over a certain percentage
- A language shall only be intergrated into the software when it's approval is over a certain percentage (similar to #3436).

michael-smt:
## Proposed changes

Show approved percentage in the object listing to have a number corresponding to the blue bar.

Fixes #9855

## Checklist

## Other information

The new column is displayed in front of the translation percentage because when approvals are active they probably become the main progress criteria.

On a smaller display resolution it will be shown in third priority after the name and 'Translated' column.
The other columns appear 100px later than before, except 'Unfinished Characters' which appears after 'Comments' at an additional 100px instead of 200px. This allows 'Untranslated' to show at the usual 1400px, so on common 1440px wide screens no information is lost.

### DIFF
--- a/weblate/templates/snippets/list-objects.html
+++ b/weblate/templates/snippets/list-objects.html
@@ -46,6 +46,16 @@
             </a>
           {% endif %}
         </th>
+        <th title="{% trans "Sort this column" %}" class="number zero-width-640 sort-cell">
+          {% if objects.paginator.num_pages > 1 %}
+            <a href="?page={{ objects.number }}&amp;sort_by={% if objects.paginator.sort_by == "approved" %}-{% endif %}approved">
+          {% endif %}
+          {% trans "Approved" %}
+          <span class="sort-icon {% if objects.paginator.sort_by == "approved" %}sort-down{% elif objects.paginator.sort_by == "-approved" %}sort-up{% endif %}" />
+          {% if objects.paginator.num_pages > 1 %}
+            </a>
+          {% endif %}
+        </th>
         <th title="{% trans "Sort this column" %}" class="number sort-cell">
           {% if objects.paginator.num_pages > 1 %}
             <a href="?page={{ objects.number }}&amp;sort_by={% if objects.paginator.sort_by == "translated" %}-{% endif %}translated">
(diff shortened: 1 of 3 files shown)


weblate/templates/snippets/list-objects.html
```diff
@@ -46,6 +46,16 @@
...
+        <th title="{% trans "Sort this column" %}" class="number zero-width-640 sort-cell">
+          {% if objects.paginator.num_pages > 1 %}
+            <a href="?page={{ objects.number }}&amp;sort_by={% if objects.paginator.sort_by == "approved" %}-{% endif %}approved">
+          {% endif %}
+          {% trans "Approved" %}
+          <span class="sort-icon {% if objects.paginator.sort_by == "approved" %}sort-down{% elif objects.paginator.sort_by == "-approved" %}sort-up{% endif %}" />
+          {% if objects.paginator.num_pages > 1 %}
+            </a>
+          {% endif %}
+        </th>
```
nijel, line 58: This column should be shown only if reviews are turned on, it will show just useless zeroes when turned off.

nijel:
Having it first makes more sense UX wise, it just needs to adjust the CSS logic to avoid hard-coding widths.

nijel:
Closing in favor of https://github.com/WeblateOrg/weblate/pull/9936.

PR #9880, opened 2023-09-07
ui: Clarify menu item for moving

nijel:
## Proposed changes

This hides both renaming and moving to different project or category, so the menu item should be clearer here.

See #9807

## Checklist

### DIFF
--- a/weblate/templates/trans/delete-form.html
+++ b/weblate/templates/trans/delete-form.html
@@ -1,6 +1,5 @@
 {% load i18n %}
 {% load crispy_forms_tags %}
-<div class="tab-pane" id="delete">
 
 <form method="post" action="{% url "remove" path=object.get_url_path %}">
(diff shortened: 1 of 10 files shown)


nijel:
This is where it points to:

image

orangesunny:
Let's join it with Remove menu and call the menu item 
**Organize or remove**

nijel:
Good idea, I've changed the PR to implement that.

m-aciek (approved):

weblate/templates/trans/delete-form.html
```diff
@@ -1,6 +1,5 @@
 {% load i18n %}
 {% load crispy_forms_tags %}
-<div class="tab-pane" id="delete">
```
m-aciek, line 3: nit: this newline seems superfluous now

Issue #9903, opened 2023-09-10
Improve other units lookup speed


nijel:
### Describe the problem

Loading of "Other units" tab currently takes most of the time on translating strings what is not acceptable.

### Describe the solution you'd like

Improve the lookup performance.

### Describe alternatives you've considered

Load this tab on demand similarly as we do with other ones which take longer to load.

### Additional context

It can be seen measured at https://sentry.weblate.org/organizations/weblate/performance/weblate:716f79e6d30342e78770ec8aa0b711d5/, where this took 3 seconds for https://hosted.weblate.org/translate/weblate/application/tok/

74e62740a9a3add3b4e644d9efd147d71609feda improves the situation a bit as it reduces pressure on the database and 2e54bbb7351bc069fb063e4d7b1b14f00de13ddd adds sentry span for this operation so that it can be measured together with Python processing.

tbodt:
I'm curious what the sentry measurement looks like, could you post a screenshot?

nijel:
This is the relevant part:

![image](https://github.com/WeblateOrg/weblate/assets/212189/ac0bf151-b50c-49eb-8e5b-e951668b08ad)

tomkolp:
> @tomkolp [0c9b631](https://github.com/WeblateOrg/weblate/commit/0c9b631fb1de1a15af20cce080e202aa71042b46) could improve this as well, looking forward for outcome in your env.

It looks like it's working. I did a dozen or so tests, taking the average of loading 20 strings. The result is 2.6s - 3.4s. Previously it was an average of 3.25s per string.

tomkolp:
https://github.com/WeblateOrg/weblate/commit/8c9f3c896ac6545507f70dcf885784cf5e4d35da
This correction gave me a result of 2.35s today

nijel:
Looking at metrics, other units lookup is now there under 100ms, so I don't see much room for improvement here.

tomkolp:
> Looking at metrics, other units lookup is now there under 100ms, so I don't see much room for improvement here.

Everything is OK. The current loading time of the string is satisfactory.

tbodt:
Have the changes been deployed to hosted.weblate.org?

nijel:
Not yet, it will be updated during scheduled maintenance this Saturday.

PR #9909, opened 2023-09-11
 translate: improve other units performance 

nijel:
## Proposed changes

This seems to perform better in many situations. But needs more testing.

See #9903

## Checklist

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -7,6 +7,8 @@ Not yet released.
 
 **Improvements**
 
+* Translate page performance.
+
 **Bug fixes**
 
 **Compatibility**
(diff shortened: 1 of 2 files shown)


nijel:
@tomkolp I would appreciate it if you could check the performance impact of this on your server. On some test data, this heavily improves performance, while on others it is about same as the original code, but I definitely do not want to merge something what makes it perform worse in some situations.

tomkolp:
> @tomkolp I would appreciate it if you could check the performance impact of this on your server. On some test data, this heavily improves performance, while on others it is about same as the original code, but I definitely do not want to merge something what makes it perform worse in some situations.

I'll check as soon as the new docker image appears.
Currently, the next string loads on average in 3.25s, I don't know if I will be able to measure the improvement (if it is minimal).
Alternatively, when I upload the amendment, you can check it in sentry (I don't have access because I didn't receive any invitation via e-mail)

tomkolp:
@nijel Now it's slower: 4.65s (previously 3.25s)

PR #9934, opened 2023-09-13
correction

emmapeel2:
this is the url that works for me, for example at https://hosted.weblate.org/api/components/tor/tor-browser%252Ftb-aboutdialogdtd/statistics/


nijel:

PR #9936, opened 2023-09-13
ui: Add approved percentage to the object listing (v2)

michael-smt:
## Proposed changes

Show approved percentage in the object listing to have a number corresponding to the left most bar when reviews are enabled.

Only shows the column if reviews are enabled in the project.

A simpler implementation compared to MR #9864.

Resolves WeblateOrg#9855

## Checklist

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -6,6 +6,7 @@ Not yet released.
 **New features**
 
 **Improvements**
+* :ref:`project-translation_review` also shows the approval percentage in object listings.
 
 **Bug fixes**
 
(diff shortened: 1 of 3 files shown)


docs/changes.rst
```diff
@@ -29,6 +29,7 @@ Released on September 14th 2023.
 
 * Translate page performance.
 * Search now looks for categories as well.
+* :ref:`project-translation_review` also shows the approval percentage in object listings.
```
nijel, line 32: This needs to be moved to the 5.1 section now.

PR #9950, opened 2023-09-14
ui: add option to hide linked repositories in admin repo status page

michael-smt:
When there are a lot of linked repositories it can become difficult to get a overview of the status of the real repos.

## Proposed changes

Adds a basic checkbox that toggles the display of linked repositories in pure CSS.

## Checklist

## Other information

Input needs to be on same level as components to make the CSS selector work. Because there is no way in CSS to select a parent element.


nijel:

Issue #9786, opened 2023-08-21
Replace misaka with mistletoe


nijel:
### Describe the problem

mistletoe is used for parsing Markdown in translate-toolkit so it is requirement we can not avoid. misaka is used internally by Weblate. So there are now two Markdown parsers required to run Weblate.

### Describe the solution you'd like

Replace usage of misaka with mistletoe. It should not be hard, our customization is pretty minimal: https://github.com/WeblateOrg/weblate/blob/main/weblate/utils/markdown.py

### Additional context

https://github.com/FSX/misaka:

> Misaka is not being maintained anymore. I've not been able to find time to to properly maintain this project

itlallou:
Hello!I'd like to contribute to this

nijel:
@itlallou You're welcome!

subhajit20:
@nijel Hey this has been open for 3 weeks.
So I am taking this and coming up with a PR.

nijel:
You are welcome!

subhajit20:
@nijel Made a pr.
Can you check and let me know for further corrections needed?

nijel:
Look at https://github.com/miyuchina/mistletoe documentation and reimplement existing behavior using it.

Maddy0027:
@nijel if @subhajit20 is not working on this issue i would like to work on it

JanMalte:
> @nijel if @subhajit20 is not working on this issue i would like to work on it

You are welcome to make a new PR by yourself. Just mention this issue in the description, so it will get linked here.

PR #9951, opened 2023-09-14
Replaced misaka with mistletoe

subhajit20:
## Proposed changes
Replaced misaka with mistletoe

Issue - #9786

### DIFF
--- a/weblate/utils/markdown.py
+++ b/weblate/utils/markdown.py
@@ -5,7 +5,7 @@
 import re
 from functools import reduce
 
-import misaka
+import mistletoe
 from django.db.models import Q
 from django.utils.safestring import mark_safe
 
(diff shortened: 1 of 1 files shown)


weblate/utils/markdown.py
```diff
@@ -24,7 +24,7 @@ def get_mention_users(text):
     )
 
 
-class WeblateHtmlRenderer(misaka.SaferHtmlRenderer):
+class WeblateHtmlRenderer(mistletoe.SaferHtmlRenderer):
```
nijel, line 27: The API is different, it will need more work than just replacing library name...

subhajit20: do I have to use HtmlRenderer which comes from mistletoe itself instead of using SaferHtmlRenderer?

nijel: Yes, but it will probably need additional tweaking to behave same - it doesn't allow HTML to be used in the markup.

PR #9958, opened 2023-09-15
Add "nb" locale mapping to Google Translate

raymondelooff:
## Proposed changes

Currently only the "nb_NO" (locale with country code) is mapped to "no", but no mapping exists for just the locale "nb". This "nb" locale doesn't exist in Weblate by default, but can be created through the Django admin interface via the "Languages" section. Because of this, the Google Translate automatic translation provider currently doesn't recognize the "nb" locale. This change will add a mapping for the "nb" locale as well.

## Checklist


nijel:

PR #9990, opened 2023-09-19
settings: add support for multiple vcs api host credentials from env var

### LINKED ISSUE #8132: docker: Enable WEBLATE_GITLAB_CREDENTIALS environment variable
### Describe the problem

Right now it seems I can use gitlab_username and gitlab_token variables. But when I try to use gitlab_credentials:

>       WEBLATE_GITLAB_CREDENTIALS: "git.duniter.org": {username: weblate,token: XXXXXXXXXXXXXXX}

I get this error:

> ERROR: yaml.scanner.ScannerError: mapping values are not allowed here
>   in "./docker-compose.override.yml", line 17, column 52

Add weblate_gitlab_credentials support

_No response_

### Screenshots

_No response_

### Additional context

_No response_

michael-smt:
## Proposed changes

Resolves #8132

## Checklist

## Other information

The env variable is a "Python dict as a string" because it most closely matches the [docs](https://docs.weblate.org/en/latest/admin/config.html#gitlab-credentials). Could also be JSON or YAML instead.

### DIFF
--- a/docs/admin/install/docker.rst
+++ b/docs/admin/install/docker.rst
@@ -751,6 +751,7 @@ Generic settings
 .. envvar:: WEBLATE_GITHUB_USERNAME
 .. envvar:: WEBLATE_GITHUB_TOKEN
 .. envvar:: WEBLATE_GITHUB_HOST
+.. envvar:: WEBLATE_GITHUB_CREDENTIALS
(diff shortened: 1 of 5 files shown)


nijel:

Issue #8382, opened 2022-11-25
False postives in checks while source has plural forms


Geeyun-JY3:
### Describe the issue

It seems that checks are currently applied between each plural form of translation and the first plural form of the source.
It produces many false positives.

Only tested in projects whose source language is English.

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar issues in this repository.

### Steps to reproduce the behavior

1. Go to a string whose source has plural forms
2. Enable some checks
3. See the false positive checks

### Expected behavior

There is no false positive on string whose source has plural forms.

I think checks should be applied between each plural form of translation and the corresponding plural form of the source if available.
Not sure if the matching of plural form between source and target language would be complicated.

### Screenshots

Example 1:

- Source language (English): https://hosted.weblate.org/translate/koreader/koreader/en/?checksum=3a041d774ef3f333#translations
  ![image](https://user-images.githubusercontent.com/49396421/203926605-c3de2de0-b57d-4d83-b04b-d4e0b364c097.png)
- Target language 1 (Chinese (Simplified)): https://hosted.weblate.org/translate/koreader/koreader/zh_Hans/?checksum=3a041d774ef3f333#translations
  ![image](https://user-images.githubusercontent.com/49396421/203927107-3ab73aea-e5c7-40b8-bc39-af1cd865ebee.png)
- Target language 2 (Hungarian): https://hosted.weblate.org/translate/koreader/koreader/hu/?checksum=3a041d774ef3f333#translations
  ![image](https://user-images.githubusercontent.com/49396421/203940350-fc94f667-3832-4db3-a538-cba80775a6a3.png)
- Target language 3 (Moldavian): https://hosted.weblate.org/translate/koreader/koreader/ro_MD/?checksum=3a041d774ef3f333#translations
  ![image](https://user-images.githubusercontent.com/49396421/203936872-85013948-6ce4-4a7b-8110-e08ab0a5c677.png)

Example 2:

- Source language (English): https://hosted.weblate.org/translate/liberapay/core/en/?checksum=a947b1502cd1e4d6#translations
  ![image](https://user-images.githubusercontent.com/49396421/203938274-7a98414d-efff-418b-b183-6f7385114ba0.png)
- Target language 1 (Chinese (Simplified)): https://hosted.weblate.org/translate/liberapay/core/zh_Hans/?checksum=a947b1502cd1e4d6#translations
[...]

PR #10022, opened 2023-09-24
Improve accuracy of checks on plural messages

Changaco:
## Proposed changes

This commit modifies the `Check.check_target_unit()` method to improve how source and target strings are paired for comparison, which should reduce both false positives and false negatives. It fixes <https://github.com/liberapay/salon/issues/581>.

## Checklist

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -7,9 +7,11 @@ Not yet released.
 * :ref:`mt-yandex-v2` machine translation service.
 
 **Improvements**
+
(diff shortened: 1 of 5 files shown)


Geeyun-JY3:
Related: #8382

Changaco:
I'm not sure about the impact of this change on performance, because I don't know if the `PluralMapper.instances` cache is effective. If it isn't, then a lot of CPU cycles are wasted on doing the same work over and over.

nijel:

PR #10077, opened 2023-10-01
refactor: Utilize expression index for auth

### LINKED ISSUE #9973: Utilize expression index for auth
### Describe the problem

Weblate does manually create index using `CREATE UNIQUE INDEX` during the `weblate.auth` model migration.

This should be changed to use Django standard expression index `UniqueConstraint(...Upper(...))` instead for better readability and portability.

_No response_

### Screenshots

_No response_

### Additional context

Need Django 4.0+ (Weblate already requires 4.2).

See also https://stackoverflow.com/a/69752763/225718 and https://stackoverflow.com/a/66966391/225718

mohebmithani:
- resolves #9973 
- Have used Django standard expression index UniqueConstraint(...Upper(...)) for username and email fields of User table. I've removed manual index creation from migration.

## Checklist


nijel (commented):
I don't think the migrations will work this way. Existing installations have index created manually and you add another migration to create it. If both indexes are really the same, you should replace the `RunPython` step with `AddConstraint` one. If Django generated inxex expression is different, just mark the `RunPython` step as elidable (so that it can be removed next time the migrations are squashed) and add a migration to execute `DROP INDEX` manually before re-creating them.

mohebmithani:
> I don't think the migrations will work this way. Existing installations have index created manually and you add another migration to create it. If both indexes are really the same, you should replace the `RunPython` step with `AddConstraint` one. If Django generated inxex expression is different, just mark the `RunPython` step as elidable (so that it can be removed next time the migrations are squashed) and add a migration to execute `DROP INDEX` manually before re-creating them.

Okay i'll look into it and fix it. Thanks for the review

nijel:
You can see the migration failure for existing installs at https://github.com/WeblateOrg/weblate/actions/runs/6372487199/job/17304733816?pr=10077

mohebmithani:
> I don't think the migrations will work this way. Existing installations have index created manually and you add another migration to create it. If both indexes are really the same, you should replace the `RunPython` step with `AddConstraint` one. If Django generated inxex expression is different, just mark the `RunPython` step as elidable (so that it can be removed next time the migrations are squashed) and add a migration to execute `DROP INDEX` manually before re-creating them.

@nijel I've implemented your first suggestion.

nijel:

mohebmithani:
> Merged, thanks for your contribution!

Thanks.

Issue #7604, opened 2022-05-06
Web Hooks


VygantasOBDeleven:
### Describe the problem

The issue that I've encountered is due to the complexity of systems that integrates with the Weblate through the REST API. Right now we need to go through every single translation in the project to sync with the external system.

### Describe the solution you'd like

The solution could be the implementation of WebHooks. Every hook could have a type which could be something like "Translation Added", "Translation Approved", "Suggestion Added", "Suggestions Accepted", etc. On every hook type, Weblate could call a configured endpoint with a specific payload of the change.

In addition to that, there could be a configurable retry mechanism and request batching.

nijel:
Things to decide before we start implementing this:

* Should webhooks be configured at component, project, or both?
* What all should be configured for webhooks (GitHub has URL, payload type (form/JSON), secret, SSL verification, event selection)?
* Should we log deliveries?
* Do we want to use add-ons for this or implement webhooks standalone?


There is also https://github.com/WeblateOrg/fedora_messaging/ which is slightly related (it allows notification of Weblate events to external systems).

ParthS007:
_discussion dump_

- Keep in mind:
  - It should be generic
  - Will be extended in future

- Implementation
  -  > Should webhooks be configured at component, project, or both?
     - Both  
  - Hooks UI would be inspired from Github UI
     - check addons UI
  - There will be 1 task for scheduling the deliveries and there will be 1 task which will be doing delivery for the event.
  - Webhook Service: `get the content and construct the json and trigger the webhook`
     - check fedora messaging app
  - Final: **Integrations app**
     - Models: similar to addons
     - base.py
     - forms.py

- [ ] Step 1: Add the models.py of `integrations` app

ParthS007:
Moving ahead on issue. 🤞 

- Integration App
   - Integration (There will be json configuration to store the integration config.) <-> Integration Form
   - Integration web hook service
     - Gets the content of change, Schedule the delivery
     - Executing delivery of tasks to external service

- Flow to add integration:
  - Fill the Integration Form to watch for changes.
  
- Flow to update/delete integration:
  - Change in integration form - Triggers `store_change` in Integration - (Change/Remove) -> Integration config changed -> Integration service trigger -> get and process trigger data -> scheduling payload to external API which contains changed `ACTION`

nijel:
[...]

PR #10100, opened 2023-10-04
addons: enable project-wide addons

### LINKED ISSUE #10051: Project-wide installed add-ons
### Describe the problem

In many cases, it is desired to install add-on on a project level. Especially with a lot of components, it is cumbersome to keep the add-on configuration in sync.

* Add `project` foreign key field to the `Addon` model
* Adjust `AddonQuerySet.filter_component` to filter based on the project
* Change all `project_scope` add-on instances have a valid project reference and make sure that `project_scope` flagged add-ons are installed that way
* Remove `project_scope` from the database model (once existing instances are migrated)
* Add project level UI for add-ons

_No response_

### Screenshots

_No response_

### Additional context

See discussion at https://github.com/WeblateOrg/weblate/issues/7604 and https://github.com/WeblateOrg/weblate/discussions/9937

### LINKED ISSUE #10052: Site-wide add-ons
### Describe the problem

With upcoming external notification add-ons (like https://github.com/WeblateOrg/weblate/issues/7604 or https://github.com/WeblateOrg/fedora_messaging/), it will be useful to install them site-wide.

* Allow null value for `component` and `project` (see https://github.com/WeblateOrg/weblate/issues/10051) fields on the `Addon` model
* Adjust `AddonQuerySet.filter_component` to include add-ons with both being null
* Add site level UI for add-ons at /manage

_No response_

### Screenshots

_No response_

### Additional context

See discussion at https://github.com/WeblateOrg/weblate/issues/7604 and https://github.com/WeblateOrg/weblate/discussions/9937

ParthS007:
closes #10051 
closes #10052 

## Proposed changes

#10051 

- [x] Add project foreign key field to the Addon model
- [x] Adjust AddonQuerySet.filter_component to filter based on the project.
- [x] Change all `project_scope` add-on instances have a valid project reference and make sure that project_scope flagged add-ons are installed that way.
- [x] Remove `project_scope` from the database model.
- [x] Add project level UI for add-ons

#10052 

- [x] Allow null value for `component` and `project`fields on the Addon model.
- [x] Adjust `AddonQuerySet.filter_component` to include add-ons with both being `null`
- [x] Add site level UI for add-ons at /manage

## Checklist

Related to #7604 

- #10051 
Screenshot 2023-10-09 at 11 29 55
Screenshot 2023-10-09 at 12 26 17
 

- #10052 

Screenshot 2023-11-11 at 14 33 18

Screenshot 2023-11-11 at 14 38 29


weblate/addons/models.py
```diff
@@ -48,12 +48,20 @@
...
+            | (
+                Q(component__linked_component=component)
+                & Q(repo_scope=True)
+                & Q(project=project)
+            )
+            | (
+                Q(component=component.linked_component)
+                & Q(repo_scope=True)
+                & Q(project=project)
+            )
```
nijel, line 64: The final query here should be:

```suggestion
            Q(component=component)
            | Q(project=project)
            | (Q(component__linked_component=component) & Q(repo_scope=True))
            | (Q(component=component.linked_component) & Q(repo_scope=True))
```

This needs dropping `project_scope` and using `project` properly instead.

* `component` would be NULL for project scoped add-ons
* `project` would be NULL for component scoped add-ons

weblate/addons/models.py
```diff
@@ -62,9 +70,11 @@ def filter_event(self, component, event):
 
 class Addon(models.Model):
     component = models.ForeignKey(Component, on_delete=models.deletion.CASCADE)
```
nijel, line 72:
```suggestion
    component = models.ForeignKey(Component, on_delete=models.deletion.CASCADE, null=True)
```

weblate/addons/base.py
```diff
@@ -116,12 +116,14 @@ def configure(self, settings):
 
     def post_configure(self, run: bool = True):
         component = self.instance.component
+        project = self.instance.project
 
         # Configure events to current status
         component.log_debug("configuring events for %s add-on", self.name)
+        project.log_debug("configuring events for %s add-on", self.name)
```
nijel, line 123: These can be None, so something like this is needed:

```suggestion
        (component or project).log_debug("configuring events for %s add-on", self.name)
```

weblate/addons/base.py
```diff
@@ -116,12 +116,14 @@ def configure(self, settings):
...
         self.instance.configure_events(self.events)
 
-        if run:
+        if run and component:
```
nijel, line 126: This should run on each component from `project.component_set.all()` in case of project wide.

weblate/addons/models.py
```diff
@@ -48,24 +48,32 @@
...
 
 class AddonQuerySet(models.QuerySet):
-    def filter_component(self, component):
-        return self.prefetch_related("event_set").filter(
-            (Q(component=component) & Q(project_scope=False))
-            | (Q(component__project=component.project) & Q(project_scope=True))
-            | (Q(component__linked_component=component) & Q(repo_scope=True))
-            | (Q(component=component.linked_component) & Q(repo_scope=True))
-        )
+    def filter_component(self, component=None, project=None):
```
nijel, line 51: This is called filter_**component** so let's stick with filtering components only. Please add a separate method for filtering projects instead.

weblate/addons/models.py
```diff
@@ -81,7 +89,6 @@ def save(
         self, force_insert=False, force_update=False, using=None, update_fields=None
     ):
         cls = self.addon_class
-        self.project_scope = cls.project_scope
```
nijel, line 84: It should set `project` instead of `component` in this case (similar like changing the component for the repo_scope below).

nijel (commented):
There is a missing step in the migrations. Add `project_scope` migrations need to be adjusted to have `project = component.project` and `component = None`.

weblate/addons/migrations/0002_remove_addon_project_scope_addon_project_and_more.py
```diff
@@ -0,0 +1,49 @@
...
+
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+def populate_project_from_component(apps, schema_editor):
+    """Migrate Addons with project_scope from using component to using project."""
+    Addon = apps.get_model("addons", "Addon")
+    for addon in Addon.objects.all():
+        if addon.component and addon.project_scope:
```
nijel, line 13: Not needed, `project_scope` filtering should be done in the datababase (see above suggestion) and component can not be NULL at this point:

```suggestion
```

weblate/addons/models.py
```diff
@@ -81,15 +92,23 @@ def save(
...
         self.repo_scope = cls.repo_scope
+        self.project_scope = cls.project_scope
+
+        if self.project_scope:
+            self.project = self.component.project
+            self.component = None
```
nijel, line 100: Do not set non-existing attribute on a class:

```suggestion
        original_component = None
        if cls.project_scope:
            original_component = self.component
            self.project = self.component.project
            self.component = None
```

weblate/addons/models.py
```diff
@@ -81,15 +92,23 @@ def save(
...
 
         # Reallocate to repository
         if self.repo_scope and self.component.linked_component:
             self.component = self.component.linked_component
```
nijel, line 104:
```suggestion
            original_component = self.component
            self.component = self.component.linked_component
```

weblate/addons/models.py
```diff
@@ -81,15 +92,23 @@ def save(
...
 
         # Clear add-on cache
-        self.component.drop_addons_cache()
+        if self.component:
+            self.component.drop_addons_cache()
+        elif self.project:
+            for component in self.project.component_set.all():
+                component.drop_addons_cache()
```
nijel, line 111: The cache is in-memory only, so only currently affected objects need to be purged (this avoids possibly expensive loading of all components):

```suggestion
        if self.component:
            self.component.drop_addons_cache()
        if original_component:
            original_component.drop_addons_cache()
```

weblate/addons/models.py
```diff
@@ -110,13 +129,23 @@ def get_absolute_url(self):
...
+            )
+        elif self.project:
+            for component in self.project.component_set.all():
+                Change.objects.create(
+                    action=action,
+                    user=self.project.acting_user,
+                    component=component,
+                    target=self.name,
+                    details=self.configuration,
+                )
```
nijel, line 148: Bind this to project instead:

```suggestion
                Change.objects.create(
                    action=action,
                    user=self.project.acting_user,
                    project=project,
                    target=self.name,
                    details=self.configuration,
                )
```

weblate/addons/models.py
```diff
@@ -136,7 +165,12 @@ def delete(self, using=None, keep_parents=False):
         self.store_change(Change.ACTION_ADDON_REMOVE)
         # Delete any addon alerts
         if self.addon.alert:
-            self.component.delete_alert(self.addon.alert)
+            if self.component:
+                self.component.delete_alert(self.addon.alert)
+            elif self.project:
+                for component in self.project.component_set.all():
+                    component.delete_alert(self.addon.alert)
```
nijel, line 172: This is costly as this triggers loading all alerts for all components and there are projects with thousands of components. Delete them in the database instead:

```suggestion
                Alert.objects.filter(component__in=self.project.component_set.all(), name=self.addon.alert).delete()
```

weblate/addons/views.py
```diff
@@ -35,40 +42,64 @@ def redirect_list(self, message=None):
 
     def get_context_data(self, **kwargs):
         result = super().get_context_data(**kwargs)
-        component = self.kwargs["component_obj"]
-        result["object"] = component
+        target = self.kwargs.get("component_obj") or self.kwargs.get("project_obj")
```
nijel, line 45: Why not use `path_object` here instead?

nijel (commented):
Generally, try to avoid project.component_set.all() unless really needed. These can be really expensive (for example, https://weblate.eso-spolszczenie.eu/api/projects/eso-spolszczenie/components/ has nearly 3000 components).

weblate/addons/base.py
```diff
@@ -115,16 +116,21 @@ def configure(self, settings):
...
-            # Flush cache in case this was eager mode
-            component.repository.clean_revision_cache()
+            if isinstance(target, Project):
+                components = target.component_set.all()
+                for component in components:
+                    postconfigure_addon.delay(component.pk)
+                    component.repository.clean_revision_cache()
+            else:
+                postconfigure_addon.delay(target.pk)
+                component.repository.clean_revision_cache()
```
nijel, line 133: `postconfigure_addon` received AddOn instance, not a component instance, so the loop is not needed here. I'm rewriting this code anyway in https://github.com/WeblateOrg/weblate/pull/10149:

```suggestion
            postconfigure_addon.delay(self.instance.pk)
            # Flush cache in case this was eager mode
            component.repository.clean_revision_cache()
```

weblate/addons/migrations/0002_remove_addon_project_scope_addon_project_and_more.py
ParthS007: Thanks for pointing this out 👍

weblate/addons/migrations/0002_remove_addon_project_scope_addon_project_and_more.py
```diff
@@ -0,0 +1,49 @@
...
+        migrations.AlterField(
+            model_name="addon",
+            name="component",
+            field=models.ForeignKey(
+                null=True,
+                on_delete=django.db.models.deletion.CASCADE,
+                to="trans.component",
+            ),
+        ),
+        migrations.RunPython(populate_project_from_component),
```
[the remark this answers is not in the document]
ParthS007: good to know. Interesting, will read more about it.

weblate/addons/views.py
```diff
@@ -35,63 +42,88 @@ def redirect_list(self, message=None):
...
+        elif isinstance(self.path_object, Project):
+            # Do I need to check permissions here for compnenets?
+            result["available"] = sorted(
+                (
+                    x
+                    for x in ADDONS.values()
+                    if x.project_scope and (x.multiple or x.name not in installed)
+                ),
+                key=lambda x: x.name,
+            )
```
nijel, line 68:
```suggestion
            result["available"] = sorted(
                (
                    x
                    for x in ADDONS.values()
                    if (x.multiple or x.name not in installed)
                ),
                key=lambda x: x.name,
            )
```

I think no filtering is needed here. The add-on execution should skip unsupported components, though. All `@receiver` in `weblate/addons/models.py` should get something like this in the loop:

```python
    if addon.project and not addon.addon.can_install(component):
       continue
```

I've made component variable available in all of these in 0df9af9b0963c7fc16f8c7a3244eeb23a3ff896c. This code that would deserve some refactoring as it is nearly the same for all the events, but let's not do it here – https://github.com/WeblateOrg/weblate/issues/10172

weblate/addons/views.py
```diff
@@ -105,26 +137,31 @@ def get_form(self, form_class=None):
...
-        return reverse("addons", kwargs={"path": self.object.component.get_url_path()})
+        target = self.object.component or self.object.project
+        return reverse("addons", kwargs={"path": target.get_url_path()})
 
     def get_object(self):
         obj = super().get_object()
         if not self.request.user.has_perm("component.edit", obj.component):
             raise PermissionDenied("Can not edit component")
+        elif not self.request.user.has_perm("project.edit", obj.project):
+            raise PermissionDenied("Can not edit project")
```
nijel, line 157:
```suggestion
        if obj.component and not self.request.user.has_perm("component.edit", obj.component):
            raise PermissionDenied("Can not edit component")
        if obj.project and not self.request.user.has_perm("project.edit", obj.project):
            raise PermissionDenied("Can not edit project")
```

nijel:
Once it is merged, can you please rebase on https://github.com/WeblateOrg/weblate/pull/10807 so that the changes in the signal handling does not have to be duplicated. Also please look at the test failures.

nijel:
Can you please look at the test failures?

weblate/addons/forms.py
```diff
@@ -459,7 +459,9 @@ class AutoAddonForm(AutoForm, AddonFormMixin):
     def __init__(self, user, addon, instance=None, **kwargs):
         self.user = user
         self._addon = addon
-        super().__init__(obj=addon.instance.component, **kwargs)
+        # send addon instance as obj in case of a project addon
+        obj = addon.instance.component or addon.instance
+        super().__init__(obj=obj, **kwargs)
```
nijel, line 464: `AutoForm` expects `obj` to be a `Component`, passing `AddOn` instead cannot work.

weblate/addons/migrations/0002_remove_addon_project_scope_addon_project_and_more.py
```diff
@@ -0,0 +1,48 @@
...
+from django.db import migrations, models
+
+
+def populate_project_from_component(apps, schema_editor):
+    """Migrate Addons with project_scope from using component to using project."""
+    Addon = apps.get_model("addons", "Addon")
+    for addon in Addon.objects.filter(project_scope=True):
+        addon.project = addon.component.project
+        addon.component = None
+        addon.save(update_fields=["project", "component"])
```
nijel, line 15: Use `bulk_update` for better performance:

```suggestion
    updates = []
    for addon in Addon.objects.filter(project_scope=True):
        addon.project = addon.component.project
        addon.component = None
        updates.append(addon)
        if len(updates) > 1000:
             Addon.objects.bulk_update(updates, ["project", "component"])
             updates = []
    if updates:
         Addon.objects.bulk_update(updates, ["project", "component"])
```

weblate/addons/models.py
```diff
@@ -101,12 +122,22 @@ def get_absolute_url(self):
...
+                details=self.configuration,
+            )
+        elif self.project:
+            Change.objects.create(
+                action=action,
+                user=self.project.acting_user,
+                project=self.project,
+                target=self.name,
+                details=self.configuration,
+            )
```
nijel, line 140: Can be simplified:

```suggestion
        Change.objects.create(
            action=action,
            user=self.component.acting_user,
            project=self.project,
            component=self.component,
            target=self.name,
            details=self.configuration,
        )

```

nijel (commented):
Submitting partial review, I didn't yet get to the views part...

weblate/addons/base.py
```diff
@@ -114,24 +119,36 @@ def configure(self, configuration):
...
         self.instance.configure_events(self.events)
 
         if run:
             if settings.CELERY_TASK_ALWAYS_EAGER:
-                postconfigure_addon(self.instance.pk, self.instance)
+                postconfigure_addon(self.instance.pk)
+                # Flush cache in case this was eager mode
+                if self.instance.component:
+                    # Project does not have repository attribute
+                    target.repository.clean_revision_cache()
```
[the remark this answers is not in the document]
ParthS007: I remember there was an error in CI when flushing the cache in case of project addon. I will recheck just in case.

weblate/addons/base.py
```diff
@@ -72,27 +74,38 @@ def get_identifier(cls):
         return cls.name
 
     @classmethod
-    def create_object(cls, component, **kwargs):
+    def create_object(
+        cls, *, component: Component = None, project: Project = None, **kwargs
```
nijel, line 78: The type hints should indicate this is an optional arg:

```suggestion
        cls, *, component: Component | None = None, project: Project | None = None, **kwargs
```

weblate/addons/base.py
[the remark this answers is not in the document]
nijel: The instance is passed exactly for the purpose of correctly purging caches when not executed in the background.

weblate/trans/forms.py
```diff
@@ -904,12 +904,21 @@ def __init__(self, obj, user=None, *args, **kwargs):
...
+                project__contribute_shared_tm=True,
+            ).exclude(project=obj.project)
+
+        if hasattr(obj, "project") and hasattr(obj.project, "source_language_ids"):
+            self.components = obj.project.component_set.filter(
+                source_language_id__in=obj.project.source_language_ids
+            ) | Component.objects.filter(
+                source_language_id__in=obj.project.source_language_ids,
+                project__contribute_shared_tm=True,
+            ).exclude(project=obj.project)
```
nijel, line 921: This seems weird: a `Component` instance matches both if clauses here, so the first one is effectively never used. Maybe test using `isinstance(obj, Component)` to make the intention more clear?

weblate/trans/models/component.py
```diff
@@ -931,7 +931,8 @@ def install_autoaddon(self):
                 continue
 
             if (
-                addon.project_scope
+                hasattr(addon, "project")
+                and addon.project
```
nijel, line 935: I think this code should have stayed as it was as it doesn't work with a model here.

weblate/urls.py
```diff
@@ -186,6 +186,12 @@
         weblate.lang.views.CreateLanguageView.as_view(),
         name="create-language",
     ),
+    path(
+        # Site-wide add-ons
+        "addons/",
```
nijel, line 191: This should probably go into manage namespace:

```suggestion
        "manage/addons/",
```

weblate/trans/models/component.py
ParthS007: CI was failing for not having project_scope attribute for the class and throwing AttributeError. Added this change by removing current state for clarity and avoiding AttributeError.
```suggestion
            getattr(addon, "project_scope", False)        
```

weblate/trans/forms.py
ParthS007: thanks, Good catch. I have added a check for Component in first if clause. Does it make sense to have `if`, `elif` here since in the case when `Component` obj has `project` attr, it will replace `self.components` from the earlier `if` clause. 

I assume, what we want here in case of `project` addon.

1. Filter the components of the `obj`'s project that have a `source_language_id` in `obj.project.source_language_ids`.

2. The combine the queryset with the one that filters all `Component` objects that have a `source_language_id` in `obj.project.source_language_ids` and belong to a project where `contribute_shared_tm` is `True`, excluding the components of `obj`'s project.

weblate/addons/base.py
```diff
@@ -29,6 +29,8 @@
 if TYPE_CHECKING:
     from django_stubs_ext import StrOrPromise
 
+    from weblate.trans.models import Component, Project
```
ParthS007, line 32: I am not sure why `ruff` moves the import inside `if`. What might be the reasoning here?

nijel: To avoid importing it at runtime when it is needed only for type checking.

weblate/trans/models/component.py
```diff
@@ -931,7 +931,7 @@ def install_autoaddon(self):
                 continue
 
             if (
-                addon.project_scope
+                getattr(addon, "project_scope", False)
```
nijel, line 934: This really should not be needed. `addon` here should be a `BaseAddon` subclasss which has this attribute.

weblate/wladmin/views.py
```diff
@@ -77,6 +77,7 @@
     ("appearance", "manage-appearance", gettext_lazy("Appearance")),
     ("tools", "manage-tools", gettext_lazy("Tools")),
     ("machinery", "manage-machinery", gettext_lazy("Automatic suggestions")),
+    ("addons", "manage-addons", gettext_lazy("Addons")),
```
nijel, line 80: Let's be consistent in naming:

```suggestion
    ("addons", "manage-addons", gettext_lazy("Add-ons")),
```

weblate/urls.py
```diff
@@ -186,6 +186,12 @@
         weblate.lang.views.CreateLanguageView.as_view(),
         name="create-language",
     ),
+    path(
+        # Site-wide add-ons
+        "manage/addons/",
+        weblate.addons.views.AddonList.as_view(),
+        name="addons",
+    ),
```
nijel, line 194: This is now duplicated with `name="manage-addons"` below. Are both really needed?

weblate/trans/forms.py
```diff
@@ -904,12 +904,21 @@ def __init__(self, obj, user=None, *args, **kwargs):
...
```
nijel, line 921: This needs more design decisions how it should behave. Presently, `obj` will be either `Component` or `None` because that is passed here:

https://github.com/WeblateOrg/weblate/blob/31358338c3613433794d19308c86269514232eda/weblate/addons/forms.py#L459

So this needs to be changed to pass `addon.instance.component or addon.instance.project` instead.

This form is also used for directly invoked automatic translation, where `obj` is always `Component` and the existing behavior should be preserved.

I think in the end we need something like this:

```suggestion
        if isinstance(obj, Component):
            obj = obj.project

        if obj is not None:
            # This is for both component and project scope
            self.components = obj.component_set.filter(
                source_language=obj.source_language
            ) | Component.objects.filter(
                source_language_id=obj.source_language_id,
                project__contribute_shared_tm=True,
            ).exclude(project=obj)
        else:
            # This is for site-wide
            self.components = Component.objects.all()
```

ParthS007: After discussing, what we want for site wide addons

> we only offer site-wide machine translation engines in that case.

How to check for sitewide memory translation engine in case of Site-wide addons?

1. MachineryConfiguration has `sitewide` attribute - `sitewide` is a boolean field
2. machinery_settings is a dictionary of machinery settings - `sitewide` is a key in machinery_settings and what is the schema of machinery_settings?

weblate/addons/models.py
```diff
@@ -66,21 +77,31 @@ class Meta:
         verbose_name_plural = "add-ons"
 
     def __str__(self):
-        return f"{self.addon.verbose}: {self.component}"
+        return f"{self.addon.verbose}: {self.project or self.component}"
```
nijel, line 80: Nicer representation of site-wide add-ons:

```suggestion
        return f"{self.addon.verbose}: {self.project or self.component or 'site-wide'}"
```

weblate/addons/views.py
```diff
@@ -9,21 +9,29 @@
...
+                raise PermissionDenied("Can not edit component")
+            self.kwargs["component_obj"] = self.path_object
+            return Addon.objects.filter_component(self.path_object)
+        if isinstance(self.path_object, Project):
+            if not self.request.user.has_perm("project.edit", self.path_object):
+                raise PermissionDenied("Can not edit project")
+            self.kwargs["project_obj"] = self.path_object
+            return Addon.objects.filter_project(self.path_object)
+
+        return Addon.objects.filter_component(None)
```
nijel, line 34:
```suggestion
        return Addon.objects.filter_sitewide()
```

docs/admin/addons.rst
```diff
@@ -5,7 +5,7 @@ Add-ons
 
 Add-ons provide ways to customize and automate the translation workflow.
 Admins can add and manage add-ons from the :guilabel:`Manage` ↓ :guilabel:`Add-ons` menu of each respective
-translation component.
+translation project or component.
```
nijel, line 8: Mention site-wide add-ons:

```suggestion
translation project or component. Add-ons can be also installed site-wide in :ref:`management-interface`.
```

weblate/addons/base.py
```diff
@@ -72,27 +74,46 @@ def get_identifier(cls):
...
         result = cls(storage)
         result.post_configure(run=run)
         return result
 
     @classmethod
-    def get_add_form(cls, user, component, **kwargs):
+    def get_add_form(
+        cls,
+        user,
+        component: Component | None = None,
```
nijel, line 109: Make these kwargs only so that it is clear whether component or project is passed:

```suggestion
        *,
        component: Component | None = None,
```

nijel (commented):
We're getting close! Let's finish this for the 5.5 release. Besides the suggestions I've made, I see the following bit missing.

**Documentation**

* Changelog entry in the documentation
* Mention of site-wide add-ons in docs/admin/admin.rst
* There is an issue with the documentation build, I will look into that

**Site-wide add-ons**

As this is a recent addition to this PR, there are some uncovered bits:

* I'd like to have tests for installing and triggering site-wide add-ons in weblate/addons/tests.py.
* I doubt that site-wide add-on would get ever triggered with current code. The problem is that `filter_component` is used for both listing and execution. For execution (see below for code) we want to filter all applicable add-ons, while for listing only the ones installed in the scope.
* Daily project/site-wide add-ons will not get valid scope when executing (see below for code). It will probably need moving inner code to a separate function and perform it in a loop over all (project) components in that case.
* The project/site-wide add-ons do not perform component validation when installing (obviously), so these should do it at execution time and skip unsupported components silently (instead of getting uninstalled on error, see execution code below).


**Related code**

Loading of add-ons for execution:

https://github.com/WeblateOrg/weblate/blob/2bfbc3c0ec2bb3ae35b71a0daded22f103d7310e/weblate/trans/models/component.py#L3682-L3694

Add-on execution code:

https://github.com/WeblateOrg/weblate/blob/2bfbc3c0ec2bb3ae35b71a0daded22f103d7310e/weblate/addons/models.py#L218-L253

nijel:
The documentation build should be addressed by https://github.com/WeblateOrg/weblate/pull/11150, but it will have a few conflicts with this PR, sorry.

weblate/api/serializers.py
```diff
@@ -1342,19 +1342,36 @@ class AddonSerializer(serializers.ModelSerializer):
...
             "name",
             "id",
             "configuration",
             "url",
         )
         extra_kwargs = {"url": {"view_name": "api:addon-detail"}}
 
+    @staticmethod
+    def check_addon(name, target, filter_method):
+        installed = set(filter_method(target).values_list("name", flat=True))
```
nijel, line 1366:
```suggestion
    def check_addon(name, queryset):
        installed = set(queryset.values_list("name", flat=True))
```

weblate/api/serializers.py
```diff
@@ -1378,25 +1396,24 @@ def validate(self, attrs):
...
-            )
-            available = {
-                x.name for x in ADDONS.values() if x.multiple or x.name not in installed
-            }
-            if name not in available:
-                raise serializers.ValidationError(
-                    {"name": f"Add-on already installed: {name}"}
-                )
+            if component:
+                self.check_addon(name, component, Addon.objects.filter_component)
```
nijel, line 1402:
```suggestion
                self.check_addon(name, Addon.objects.filter_component(component))
```

weblate/api/serializers.py
```diff
@@ -1378,25 +1396,24 @@ def validate(self, attrs):
...
+                if not addon.can_install(component, None):
+                    raise serializers.ValidationError(
+                        {"name": f"could not enable add-on {name}, not compatible"}
+                    )
+            if project:
+                self.check_addon(name, project, Addon.objects.filter_project)
```
nijel, line 1408:
```suggestion
                self.check_addon(name, Addon.objects.filter_project(project))
```

ParthS007:
- Documentation
   - [x] Change log entry in the documentation
   - [x] Mention of site-wide add-ons in docs/admin/admin.rst
   - [x] There is an issue with the documentation build, I will look into that

- Site-wide add-ons
   - [ ] **WIP tests for installing and triggering site-wide add-ons.**
   - [x] Add `filter_for_execution` for all applicable add-ons while for listing only the ones installed in the scope.
   - [x] Add `execute_addon_event` for `Daily project/site-wide add-ons` wiith a valid scope.
   - [x] Validation of unsupported components of `project/site-wide add-ons` at execution time and skipping unsupported components silently.

weblate/addons/models.py
```diff
@@ -42,24 +42,42 @@
...
+            | Q(project=component.project)
+            | Q(component__project=component.project)
+        )
+
+    def filter_project(self, project):
+        return self.prefetch_related("event_set").filter(project=project)
+
+    def filter_sitewide(self):
+        return self.prefetch_related("event_set").filter(
+            Q(component__isnull=True, project__isnull=True)
```
nijel, line 66: Q is not needed here:

```suggestion
            component__isnull=True, project__isnull=True
```

weblate/addons/models.py
```diff
@@ -68,22 +86,32 @@ class Meta:
         verbose_name = "add-on"
         verbose_name_plural = "add-ons"
 
-    def __str__(self) -> str:
-        return f"{self.addon.verbose}: {self.component}"
+    def __str__(self):
```
nijel, line 89: Type hint gone missing:

```suggestion
    def __str__(self) -> str: 
```

weblate/addons/models.py
```diff
@@ -194,6 +235,54 @@ class Meta:
         prefix = ""
 
 
+def execute_addon_event(
+    addon: Addon,
+    component: Component,
+    scope: Project | Component | str,
```
nijel, line 241: What it really should be:

```suggestion
    scope: Translation | Component,
```

We don't need to go up here:

* If translation is passed, use translation
* Otherwise use component

It should match https://github.com/WeblateOrg/weblate/pull/10100/files#diff-815de81bffc610b75def2de884c550297811d0f3b4ec3ceb8538c9d261574aa5R296-R297

weblate/addons/models.py
```diff
@@ -194,6 +235,54 @@ class Meta:
...
+    event: AddonEvent,
+    method: str | Callable,
+    args: tuple | None = None,
+):
+    with transaction.atomic():
+        scope.log_debug("running %s add-on: %s", event.label, addon.name)
+        # Skip unsupported components silently
+        if hasattr(addon.addon, "can_install") and not addon.addon.can_install(
```
nijel, line 249: Skipping should happen only for component installed add-ons:

```suggestion
        if addon.component and not addon.addon.can_install(
```

weblate/addons/models.py
```diff
@@ -216,44 +305,22 @@ def handle_addon_event(
...
+            scope = "site-wide"  # there is a better way to handle this?
+
+        if isinstance(scope, component):
+            execute_addon_event(addon, component, scope, event, method, args)
+        if isinstance(scope, Project):
+            for _component in component.project.component_set.iterator():
+                execute_addon_event(addon, _component, scope, event, method, args)
+        else:  # site-wide
+            for _component in Component.objects.all().iterator():
+                execute_addon_event(addon, _component, scope, event, method, args)
```
nijel, line 323: I think this should work:

```suggestion
        if not auto_scope:
            execute_addon_event(addon, component, scope, event, method, args)
        else:
            if addon.component:
                components = [addon.component]
            elif addon.project:
                components = addon.project.component_set.iterator()
            else:
                components = Component.objects.iterator()
            
            for component in components:
                execute_addon_event(addon, component, component, event, method, args)
```

weblate/addons/tasks.py
```diff
@@ -78,7 +78,7 @@ def cdn_parse_html(files: str, selector: str, component_id: int) -> None:
 )
 def language_consistency(addon_id: int, language_ids: list[int]) -> None:
     addon = Addon.objects.get(pk=addon_id)
-    project = addon.component.project
+    project = addon.project
```
nijel, line 81: This now needs to get project_id as a parameter, as it can no longer can be inferred from the add-on for site-wide add-ons.

weblate/addons/models.py
```diff
@@ -42,24 +42,42 @@
 
 
 class AddonQuerySet(models.QuerySet):
+    def filter_for_execution(self, component=None, project=None):
+        queryset = self.filter_sitewide()
+        # Include by component or project
+        if component:
+            queryset |= self.filter_component(component)
```
nijel, line 49: It needs to include the repo_scope logic here removed from `filter_component`.

weblate/addons/models.py
```diff
@@ -42,24 +42,42 @@
...
     def filter_component(self, component):
         return self.prefetch_related("event_set").filter(
-            (Q(component=component) & Q(project_scope=False))
-            | (Q(component__project=component.project) & Q(project_scope=True))
-            | (Q(component__linked_component=component) & Q(repo_scope=True))
-            | (Q(component=component.linked_component) & Q(repo_scope=True))
+            Q(component=component)
+            | Q(project=component.project)
+            | Q(component__project=component.project)
+        )
```
nijel, line 59: I'm still uncertain here. Do we want to list elsewhere installed add-ons on the component site? If no, it should be just `component=component` here. If yes, the repo_scope logic should stay.

ParthS007: After thinking through, It would be more intuitive to not list elsewhere installed add-ons on the component site.

The AddonQuerySet would look like this

```
class AddonQuerySet(models.QuerySet):
    def filter_for_execution(self, component):
        return self.prefetch_related("event_set").filter(
            (Q(component=component) & Q(project_scope=False))
            | (Q(component__project=component.project) & Q(project_scope=True))
            | (Q(component__linked_component=component) & Q(repo_scope=True))
            | (Q(component=component.linked_component) & Q(repo_scope=True))
        )

    def filter_component(self, component):
        return self.prefetch_related("event_set").filter(component=component)

    def filter_project(self, project):
        return self.prefetch_related("event_set").filter(project=project)

    def filter_sitewide(self):
        return self.prefetch_related("event_set").filter(
            component__isnull=True, project__isnull=True
        )

    def filter_event(self, component, event):
        return component.addons_cache[event]
 ``` 
 
 wdyt?

Comment (nijel) [reply weblate/addons/models.py:51]: I agree. Just your `filter_for_execution` misses site-wide addons, so additional `| Q(component__isnull=True) & Q(project__isnull=True) is needed there.

**Review by ParthS007 — COMMENTED** [review]
@nijel Here are some questions which are blocking me. Please have a look. Thanks

File: weblate/addons/views.py [hunk 46]
```diff
@@ -35,40 +43,66 @@ def redirect_list(self, message=None):
...
+                ),
+                key=lambda x: x.name,
+            )
+
         return result
 
     def post(self, request, **kwargs):
-        component = self.path_object
-        component.acting_user = request.user
+        obj = self.path_object
```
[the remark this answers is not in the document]
nijel: No, the whole `acting_user` logic is hacky: it is really a workaround for passing the `User` object with each operation. In this case, it is needed for generating change when `Addon.save()` is called. What could be done is adding `acting_user` attribute to the `Addon` class as well, set it there and use it from there in `Addon.store_change`.

weblate/addons/forms.py
```diff
@@ -456,7 +456,9 @@ def clean_intermediate_template(self):
 class AutoAddonForm(BaseAddonForm, AutoForm):
     def __init__(self, user, addon, instance=None, **kwargs) -> None:
         BaseAddonForm.__init__(self, user, addon)
-        AutoForm.__init__(self, obj=addon.instance.component, **kwargs)
+        AutoForm.__init__(
+            self, obj=addon.instance.component or addon.instance.project, **kwargs
+        )
 
 
 class BulkEditAddonForm(BaseAddonForm, BulkEditForm):
```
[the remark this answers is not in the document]
nijel: I'd change it to:

```
        BulkEditForm.__init__(
            self, obj=component, project=component.project if component else None, user=None, **kwargs
        )
```

The `BulkEditForm` accepts `None` as `project` now: f61fea66282be6d57ac61b48849d7905f48b56e8

weblate/addons/base.py
```diff
@@ -119,10 +140,11 @@ def configure(self, configuration) -> None:
         self.post_configure()
 
     def post_configure(self, run: bool = True) -> None:
-        component = self.instance.component
+        # What target to have in site-wide add-ons?
```
ParthS007, line 143: In `post_configure`

```
# Where to log for site-wide add-ons?
target = self.instance.component or self.instance.project
```

nijel: I think it should use a global logger, something like:

```py
import logging

logging.getLogger("weblate.addons")
```

The annoying thing is that `LoggerMixin` (used in `Project` and `Component`) has a different API.

weblate/addons/tests.py
```diff
@@ -95,6 +95,13 @@ def test_create_project_addon(self) -> None:
         self.assertEqual(self.component.project.addon_set.count(), 1)
         self.assertEqual("Test add-on: Test", str(addon.instance))
 
+    def test_create_site_wide_addon(self) -> None:
+        addon = TestAddon.create(acting_user=self.component.acting_user)
```
nijel, line 99: `self.component.acting_user` is probably None at this point, better use self.user instead.

weblate/addons/tests.py
```diff
@@ -1059,7 +1082,7 @@ def test_language_consistency(self) -> None:
         # Add one language
         language = Language.objects.get(code="af")
         self.component.add_new_language(language, None)
-        # To check: Failing on BulkForm
+        # To check #
```
nijel, line 1085: Probably, comment can be removed now?

weblate/addons/tests.py
```diff
@@ -682,6 +698,14 @@ def test_add_simple_project_addon(self) -> None:
         )
         self.assertContains(response, "1 add-on installed")
 
+    def test_add_simple_site_wide_addon(self) -> None:
+        response = self.client.post(
+            reverse("manage-addons"),
+            {"name": "weblate.consistency.languages"},
+            follow=True,
+        )
+        self.assertContains(response, "1 add-on installed")
```
nijel, line 707: Please add one test for site-wide add-on with config as well.

weblate/addons/tests.py
ParthS007: This is the only test that is failing, I looked into it and when we add one language `af`

The `<Translation: Test/TS — Afrikaans>` is missing from `Translation.objects` that is causing it to fail.

nijel: That means that the add-on is not properly executed as it should be. It didn't create the additional translations.

nijel: I currently don't see an error, but the problem can be in following paths:

* The project scoped add-on is not correctly installed within project.
* The project scoped add-ons are not picked up when filtering for execution.
* The language consistency add-on execution is broken.

ParthS007: Yes, Thanks for the comment. I am looking into it.

weblate/addons/models.py
```diff
@@ -42,25 +43,41 @@
 
 
 class AddonQuerySet(models.QuerySet):
-    def filter_component(self, component):
+    def filter_for_execution(self, component):
         return self.prefetch_related("event_set").filter(
-            (Q(component=component) & Q(project_scope=False))
-            | (Q(component__project=component.project) & Q(project_scope=True))
+            Q(component=component)
+            | Q(component__project=component.project)
```
nijel, line 49: I think this is extra:

```suggestion
```

It would cause execution of other component add-ons on all components in a project.

weblate/addons/models.py
```diff
@@ -42,25 +43,40 @@
...
+    component = models.ForeignKey(
+        Component, on_delete=models.deletion.CASCADE, null=True
+    )
+    project = models.ForeignKey(Project, on_delete=models.deletion.CASCADE, null=True)
     name = models.CharField(max_length=100)
     configuration = models.JSONField(default=dict)
     state = models.JSONField(default=dict)
-    project_scope = models.BooleanField(default=False, db_index=True)
     repo_scope = models.BooleanField(default=False, db_index=True)
+    acting_user = models.ForeignKey(User, on_delete=models.deletion.CASCADE, null=True)
```
nijel, line 79: Should be attribute only, not stored in the database.

```suggestion
```

nijel:

PR #10110, opened 2023-10-05
category: preserve the category id when creating a new component

joubu:
When you are inside a category and want to create a new component, you can either click "Manage > Add new translation component", or "+" (Add new translation component link)

This + link is on the left of the table's headers of the component list

There is a typo "categry" vs "category" which leads to an empty value on the second screen of the creation form (ie. the category is not pre-selected)


nijel:

PR #10136, opened 2023-10-10
translate: Show plural form examples more prominently

### LINKED ISSUE #9675: Don't use the misleading name "One" for a singular form used for numbers other than 1
### Describe the problem

Consider the following English source string with plural form (from a PO file):
```
msgid "One apple"
msgid_plural "{count} apples"
```
As you see, msgid uses "One apple" instead of "{count} apple", which is acceptable in English.

In other languages, however, the singular form may be used not only for 1. For example, in Russian, Ukrainian, Belarusian, Serbian, and Croatian the singular form is also used for numbers ending with 1. The problem with Webalte is that although it shows the right hint when you hover over the circled "i", the entry name literally says "One", which is simply not true for the mentioned languages:
Untitled

Naturally, this misguides translators: they think that the message will be used for exactly 1 item, and create a translation that doesn't work with 21, 31, 41, etc. - that is, in the example source string above, they will use the analog of "One apple" instead of "{count} apple". After all, if the entry clearly says "one", why would anyone check the hint button?

The entry label should probably not say "One", but instead include the example text currently hidden in (i).

_No response_

### Screenshots

_No response_

### Additional context

_No response_

nijel:
## Proposed changes

Instead of the tooltip, the text is rendered in the page now:

image

Fixes #9675

## Checklist

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -18,6 +18,7 @@ Not yet released.
 * Enage page better shows stats.
 * Strings which can not be saved to a file no longer block other strings to be written.
 * Fixed some API URLs for categorized components.
+* Show plural form examples more prominently.
 
 **Bug fixes**
 
--- a/weblate/lang/models.py
+++ b/weblate/lang/models.py
@@ -33,14 +33,13 @@ from weblate.logger import LOGGER
 from weblate.trans.defines import LANGUAGE_CODE_LENGTH, LANGUAGE_NAME_LENGTH
 from weblate.trans.mixins import CacheKeyMixin
 from weblate.trans.util import sort_objects, sort_unicode
-from weblate.utils.templatetags.icons import icon
 from weblate.utils.validators import validate_plural_formula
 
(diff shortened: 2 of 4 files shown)


t-cordonnier:
> Fixes https://github.com/WeblateOrg/weblate/issues/9675

Actually if I read the text of the ticket, no:

> In other languages, however, the singular form may be used not only for 1. For example, in Russian, Ukrainian, Belarusian, Serbian, and Croatian the singular form is also used for numbers ending with 1.

So if you want to follow this logic this should be
One (1, 21, 31, 121, ...)

nijel:
In Czech it is typically only 1. You can also use singular for anything ending 1 (except 11), but that is not widely used anymore. CLDR definition matches that, see https://www.unicode.org/cldr/charts/40/supplemental/language_plural_rules.html#cs.

This PR shows actual examples matching plural form. You can see these examples on language pages, for example at https://hosted.weblate.org/languages/hr/#information

Issue #9998, opened 2023-09-21
Mobile website width issues


Daviteusz:
### Describe the issue

The scroll bar for tables is missing making the page not fit the screen

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar issues in this repository.

### Steps to reproduce the behavior

1. Open Chrome, Firefox or similar browser
2. Go to project information or history tab
3. Zoom out

### Expected behavior

The site should fit the width of the screen in every possible place.

### Screenshots

<img src=https://github.com/WeblateOrg/weblate/assets/11527277/12392233-78f8-45df-aa63-3342a3d4a27d width=30%>
<img src=https://github.com/WeblateOrg/weblate/assets/11527277/c4818cfa-f720-42d3-b951-902aa473d45d width=30%>

### How do you run Weblate?

weblate.org service

Maddy0027:
@nijel i like to work on this issue but i am new to this so i would like if you help me with this my tech stack are HTML CSS javascript c++ and python

Maddy0027:
@nijel do you want to adjust overflow that is cause due reduce in screen size when we shift form desktop to mobile or tab version

nijel:
There are more issues on the page, for start just pick one of them to address.

Quickly I've identified two of them:

* Strings stats table, that probably needs horizontal scrollbar
* Activity chart should probably use narrower bars on smaller screens

Maddy0027:
is there is discord or slack channel where i can get in touch with you

nijel:
Sorry, I'm not anywhere like that currently.

PR #10146, opened 2023-10-11
activity: The bars width adpts to the width of screen

Maddy0027:
#issue #9998 
In this we have made the Activity chart more responnsive.
The bars width  adpts to the width of screen


nijel:

PR #10161, opened 2023-10-11
utils: provide lock name in timeout error message for better debugging

michael-smt:
## Proposed changes

Include the lock's name in the `WeblateLockTimeoutError` message to help with pinpointing which lock is affected.

## Checklist

## Other information

This is only for redis locks.


nijel:

Issue #10172, opened 2023-10-12
Refactor add-on event handling


nijel:
### Describe the problem

The add-on event handling code is repeated in each `@receiver`:

https://github.com/WeblateOrg/weblate/blob/0df9af9b0963c7fc16f8c7a3244eeb23a3ff896c/weblate/addons/models.py#L211-L404

### Describe the solution you'd like

I'd like to have shared code that would cover the logic and the individual receivers would just parametrize that.

### Additional context

Discovered when reviewing https://github.com/WeblateOrg/weblate/pull/10100

Shashankpantiitbhilai:
I am interested please assign it to me

nijel:
@Shashankpantiitbhilai you're welcome to work on the issue.

rawatsushil:
@nijel  I see the inactivity here , hence tried solving this and has raised PR also . https://github.com/WeblateOrg/weblate/pull/10203
Can you please see this and let me know if you need to optimize it further

rawatsushil:
@Shashankpantiitbhilai are you still working on it ? I

Shashankpantiitbhilai:
Yes

PR #10175, opened 2023-10-13
Maddy0027-patch-2

Maddy0027:
issue https://github.com/WeblateOrg/weblate/issues/10172
#10172 
created a function handle_addon_event to make the code more reusable

PR #10182, opened 2023-10-13
Update models.py

Maddy0027:
issue https://github.com/WeblateOrg/weblate/issues/10172
created a function handle_addon_event to make the code more reusable

### DIFF
--- a/weblate/addons/models.py
+++ b/weblate/addons/models.py
@@ -207,198 +207,66 @@ def handle_addon_error(addon, component):
     if not addon.addon.can_install(component, None):
         addon.disable()
 
+def handle_addon_event(component, event_type, addon_method, description):
+    addons = Addon.objects.filter_event(component, event_type)
+    
+    for addon in addons:
+        component.log_debug(f"running {event_type} add-on: {addon.name}")
+        try:
+            with sentry_sdk.start_span(op=f"addon.{event_type}", description=addon.name):
+                addon_method(addon.addon, component)
+        except DjangoDatabaseError:
+            raise
+        except Exception:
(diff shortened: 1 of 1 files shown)


weblate/addons/models.py
```diff
@@ -208,34 +208,42 @@ def handle_addon_error(addon, component):
...
+def handle_addon_event(component, event_type, addon_method, description):
+    addons = Addon.objects.filter_event(component, event_type)
+
+    for addon in addons:
+        component.log_debug(f"running {event_type} add-on: {addon.name}")
         try:
-            with sentry_sdk.start_span(op="addon.pre_push", description=addon.name):
-                addon.addon.pre_push(component)
+            with sentry_sdk.start_span(
+                op=f"addon.{event_type}", description=addon.name
```
nijel, line 218: https://github.com/WeblateOrg/weblate/pull/10175#discussion_r1357848079 still applies.

weblate/addons/models.py
```diff
@@ -208,34 +208,42 @@ def handle_addon_error(addon, component):
...
-            component.log_debug("completed pre_push add-on: %s", addon.name)
+            component.log_debug(f"completed {event_type} add-on: {addon.name}")
+
+
+@receiver(vcs_pre_push)
+def pre_push(sender, component, **kwargs):
+    handle_addon_event(
+        component,
+        EVENT_PRE_PUSH,
+        lambda addon, comp, component=component: addon.pre_push(comp),
```
nijel, line 234: Using lambdas here is not really a practical approach, as it makes the code hard to read, especially due to unused arguments for these.

Issue #6170, opened 2021-06-11
Support for Azure DevOps Pull Request


warlof:
**Is your feature request related to a problem? If so, please describe.**

When policies are enabled on a branch synched with Weblate using Azure DevOps, it is not able to push changes without using a pull request.

**Describe the solution you'd like**

The same way Github or Gitlab have the ability to submit changes using a pull request / merge request, Weblate should be able to submit changes to Azure DevOps using a pull request.

https://docs.weblate.org/en/latest/vcs.html#pushing-changes-to-github-as-pull-requests

**Screenshots**

![image](https://user-images.githubusercontent.com/648753/121654846-6f8bab80-ca9e-11eb-9226-7a239c383b1b.png)
![image](https://user-images.githubusercontent.com/648753/121654855-72869c00-ca9e-11eb-92b0-8ff87532aa2a.png)

PR #10227, opened 2023-10-18
Added support for Azure DevOps Pull Requests

boginw:
## Proposed changes

This pull request adds support for creating pull requests on Azure DevOps both from the same origin and from forks, thereby closing #6170. The introduced DevOpsRepository can also attach work items to the newly created pull requests. We're using DevOps at my organization and this will streamline our continuous localization efforts.

## Checklist

## Other information

An important part of DevOps is the notion of organizations and projects. AFAIK every repository in Azure DevOps resides in a project, which in turn resides in an organization. All API calls to DevOps all three identifiers (organization, project, and repository). This poses a challenge for the current credentials conventions of `X_HOST`, `X_USERNAME`, and `X_TOKEN`. This PR's solution is to just populate the credentials with the required pieces of information. I'd love to hear your opinions regarding this.

Also, I'm unsure how pull requests with forks would work across organizations in DevOps, that's not a workflow I'm familiar with on that platform. Since I only have access to one organization with one project, I'm also unable to test that part. However, creating pull requests on the same project is working as expected.

Finally, no docs have been added yet. I'd love to have the code in a state where it's mergeable before writing docs.


nijel:
On a first look, the code looks good, even though it is quite complex compared to other pull request backends.

Including additional bits in the credentials is definitely the way, it was intended to be used like this (it's just a coincidence that none backend needed it so far).

If you are talking about Docker configuration, you can use `WEBLATE_DEVOPS_CREDENTIALS` instead of the split variables. These exist to make the most usual setup simpler. Support for `_ORGANIZATION` suffix can also be added to `get_env_credentials` to make it easier to set up…

nijel:
As for the credentials, I've added validation of the configuration in https://github.com/WeblateOrg/weblate/pull/10228, so you might want to extend `REQUIRED_CONFIG` and `OPTIONAL_CONFIG` to cover what this integration needs..

boginw:
Thanks for your review :)

I've tried to simplify things a bit and added more tests, but DevOps seems to do things in a way that makes it complicated.
I've also extended `REQUIRED_CONFIG` and `OPTIONAL_CONFIG`.

boginw:
I had some issues with git and rebasing on top of the upstream, which led to many force-pushes. I've added support for the `*_ORGANIZATION` env key, added a few more tests for the failure states, and also included some documentation. I don't have any experience writing in `.rst` files, so any feedback is much appreciated :)

docs/admin/config.rst
```diff
@@ -831,6 +831,33 @@ List for credentials for Bitbucket servers.
...
+
+    DEVOPS_CREDENTIALS = {
+        "dev.azure.com": {
+            "username": "project-name",
+            "token": "your-api-token",
+            "organization": "organization-name",
+        },
+    }
+
+.. include:: /snippets/vcs-credentials-devops.rst
```
nijel, line 853: The sole purpose of the snippet is to be able to reuse it. If you can't use the generic description here, just include it here directly.

nijel (commented):
Looks great! The only thing where I'm not sure is naming. Maybe it would be better to name this as *Azure DevOps* instead of just *DevOps*?

[DevOps](https://en.wikipedia.org/wiki/DevOps) is a generic term, and people not knowing this Azure service will think it's some generic integration.

docs/admin/config.rst
nijel: I've rewritten this to better work with different services, see https://en.wikipedia.org/wiki/DevOps. That should allow you to reference existing docs while documenting service specifics properly.

boginw: Thanks for taking a look, I knew I was missing something. I've inlined this in the appropriate place and tried to rewrite the section so it reads better and describes the options adequately :)

boginw:
Thanks for your feedback! I see what you mean by DevOps being misconstrued, I've renamed everything from DevOps to Azure DevOps and verified that everything still works :)

nijel:

warlof:
@boginw thank you for this enhancement
I noticed you required the project in devops credentials - can it be an empty string ?

An Azure DevOps tenant can host multiple project, and repositories are attached to them.

If the project is mandatory, that mean you can connect a Weblate instance only to one single project :o

boginw:
Hi @warlof :)

You're right, and that's a giant oversight from my side, sorry about that. I originally required project-name as every API endpoint needed it, but I realize that as long as you have access to the repo URL you can derive both the organization and the project. But I'm having trouble figuring out one thing. Given more than one project on an Azure DevOps tenant and let's say that a project is configured on Weblate to fork the repository, then which project should we fork the repository into? The same project or should that be configurable?

Besides the above-mentioned it shouldn't be an issue. I'll look into supporting multi-project single-tenant repos :)

warlof:
Can it be a config element on the component related project ?
Otherwise, maybe it can be extract from the repository url ? either pull or push - the project will live in it.

actually, I know URL follow this pattern :

- `https://dev.azure.com/{organization}/{project_name}` : project url
- `https://{organization}@dev.azure.com/{organization}/{project_name}/_git/{repository_name}` : https git remote url
- `git@ssh.dev.azure.com:v3/{organization}/{project_name}/{repository_name}` : ssh git remote url

I read some old tenants might use variation - however I think they since been all migrate to follow this format

As of the fork issue - if using a fork, both repository and project will leave in distinct organization - isn't it ?

nijel:
Following the same logic as Weblate uses for push URLs, could it that if a project is configured, it will fork to that project; otherwise the fork will be in the same project as the original repo? See https://docs.weblate.org/en/latest/admin/continuous.html#pushing-changes-from-weblate

warlof:
> Following the same logic as Weblate uses for push URLs, could it that if a project is configured, it will fork to that project; otherwise the fork will be in the same project as the original repo? See https://docs.weblate.org/en/latest/admin/continuous.html#pushing-changes-from-weblate

so with the design in your mind, you'll did a dedicated weblate project and fork all translated repository to that project ?
I think it might work - even if I never play with fork with Azure DevOps

python is currently out of my skill to design such changes in Weblate, but I trust on @boginw ^^

Mathyn:
@boginw I am currently integrating Azure DevOps in my Weblate setup. We've got an Azure DevOps setup with multiple projects. Are multiple projects currently supported or not in Weblate? I'm running into some errors suggesting they are not.

Also see my Q&A ticket -> https://github.com/WeblateOrg/weblate/discussions/12002

PR #10237, opened 2023-10-19
feat: Add Alibaba machine translate service

### LINKED ISSUE #7659: Add Alibaba Translate machine translate service
### Basic details

Provider: Alibaba Translate

**Description**

1. This provider offer free quota, 1 million characters per month
2. they've several services that collaborate with real translators for many languages so ofcourse MT quality might be better (or in future) than to other popular providers that get developed using public people's inputs etc.

https://www.alibabacloud.com/help/en/machine-translation
https://www.alibabacloud.com/help/en/machine-translation/latest/product-pricing-introduction

- _https://translate.alibaba.com/ (note: to avoid confuse, if you click API access on this site, it will redirect to the local alibaba translate; aliyun site, but you can use above alibabacloud (en site) for easy)_
- _https://alynx.alifanyi.com/ https://alynx.alifanyi.com/pay.html
(also it has a free plan and part of alibaba translate)_
- _https://www.zuodao.com/crowdsourcing/homePage.html
(part of alibaba translate)_

hussainshaikh12:
## Proposed changes

Fixes #7659
## Checklist

## Other information
1. Tested the feature locally and it works perfectly.
Screenshot from 2023-10-20 00-05-48


hussainshaikh12:
Hey @nijel @orangesunny do i need to do anything else here ?

docs/admin/machine.rst
```diff
@@ -26,6 +26,27 @@ The services translate from the source language as configured at
 
    :ref:`machine-translation`
 
+.. _mt-alibaba:
+
+Alibaba
+---
```
nijel, line 32:
```suggestion
Alibaba
-------
```

requirements-optional.txt
```diff
@@ -1,4 +1,7 @@
 -r requirements.txt
+# Alibaba
+aliyun-python-sdk-alimt==3.2.0
+aliyun-python-sdk-core-v3==2.13.33
```
nijel, line 4: `aliyun-python-sdk-alimt` depends on `aliyun-python-sdk-core`, why is `aliyun-python-sdk-core-v3` needed?

Also I suggest using less strict dependency pin here, something like:

```suggestion
aliyun-python-sdk-alimt>=3.2.0,<4.0.0
```

nijel (commented):
Besides above feedback, can you please add tests for this backend?

requirements-optional.txt
hussainshaikh12: Ok got it. `aliyun-python-sdk-core-v3` was mentioned in the docs, if using python > 3.0.

hussainshaikh12:
> Besides above feedback, can you please add tests for this backend?

How should i write tests ? as they dont have a specific url which to fetch and i am using their sdk. It is similar to aws where they are using a stubber but i guess that same stubber wont work here as it is aws specific

nijel:
As for the tests, the library seems to be built on top of requests, so you can use existing mocking by `responses`, just define mocked responses in `mock_response` method of the test. You can use IBM service as an example:

https://github.com/WeblateOrg/weblate/blob/1a4fdc462beb59f4cd79397137f87d82bdcf4bd2/weblate/machinery/tests.py#L1225-L1258

hussainshaikh12:
> As for the tests, the library seems to be built on top of requests, so you can use existing mocking by `responses`, just define mocked responses in `mock_response` method of the test. You can use IBM service as an example:
> 
> https://github.com/WeblateOrg/weblate/blob/1a4fdc462beb59f4cd79397137f87d82bdcf4bd2/weblate/machinery/tests.py#L1225-L1258

Tried this, but it still makes the request and gives InvalidAccessKeyId.NotFound error

ERROR: test_translate (weblate.machinery.tests.AlibabaTranslationTest)
  File "/home/hussain/Desktop/open/weblate/weblate/machinery/base.py", line 409, in _translate
    result = [
  File "/home/hussain/Desktop/open/weblate/weblate/machinery/base.py", line 409, in <listcomp>
    result = [
  File "/home/hussain/Desktop/open/weblate/weblate/machinery/alibaba.py", line 275, in download_translations
    response = self.client.do_action_with_exception(request)
  File "/home/hussain/Desktop/open/venv/lib/python3.10/site-packages/aliyunsdkcore/client.py", line 501, in do_action_with_exception
    raise exception
aliyunsdkcore.acs_exception.exceptions.ServerException: HTTP Status: 404 Error:InvalidAccessKeyId.NotFound Specified access key is not found. RequestID: D4E7621B-5052-55EE-B947-E4CDD671FDE3

nijel:
Ah, they are using a vendored requests instead of a dependency. It should be possible to patch that as well using `responses`, but there is really no documentation on that, see https://github.com/getsentry/responses/issues/188

Looking at the `responses` code, the following might work (not tested at all):

```python


class AlibabaTranslationTest(BaseMachineTranslationTest): 
    ....
    mock = responses.RequestsMock(target="aliyunsdkcore.vendored.requests.adapters.HTTPAdapter.send")
    
     def mock_response(self):
         self.mock.add(...)

     @mock.activate
     def test_translate(self, **kwargs):
         super().test_translate(**kwargs)

hussainshaikh12:
Followed the approach from GoogleV3Translation and used the patch.object to override the call for the do_action_with_exception method and this works finally.

Issue #10222, opened 2023-10-17
Cannot assign a team/group administrator using the REST API


joubu:
### Describe the problem

It's not possible to grant administration to a group/team using the REST API

### Describe the solution you'd like

It does not seem easy to implement, but just logging the problem in an issue to report the problem.

rawatsushil:
@joubu  Can you describe the problem please

joubu:
Using the UI you can "Grant administration" or "Revoke administration" on the following view: http://localhost:8888/teams/GROUP_ID/#users
![image](https://github.com/WeblateOrg/weblate/assets/671757/1192a401-749c-4a04-beb3-03a4953d6af2)

I have not found a way to do that using the REST API.

nijel:
Indeed it is not exposed in the API. Adding that should be pretty easy to the existing API endpoints: https://docs.weblate.org/en/latest/api.html#groups

nidhi-kala:
@nijel hey! I would like to work on this.
I'll add a new endpoint to the GroupViewSet in the api. Is there a preference for name of the endpoint?

nijel:
Sorry for missing your question. See how components endpoint works at GroupViewSet, this should be similar.

PR #10292, opened 2023-10-27
Adding grant revoke admin endpoints

nidhi-kala:
Addressing #10222 

This PR will add API endpoints for granting and revoking admin rights.
I have added tests for the endpoints.

## Checklist

### DIFF
--- a/weblate/api/views.py
+++ b/weblate/api/views.py
@@ -461,10 +461,16 @@ class GroupViewSet(viewsets.ModelViewSet):
     def get_queryset(self):
         if self.request.user.has_perm("group.edit"):
             return Group.objects.order_by("id")
-        return self.request.user.groups.order_by("id")
+        return self.request.user.groups.order_by(
+            "id"
+        ) | self.request.user.administered_group_set.order_by("id")
 
-    def perm_check(self, request) -> None:
-        if not request.user.has_perm("group.edit"):
+    def perm_check(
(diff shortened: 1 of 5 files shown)


nijel (commented):
To be consistent with other similar API endpoints, it should be:

* /admins/ endpoit - GET to list admins, POST to add
* /admins/(int:user_pk)/ - DELETE To delete

See how /components is handled at GroupViewSet.

nidhi-kala:
Thank you for guiding me @nijel !
I have made the naming consistent for grant and revoke admin endpoints and updated the tests.

weblate/api/views.py
```diff
@@ -631,6 +631,35 @@ def delete_components(self, request, id, component_id):
         obj.components.remove(component)
         return Response(status=HTTP_204_NO_CONTENT)
 
+    @action(detail=True, methods=["post"], url_path="admins")
+    def grant_admin(self, request, id):
+        group = self.get_object()
+        self.perm_check(request)
```
nijel, line 637: The perm_check should be extended to optionally accept object, so that group admins can also use this interface.

nijel (changes requested):
Thanks, the code looks good, but there are few things which I'd like to see improved:

* The documentation is missing.
* The admins API will not work for group admins, see my comment on permission check.
* Can test for group admin doing changes be added?

PR #10302, opened 2023-10-29
Update install.rst - Celery not compatible with Python 3.12

JohnRDOrazio:
See discussion #10301 , Celery is not currently compatible with Python 3.12

I have instead successfully instantiated with Python 3.11.


nijel:
As there is other issue with Python 3.12 as well, I've done an alternative PR at https://github.com/WeblateOrg/weblate/pull/10303

JohnRDOrazio:
Superseded by PR #10303

PR #10337, opened 2023-11-02
Add formality parameter to DeepL API integration

tipa:
## Proposed changes

I added an option to the DeepL API integration to specify the formality.
Link to DeepL API: https://www.deepl.com/docs-api/translate-text

This PR adds a dropdown-setting to the configuration screen of the DeepL integration, with 3 options:
Default (which is the default and sets `formality` to `default`)
Informal (this sets `formality` to `prefer_less`)
Formal (this sets `formality` to `prefer_more`)

This PR also adjusts the `mtid` field of the `MachineTranslation` to make sure the caching works correct (informal/formal/default texts of DeepL suggestions are cached separately).

The `@FORMAL` and `@INFORMAL` language id endings have precedence over this new setting.

## Checklist

## Other information
I am not very familiar with Python or Django, so I didn't know what the best way is to translate the config options and label/help text for the new `CharField`

### DIFF
--- a/weblate/machinery/deepl.py
+++ b/weblate/machinery/deepl.py
@@ -62,7 +62,9 @@ class DeepLTranslation(MachineTranslation):
 
     # using custom cache key to ensure that formal and informal suggestions are cached separately
     def translate_cache_key(self, source, language, text, threshold):
-        return super().translate_cache_key(source, language, text, threshold) + self.settings.get("formality", "default")
+        return super().translate_cache_key(
+            source, language, text, threshold
+        ) + self.settings.get("formality", "default")
 
     def download_translations(
         self,
--- a/weblate/machinery/deepl.py
+++ b/weblate/machinery/deepl.py
@@ -26,7 +26,8 @@ class DeepLTranslation(MachineTranslation):
 
     def __init__(self, settings: dict[str, str]):
         super().__init__(settings)
-        self.mtid = self.mtid + "_" + settings['formality']
+        self.formality =  self.settings["formality"] if "formality" in self.settings else "default"
+        self.mtid = self.mtid + "_" + self.formality
 
     def map_language_code(self, code):
         """Convert language to service specific code."""
@@ -78,7 +79,7 @@ class DeepLTranslation(MachineTranslation):
             "text": text,
             "source_lang": source,
             "target_lang": language,
-            "formality": self.settings["formality"] if "formality" in self.settings else "default",
+            "formality": self.formality,
             "tag_handling": "xml",
             "ignore_tags": "x",
         }
--- a/weblate/machinery/deepl.py
+++ b/weblate/machinery/deepl.py
@@ -24,13 +24,6 @@ class DeepLTranslation(MachineTranslation):
     hightlight_syntax = True
     settings_form = DeepLMachineryForm
 
-    def __init__(self, settings: dict[str, str]):
-        super().__init__(settings)
-        self.formality = (
-            self.settings["formality"] if "formality" in self.settings else "default"
-        )
-        self.mtid = self.mtid + "_" + self.formality
-
     def map_language_code(self, code):
         """Convert language to service specific code."""
         return super().map_language_code(code).replace("_", "-").upper()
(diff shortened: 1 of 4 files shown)


weblate/machinery/deepl.py
```diff
@@ -24,6 +24,13 @@ class DeepLTranslation(MachineTranslation):
     hightlight_syntax = True
     settings_form = DeepLMachineryForm
 
+    def __init__(self, settings: dict[str, str]):
+        super().__init__(settings)
+        self.formality = (
+            self.settings["formality"] if "formality" in self.settings else "default"
+        )
+        self.mtid = self.mtid + "_" + self.formality
+
```
nijel, line 33: I'd move this out of init:

```suggestion
```

* formality can be handled at request time
* `mtid` is used for caching list of languages and rate limiting, none of these is affected by formality

weblate/machinery/deepl.py
```diff
@@ -74,6 +81,7 @@ def download_translations(
             "text": text,
             "source_lang": source,
             "target_lang": language,
+            "formality": self.formality,
```
nijel, line 84:
```suggestion
            "formality": self.settings.get("formality", "default"),
```

weblate/machinery/forms.py
```diff
@@ -217,3 +217,19 @@ class DeepLMachineryForm(KeyURLMachineryForm):
...
+            "Uses the specified formality if language is not specified as (in)formal"
+        ),
+        widget=forms.Select(
+            choices=(
+                ("default ", "Default"),
+                ("prefer_more", "Formal"),
+                ("prefer_less", "Informal"),
+            )
+        ),
+        initial="default",
```
nijel, line 234:
```suggestion
        initial="default",
        required=False,
```

Without this, existing configuration will be suddenly invalid (and this is why the tests fail).

nijel:
Also, https://github.com/WeblateOrg/weblate/blob/main/docs/admin/machine.rst?plain=1 needs update, you can generate the parameters table using `./manage.py list_machinery`.

weblate/machinery/deepl.py
tipa: Should I just commit this suggestion with the others or did you just select/highlight this code in order to comment? I don't see where the `self.mtid` field would be written in your suggested change

tipa:
> Also, https://github.com/WeblateOrg/weblate/blob/main/docs/admin/machine.rst?plain=1 needs update, you can generate the parameters table using `./manage.py list_machinery`.

Sorry, I couldn't figure out how to run this. I don't have an dev environment for python set up on my (Windows) machine and can't figure how to execute this command without error. Would it be possible for you to run it and add the updated file to this PR?

nijel:
You can edit it manually as well, just add the newly added parameter to the parameters table.

weblate/machinery/deepl.py
tipa: So should I overwrite that method in `deepl.py` as follows?
```
def translate_cache_key(self, source, language, text, threshold):
    if not self.cache_translations:
        return None
    return super().translate_cache_key(source, language, text, threshold) + ":" + self.settings.get("formality", "default")
```

tipa: I have pushed another commit with that change.
I also moved the check for the `cache_translations` variable to the `get_cached` method to avoid code duplication

weblate/machinery/deepl.py
```diff
@@ -60,6 +60,12 @@ def is_supported(self, source, language):
         """Check whether given language combination is supported."""
         return (source, language) in self.supported_languages
 
+    # using custom cache key to ensure that formal and informal suggestions are cached separately
+    def translate_cache_key(self, source, language, text, threshold):
+        return super().translate_cache_key(
+            source, language, text, threshold
+        ) + self.settings.get("formality", "default")
```
nijel, line 67: Let's keep the key formatting:

```suggestion
        key = super().translate_cache_key(source, language, text, threshold)
        formality = self.settings.get("formality", "default")
        return f"{key}:{formality}"
```

nijel:

PR #10356, opened 2023-11-06
profile: Improve e-mail configuration

### LINKED ISSUE #10352: Explain e-mail use for commits more transparently
### Describe the problem

So currently there are two e-mail fields:

* E-mail (Choose primary e-mail from verified addresses.)
* Commit e-mail (Choose commit e-mail from verified addresses.)

This could intuitively be understood as "*e-mail* is for my account and **only** the *commit e-mail* is for commits" (if left empty there is no e-mail used in the commits).

But what happens instead is that if the *commit e-mail* field is empty, then the *e-mail* becomes the *commit e-mail*, (at least for po-file exports) which is not very transparent.

It becomes a bigger problem because once you add a translation without setting the *commit e-mail* field, Weblate will internally store your *e-mail* and use that in any po export. This will even happen if you change your *commit e-mail* later, so a user who once fell into the trap outlined above, will have their *e-mail* locked into the weblate database with no way to remove/change it. Of course addresses can be changed, but in a po export, **both** e-mail addresses will appear.

I'm not a laywer, but personal data transparency could become a [GDPR](https://en.wikipedia.org/wiki/General_Data_Protection_Regulation) issue. 

Short notice: Add some very transparent text blocks to the Profile/Account page, explaining what happens with the e-mail addresses you enter, and potentially a warning that your e-mail address could potentially turn up publicly in the project's git repo that you are contributing to if you fill the fields incorrectly.

This is just a band-aid though, see below for better alternatives.

Longer term: Add an option to edit your personal data that is stored inside the Weblate database with the option to remove/change your address after it has been added. This would solve the problem that someone could have missed the info block and still added an e-mail address they want removed or updated later.

Alternative longer term: Link the e-mail address to the account and propagate any changes on the account page to the database, so that if someone modifies their e-mail address, it will also be updated in the database. Current behavior is to just add new addresses while keeping the old ones in the database

Additional option: Give a user the option to keep their address anonymous, e.g. if no commit address is given. This could either be an omission of the email address or an anonymization like a***.b***@gmail.com. In addition to user choice this could also be a global setting for admins to set if they don't want any trouble with GDPR whatsoever, so they can avoid committing any personal information to their project repo.

### Screenshots

_No response_

### Additional context

_No response_

nijel:
## Proposed changes

- use radio instead of select to make all choices directly visible (there won't be much of them)
- clarify how each e-mail address is used (fixes #10352)
- add shared text about e-mail verification
- rename "add e-mail" to "verify e-mail" as that better describes what it does

## Checklist

## Other information

Screenshot:

image

### DIFF
--- a/weblate/accounts/forms.py
+++ b/weblate/accounts/forms.py
@@ -30,7 +30,7 @@ from weblate.accounts.utils import (
     get_all_user_mails,
     invalidate_reset_codes,
 )
-from weblate.auth.models import Group, User, get_auth_keys
+from weblate.auth.models import Group, User
 from weblate.lang.models import Language
 from weblate.logger import LOGGER
 from weblate.trans.defines import FULLNAME_LENGTH
@@ -173,8 +173,11 @@ class CommitForm(ProfileBaseForm):
     commit_email = forms.ChoiceField(
         label=gettext_lazy("Commit e-mail"),
         choices=[("", gettext_lazy("Use account e-mail address"))],
-        help_text=gettext_lazy("Choose commit e-mail from verified addresses."),
+        help_text=gettext_lazy(
+            "Used in version control commits. The address will stay in the repository forever once changes are commited by Weblate."
+        ),
         required=False,
+        widget=forms.RadioSelect,
     )
 
     class Meta:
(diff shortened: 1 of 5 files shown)


weblate/accounts/forms.py
```diff
@@ -173,8 +173,11 @@ class CommitForm(ProfileBaseForm):
     commit_email = forms.ChoiceField(
         label=gettext_lazy("Commit e-mail"),
         choices=[("", gettext_lazy("Use account e-mail address"))],
-        help_text=gettext_lazy("Choose commit e-mail from verified addresses."),
+        help_text=gettext_lazy(
+            "Used in version control commits. The address will stay in the repository forever once changes are commited by Weblate."
```
Cerno-b, line 177: 👍

PR #10386, opened 2023-11-09
add-ons: generate MO can optionally include strings needing editing

### LINKED ISSUE #9418: Allow GenerateMoAddon to export fuzzy strings
### Describe the problem

We use the 'fuzzy' marking to mark texts that need to be reviewed by our translators. However, we would like to use those texts in our application as soon as possible.

Right now, the addon `GenerateMoAddon` is programmed such that only units with a state greater than or equal to `STATE_TRANSLATED` are exported. 

It would be great for us if a configuration item could be added that allows us to export units with state `STATE_FUZZY` or greater.

I could investigate using the review workflow, but this is a project-wide settings and I would like to use this on a component level.

### Screenshots

_No response_

### Additional context

_No response_

nijel:
## Proposed changes

Fixes #9418

## Checklist

### DIFF
--- a/docs/admin/addons.rst
+++ b/docs/admin/addons.rst
@@ -660,9 +660,11 @@ Generate MO files
 -----------------
 
 :Add-on ID: ``weblate.gettext.mo``
-:Configuration: +----------+---------------------------+-------------------------------------------------------------+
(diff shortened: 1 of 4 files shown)


jorrit:
Thank you!

Issue #3006, opened 2019-09-10
Use of soft hyphen for better automatic line break control in translations


rhofer:
Is there a good way to deal with soft hyphen in weblate? (Or any other recommendation?)

nijel:
The only related thing Weblate does right now is to check for existence of [zero width space](https://docs.weblate.org/en/latest/user/checks.html#zero-width-space).

Can you please clarify what are you looking for?

rhofer:
It's about resulting visualization of a translation in application's UI. E.g.
- English: Display Settings
- German: Anzeigeeinstellungen

Per it's nature, English is not dealing with very long words, constructed and concatenated as it is possible in german. Therfore in some places the according visualization widget needs to break the text into two lines. What works well for english above, since there two words are separated with white space, resulting in line 1: Display, line 2: Settings.

Whereas for german, the out of box result would be something like: line 1: Anzeigeinstel, line 2: lungen. Without a hyphen, just arbitrarily wrapped. 

With soft hyphen this could be controlled better.
- HTML style: `Anzeige&shy;einstellungen`
- Result: 
  - line 1: Anzeige-
  - line 2: einstellungen

Question is whether from weblate point of view there is a way to deal with this.
Drawback is, soft hyphen is technology specific (html5, xml, ...)

rhofer:
The problem starts with editing, how a soft hyphen could be entered. There is a unicode character for this: U+00AD, see also https://de.wikipedia.org/wiki/Weiches_Trennzeichen

In windows you can "write" this sign by press+hold : ALT + 0173. The following is an example where after each word a soft hyphen is entered:

This­Is­A­Very­Long­Sentence­Where­After­Each­Word­A­Soft­Hyphen­Is­Entered­So­That­By­Exceeding­A­Line­It­Automatically­Breaks­At­The­Desired­Position­And­Renders­A­Hyphen­Accordingly.

Drawback is that this is a non-displayable character unless we come to a line end where at the position of a soft hyphen the line breaks, and only then this character becomes visible as a hyphen.

For editing in weblate it could be helpful to treat soft hyphen specially, as:

![shy-editor](https://user-images.githubusercontent.com/8058166/64759698-3d02f680-d538-11e9-9a4c-30c021aee2e8.png)

So that we have besides the special char buttons an additional one for soft hyphen (shy). When pressing this button and only the unicode char is injected this still is invisible. So may to enrich the input field with a special visualization for the soft hyphen char.

[...]

PR #10441, opened 2023-11-20
specialchars: add soft hyphen (U+00AD) button

michael-smt:
Related to #3006

## Proposed changes

When a languge has hyphens then show a button in the visual keyboard to insert a [soft hyphen](https://en.wikipedia.org/wiki/Soft_hyphen). 

## Checklist

## Other information

Using `WHITESPACE_REGEX` to get a red highlight in the editor where there are soft hyphens.

### DIFF
--- a/weblate/static/loader-bootstrap.js
+++ b/weblate/static/loader-bootstrap.js
@@ -405,7 +405,7 @@ function initHighlight(root) {
       let whitespace_regex = new RegExp(
         [
           "  +|(^) +| +(?=$)| +\n|\n +|\t|",
-          "\u00A0|\u1680|\u2000|\u2001|",
+          "\u00A0|\u00AD|\u1680|\u2000|\u2001|",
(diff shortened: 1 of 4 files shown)


nijel:
Thanks, but this only fixes part of the problem — it adds a button to add it, but doesn't make it appear in the editor when the hyphen is not needed in the editor to wrap the text. So, eventually, it looks like the button is not working.

Soft hyphen added shows nothing:

image

But when I extend the text, it appears:

image

Have you looked at options of improving this as well? Without this, this doesn't fully resolve #3006, but still is useful enough to merge.

PS: In 56e4ce6e14b72bea4d7246b01b04917791364e02 I've improved its appearance in the toolbar.

nijel:
I've tested that in Firefox, and it doesn't show it. The regexp makes it appear with a border only when rendered for me. Anyway, I think we can merge this as it is now as it clearly makes the situation better.

PR #10453, opened 2023-11-22
API: project endpoint to get and create labels

vinceh121:
## Proposed changes

This change adds the `GET,POST /api/projects/:projectSlug/labels/` endpoint, which allows listing labels, and users with the `project.edit` permission to create labels.

## Checklist

## Other information

This change induces an inconsistency with Unit serialization, as they currently serialize labels as strings. Should they now be serialized as objects?


nijel:
I think unit labels serializer should be changed, even if it is a breaking change.

Can you please also add tests for this new API?

vinceh121:
I've added tests, modified the tests on units, and now the `GET /api/units/:unitId/` endpoint serializes the labels as objects with id, name and color.

nijel:
Looks good, can you please update the documentation as well_

vinceh121:
I've added the two endpoints to the documentation.

PR #10574, opened 2023-12-08
specialchars: use iso symbol as soft hyphen button for better clarity

michael-smt:
## Proposed changes

A button with the text "SHY" might not be that clear for non-technical people. Replace it with something that looks similar to the ISO symbol for the soft hypen.

## Checklist

## Other information

Icon from [Wikipedia](https://en.wikipedia.org/wiki/Soft_hyphen) and [this proposal to Unicode](https://www.unicode.org/L2/L2017/17072-symbols-9995.pdf#page=5). U+2433 never got included in the [Control Pictures Unicode block](https://en.wikipedia.org/wiki/Control_Pictures), so no font has a representation for this.

### Comparison:

Before:
image

After:
image


nijel:

PR #10606, opened 2023-12-13
ui: include readonly in approved percentage listing to match with progress bars

michael-smt:
## Proposed changes

Count read-only strings to the approved percentage number shown in object listings.

## Checklist

## Other information

The progress bars display both approved and read-only strings as "Approved".
Also in the string status page both read-only strings and approved strings have the same colour.

Since #9936 the percentage of approved strings is shown as a number in the object listing, but without considering read-only strings.

This leads to a visual mis-match between the progress bars and the number:

image


nijel:

PR #10612, opened 2023-12-13
search: fix replacement limit notice

michael-smt:
## Proposed changes

Fix `limited` so the notice that only the first 250 matches are affected is shown.

## Checklist


nijel:

Issue #10651, opened 2023-12-18
No notification when source strings are added or updated


laoneo:
### Describe the issue

When a source strings get's updated or added in a component through a git commit in the en-GB "Monolingual base language file", then no notification mail goes out. In the notification settings I'v set "New string is available for translation" and "String was changed" to instant notification. Do I miss here something or is this a bug?

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [x] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

1. Choose a translated language in the user settings
2. Watch a project in the user settings
3. Set all notifications to Instant in the user settings
4. Commit a new string and an update string in a base monolingual file

### Expected behavior

A notification mail goes out to the user.

### Screenshots

![image](https://github.com/WeblateOrg/weblate/assets/251072/39ffcda6-ba13-438b-8385-8460e5f5b66a)

### How do you run Weblate?

Docker container

### Weblate versions

5.3

nijel:
There is currently no notification for source string being changed. Closest to that is "Unfinished strings exist" as that will trigger when some string will need a translation.

laoneo:
Any plan to integrate this? Coming from Transifex this is one of the key features as rarely no translators are polling weblate for changes. It keeps them engaged.

nijel:
This seems like a duplicate of https://github.com/WeblateOrg/weblate/issues/5602. Can you please review that and close this if this is really the case?

laoneo:
Not really, mine is more about notifications and not about pending tasks.

laoneo:
Or is there another way a translator who watches a project gets a notification when a source string has change or is added?

Aga-C:
@nijel @laoneo I can confirm that it's a bug. I haven't changed anything in the settings, but some time ago, I stopped receiving any emails about new strings to translate or that someone translated something. I am only receiving notifications about new comments or new components.

nijel:
@laoneo The notification for new string is there (New string is available for translation), the notification when source string changes is not there.

@Aga-C Unfortunately, we're not talking about a bug here, but about a missing feature. If you see a bug, please open a separate issue for that and provide more information (especially what notifications you have configured).

laoneo:
Thanks @nijel , but I do not get a mail when a new string is added either after a git rebase from the github webhook.

laoneo:
[...]

PR #10659, opened 2023-12-19
change: add separate event for new string from the repo

nijel:
## Proposed changes

This makes the history more clear (it doesn't start with "String changed in the repository") and allows to trigger new string notification on such event.

See #10651

## Checklist


laoneo:
Thanks, how can this be configured? Is it a new setting in the notification settings of the user profile?

nijel:
No, the notification setting is the same.  It will now trigger on new strings from the repository as well.

PR #11004, opened 2024-02-14
basic languages: factor in the default variant of languages without a base form

ray2c:
A regression was introduced a while back when a couple of languages were split into their respective couple of variants in [language-data](https://github.com/WeblateOrg/language-data/commits/main/weblate_language_data/languages.py?since=2023-04-18&until=2023-04-18) and, thereby, here got filtered out of the default list of basic languages altogether since they no longer present themselves in their base form.

## Proposed changes

Instead of adding them to `UNDERSCORE_EXCEPTIONS` to resurrect their pre-split position, this PR tries to adopt a more future-proof approach to the default list of basic languages to include the default variant of languages without a base form (when an alias for which is present) such that they would not be filtered out of the default list of basic languages altogether just because, some point in time, they cease to stand alone 'bare feet' in Weblate's list of languages.

## Checklist

## Other information

Subject to manual overrides, the list of basic languages is presented to the user in the "Start new translation" page, which says they would need to contact the project maintainer if they wish to add a language not on that list.

### DIFF
--- a/weblate/lang/data.py
+++ b/weblate/lang/data.py
@@ -4,6 +4,7 @@
 
 from django.utils.translation import pgettext_lazy
 from weblate_language_data import languages
+from weblate_language_data.aliases import ALIASES
 from weblate_language_data.ambiguous import AMBIGUOUS
 
 NO_CODE_LANGUAGES = {lang[0] for lang in languages.LANGUAGES}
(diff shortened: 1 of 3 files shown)


nijel:
Can you please also add tests for this behavior?

ray2c:
Hi, thanks for the prompt response!  (and the great software of course)
I'm sorry that this's actually my first PR (and first time coding Python), so please kindly bear with me : )
May I ask the silly question if there's any quick pointer for the tests which I might look at?
Cheers,
with best regards

ray2c:
Thanks for the prompt response!  Kindly bear with me as this is my first time coding in Python and submitting a PR : )
I've poked around a bit the tests; it seems there isn't a direct test on `BASIC_LANGUAGES` (forgive me if I'm wrong) but I've included an indirect test: I gather the 'add new language' form presented to the user (other than the owner, I think) by default is restricted to `BASIC_LANGUAGES` (if I'm not mistaken), adding the default variant of a language without a base form should succeed if it is included as one of the `BASIC_LANGUAGES`.
Kindly let me know if you've any insights or pointers on any improvement.  Thank you!

nijel:
Indeed, there is no direct test for this. Maybe it would be better to have one because this way it is not apparent what you are testing.

ray2c:
Sounds like a good idea.  A separate test case for the default list of basic languages has been added which hopefully might be more comprehensive and spell out the logic with some examples.

nijel:

ray2c:
Thank you very much for your comments and review!

PR #11064, opened 2024-02-21
improved description of Project website field

orangesunny:
## Checklist


nijel:
This would duplicate content from the translator instructions field (which used to be a URL field as well).

PR #11079, opened 2024-02-22
component: Improve duplication of components

### LINKED ISSUE #10896: Issues when duplicating a component
### Describe the problem

Creating a component that's a copy of another component feels more difficult than it needs to be.

1. It's not obvious that a component can be duplicated. When viewing a component there's no option to duplicate it. You have to create a new component, and choose the option to base the new component on an existing component. This is not discoverable and quite surprising.
2. The new component does not duplicate key settings from the source component. In my testing so far the settings that are not duplicated are at least:
a. The license
b. The contributor agreement text
c. The commit messages

These all have to be manually copied from an existing component to the new component.

I'd make it more obvious that duplication is possible with a copy icon (e.g,. something like https://www.flaticon.com/free-icon/copy_126498?term=copy&page=1&position=2&origin=tag&related_id=126498) next to the existing "browse" icon.

The lack of duplication of the other settings feels like a bug.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

ParthS007:
## Proposed changes

closes #10896 

## Checklist

- [x] I have tested that my improvement is effective and works.

**Duplicate Functionality in Dropdown**

Screenshot 2024-05-10 at 11 13 14

**Pre-selected Component in dropdown**
Screenshot 2024-05-29 at 12 27 13

## Release Notes

- **New Features**
  - Enhanced duplication process for components, allowing automatic cloning of additional settings.
  - Added a direct link for users with edit permissions to duplicate existing components in the management interface.
  
- **Improvements**
  - Updates to the form handling for language translations, improving metadata management.
  - Enhanced component selection during creation, making it easier to duplicate components with relevant attributes retained.

These updates aim to streamline workflows and improve user experience.

### DIFF
--- a/weblate/templates/component.html
+++ b/weblate/templates/component.html
@@ -96,6 +96,7 @@
       <li><a href="{% url 'manage-access' project=object.project.slug %}">{% trans "Users" %}</a></li>
       {% endif %}
       {% if user_can_edit_component %}
+      <li><a href="{% url 'create-component' %}?component={{ object.id }}#existing">{% trans "Duplicate Component" %}</li>
       <li><a href="{% url 'guide' path=object.get_url_path %}">{% trans "Community localization checklist" %}</a></li>
       <li><a href="{% url 'addons' path=object.get_url_path %}">{% trans "Add-ons" %}</a></li>
       {% endif %}
--- a/weblate/trans/forms.py
+++ b/weblate/trans/forms.py
@@ -1686,10 +1686,27 @@ class ComponentCreateForm(SettingsBaseForm, ComponentDocsMixin, ComponentAntispa
             "language_regex",
             "source_language",
             "is_glossary",
+            "agreement",
+            "merge_style",
+            "commit_message",
+            "add_message",
+            "delete_message",
+            "merge_message",
+            "addon_message",
+            "pull_message",
         ]
         widgets = {
             "source_language": SortedSelect,
             "language_code_style": SortedSelect,
+            "license": forms.HiddenInput(),
+            "agreement": forms.HiddenInput(),
+            "merge_style": forms.HiddenInput(),
+            "commit_message": forms.HiddenInput(),
+            "add_message": forms.HiddenInput(),
+            "delete_message": forms.HiddenInput(),
+            "merge_message": forms.HiddenInput(),
+            "addon_message": forms.HiddenInput(),
+            "pull_message": forms.HiddenInput(),
         }
 
 
--- a/weblate/trans/tests/utils.py
+++ b/weblate/trans/tests/utils.py
@@ -24,7 +24,7 @@ from django.utils.functional import cached_property
 from weblate.auth.models import User
 from weblate.configuration.models import Setting
 from weblate.formats.models import FILE_FORMATS
-from weblate.trans.models import Component, Project
+from weblate.trans.models import Category, Component, Project
 from weblate.utils.files import remove_tree
 from weblate.vcs.models import VCS_REGISTRY
 
@@ -167,6 +167,12 @@ class RepoTestMixin:
         self.addCleanup(remove_tree, project.full_path, True)
         return project
 
+    def create_category(self, project, **kwargs):
+        """Create test category."""
+        return Category.objects.create(
+            name="Test", slug="test", project=project, **kwargs
+        )
+
     def format_local_path(self, path):
         """Format path for local access to the repository."""
         if sys.platform != "win32":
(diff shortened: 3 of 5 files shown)


weblate/trans/views/create.py
```diff
@@ -521,6 +521,9 @@ def form_valid(self, form):
                 slug=form.cleaned_data["slug"],
                 vcs=component.vcs,
                 source_language=component.source_language.pk,
+                license=component.license,
+                agreement=component.agreement,
+                commit_message=component.commit_message,
```
nijel, line 526: Probably category should be kept as well. And there are more message configurations -            "add_message",             "delete_message",              "merge_message",             "addon_message",             "pull_message".

weblate/templates/snippets/list-objects.html
```diff
@@ -151,6 +151,9 @@
     <td class="actions" rowspan="2">
       <a href="{{ category.get_absolute_url }}" class="btn btn-link green" title="{% trans "Browse" %}" alt="{% trans "Browse" %}">{% icon "folder-multiple-outline.svg" %}</a>
     </td>
+    <td class="actions" rowspan="2">
+      <a href="{{ category.get_absolute_url }}" class="btn btn-link green" title="{% trans "Duplicate" %}" alt="{% trans "Duplicate" %}">{% icon "copy.svg" %}</a>
+    </td>
```
nijel, line 156: I'm not convinced this is a good place for such an action. At least, it should be only shown to users who have permission for that operation.

Moreover, adding another `<td>` will break the layout, it should be included in the existing actions `<td>`.

ParthS007: okay, thanks for the comments. Do we need duplicate item button on the `All Components` Page?

nijel: I think the Manage menu of a component might be a better place.

weblate/templates/component.html
```diff
@@ -96,6 +96,7 @@
       <li><a href="{% url 'manage-access' project=object.project.slug %}">{% trans "Users" %}</a></li>
       {% endif %}
       {% if user_can_edit_component %}
+      <li><a href="{% url 'create-component' %}#existing">{% trans "Duplicate Component" %}</li>
```
nijel, line 99: I'd expect the current component to be selected when clicking on the link here.

weblate/trans/views/create.py
```diff
@@ -520,10 +520,20 @@ def form_valid(self, form):
             return self.redirect_create(
                 repo=component.get_repo_link_url(),
                 project=component.project.pk,
+                category=component.category.pk,
```
nijel, line 523: Category can be `None` and then this fails with: `AttributeError: 'NoneType' object has no attribute 'pk'`

weblate/trans/views/create.py
```diff
@@ -520,10 +520,20 @@ def form_valid(self, form):
...
                 source_language=component.source_language.pk,
+                license=component.license,
+                agreement=component.agreement,
+                merge_style=component.merge_style,
+                commit_message=component.commit_message,
+                add_message=component.add_message,
+                delete_message=component.delete_message,
+                merge_message=component.merge_message,
+                addon_message=component.addon_message,
+                pull_message=component.pull_message,
```
nijel, line 536: While you include these in the URL, there is nothing to process these (`ComponentCreateForm` doesn't accept these). The additional fields either should be hidden inputs in the create form, or it should have tabs/folding for the additional parameters.

weblate/templates/component.html
ParthS007: Okay, To have `initial` value for `component` field in `get_form`, where I can set `initial` of `form` in `fetch_params` from the request?

```
form.fields['component'].initial = form.initial.get('component')
```

weblate/trans/views/create.py
ParthS007: We can keep the fields as hidden inputs in create form.

weblate/trans/forms.py
```diff
@@ -1686,10 +1686,27 @@ class Meta:
...
+            "add_message",
+            "delete_message",
+            "merge_message",
+            "addon_message",
+            "pull_message",
         ]
         widgets = {
             "source_language": SortedSelect,
             "language_code_style": SortedSelect,
+            "license": forms.HiddenInput(),
```
nijel, line 1701: Why this is hidden input now? This is not supposed to remove choices which were present before.

weblate/trans/tests/utils.py
```diff
@@ -188,6 +194,8 @@ def _create_component(
             raise SkipTest(f"File format {file_format} is not supported!")
         if "project" not in kwargs:
             kwargs["project"] = self.create_project()
+        if "category" not in kwargs:
+            kwargs["category"] = self.create_category(project=kwargs["project"])
```
nijel, line 198: It is okay for category to be not set. If you need it for a specific test, just create it there.

```suggestion
```

nijel:
Superseeded by https://github.com/WeblateOrg/weblate/pull/12320

PR #11128, opened 2024-03-02
Bots should avoid using site commit email

vinceh121:
## Proposed changes

This fix prevents bot users to commit using the email based on their username especially when `PRIVATE_COMMIT_EMAIL_OPT_IN = False`
Because bot users have a colon in their username, this results in invalid emails and 500 errors when committing and sending notifications.

## Checklist


nijel:

PR #11137, opened 2024-03-04
Add port 60000 to an scripts/test-database.sh

m-aciek:
docker-compose-postgresql.yml exposes db on host on port 60000. Small improvement for development setup docs. The content of the bash script is inlined in `contributing/tests.rst`.

## Checklist

- ~[ ] I have added tests that prove my fix is effective or that my feature works.~
- ~[ ] I have added documentation to describe my feature.~


nijel:
This snippet is also used in other setups where the default PostgreSQL port is used, so it needs someting more complex.

nijel:
I've instead inlined that in the documentation in 5d162af6fd.

Issue #671, opened 2015-03-06
Bulk handling for suggestions


paulpach:
A person came to my site and added hundreds of suggestions.

Currently I have to review each suggestion one at a time, and accept them one at a time.

It would be great if there was a "zen mode"  for reviewing large number of suggestions.
##

Frexuz:
Would like to see this too!

derikson:
This looks to me like a duplicate of #849.

mark-yougov:
It would be great if there was a "zen mode" for not only reviewing a large number of suggestions but also bulk approving all the suggestions in the language. ie having an "Approve All Suggestions" button.

exzhawk:
My implementation https://github.com/exzhawk/weblate/commit/ce5106970502e43825340ece0608beaa90b74c84
Known issue: No vote/spam function.
Screenshot
![image](https://user-images.githubusercontent.com/7038938/110248404-da16e200-7fab-11eb-9adf-3a1476fb4ccd.png)

comradekingu:
@exzhawk Avatars is a great idea.
I think it would be better to put them left-aligned (with a mouseover for the name) so that the text has a clear A-B comparison instead of being offset.

comradekingu:
@exzhawk Avatars is a great idea.
I think it would be better to put them left-aligned (with a mouseover for the name) so that the text has a clear A-B comparison instead of being offset.

nijel:
Related: https://github.com/WeblateOrg/weblate/discussions/8554#discussioncomment-4612940

juliancoffee:
Is it something that is accepted as a feature? If it just needs an actual implementation, I could try to create a PR.

nijel:
@juliancoffee Yes, it is accepted that we need such a feature. The exact design was not yet discussed, so please post your proposal here before implementing so that feedback can be provided.

juliancoffee:
Well, I quite like the design in https://github.com/WeblateOrg/weblate/issues/671#issuecomment-841025994.

It seems to include both a "Suggest" button with an exclamation sign, as well as all suggestions with accept/accept and edit/trash buttons. And looks quite clean. Except, I don't think we need to have suggestions to be different colours or smth, having an avatar and all the buttons gives IMO enough information to differentiate between current translation and suggestions.

And maybe add a vote button along with accept/accept and edit/trash.

nijel:
[...]

PR #11187, opened 2024-03-13
Draft: Suggestions in Zen mode

juliancoffee:
## Proposed changes

Solves #671, kind of.
At the time of writing, it is functional, but probably need to move most events to javascript to avoid resending Zen each time.
Also, I went via always show suggestions route, but I think that probably may be configured.

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced permission checks for user interactions on suggestions in the translation unit interface.
	- Added a form for users to submit suggestions, enhancing user engagement.
- **Bug Fixes**
	- Adjusted the layout of the voting buttons for improved visual positioning.

### DIFF
--- a/weblate/templates/snippets/suggestions.html
+++ b/weblate/templates/snippets/suggestions.html
@@ -25,7 +25,7 @@
         {% else %}
         {% perm 'suggestion.delete' suggestion as user_can_delete_suggestion %}
         {% if user_can_vote_suggestion or user_can_accept_suggestion or user_can_delete_suggestion %}
-        <div class="btn-float pull-right flip">
+        <div class="btn-float pull-right flip" style="top: -20px">
             {% if user_can_vote_suggestion %}
             <button type="submit" class="btn btn-link green" name="upvote" value="{{ suggestion.id }}" title="{% trans "Vote for" %}">{% icon "thumb-up.svg" %}</button>
             <button type="submit" class="btn btn-link red" name="downvote" value="{{ suggestion.id }}" title="{% trans "Vote against" %}">{% icon "thumb-down.svg" %}</button>
--- a/weblate/trans/views/edit.py
+++ b/weblate/trans/views/edit.py
@@ -893,6 +893,32 @@ def zen(request, path):
 
     search_result, unitdata = get_zen_unitdata(obj, project, unit_set, request)
     sort = get_sort_name(request, obj)
+    if (
+        request.method == "POST"
+        and "merge" not in request.POST
+        and (
+            "accept" in request.POST
+            or "accept_edit" in request.POST
+            or "accept_approve" in request.POST
+            or "delete" in request.POST
+            or "spam" in request.POST
+            or "upvote" in request.POST
+            or "downvote" in request.POST
+        )
+    ):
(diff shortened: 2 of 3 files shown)


weblate/templates/snippets/suggestions.html
```diff
@@ -25,7 +25,7 @@
         {% else %}
         {% perm 'suggestion.delete' suggestion as user_can_delete_suggestion %}
         {% if user_can_vote_suggestion or user_can_accept_suggestion or user_can_delete_suggestion %}
-        <div class="btn-float pull-right flip">
+        <div class="btn-float pull-right flip" style="top: -20px">
```
nijel, line 28: Please avoid inline styles.

weblate/trans/views/edit.py
```diff
@@ -893,6 +893,32 @@ def zen(request, path):
...
+            or "delete" in request.POST
+            or "spam" in request.POST
+            or "upvote" in request.POST
+            or "downvote" in request.POST
+        )
+    ):
+        # Handle accepting/deleting suggestions
+        _obj, unit_set, _context = parse_path_units(
+            request, path, (Translation, ProjectLanguage, CategoryLanguage)
+        )
```
nijel, line 912: This is already done at top of this function.

weblate/trans/views/edit.py
```diff
@@ -893,6 +893,32 @@ def zen(request, path):
...
+        and "merge" not in request.POST
+        and (
+            "accept" in request.POST
+            or "accept_edit" in request.POST
+            or "accept_approve" in request.POST
+            or "delete" in request.POST
+            or "spam" in request.POST
+            or "upvote" in request.POST
+            or "downvote" in request.POST
+        )
```
nijel, line 907: This is essentially a copy of code from `translate` function, it should be shared.

weblate/trans/views/edit.py
```diff
@@ -893,6 +893,32 @@ def zen(request, path):
...
+
+        checksum_form = ChecksumForm(unit_set, request.POST)
+        if not checksum_form.is_valid():
+            show_form_errors(request, checksum_form)
+            return HttpResponseBadRequest("Invalid checksum")
+
+        unit = checksum_form.cleaned_data["unit"]
+        # We are ignoring any redirects and responses here
+        _response = handle_suggestions(request, unit, "", "")
```
nijel, line 921: If the return value is not needed, there is no need to store it in a variable.

juliancoffee:
Thanks for the review! I want to continue working on it, just got interrupted somewhat.

weblate/templates/snippets/suggestions.html
juliancoffee: yeah, I don't like it too, I wonder what should I do instead? Changing the original class doesn't sound like a good idea either, I guess I could create new one?

weblate/trans/views/edit.py
juliancoffee: Can I just put such a function in the same file, or it should go to some utils.py file?

juliancoffee:
meow, bot

Issue #11168, opened 2024-03-11
Remove empty stale glossaries


nijel:
### Describe the problem

The glossary languages are automatically added when a new translation language is used. There is no logic in place to remove such automatically added glossary languages when they are no longer needed.

### Describe the solution you would like

Cleanup of not needed empty glossaries should happen automatically in `sync_glossary_languages` and when translations are removed.

AmanSal1:
@nijel  Can I give it a shot ?

nijel:
Sure, you're welcome.

AmanSal1:
Is this the right documentation to start ?
https://docs.weblate.org/en/latest/contributing/start.html

nijel:
Weblate will not work on Windows, so using Docker is recommended.

AmanSal1:
@nijel I have created a pull request can you review it ?

PR #11189, opened 2024-03-13
Implement automated cleanup of unused glossary languages

AmanSal1:
## Proposed changes

This commit implements a solution to identify and remove unnecessary empty glossary languages in the translation tool. It calculates a set of empty glossary languages by comparing the set of all glossary languages associated with a component to the set of languages with non-empty glossary entries in its translations. The identified empty glossary languages are then removed from the component.

Additionally, this commit addresses the removal of translation languages and their associated glossary languages. It identifies removed translation languages and iterates through their associated glossary languages, removing them from the component.

These changes streamline the management of glossary languages and ensure that only relevant languages are retained in the system.

## Checklist

## Other information

Fixes: #11168

### DIFF
--- a/weblate/glossary/tasks.py
+++ b/weblate/glossary/tasks.py
@@ -31,16 +31,23 @@ def sync_glossary_languages(pk: int, component: Component | None = None):
 
     # Identify unnecessary empty glossary languages
     empty_glossary_languages = set(component.glossary_languages.all()) - set(
-        component.translations.filter(glossary_entries__isnull=False).values_list("language", flat=True))
+        component.translations.filter(glossary_entries__isnull=False).values_list(
+            "language", flat=True
+        )
+    )
 
     # Remove unnecessary empty glossary languages
     for glossary_language in empty_glossary_languages:
         component.remove_language(glossary_language)
 
     # Identify removed translation languages and their associated glossary languages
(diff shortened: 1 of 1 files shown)


weblate/glossary/tasks.py
```diff
@@ -28,6 +28,29 @@ def sync_glossary_languages(pk: int, component: Component | None = None):
     )
     if not missing:
         return
+
+    # Identify unnecessary empty glossary languages
```
nijel, line 32: This would be executed only in case some languages are missing, see the above condition.

weblate/glossary/tasks.py
```diff
@@ -28,6 +28,29 @@ def sync_glossary_languages(pk: int, component: Component | None = None):
...
+    empty_glossary_languages = set(component.glossary_languages.all()) - set(
```
nijel, line 33: AttributeError: 'Component' object has no attribute 'glossary_languages'.

weblate/glossary/tasks.py
```diff
@@ -28,6 +28,29 @@ def sync_glossary_languages(pk: int, component: Component | None = None):
...
+        component.translations.filter(glossary_entries__isnull=False).values_list(
```
nijel, line 34: There is no such field `glossary_entries`.

nijel (changes requested):
Can you please also add a test case to verify your fix is effective?

Issue #11239, opened 2024-03-20
Allow project wide authenticated requests to download translations at once


viniciusmoreria:
### Describe the problem

Today we are able to download all translations of a project using this method: https://docs.weblate.org/en/latest/user/files.html#downloading-components-categories-or-projects 

That makes a request to this path: {{ host }}/download/{{ project }}/?format=zip:csv

But this requires us to leave our projects and components public, since this request is made by a service we have, that fetches all translations and upload them to S3 to be consumed elsewhere. 

### Describe the solution you would like

Being able to download all translations of a project in one request, like the one provided in the File menu, but expose it to the api so we can make authenticated requests to it, removing the necessity of making our project public.

### Describe alternatives you have considered

The alternative right now is to make a request to https://docs.weblate.org/en/latest/api.html#get--api-components-

And iterate on every component making new separated requests to https://docs.weblate.org/en/latest/api.html#get--api-translations-(string-project)-(string-component)-(string-language)-

PR #11258, opened 2024-03-23
feat(api): allow project wide translations download

viniciusmoreria:
## Proposed changes

Allows authenticated and permissioned users to download all translations for a specific project.

This exposes the `{{hosted}}/api/projects/{{project}}/file/` endpoint, accessible via a GET request, as proposed in https://github.com/WeblateOrg/weblate/issues/11239

## Checklist

## Other information

Closes https://github.com/WeblateOrg/weblate/issues/11239

### DIFF
--- a/weblate/api/tests.py
+++ b/weblate/api/tests.py
@@ -1811,6 +1811,52 @@ class ProjectAPITest(APIBaseTest):
         component = Component.objects.get(slug="local-project")
         self.assertEqual(component.enforced_checks, ["same"])
 
+    def test_download_project_translations_private(self) -> None:
+        project = self.component.project
+        project.access_control = Project.ACCESS_PRIVATE
+        project.save(update_fields=["access_control"])
+        self.do_request(
+            "api:project-file",
+            self.project_kwargs,
+            method="get",
+            code=404,
+            request={"format": "zip"},
+        )
+
+    def test_download_project_translations_prohibited(self) -> None:
+        self.authenticate()
+        self.user.groups.clear()
+        self.user.clear_cache()
+        self.do_request(
+            "api:project-file",
+            self.project_kwargs,
+            method="get",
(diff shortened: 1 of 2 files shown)


viniciusmoreria:
hey @nijel, this is one of my first non mobile pull requests, so I would like to confirm a few points before actually asking for review

i tried to follow the pattern of exposing the `file` path as i have found in other places of the codebase, it made sense to include it in the `ProjectViewSet`, does that seem right? if so, could you help me firing up the documentation changes?

thanks

weblate/api/tests.py
```diff
@@ -1811,6 +1811,63 @@ def test_create_component_enforced(self) -> None:
...
+        self.assertEqual(response.headers["content-type"], "application/zip")
+
+    def test_download_project_translations_target_language(self) -> None:
+        response = self.do_request(
+            "api:project-file",
+            self.project_kwargs,
+            method="get",
+            code=200,
+            superuser=True,
+            request={"format": "zip", "language_code": "en"},
```
nijel, line 1867: The test passes `language_code` in request, while the endpoint expects it in URL.

weblate/api/views.py
```diff
@@ -826,6 +829,31 @@ def destroy(self, request, *args, **kwargs):
...
+        instance = self.get_object()
+
+        if not request.user.has_perm("translation.download", instance):
+            raise PermissionDenied
+
+        components = instance.component_set.filter_access(request.user)
+        requested_format = request.query_params.get("format", "zip")
+
+        if language_code:
+            translations = Translation.objects.filter(
```
nijel, line 843: This is not covered by the tests (see above).

nijel (commented):
Thanks for your contribution. Can you please also add the new endpoint to the API documentation?

nijel:

Issue #11279, opened 2024-03-26
Changes missing after merge with upstream


bastianbeischer:
### Describe the issue

With this git commit graph:

```
(base) -> A (weblate)
       -> B (upstream)
```

where the relevant po file in the repository was modified in both A and B (but in a way which doesn't result in a conflict), modifications in A were done in weblate, modifications in B were done directly in the upstream repository.

weblate merged the remote repository (I have setup the weblate component to use merge style 'merge' and not rebase) and created the following graph:

```
(base) -> A (weblate old)  >
                             -> M (weblate new)
       -> B (upstream)     >
```

where M is a merge commit with parents A and B. So far so good.

However in the actual po file in commit M the changes in commit A are gone. The files in the commit M are identical to those in commit B (the git tree is the same). It looks like weblate was using some weird merge strategy which lead to changes in A being ignored?

When manually recreating the merge by creating a new branch on B and merging A into it, I get a different commit M2 which does not have this problem. Here changes from both A and B are present after the merge.

Any ideas how I could debug this further? I have already looked at the git reflog of the weblate internal repository (on the server), but it didn't provide any insights, except that apparently the merge was made with the 'ort' strategy, which seems to be reasonable...

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

Might need specific commits in the repository in question, so didn't try to build a reproducer

### How do you run Weblate?

Docker container

### Weblate versions

5.4.2

bastianbeischer:
Relevant part of git reflog:

```
7a2bc2d HEAD@{2024-03-18 11:04:27 +0000}: merge weblate-merge-tmp: Fast-forward
b13a2f1 HEAD@{2024-03-18 11:04:27 +0000}: checkout: moving from weblate-merge-tmp to dev
7a2bc2d HEAD@{2024-03-18 11:04:27 +0000}: merge dev: Merge made by the 'ort' strategy.
a1fe442 HEAD@{2024-03-18 11:04:27 +0000}: checkout: moving from dev to weblate-merge-tmp
b13a2f1 HEAD@{2024-03-18 11:04:27 +0000}: checkout: moving from dev to dev
```

but if I manually checkout `a1fe442` and then merge `b13a2f1` I get a commit which is different two `7a2bc2d` (in terms of git tree content, obviously)

nijel:
This is what Weblate does on merge:

[...]

PR #11280, opened 2024-03-26
Fix merging of po files with "git merge-file".

bastianbeischer:
Move output to correct output file destination

See #11279


nijel (approved):
Good catch, thanks for your contribution!

PR #11290, opened 2024-03-27
Fixed #10296: Tolerated interrobang swapping.

### LINKED ISSUE #10296: Automatic checker to tolerate interrobang swapping (!?/?!)
### Describe the problem

Languages have differing preference in the order of symbols representing astonishment. Namely, English tends to use _?!_ ("What?!") while Japanese _!?_ ("Nani!?"). The system checks unmatched punctuation between source and target, but every time the composite punctuation appears in such language pairs, it reports false positive warning.

Do not warn, or make an option that does not warn when `?!` and `!?` are swapped. Also for fullwidth `？！` and `！？`, and possibly for pre-combined `⁈` (U+2048) and `⁉` (U+2049) as well.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

manuelmcruz:
The updated code now allows for the swapping of "?!/!?" while translating and checks for the presence of interrobangs in both the source and translation. Tests have been implemented to validate the changes made to the code.

## Proposed changes

"EndQuestionCheck" and "EndExclamationCheck" do not produce warnings if an interrobang (?!/!?) is present. I added "EndInterrobangCheck" to verify if there is an interrobang in the original text or the translation (through the check_interrobang function) and generate warnings in case of mismatch. This function also permits the swapping of (?!/!?) without triggering any warnings. Tests have been implemented to validate the modifications made to the code.

Fixes: #10296 

## Checklist

### DIFF
--- a/weblate/checks/base.py
+++ b/weblate/checks/base.py
@@ -132,6 +132,17 @@ class Check:
             return False
 
         return (src in chars) != (tgt in chars)
+    
+    def check_interrobang(self, source, target):
+        """Ckeck whether interrobang is present."""
+        strings = ("!?","?!")
+        try:
+            src = source[-2:]
+            tgt = target[-2:]
+        except IndexError:
+            return False
+
+        return (src in strings) != (tgt in strings)
 
     def get_doc_url(self, user=None):
         """Return link to documentation."""
--- a/weblate/checks/base.py
+++ b/weblate/checks/base.py
@@ -132,10 +132,10 @@ class Check:
             return False
 
         return (src in chars) != (tgt in chars)
-    
+
     def check_interrobang(self, source, target):
         """Ckeck whether interrobang is present."""
-        strings = ("!?","?!")
+        strings = ("!?", "?!")
         try:
             src = source[-2:]
             tgt = target[-2:]
--- a/weblate/checks/base.py
+++ b/weblate/checks/base.py
@@ -133,17 +133,6 @@ class Check:
 
         return (src in chars) != (tgt in chars)
 
-    def check_interrobang(self, source, target):
-        """Ckeck whether interrobang is present."""
-        strings = ("!?", "?!")
-        try:
-            src = source[-2:]
-            tgt = target[-2:]
-        except IndexError:
-            return False
-
-        return (src in strings) != (tgt in strings)
-
     def get_doc_url(self, user=None):
         """Return link to documentation."""
         return get_doc_url("user/checks", self.doc_id, user=user)
--- a/weblate/checks/chars.py
+++ b/weblate/checks/chars.py
@@ -248,7 +248,7 @@ class EndQuestionCheck(TargetCheck):
     def check_single(self, source, target, unit):
         if not source or not target:
             return False
-        if source[-2:]  in {"?!", "!?"} or target[-2:] in {"?!", "!?"}:
+        if source[-2:] in {"?!", "!?"} or target[-2:] in {"?!", "!?"}:
             return False
         if unit.translation.language.is_base(("jbo",)):
             return False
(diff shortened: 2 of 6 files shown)


weblate/checks/base.py
```diff
@@ -133,6 +133,17 @@ def check_chars(self, source, target, pos, chars):
...
+    def check_interrobang(self, source, target):
+        """Ckeck whether interrobang is present."""
+        strings = ("!?", "?!")
+        try:
+            src = source[-2:]
+            tgt = target[-2:]
+        except IndexError:
+            return False
+
+        return (src in strings) != (tgt in strings)
```
nijel, line 145: I don't think this belongs to the base class.

nijel (commented):
Can you please also add a documentation for the check? `/manage.py list_check` will help you to generate the basic structure.

weblate/checks/base.py
manuelmcruz: Would you like me to add to an existing class or create a new one for the check?

weblate/checks/chars.py
```diff
@@ -290,6 +294,21 @@ def check_single(self, source, target, unit):
...
+    check_id = "end_Interrobang"
+    name = gettext_lazy("Mismatched interrobang")
+    description = gettext_lazy(
+        "Source and translation do not both end with an interrobang expression"
+    )
+
+    def check_single(self, source, target, unit):
+        if not source or not target:
+            return False
+        return self.check_interrobang(source, target)
```
nijel, line 309: Move it here (and simplify):

```suggestion
        strings = ("!?", "?!")
        return (source[-2:] in strings) != (target[-2:] in strings)
```

weblate/checks/base.py
nijel, line 145: I'd move it directly to the `EndInterrobangCheck` class, as it is unlikely this would be ever reused.

```suggestion
```

weblate/checks/chars.py
nijel: Adding support for pre-combined ⁈ (U+2048) and ⁉ (U+2049) might be reasonable as well (see https://github.com/WeblateOrg/weblate/issues/10296).

weblate/checks/tests/test_chars_checks.py
```diff
@@ -217,6 +218,21 @@ def test_eu(self) -> None:
         self.do_test(False, ("Text!", "¡Texte!", ""), "eu")
```
nijel, line 218: Please also add tests to `EndQuestionCheckTest` and `EndExclamationCheckTest` to verify interrobang skipping.

weblate/checks/chars.py
```diff
@@ -248,6 +249,11 @@ def _check_my(self, source, target):
     def check_single(self, source, target, unit):
         if not source or not target:
             return False
+        if (
+            source[-2:] in {"?!", "!?", "？！", "！？"}
+            or target[-2:] in {"?!", "!?", "？！", "！？"}
+        ) or (source[-1] in {"⁈", "⁉"} or target[-1] in {"⁈", "⁉"}):
```
nijel, line 255: This can be much simpler:

```suggestion
        if source.endswith(interrobangs) or target.endswith(interrobangs):
```

weblate/checks/chars.py
```diff
@@ -290,6 +296,28 @@ def check_single(self, source, target, unit):
...
+
+    def check_single(self, source, target, unit):
+        if not source or not target:
+            return False
+
+        interrobang_sets = [("!?", "?!"), ("？！", "！？")]
+
+        for sets in interrobang_sets:
+            if (source.endswith(sets)) != (target.endswith(sets)):
+                return True
```
nijel, line 316: Why are full width and standard width separated here? I think it is expected to use full width in full width languages like Chinese and standard width in Latin languages. So having source in English and target in Chinese, the interrobang width is expected to differ.

manuelmcruz: I thought every interrobang expression could be used in any language, and it was supposed to ensure with the code that both source and translation belonged to the same set. I fixed it to allow full width and standard width to be used interchangeably. I hope it's correct now!

nijel:

Issue #9089, opened 2023-04-15
Clickable comments in project view


comradekingu:
### Describe the issue

Can't click comments here
https://hosted.weblate.org/projects/lomiri/

Can click comments here
https://hosted.weblate.org/projects/lomiri/lomiri-camera-app/ 

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar issues in this repository.

### Steps to reproduce the behavior

Try to click comments from https://hosted.weblate.org/projects/lomiri/

### Expected behavior

Opening all comments for the respective component.
https://hosted.weblate.org/search/lomiri/lomiri-camera-app/?q=has%3Acomment&sort_by=-priority%2Cposition&checksum= , but in the actual editor (so the comments can be read).

### Screenshots

UBPorts/Lomiri
![bilde](https://user-images.githubusercontent.com/13802408/232194191-60911c1c-78f5-4acf-b62e-384725a87e26.png)
UBPorts/Lomiri/CameraApp
![bilde](https://user-images.githubusercontent.com/13802408/232194157-f28ae863-5c48-4b01-a35e-b531e88112dc.png)

### How do you run Weblate?

weblate.org service

### Additional context

Would be nice with (unsolved) comments for the entire project
https://hosted.weblate.org/search/lomiri/?q=has%3Acomment&sort_by=-priority%2Cposition&checksum=
by clicking the top field.
, or language (without any projects selected).

I don't think people are awere comments are language-specific.

themohammadsa:
Hey @comradekingu, I would like to pick this issue up. 

Issue #1:
Unable to click the comment on 'untranslated' links ie. the red ones.

Have a doubt though, Is the above the only issue?

nijel:
@ashishnagar19 You're welcome. Don't hesitate to ask if something is unclear.

ashishnagar19:
sir could you explain additional context more clearly.

ashishnagar19:
i did some changes how to run your website locally so i can see changes

ashishnagar19:
<a href="{{ translate_url }}{{ object.component.project.name }}?q=has:comment#comments" class="btn btn-link" alt="{% trans "Comments" %}">{% icon "pencil.svg" %}
     {% trans "Comments" %}
  </a>
          is this the correct way, i have make changes in list-objecst.html.

ashishnagar19:
[...]

PR #11292, opened 2024-03-27
Add clickable list object numbers using search URLs

BC46:
## Proposed changes

~~Added an additional list object component to `list-objects.html` which gets displayed when the project view is open.~~

Updated `list-objects.html` to add a search URL to every list objects number entry. `list-objects-number.html` has also been modified to ensure that the number can be clicked if the search URL is defined. This allows the user to click on all numbers within the table. This also includes comments as described in https://github.com/WeblateOrg/weblate/issues/9089.

## Checklist

## Other information

Resolves https://github.com/WeblateOrg/weblate/issues/9089.

In the repository I could not find any examples of frontend-related tests for specific UI elements. Since this pull request only includes frontend changes, I did not add any tests for it. Also since the proposed change is a very minor UX update which was previously already available in other views, additional documentation has not been added.

Please let me know if tests and/or documentation changes are needed regardless.

### DIFF
--- a/weblate/templates/snippets/list-objects.html
+++ b/weblate/templates/snippets/list-objects.html
@@ -159,13 +159,13 @@
     {% endif %}
     {% include "snippets/list-objects-percent.html" with percent=category.stats.translated_percent value=category.stats.translated query="q=state:>=translated" all=category.stats.all %}
     {% if not hide_details %}
-    {% include "snippets/list-objects-number.html" with value=category.stats.todo query="q=state:<translated" class="zero-width-640" %}
-    {% include "snippets/list-objects-number.html" with value=category.stats.todo_words query="q=state:<translated" class="zero-width-720" %}
-    {% include "snippets/list-objects-number.html" with value=category.stats.todo_chars query="q=state:<translated" class="zero-width-1200" %}
-    {% include "snippets/list-objects-number.html" with value=category.stats.nottranslated query="q=state:empty" class="zero-width-1400" %}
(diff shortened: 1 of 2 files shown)


weblate/templates/snippets/list-objects.html
```diff
@@ -260,7 +260,11 @@
 {% include "snippets/list-objects-number.html" with value=object.stats.nottranslated query="q=state:empty" class="zero-width-1400" %}
 {% include "snippets/list-objects-number.html" with value=object.stats.allchecks query="q=has:check" class="zero-width-768" %}
 {% include "snippets/list-objects-number.html" with value=object.stats.suggestions query="q=has:suggestion#suggestions" class="zero-width-900" %}
-{% include "snippets/list-objects-number.html" with value=object.stats.comments query="q=has:comment#comments" class="zero-width-1000" %}
+{% if name_source == "name" %}
+  {% include "snippets/list-objects-number-search-url.html" with value=object.stats.comments object=object query="q=has:comment#comments" class="zero-width-1000" %}
+{% else %}
+  {% include "snippets/list-objects-number.html" with value=object.stats.comments query="q=has:comment#comments" class="zero-width-1000" %}
+{% endif %}
```
nijel, line 267: Instead of fixing this just for comments, I think it should fix it for all the fields and make `snippets/list-objects-number.html` use the search URL if translate_url is not available.

BC46: Updated. Please let me know if you have any other suggestions.

nijel (approved):

Issue #3745, opened 2020-04-17
Do not commit changes for a translation that needs approval from a reviewer


goetas:
**Is your feature request related to a problem? Please describe.**

We are using a workflow there translations are reviewed by other translators before going online.

Current status:

1. Translator saves a translation.
2. Weblate updates the file (XLIFF)
3. Weblate or an user commits and pushes the changes 
4. App/project gets the new translation (since the changes have been already committed)
5. Reviewer rejects the translation because quality was low
6. go to step 1

A this stage the app was already deployed with a translation that was not approved.

Of course it is possible to leverage the `approved` XLIFF attribute, but no framework does it....

**Describe the solution you'd like**

Do not commit translations unless they are approved when using a workflow with Translation reviews.

Weblate should keep translations in its own database and write them to FS only when approved.

**Describe alternatives you've considered**
I've tried to use the "suggest" feature, but is less powerful than doing actual translations (all the translation progress gets lost).

This applies only when using a workflow with Translation reviews.

goetas:
The issue is that the XLIFF files have a single state, and as soon a file has been written, there is a potential that those changes might land in production. 

This makes kinda useless the review process... 

I saw the automatic comment added by the bot. Is there something more that I can provide?

nijel:
The review status is exposed in the XLIFF files exactly for this reason, but I can understand that it might be complicated to change other tooling to understand that.

goetas:
Even if changing the other tooling, there will be a moment when a wrong translation is on production.

Case:
1. dev uploads translation
2. translator translates
3. reviewer approves the translation
4. correct translation goes on production
5. translator updates the translation (status goes back to `approved=no`)
6. system deploys the translations, since `approved=no` the tooling framework will have to fallback to another language as the translation looks not approved :worried: 
7. reviewer approves the translation, `approved` is set to `yes`,  on the next deploy the text is updated.

With this, at step 6, even if the tolling supports the `approved` flag (or the other statuses), there will be always a moment when some wrong translation goes online.

Having tens of languages, it is hard to coordinate with translators and reviewers what is ready and what is not.

I'm not sure if XLIFF supports this, what seems needed here is to be able to have two xliff entries with the same key, where one could be approved, and the other no, I this way the frameworks should prefer the xliff translation with approved=yes.

nijel:
[...]

PR #11390, opened 2024-04-11
Allow custom key separators for JSON based file formats and exporters.

mkarol-neurosys:
## Proposed changes
During discussion in https://github.com/translate/translate/pull/5179 I realized that Weblate during export is losing key structure by not escaping dots in the natural language keys. Instead of escaping and unescaping separators throughout whole application I've implemented support for custom key separator allowing for simple import and export of files containing non-default key separators.

## Checklist

## Other information
Setting up key separator when creating component from scratch. Option is also available when creating component from other sources as well it is possible to change key separator in component settings.
image

Have a great day!
Michał Karol


weblate/templates/component.html
```diff
@@ -98,7 +98,6 @@
       {% if user_can_edit_component %}
       <li><a href="{% url 'guide' path=object.get_url_path %}">{% trans "Community localization checklist" %}</a></li>
       <li><a href="{% url 'addons' path=object.get_url_path %}">{% trans "Add-ons" %}</a></li>
-      <li><a href="{% url 'settings' path=object.get_url_path %}">{% trans "Settings" %}</a></li>
```
nijel, line 101: This seems wrong.

nijel (commented):
Thanks for your contribution, unfortunately I see several issues with it:

* The setting is done at the component level, but doesn't affect file loading when parsing files from the repository. That is confusing.
* When introducing parameters for file formats, we would rather need something more generic, not adding arbitrary parameters for few formats. See https://github.com/WeblateOrg/weblate/issues/1897
* Hard-coding separator to `.` would break YAML as it historically uses `->`, see https://github.com/translate/translate/blob/fea03126d2421454bb6d7f029eff2342c681dc30/translate/storage/yaml.py#L32 (I see you've limited this to JSON subclasses only, but this makes the setting behavior inconsistent)
* Generally, Weblate preserves whatever is in the file in the round-trip. What is the motivation to add this specific exporter and not just save changed files within Weblate?

weblate/templates/component.html
mkarol-neurosys: True. It should be other PR. In component overview you can have Settings menu item just right next to other Setttings menu item (from line 106), both pointing to the same url.

nijel: Ah, I see, fixed that in 76379e559e051536d5617209f9009f20ef6fbd38.

mkarol-neurosys:
Ad 1) When loading from vsc, after setting the source, there should be and option to specify key separator on `ComponentCreateForm`. Later, when celery task is performing `weblate/trans/models/component.py:2448 check_sync`  function it uses `weblate/trans/models/translation.py:268 self.component.file_format_cls` which uses wrapped file format class from component.

Ad 2) This is currently rather quick fix, I am planning to add changes to `translate` to allow for key separator customization.  If you want I can start with changes to the `translate` and utilizing those changes in weblate later on.

Ad 3) Like I said, it is first step. 

Ad 4) Unfortunately when some edits are made to the translations, there is no possibility to export changed i18nextv4 translations. JSON and JSON nested formats are stripping out pluras, and i18nextv4 exporter alone will nest keys in resulting JSON regardless whether the dot was inside the key eg. `{ "Natural language key. Will be nested": "some translation" }`.

nijel:
> Unfortunately when some edits are made to the translations, there is no possibility to export changed i18nextv4 translations.

You don't need to convert the format via exporters when changing existing i18nextv4 translation. Just download the file as it is. The format exporting is there for interoperability with a few formats, and not as a generic framework for converting anything to anything.

mkarol-neurosys:
Unfortunately there is no other way to download only reviewed and accepted translations. Exporters allow for selecting subset of translations without downloading whole file.

nijel:
I understand your need, but I see this more like a workaround than a solution.

* We need a generic solution for file format parameters, adding a database field for a single parameter for a few file formats is not going to scale. This is what https://github.com/WeblateOrg/weblate/issues/1897 describes.
* There is https://github.com/WeblateOrg/weblate/issues/3745 to let Weblate commit only approved strings what might be a better approach to your problem.

PR #11466, opened 2024-04-23
widgets: add PNG status badge

### LINKED ISSUE #6221: Add png status badge

On certain sites, we aren't allowed to use SVG files unless they're from trusted sources.
For my particular case, it's the [Visual Studio Code Marketplace](https://marketplace.visualstudio.com/), which disallows SVGs for security reasons. ([More Info](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#publishing-extensions))

Could we have a separate URL that generates a PNG version of `SVGBadgeWidget`? 

Then there could be 2 versions:
* SVG - `/widgets/{project-name}/-/status-badge.svg` / `/widgets/{project-name}/-/svg-badge.svg`
* PNG - `/widgets/{project-name}/-/status-badge.png`

**Additional context**

I think a nice way might be to add a utility in the `weblate.utils` module to convert an SVG to PNG in memory?
This way it's really efficient to build the image because it's just inserting variables into XML, and can be converted into the PNG as a final step.

Otherwise, it could be fine to just build the image like any of the other raster/bitmap widgets?

<details>
   <summary>The rest is only relevant if there is a way to convert an SVG with CSS/JavaScript to a PNG. I was just playing with SVGs, but had issues rasterizing them. Just noting the thoughts here.</summary>

---

I've tried looking a bit, but it seems most aren't able to handle things like `<style>` or `<foreignObject>`.
If there is a way to achieve this, then this approach could be used to resolve other issues in the repository as well.

https://github.com/WeblateOrg/weblate/issues/4198#issuecomment-662497131 - If this used an SVG base, then one can just use [getComputedTextWidth](https://developer.mozilla.org/en-US/docs/Web/API/SVGTextContentElement) to size the button according to the text.

```svg
<svg version="1.1" viewBox="0 0 512 512"
  xmlns="http://www.w3.org/2000/svg">
  <title>My Cool SVG</title>
  <desc>Base for Open Graph Thing</desc>

  <rect id="rect" rx="16" fill="grey"/>
  <text id="text" x="64" y="64" font-size="16" fill="white">Use Weblate</text>

  <script>
    const PADDING = 16;

    const rect = document.getElementById('rect');
    const text = document.getElementById('text');

    const textLength = text.getComputedTextLength();
    
    const textX = parseInt(text.getAttribute('x'));
    const textY = parseInt(text.getAttribute('y'));
    const textSize = parseInt(text.getAttribute('font-size'));

    const x = textX - PADDING;
    const y = textY - (textSize) - (PADDING / 2);
    const width = textLength + (PADDING * 2);
    const height = textSize + PADDING;

    rect.setAttribute('x', x);
    rect.setAttribute('y', y);
    rect.setAttribute('width', width);
    rect.setAttribute('height', height);
  </script>
</svg>
```

#6219 - If this used an SVG base, then one can just use CSS to truncate and use an ellipsis.

```svg
<svg version="1.1" viewBox="0 0 512 512"
  xmlns="http://www.w3.org/2000/svg">
  <title>My Cool SVG</title>
  <desc>Base for Open Graph Thing</desc>

  <style>
    div {
      width: 256px;
      overflow: hidden;
      text-overflow: ellipsis;
      white-space: nowrap;
    }

    #project-name {
      font-weight: bold;
    }
  </style>

  <foreignObject x="0" y="0" width="512" height="512">
    <div xmlns="http://www.w3.org/1999/xhtml">
      <span id="project">Project</span>
      <span id="project-name">Very Long Project Name</span>
    </div>
  </foreignObject>
</svg>
```
</details>

nijel:
## Proposed changes

It is SVG rendered into PNG using rsvg, which is already pulled as a dependency for Pango/Cairo stack.

Fixes #6221

## Checklist


tipa:
@nijel after this change, I get tons of error reports because Weblate throws errors when trying to get the PNG status badge:
```
Exception Type: AttributeError at /widget/my_app/status-badge.png Exception Value: 'gi.repository.Rsvg' object has no attribute 'Rectangle'
Raised during: weblate.trans.views.widgets.render_widget
```

seems like that `rsvg` dependency is not automatically pulled in for me, even though I have `pycairo-1.26.0` installed (via pip)

nijel:
Ah, this interface was added in librsvg 2.46 (see https://lazka.github.io/pgi-docs/Rsvg-2.0/classes/Handle.html#Rsvg.Handle.render_document), so you probably have an older version. I've adjusted the required version in e58b4099dbb1d978ca65d71cb4583f0c2656bade.

nijel:
f3a94caef51bc3ebf97c1866038260c7b161bd63 should make it work with older versions as well.

tipa:
That worked - thanks for the quick help!

PR #11512, opened 2024-04-27
Update documentation on file upload permissions

danimoh:
## Proposed changes

- I updated the documentation with which permissions are required for the individual file upload methods.
- I also added documentation for the `approve` method, which was previously missing. Please confirm that its accurate and actually what this method does as I just assumed that this is what it does, based on what the `suggestion` method does.

## Checklist

## Other information

I did not build the documentation locally, or tested that the references/links are correct. I'd appreciate if you could do that on your side.


nijel:

PR #11513, opened 2024-04-27
Allow replacing source file with appropriate unit permissions

danimoh:
## Proposed changes

In addition to previous permission `component.edit`, now also allow the `replace` upload method on source file upload, if the `unit.add`, `unit.delete` and `unit.edit` permissions are present.

This allows to give users access to replacing the source file, without having to give them full `component.edit` permissions.

Background: I think, the previous requirement of `component.edit` is rather confusing and not intuitive at all, as I struggled to setup Weblate in a way that other users without admin rights can replace the monolingual source file, either via UI or API. I've tried to achieve that with different settings like "Manage strings", "Edit base file", even creating new user roles via the Django interface with different permissions like "Add new string", "Remove a string", "Edit source strings", "Overwrite existing strings with upload" etc., all without success.
The documentation was not much help (which is why I also opened PR #11512), and only digging into Weblate's code unveiled what user permissions are required to achieve my goal.

## Checklist

## Other information

- This PR is meant as a draft / inspiration. I would appreciate, if you could take further changes from here, if needed, as I lack any in-depth knowledge on Weblate.
- For example, I was not sure whether any other permissions should be added into the mix, like `unit.override` or `upload.overwrite`.
- Also, I was not sure, whether the proposal fully covers correct behavior for both, monolingual and bilingual files, and "Manage strings" and/or "Edit base file" being enabled/disabled in the component settings, whether a source or translation file is being updated, and whether the file is uploaded via the UI or via the API.
- I have not run or tested this feature locally. I would appreciate, if you could do that on your side.


nijel:

Issue #10596, opened 2023-12-12
Gerrit integration: configure gitreview.username from ssh://user@gerrit... push-URL / provide way to set username from UI


cloph:
### Describe the problem

gerrit integration ignores the username provided in push-URL and instead uses the system user's name, at least in the presence of a .gitreview file specifying host, port, project and branch but omitting a username.

That then obviously fails due to lack of permissions and git-review used by weblate asks for a username to try again (and since it is run non-interactive mode it only gets EOF)

### Describe the solution you'd like

Weblate should set up gitreview.username configuration when setting up a component/new repository. Either by adding a dedicated input field in the VCS settings, or more straightforward reading it from the pushurl if in ssh://user@..... format

### Describe alternatives you've considered

Alternatively it can be solved by telling the user in documentation to manually set the git config in the data/vcs... clone

nijel:
This should be pretty easy to implement by customizing `configure_remote` in the `GitWithGerritRepository` class.

harriebird:
Hi guys, check my PR #11585 for this issue.

PR #11585, opened 2024-05-06
Set the value of gitreview.username from the gerrit push URL

harriebird:
## Proposed changes
This will automatically get the username from the push URL when using Gerrit as VCS
This solves issue #10596

## Checklist

### DIFF
--- a/weblate/vcs/git.py
+++ b/weblate/vcs/git.py
@@ -548,6 +548,18 @@ class GitWithGerritRepository(GitRepository):
         """Return VCS program version."""
         return cls._popen(["review", "--version"], merge_err=True).split()[-1]
 
+    def get_username_from_url(self, url) -> str:
+        if url is not None:
+            if url.startswith("git@"):
+                return url.split(":")[-1].split("/")[0]
+            elif url.startswith(("ssh://", "https://")) and '@' in url:
+                return url.split("//")[-1].split("@")[0]
+            elif url.startswith(("ssh://", "https://")):
+                return url.split("//")[-1].split("/")[1]
+            else:
+                return ""
+        return ""
+
     def push(self, branch) -> None:
         if self.needs_push():
             try:
@@ -573,7 +585,7 @@ class GitWithGerritRepository(GitRepository):
         Gets the gerrit username from push URL and
         sets it as the value of gitreview.username.
         """
-        gerrit_user = push_url.split("@")[0].split("//")[1]
+        gerrit_user = self.get_username_from_url(push_url)
         self.config_update(("gitreview", "username", gerrit_user))
         super().configure_remote(pull_url, push_url, branch, fast)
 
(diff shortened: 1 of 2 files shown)


weblate/vcs/git.py
```diff
@@ -564,6 +564,19 @@ def get_remote_configure(
...
+    def configure_remote(
+        self, pull_url: str, push_url: str, branch: str, fast: bool = True
+    ) -> None:
+        """
+        Configure remote repository.
+
+        Gets the gerrit username from push URL and
+        sets it as the value of gitreview.username.
+        """
+        gerrit_user = push_url.split("@")[0].split("//")[1]
```
nijel, line 576: This needs to be more robust to handle all possible URL schemes. Your code will crash on SSH push URL like `git@github.com:WeblateOrg/weblate.git`.

See `parse_repo_url` method above in this file for an example that work well.

nijel (commented):
Can you please also add tests to verify this actually behaves as intended?

weblate/vcs/git.py
nijel: Actually it even fails within the testsuite: https://github.com/WeblateOrg/weblate/actions/runs/8979415860/job/24661366078?pr=11585#step:7:4068

harriebird: I will improve this  to handle possible URL schemes. It was initially assumed that URLs coming from Gerrit will look like this:
ssh://<user>@<host>:<port>/<repo>.git

Actually, it has different URL schemes.

nijel: AFAIK this can be any URL Git supports.

weblate/vcs/git.py
```diff
@@ -564,6 +595,24 @@ def get_remote_configure(
...
+
+        Gets the gerrit username from push URL and
+        sets it as the value of gitreview.username.
+        """
+        gerrit_user = ""
+        if push_url is not None:
+            if push_url.startswith("ssh://"):
+                gerrit_user = push_url.split("@")[0].split("//")[-1]
+            elif push_url.startswith("git@") or push_url.startswith("https://"):
+                _, _, gerrit_user, _ = self.parse_repo_url(push_url)
```
nijel, line 612: I don't think that owner as returned by parse_repo_url is what you want here - it is owner of the repo at GitHub (for `https://github.com/WeblateOrg/weblate.git` it would be `WeblateOrg`).

harriebird: Hi @nijel, I updated how it was implemented. I also added test with different URLs.

weblate/vcs/git.py
```diff
@@ -564,6 +575,19 @@ def get_remote_configure(
...
+        """
+        Configure remote repository.
+
+        Gets the gerrit username from push URL and
+        sets it as the value of gitreview.username.
+        """
+        gerrit_user = self.get_username_from_url(push_url)
+        self.config_update(("gitreview", "username", gerrit_user))
+        super().configure_remote(pull_url, push_url, branch, fast)
+
```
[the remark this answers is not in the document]
harriebird: I agree with this @nijel . I removed the override implementation of `configure_remote`  inside `GitWithGerritRepository` and transferred the setting of `gitrefview.username` value to the `get_remote_configure`.

nijel:

PR #11640, opened 2024-05-15
feat(notifications): add notification on strings needing updating

### LINKED ISSUE #10651: No notification when source strings are added or updated

nijel:
## Proposed changes

I was originally considering a separate notification for this, but at last I think that having just one here is a better approach.

Fixes #10651

## Checklist


laoneo:
So when a string in the souorce file is updated, what happens with the translations? Do they get deleted or is it just a notification? Transifex for example (where I come from) deletes all the translations. This makes very much sense to me as the translations do not reflect anymore the current source string.

Issue #11106, opened 2024-02-28
Add announcements to Categories


emmapeel2:
### Describe the problem

We can add announcements on components, projects, and soon project languages. 

We would like to have announcements on a project category as well.

### Describe the solution you would like

Announcements on categories are useful to communicate information to translators.

See discussion in https://github.com/WeblateOrg/weblate/discussions/10939

danqixu:
@emmapeel2 Hi, could you assign this issue to me? I am a student and having OSS course, would like to give a try.

emmapeel2:
@nijel

nijel:
@danqixu You can start working on the issue right now, we don't assign issues to outside collaborators.

nijel:
@harriebird This is similar to https://github.com/WeblateOrg/weblate/issues/11003 which you've recently addressed.

harriebird:
Hi @nijel, I will check this issue.

harriebird:
Hi guys, check my PR #11644 for this issue.

emmapeel2:
I tried again today to post an announcement to a category. The menu option appears and I was able to create the announcement.

BUT it didn't show only on the category: It showed on the Project level (in all components).

nijel:
https://github.com/WeblateOrg/weblate/pull/19382 should fix that (and few other missing bits around category-level announcements).

PR #11644, opened 2024-05-15
Add announcement for category page

harriebird:
## Proposed changes
Add announcement form on category page.
To achieve this, changes are made to the models `Change` and `Announcement`.
This solves issue  #11106.

## Checklist


weblate/trans/models/announcement.py
```diff
@@ -77,6 +78,13 @@ class Announcement(models.Model):
         blank=True,
         on_delete=models.deletion.CASCADE,
     )
+    project_category = models.ForeignKey(
```
nijel, line 81: Why not just `category`?

harriebird: Hi @nijel, I named it `project_category` to avoid confusion with `category` that is already existing in this model.  The existing `category` stores the color scheme (info, warning, danger, success) of the announcement.

nijel: Good point, I didn't notice it. But for consistency I'd prefer to use category for the same thing. So it's probably time to rename existing category here to something else. Maybe call that severity? That should probably be done in a separate pull request

harriebird: Yes @nijel, we can rename it to a more relevant names. We can name the existing category field as `severity ` or `type`.

nijel: I think severity is a good choice here

harriebird: Hi @nijel, the `category` field in the `Announcement` model is now renamed to `severity`. The `category` field is now a foreign key to the `Category` model.

weblate/trans/views/basic.py
```diff
@@ -452,11 +452,15 @@ def show_project(request, obj):
 def show_category(request, obj):
     user = request.user
 
-    all_changes = (
-        Change.objects.for_category(obj).filter_components(request.user).prefetch()
+    last_changes = Change.objects.last_changes(
+        user, project=obj.project, category=obj
+    ).recent()
```
nijel, line 457: `for_category` filters for changes in all components within category. These are now gone from the results, as you only filter changes having explicitly set `category`. Instead, I think that `for_category` should be adjusted to additionally include entries with matching `category`.

harriebird: Hi @nijel, I have reverted this change and fixed the `for_category`.

weblate/trans/models/change.py
```diff
@@ -43,7 +43,9 @@ def content(self, prefetch=False):
         return base.filter(action__in=Change.ACTIONS_CONTENT)
 
     def for_category(self, category):
-        return self.filter(component_id__in=category.all_component_ids)
+        if hasattr(category, "component"):
+            return self.filter(component_id__in=category.all_component_ids)
+        return self.filter(category=category)
```
nijel, line 48: It should be more something like this:

```suggestion
        return self.filter(Q(component_id__in=category.all_component_ids) | Q(category=category))
```

harriebird: Hi @nijel, the code change you suggested in the `for_category` function has been included.

weblate/trans/migrations/0017_announcement_severity_change_category_and_more.py
```diff
@@ -0,0 +1,55 @@
+# Generated by Django 5.0.6 on 2024-05-21 06:12
```
nijel, line 1: Add lincesing info:

```suggestion
# Copyright © Michal Čihař <michal@weblate.org>
#
# SPDX-License-Identifier: GPL-3.0-or-later

# Generated by Django 5.0.6 on 2024-05-21 06:12
```

nijel:

emmapeel2:
I have tried this new feature, but when I post an announcement to the Tor Browser category at https://hosted.weblate.org/projects/tor/tor-browser/ , the announcement is displayed in the whole project, not only on the category.

PR #11679, opened 2024-05-21
ci: use uv to install packages

nijel:
## Proposed changes

- use uv instead of pip for better peformance
- drop cache as uv cache is way bigger than pip

## Checklist


nijel:
CI runs faster even with caching disabled. Leaving caching out for upstream addressing, see following issues:
* https://github.com/actions/setup-python/issues/822 
* https://github.com/actions/setup-python/pull/762 
* https://github.com/astral-sh/uv/issues/2231

PR #11697, opened 2024-05-23
feat(accounts): add autocomplete hints for password field

nijel:
## Proposed changes

This makes it possible to generate new passwords instead of filling in existing one.

Thanks to @spaze for hint on this.

## Checklist


spaze:
Nice, thanks for adding that! 👍

Issue #10787, opened 2024-01-15
Delayed removal of strings


nijel:
### Describe the problem

Upon removal of a string, Weblate immediately updates translation files. This makes it slow (see https://github.com/WeblateOrg/weblate/issues/4797#issuecomment-1886341450).

### Describe the solution you would like

Removal should be delayed until the commit time, same as we do with adding new strings.

PR #11717, opened 2024-05-26
create_translations: Enable background processing by default.

mont29:
Make administrative commands like `Update`, `Reset`, `Add Language` etc. from the admin web UI reasonably responsive, by running the actual translation updates asynchronously in background tasks.

From a user PoV, it means that the execution of these tasks now returns (very) fast, and that they are redirected to the progress page, where they can follow the background task progression.

## Technical Overview

Adds a new `run_async` parameter to `create_translations` function, `False` by default.

When `False`, the update will be processed immediately, if the lock fails and Celery is available it will be scheduled asynchronously instead (this is keeping the current behavior).

When `True` (and the debug `CELERY_TASK_ALWAYS_EAGER` setting is not `True`), the update is sent to a Celery task and will happen in the background.

A new thin wrapper around the actual update logic has been added, `create_translations_task`, which just handles the locking logic. It is designed to be called directly by the synchronous mode of `create_translations`, and all the Celery tasks. This avoids some fairly confusing code flow (`create_translations` creating the Celery task which calls `create_translations` again), especially since the calls from async tasks should never trigger new async tasks (they should rather fail, and Celery will then schedule a new attempt for the task).

The `create_translations` calls from `add_new_language`, `do_update`, `do_reset`, `do_file_scan`, `reset_after_save`, `add_new_language` and `handle_store_change` are made asynchronous. Calls to `create_translations` from the addons have not been modified, and remain synchronous.

The matching views have also been updated, and will redirect to the `component_progress` url, showing real-time progress of the update.

The `component_progress` page has been updated to be able to also redirect to a translation URL, and not only the component URL. This is used by the `new_language`  view to redirect the user to the newly created language page, once the update is complete, instead of the component page.

NOTE: This commit _may_ also fix a theoretical re-entrant loop? If the call to `create_translations` from the Celery task failed again to acquire the component lock, existing code would add a new task. Adding Celery tasks from within other tasks is not recommended.

## Checklist

- ~~[x] I have added tests that prove my fix is effective or that my feature works.~~ N/A

## Other information

Current weblate code gives a fairly weird and hard to understand behavior on admins commands that update translations, when the component is big (over 30k messages and over 50 languages). In our case (Blender UI and manual translations), an update can take up to over 2h in total.

This will always reach the timeout of the WSGI server - and in the over 2h cases, silently error when the process is terminated, leaving some languages unprocessed.

However, once in a while, the Update request will return successfully almost immediately, and then in the UI we get this nice clickable 'clock' icon showing that the background update task is running - with a nice 'update' page to follow the process even. The main target of this PR is to get that last behavior systematically.


nijel:
Generally, this is something we have to in the long term do to avoid overly long HTTP requests. The main reason for those being synchronous was to bring the user to the final state. And I think we need some UI improvements to see background operations progress immediately.

For example, when adding a new translation via `add_new_language`, the user is redirected to the newly created translation. But when the processing will be async, he ends up on empty translation, which is just being populated in the background. Yes, there is a way to watch this from the UI, but it's IMHO currently not reachable from the translation page.

A similar situation is with repository operations being asynchronous. Instead of seeing a new state like now, you would still see an old state and need to reload the page until it is updated.

I think the best approach would be to use JavaScript to submit the async operation (maybe using API, not a dedicated UI endpoints), immediately show the task progress and redirect once the task has been completed. We already do something similar with the automatic translation, but the code there doesn't redirect, it just shows the progress.

mont29:
Thanks for the suggestions. I'll try to work on this asap :)

mont29:
I added a redirect to the 'progress' page once git operations are done, for the Update/Reset/Rescan operations.

This is not (yet) change to use JS/API , but it also is easier to implement from existing code.

E.g. clicking on Reset redirects to this:

image

weblate/trans/models/component.py
```diff
@@ -1678,7 +1678,7 @@ def do_update(self, request=None, method=None):
         if result:
             # create translation objects for all files
             try:
-                self.create_translations(request=request)
+                self.create_translations(request=request, run_async=True)
```
nijel, line 1681: This should no longer raise exception, so try/except block can be removed.

mont29: I think that `FileParseError` exceptions can still be raised in case `settings.CELERY_TASK_ALWAYS_EAGER` is `True`? Then `_create_translations` is still called directly.

mont29:
Think this PR is ready for final review now? Also updated it against latest main.

weblate/trans/models/component.py
nijel: That's true, this code-path still exists.

weblate/trans/models/component.py
```diff
@@ -2362,33 +2362,33 @@ def create_translations(
...
         from_link: bool = False,
         change: int | None = None,
+        run_async: bool = False,
     ) -> bool:
         """Load translations from VCS."""
-        try:
+        if not run_async or settings.CELERY_TASK_ALWAYS_EAGER:
+            # Asynchronous processing not requested or not available, run the update
+            # directly from the request processing.
+            # NOTE: In case the lock cannot be acquired, an error will be raised.
```
nijel, line 2371: One question before doing the full review here: Did you change all calls to run async now?

* If yes, then the `run_async` parameter might not be needed anymore.
* If no, then this will now raise `WeblateLockTimeoutError`, but it did fallback to background processing before and this behavior should be restored.

mont29: Sorry for the late reply, had to dig a bit in the code again ;)

Short answer is: we still need this `run_async` parameter:
* In some cases, I did not feel confident enough about the calling context of `create_translations` to switch it to async (e,g, from some addons, like `MsgmergeAddon`,  `post_commit` function from `git.py`, and other not-directly-related areas like the `new_language` view or `Translation.handle_store_change`).
* We definitively do not want async behavior when calling from the async tasks themselves.

In fact, I wonder if we should not split the `create_translations` in two separate functions:
* `create_translations`: 'public' entry point, essentially the same thing as what is in this PR, with the 'fallback to async' behavior restored. Then we can later check all remaining synchronous calls to it and maybe entirely remove the `run_async` option.
* `create_translations_task`: very thin wrapper around `_create_translations` (just the 'context' lock around it basically), called exclusively from `create_translations`, and the celery tasks code.

Think it would make the general code flow easier to follow and understand, and likely also help in my next goal: to split `_create_translations` itself into smaller bits, and try to create a Celery canvas with them (afaict, once the source language has been updated, it should be possible to update all translations in parallel?).

nijel: The add-ons should probably stay as it is now (prefer sync, but fallback to async when needed). The reason is that when doing async, it might trigger additional add-on executions which could be avoided when doing that sync.

The `new_language` view is effectively just an optimization to call `create_translations` just once when adding multiple ones, and otherwise it would get called directly via `Component.add_new_language`. That one would need a view change as upon adding user is redirected to the newly created language and expects to see the strings to translate there. These would be missing if `create_translation` did not happen. But in case sync update can not be performed, it is confusing to users even right now.

The `Translation.handle_store_change` is called upon upload which replaces whole file and when deleting a string. The first thing could be changed to async and show the progress, the second should be addressed by https://github.com/WeblateOrg/weblate/issues/10787.

Doing the translations update in parallel is not easily possible because it heavily relies on using shared component object (to cache source string, collect pending source string checks, batch checks). While it would be technically possible to collect that via Celery results, it would complicate the code, and it would probably make the performance worse (for example because of fetching source strings from the database for each translation separately).

mont29: Thanks for all the info! I'll try to update the PR accordingly this weekend.

mont29:
think this is ready for another pass of review now :)

weblate/templates/component-progress.html
```diff
@@ -28,14 +28,14 @@
...
-<a href="{% url return_url path=object.get_url_path %}" class="btn btn-primary" id="progress-return">{% trans "Return to the component" %}</a>
+<a href="{% url return_url path=return_target.get_url_path %}" class="btn btn-primary" id="progress-return">{% trans "Return to the target" %}</a>
 {% if user_can_edit_component %}
 <a href="#" class="btn btn-danger" id="terminate-task-button">{% trans "Abort the update" %}</a>
 {% endif %}
   </div>
   <div class="form-group">
     <input type="checkbox" id="progress-redirect" checked="checked">
-    <label for="progress-redirect">{% trans "Return to the component after completion" %}</label>
+    <label for="progress-redirect">{% trans "Return to the target after completion" %}</label>
```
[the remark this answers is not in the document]
mont29: Agreed, was not very happy with it either... Maybe simply "translations" ?

weblate/templates/component-progress.html
```diff
@@ -28,14 +28,14 @@
 <div class="panel-footer">
 <form class="form-inline">
   <div class="form-group">
-<a href="{% url return_url path=object.get_url_path %}" class="btn btn-primary" id="progress-return">{% trans "Return to the component" %}</a>
+<a href="{% url return_url path=return_target.get_url_path %}" class="btn btn-primary" id="progress-return">{% trans "Return to the target" %}</a>
```
nijel, line 31:
```suggestion
<a href="{% url return_url path=return_target.get_url_path %}" class="btn btn-primary" id="progress-return">{% trans "Go back" %}</a>
```

weblate/templates/component-progress.html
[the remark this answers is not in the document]
nijel:
```suggestion
    <label for="progress-redirect">{% trans "Go back after completion" %}</label>
```

nijel (commented):
How about making the messages simpler?

weblate/templates/component-progress.html
nijel: Just after I posted, I realized that this is not really “back” in some situations. Especially when a new component is being imported, you don't go back, but to the newly created component.

nijel:
```suggestion
<a href="{% url return_url path=return_target.get_url_path %}" class="btn btn-primary" id="progress-return">{% blocktranslate %}Continue to {{ return_target }}{% endblocktrans %}</a>
```

weblate/templates/component-progress.html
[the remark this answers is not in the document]
nijel:
```suggestion
    <label for="progress-redirect">{% trans "Continue after completion" %}</label>
```

nijel:
I made some changes to the strings, but still would like to see a review from @orangesunny here.

orangesunny:
> I made some changes to the strings, but still would like to see a review from @orangesunny here.

I think you made it well and would use it. I only have some itching that `Continue after completion` misses the object and could work better like `Continue <somewhere> after completion`. But it probably not essential for understanding in the UI.

weblate/trans/views/settings.py
```diff
@@ -339,18 +339,28 @@ def announcement_delete(request: AuthenticatedHttpRequest, pk):
 
 @login_required
 def component_progress(request: AuthenticatedHttpRequest, path):
-    obj = parse_path(request, path, (Component,))
+    """
+    Show component update progress.
```
orangesunny, line 343: shouldn’t it be `Shows component....`? Not sure if this is a name of a feature or a description.

weblate/trans/models/component.py
```diff
@@ -2365,33 +2365,58 @@ def create_translations(
...
+        self,
+        force: bool = False,
+        langs: list[str] | None = None,
+        request=None,
+        changed_template: bool = False,
+        from_link: bool = False,
+        change: int | None = None,
+    ) -> bool:
+        """
+        Load synchronously translations from VCS.
```
orangesunny, line 2411: `Load translations from VCS synchronously.` seems better

orangesunny (changes requested):
Good to go after solving the comments.

weblate/trans/views/settings.py
mont29: It's 'imperative mood' docstring. Doing otherwise will trigger [error D401](https://docs.astral.sh/ruff/rules/non-imperative-mood/) from ruff checks.

weblate/trans/views/basic.py
```diff
@@ -756,13 +757,21 @@ def new_language(request: AuthenticatedHttpRequest, path):
...
                             gettext(
-                                "The translation will be updated in the background."
+                                "All languages have been added, updates of translations are in progress."
                             ),
                         )
+                        result = "{}?info=1".format(
+                            reverse(
+                                "component_progress",
+                                kwargs={"path": result.get_url_path()},
+                            )
```
nijel, line 773: This redirect breaks for project and category scopes, see https://github.com/WeblateOrg/weblate/issues/12413

Issue #3558, opened 2020-03-04
Improve date selection for reports [$32]


nijel:
**Is your feature request related to a problem? Please describe.**

Current reporting period selection is not user-friendly:

![image](https://user-images.githubusercontent.com/212189/75873029-7d6c6e80-5e0f-11ea-84e4-18cd280ab98f.png)

**Describe the solution you'd like**

I'd like to see something like this:

![image](https://user-images.githubusercontent.com/212189/75872958-60d03680-5e0f-11ea-8722-c12b09e86f4c.png)

msanden:
Can I claim this issue?

nijel:
Sure, you're welcome.

rajat641:
Hi @nijel , Can I work on this. There is no update on this for a long time.

nijel:
You're welcome! Yes, there are about two hundredths of issues without an update for a long time ;-)

rajat641:
Hi @nijel ,
Today I explored on how to implement the above change and managed to find an already written JS component implementing the above feature. Here is the reference link [http://www.daterangepicker.com/#config](). 
It looked somehow like this on my local machine-
![image](https://user-images.githubusercontent.com/8374949/83105715-03473100-a070-11ea-8cd2-3bde010cf558.png)
Please ignore the styling details as it's just done for demonstration/clarification purposes to you.
It doesn't have a **This Year** and **Next Year** option(as given in the requirements image attached above) and I tried adding it but I did work out as its predefined. 

I'm not sure If I can change styling or not as its already predefined. I haven't worked on UX before so please let me know if I'm following the right path or I have to write everything from scratch or maybe learn something else. I thought of using some Bootstrap/JQuery Plugin but I am not very confident about my approach.

nijel:
Writing from scratch is certainly not needed, this one seems like a good choice (it has compatible license and it seems to be using Bootstrap, so hopefully it will not require much of additional styling).

We use yarn to manage the deps, see https://docs.weblate.org/en/latest/contributing/frontend.html for more info on that and other frontend development hints (I've just pushed some improvements to that, it will take few minutes to be fully up to date).

WofWca:
Is it _that_ not user-friendly? Nowadays browsers have pretty solid date pickers built in.

nijel:
The date selection itself is okay, but seeing directly chosen date range would be definitely nicer than current drop-down and two inputs...

akhilsahuji:
is this feature is still needed ?

[...]

Issue #11739, opened 2024-05-29
Ability to generate contributor reports for a category


nijel:
### Discussed in https://github.com/WeblateOrg/weblate/discussions/11734

<div type='discussions-op-text'>

<sup>Originally posted by **emmapeel2** May 28, 2024</sup>
I can generate reports for all my project, or for a component.

But I need to generate them for a category.

I use a category to unify all the little components that are part of Tor Browser, and I want to have a report of the translations for that category only.</div>

harriebird:
Hi guys, check my PR #11947 for this issue.

PR #11826, opened 2024-06-10
Improve daterange selection for reports

harriebird:
## Proposed changes
This improves the selection of date ranges for the generation of reports.
This uses daterangepicker.js to help implement this feature.
This solves issue #3558.

Example image:
image

## Checklist


weblate/templates/base.html
```diff
@@ -58,7 +58,7 @@
 // End Rollbar Snippet
 </script>
 {% endif %}
-  <script defer data-cfasync="false" src="{% static 'vendor/jquery.js' %}{{ cache_param }}"></script>
+  <script data-cfasync="false" src="{% static 'vendor/jquery.js' %}{{ cache_param }}"></script>
```
nijel, line 61: Why defer is no longer possible?

weblate/templates/dashboard/user.html
```diff
@@ -1,10 +1,22 @@
...
 
+{% block extra_script %}
+  {% compress js %}
+    <script defer data-cfasync="false" src="{% static 'vendor/moment.js' %}{{ cache_param }}"></script>
+    <script defer data-cfasync="false" src="{% static 'vendor/daterangepicker.js' %}{{ cache_param }}"></script>
+  {% endcompress %}
+  {% compress css %}
+    <link rel="stylesheet" href="{% static 'vendor/daterangepicker.css' %}{{ cache_param }}" />
+  {% endcompress %}
+{% endblock %}
```
nijel, line 18: This needs to be loaded on several other pages where reports are available (currently component and project, category is requested as well in https://github.com/WeblateOrg/weblate/issues/11739). I'd probably include it in the main template instead.

weblate/templates/snippets/range-field.html
```diff
@@ -0,0 +1,26 @@
...
+          "Last 7 Days": [moment().subtract(6, "days"), moment()],
+          "Last 30 Days": [moment().subtract(29, "days"), moment()],
+          "This Month": [moment().startOf("month"), moment().endOf("month")],
+          "Last Month": [moment().subtract(1, "month").startOf("month"), moment().subtract(1, "month").endOf("month")],
+          "This Year": [moment().startOf("year"), moment().endOf("year")],
+          "Last Year": [moment().subtract(1, "year").startOf("year"), moment().subtract(1, "year").endOf("year")]
+        }
+        }, function(start, end, label) {});
+    });
+  </script>
```
nijel, line 25: Inline script is not allowed by Content-Security-Policy, please place it into `weblate/static/loader-bootstrap.js`.

weblate/trans/forms.py
```diff
@@ -133,6 +133,40 @@ def to_python(self, value):
...
+        if "widget" not in kwargs:
+            kwargs["widget"] = DateRangeInput
+        super().__init__(**kwargs)
+
+    def to_python(self, value):
+        start, end = value.split(" - ")
+        return {
+            "start_date": from_current_timezone(datetime.strptime(start, "%m/%d/%Y")),
+            "end_date": from_current_timezone(datetime.strptime(end, "%m/%d/%Y")),
+        }
```
nijel, line 155: This will crash on invalid input instead of providing user a validation error.

weblate/trans/forms.py
```diff
@@ -133,6 +133,40 @@ def to_python(self, value):
...
+
+    def validate(self, value):
+        if not (
+            isinstance(value["start_date"], datetime)
+            and isinstance(value["end_date"], datetime)
+        ):
+            raise ValidationError(gettext("Invalid date!"))
```
nijel, line 162: This should also do the validation that start is before end (currently done in ReportsForm.clean).

weblate/trans/tests/test_reports.py
```diff
@@ -127,12 +127,7 @@ def get_credits(self, style):
         self.add_change()
         return self.client.post(
             reverse("credits", kwargs=self.get_kwargs()),
-            {
-                "period": "",
-                "style": style,
-                "start_date": "2000-01-01",
-                "end_date": "2100-01-01",
-            },
+            {"style": style, "period": "01/01/2000 - 01/01/2100"},
```
nijel, line 130: Please add tests for invalid period input.

weblate/trans/tests/test_reports.py
```diff
@@ -185,27 +175,45 @@ def test_counts_view_json(self) -> None:
         self.assertJSONEqual(response.content.decode(), COUNTS_DATA)
 
     def test_counts_view_30days(self) -> None:
-        response = self.get_counts("json", period="30days")
+        end = timezone.now()
+        start = timezone.now() - timedelta(days=30)
```
nijel, line 179: Avoid race condition for different range when running testsuite on midnight (same applies to other tests):

```suggestion
        start = end - timedelta(days=30)
```

weblate/templates/base.html
harriebird: I reverted the defer.

weblate/templates/dashboard/user.html
harriebird: I noticed that the reports form is used by other pages also. I have included it to the other pages using the `extra_script` block.

weblate/trans/forms.py
harriebird: To solve this, I added a exception to handle it in case the input is invalid.

weblate/trans/forms.py
harriebird: Transferred the implementation of this `validation that start is before end` inside the `DateRangeField.validate()`. A test was also added.

weblate/trans/tests/test_reports.py
harriebird: Tests were added including a test for a blank range value.

weblate/trans/tests/test_reports.py
harriebird: Changed the code according to the suggested change.

weblate/templates/snippets/range-field.html
harriebird: Transferred the JavaScript code into `weblate/static/loader-bootstrap.js`.

weblate/static/vendor/daterangepicker.css
```diff
@@ -0,0 +1,417 @@
+/*
+  Modified daterangepicker.css
+  This is to align with the weblate design and colors
+ */
```
nijel, line 4: As this is not a verbatim vendored file, please place it outside `vendor` dir. Moreover, did you check how it looks it both dark and light theme?

weblate/templates/component.html
```diff
@@ -1,10 +1,22 @@
...
+{% block extra_script %}
+  {% compress js %}
+    <script defer data-cfasync="false" src="{% static 'vendor/moment.js' %}{{ cache_param }}"></script>
+    <script defer data-cfasync="false" src="{% static 'vendor/daterangepicker.js' %}{{ cache_param }}"></script>
+  {% endcompress %}
+  {% compress css %}
+    <link rel="stylesheet" href="{% static 'vendor/daterangepicker.css' %}{{ cache_param }}" />
+  {% endcompress %}
+{% endblock %}
+
```
nijel, line 19: I'm still thinking whether including this only in selected templates is the way to go. Having it on the dashboard/project/component page makes it likely that every user will load this anyway. With upcoming https://github.com/WeblateOrg/weblate/issues/11846 even more pages will load this. On the other side, I've not looked at the performance hit caused by loading the additional JavaScript and CSS. In case it's measurable, let's keep loading separate (but let's make it reusable instead of copying the same code to every template). Otherwise, I'd just include it in the base template to make the maintenance simpler.

weblate/trans/forms.py
nijel: The invalid date path is still not covered by the tests.

weblate/static/vendor/daterangepicker.css
harriebird: I transferred the CSS file outside the vendor directory. I checked it and it still looks dark on the light theme. I improved it and now it supports light and dark themes.

weblate/templates/component.html
harriebird: Since the usage of this feature is growing across pages, it is more appropriate to transfer its loading in the base template. Their links is now transferred to the `base.html` for easier reuse and maintenance.

weblate/trans/forms.py
harriebird: The invalid test will never pass to that path since the `to_python` is already doing validation. I removed it since it is redundant and kept only the start_date < end_date check in `validate`.

weblate/static/daterangepicker-dark.css
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright © Michal Čihař <michal@weblate.org>
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+@media (prefers-color-scheme: dark) {
```
nijel, line 7: This will not work for manually selected dark theme, please place this CSS into `weblate/static/style-dark.css` which is loaded in that case.

harriebird: Deleted `daterangepicker-dark.css` and moved it contents to `style-dark.css`.

nijel: The file seems still present (the content was added to `style-dark.css`, but the file was kept here).

weblate/static/loader-bootstrap.js
```diff
@@ -1380,6 +1380,33 @@ $(function () {
...
+        Yesterday: [moment().subtract(1, "days"), moment().subtract(1, "days")],
+        "Last 7 Days": [moment().subtract(6, "days"), moment()],
+        "Last 30 Days": [moment().subtract(29, "days"), moment()],
+        "This Month": [moment().startOf("month"), moment().endOf("month")],
+        "Last Month": [
+          moment().subtract(1, "month").startOf("month"),
+          moment().subtract(1, "month").endOf("month"),
+        ],
+        "This Year": [moment().startOf("year"), moment().endOf("year")],
+        "Last Year": [
```
nijel, line 1401: Make the range labels translatable:

```suggestion
        gettext("Today"): [moment(), moment()],
        gettext("Yesterday"): [moment().subtract(1, "days"), moment().subtract(1, "days")],
        gettext("Last 7 days"): [moment().subtract(6, "days"), moment()],
        gettext("Last 30 days"): [moment().subtract(29, "days"), moment()],
        getttext("This month"): [moment().startOf("month"), moment().endOf("month")],
        gettext("Last month"): [
          moment().subtract(1, "month").startOf("month"),
          moment().subtract(1, "month").endOf("month"),
        ],
        gettext("This year"): [moment().startOf("year"), moment().endOf("year")],
        gettext("Last year"): [
```

weblate/static/loader-bootstrap.js
```diff
@@ -1380,6 +1380,33 @@ $(function () {
     });
   });
 
+  $("input[name='period']").daterangepicker(
+    {
+      autoApply: true,
+      startDate: moment(),
+      endDate: moment(),
+      alwaysShowCalendars: true,
+      opens: "left",
```
nijel, line 1389: Localize custom range label:

```suggestion
      opens: "left",
      locale: {
        customRangeLabel: gettext("Custom range"),
      },
```

nijel:

Issue #11846, opened 2024-06-13
Improved date range selection for changes


nijel:
### Describe the problem

Selecting date range while browsing changes could be more comfortable.

### Describe the solution you would like

https://github.com/WeblateOrg/weblate/pull/11826 introduces better date range selection for reports and the same should be used when browsing changes:

https://github.com/WeblateOrg/weblate/blob/4f95e0531ed723c6959dd3ae4d13c3181b76ce14/weblate/trans/forms.py#L2716-L2717

harriebird:
Hi @nijel, I will check this issue. This issue related to what PR #11826 solves.

nijel:
Yes, it's just applying the new range selector on different form.

harriebird:
Hi guys, check my PR #11894 for this issue.

PR #11894, opened 2024-06-19
Improve daterange selection for changes

harriebird:
## Proposed changes

This improves the selection of date ranges for the generation of changes using the `DateRangeField`.
This solves issue #11846.

Example image:
Screenshot 2024-06-19 at 15-40-35 Changes @ Devel Weblate

## Checklist

### DIFF
--- a/weblate/trans/tests/test_reports.py
+++ b/weblate/trans/tests/test_reports.py
@@ -167,7 +167,11 @@ class ReportsComponentTest(BaseReportsTest):
     def test_credits_blank_period(self):
         period = ""
         response = self.get_credits("json", period=period, follow=True)
-        self.assertContains(response, "Error in parameter period: Invalid date!")
+        self.assertEqual(response.status_code, 200)
+        self.assertJSONEqual(
+            response.content.decode(),
+            [{"Czech": [["weblate@example.org", "Weblate <b>Test</b>", 1]]}],
+        )
 
     def test_credits_invalid_start(self):
         end = timezone.now()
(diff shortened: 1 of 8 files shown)


weblate/trans/tests/test_reports.py
```diff
@@ -167,7 +167,11 @@ def test_credits_view_html(self) -> None:
     def test_credits_blank_period(self):
         period = ""
         response = self.get_credits("json", period=period, follow=True)
-        self.assertContains(response, "Error in parameter period: Invalid date!")
+        self.assertEqual(response.status_code, 200)
+        self.assertJSONEqual(
+            response.content.decode(),
+            [{"Czech": [["weblate@example.org", "Weblate <b>Test</b>", 1]]}],
+        )
```
nijel, line 174: Reports should not be allowed without a period for performance reasons.

harriebird: Made the `DateRangeField` in the `ReportsForm` required. I also adjusted the `DateRangeField` to disallow empty values when `required` field is set to True.

weblate/trans/views/reports.py
```diff
@@ -83,10 +83,15 @@ def get_credits(request, path=None):
         show_form_errors(request, form)
         return redirect_param(obj or "home", "#reports")
 
+    start_date = end_date = None
+    if form.cleaned_data["period"]:
+        start_date = form.cleaned_data["period"]["start_date"]
+        end_date = form.cleaned_data["period"]["end_date"]
+
```
nijel, line 90: Please revert the changes here as well, reports need period to be provided.

harriebird: Reverted the changes for checking if the `form.cleaned_data["period"]` is None.

nijel:
This introduced regression: https://github.com/WeblateOrg/weblate/issues/11954

Issue #11928, opened 2024-06-24
Suggestions table width a bit wacky


JohnRDOrazio:
### Describe the problem

When the suggested translation contains a long unbreakable string (perhaps not a good practice for translations, but in some circumstances it might happen) such as a URL, the `diff` cell in the suggestions table can become very very wide and the whole table becomes wider than the view area. Consequently, the <kbd>Clone to translation</kbd> and <kbd>Accept</kbd> buttons can wind up underneath the sidebar and become inaccessible.

### Describe the solution you would like

I'm already working on a PR that can fix the table column widths to prevent them from going beyond set limits, here is what it should look like with the fix in place:
![Weblate screenshot AFTER table fix](https://github.com/WeblateOrg/weblate/assets/231298/03f3a04d-d350-42ed-8709-8a18759709ad)

### Screenshots

![Weblate screenshot BEFORE table fix](https://github.com/WeblateOrg/weblate/assets/231298/fae926de-f5d0-47f5-bf63-0b0d38b8d298)

PR #11929, opened 2024-06-24
fix issue #11928 suggestions table width

JohnRDOrazio:
BEFORE the fix:
Weblate screenshot BEFORE table fix

AFTER the fix:
Weblate screenshot AFTER table fix


weblate/static/style-bootstrap.css
```diff
@@ -1617,10 +1617,18 @@ input.color_edit:checked + label {
...
 
+table:has(#machinery-translations) {
+  table-layout: fixed;
+}
+
 .machinery-text {
   word-wrap: anywhere;
 }
 
+.machinery-diff {
```
JohnRDOrazio, line 1628: it might even be enough to just add this CSS rule...

JohnRDOrazio: actually, the CSS rule that would work alone would be:
`word-break: break-word`

weblate/static/style-bootstrap.css
```diff
@@ -1617,10 +1617,18 @@ input.color_edit:checked + label {
...
+  word-wrap: break-word;
```
JohnRDOrazio, line 1629: changing this CSS rule to `word-break: break-word` would even work on it's own

JohnRDOrazio:
fixes issue #11928

weblate/static/style-bootstrap.css
nijel: Maybe it would be a better approach to use just this? I'm afraid that with a fixed layout, it's impossible to account for all possible situations (for example, buttons have varying length in different translations).

JohnRDOrazio:
performed a force-overwrite with PR #11938 which had the consequence of automatically closing the current PR

Issue #11714, opened 2024-05-26
Add option for sorting in case-sensitive order to add-on "Format the Java properties file"


HaaiHenkie:
### Describe the problem

In my repository I sort the strings in the Java properties files by their key in a case-sensitive order. I noticed that newly translated strings are appended to the end of the properties file. That is why I tried the add-on "Format the Java properties file" which sorts the strings. However it is doing that in a case-insensitive order. This is giving a lot of difference to process when merging the new translations into my repository. So for the moment I am better of not using the add-on "Format the Java properties file".

### Describe the solution you would like

I would like the add-on "Format the Java properties file" to have a configuration option to sort the entries in a case-sensitive order.

### Describe alternatives you have considered

Maybe the properties files for the translations could stick to the key order of the monolingual base properties file. I think that is more difficult to implement.

### Additional context

Not relevant to me, but you could add other common format options like "Align properties in columns" and "Insert space around key-value delimiter"

nijel:
Indeed, the current implementation is case-insensitive, adding an option to make it case-sensitive should be pretty easy. Related code is here: https://github.com/WeblateOrg/weblate/blob/main/weblate/addons/properties.py

harriebird:
Hi guys, check my PR  #11933 for this issue.

PR #11933, opened 2024-06-25
Add case-sensitive key sorting setting in Properties add-on

harriebird:
## Proposed changes

This adds a configuration to toggle case-sensitive key sorting to the `weblate.properties.sort` add-on.
This solves issue #11714.

Example images:

Screenshot 2024-06-25 at 16-27-23 Freeplane_Test Component @ Devel Weblate

Screenshot 2024-06-25 at 16-23-01 Freeplane_Test Component @ Devel Weblate

## Checklist

### DIFF
--- a/weblate/addons/properties.py
+++ b/weblate/addons/properties.py
@@ -18,6 +18,7 @@ from django.utils.translation import gettext_lazy
 
 from weblate.addons.base import BaseAddon
 from weblate.addons.events import AddonEvent
+from weblate.addons.forms import PropertiesSortAddonForm
 
 SPLITTER = re.compile(r"\s*=\s*")
 UNICODE = re.compile(r"\\[uU][0-9a-fA-F]{4}")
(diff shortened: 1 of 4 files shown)


weblate/addons/properties.py
```diff
@@ -113,12 +116,12 @@ def filter_lines(lines):
     return result
 
 
-def format_file(filename) -> None:
+def format_file(filename, case_sensitive) -> None:
```
nijel, line 119: Please add type annotations when changing the code:

```suggestion
def format_file(filename: str, case_sensitive: bool) -> None:
```

We're not yet at a state where the whole Weblate would be type-checkable, but we want to get there.

weblate/addons/properties.py
```diff
@@ -136,6 +139,8 @@ class PropertiesSortAddon(BaseAddon):
     description = gettext_lazy("Formats and sorts the Java properties file.")
     compat = {"file_format": {"properties-utf8", "properties", "gwt"}}
     icon = "sort-alphabetical.svg"
+    settings_form = PropertiesSortAddonForm
 
     def pre_commit(self, translation, author) -> None:
-        format_file(translation.get_filename())
+        case_sensitive = self.instance.configuration["case_sensitive"]
```
nijel, line 145: Make sure this works also with the old configurations, which didn't have this key:

```suggestion
        case_sensitive = self.instance.configuration.get("case_sensitive", False)
```

weblate/addons/properties.py
harriebird: Suggested change applied. I will take note of putting type hints to parameters.

weblate/addons/properties.py
harriebird: A good catch @nijel. This is helpful to avoid issues when Weblate is updated to the version with this change.

Issue #2144, opened 2018-07-26
Source strings corruption on monolingual translations


nijel:
### Steps to reproduce

So far this has happened once on Hosted Weblate, no clear steps to reproduce. What were the steps

1. Translation was imported with English as source language
2. Turkish translation was added
3. Turkish translation was changed
4. Source string for Norwegian translation is Turkish string from 2, though probably it was changed earlier

### Actual behaviour

The source strings for one unit of monolingual translation was used from one particular version of Turkish translation.

### Expected behaviour

The source strings for all same units should match.

comradekingu:
@nijel 
I faintly remember something like that.
Did I maybe make a comment in the English source, someone got the e-mail and started editing English thinking it was Turkish?

nijel:
This was seen on private project once, it has never happened there again. Having other occurrence might help to track down the issue.

13. září 2018 17:11:36 SELČ, "Allan Nordhøy" <notifications@github.com> napsal:
>@nijel 
>I faintly remember something like that. What project was it?
>
>Did I maybe make a comment in the English source, someone got the
>e-mail and started editing English thinking it was Turkish?

nijel:
Another occurence from https://github.com/WeblateOrg/hosted/issues/10:

https://hosted.weblate.org/translate/your-local-weather/strings/nb_NO/?checksum=c8fa1237ef890b5c

I am seeing 
"schwere Regenschauer", but when clicking to edit the source string, it says
"light rainshower"

Edit: When clicking [nearby strings](https://hosted.weblate.org/translate/your-local-weather/strings/nb_NO/?type=nottranslated#nearby), they are presented in English, but once clicked, seen overwritten in Norwegian Bokmål.

This suggests the source history gets overwritten when editing translated languages, as that is what happened most recently.

nijel:
Thank you for your report, the issue you have reported has just been fixed. 

* In case you see problem with the fix, please comment on this issue. 
* In case you see similar problem, please open separate issue.
* If you are happy with the outcome, consider supporting Weblate by [donating](https://weblate.org/donate/).

nijel:
@sm4rk0  The source seems correct there (it reads "Abbrev N↓, NE↙, …")

PR #11937, opened 2024-06-25
fix(formats): do not copy template notes for new translations

### LINKED ISSUE #11133: Updating comment/note in source template file does not correctly update comment/note in translations
### Describe the issue

Seen when using Fluent, but I imagine it would effect other formats that use a template.

Whenever the source file changes the comment/note on one of its strings, the translation files do not pick up on this change, and Weblate shows the "string description" as a concatenation of the old and new strings.

### I already tried

Tested with my local weblate instance, but I've seen this for the Tor Project before:

1. Have a Fluent component with a template file "en.ftl" containing a Fluent string with a comment "Original comment.", and translate it to "de.ftl".
2. Change the comment in the locally stored file "en.ftl" to be "New comment.".
3. Force weblate to pick up the new changes.

Weblate will show the en string with the string description: "New comment.". But the de string will have the string description: "New comment. Original Comment.". Moreover, the "de.ftl" file will only contain the old comment: "Original Comment.".

NOTE: The concatenation comes from this line: https://github.com/WeblateOrg/weblate/blob/a808cf013882ae41987921cf93301c0055e3cdcc/weblate/formats/ttkit.py#L106-L110 and seems to be a result of the translated file not picking up on the comment change.

### Expected behavior

If the notes on a template unit changes, then all translation units with the same original note should update their note to match the new value. I.e. the new comment should entirely replace the old comment on Weblate, and in the translated files.

### Screenshots

_No response_

### Exception traceback

_No response_

### How do you run Weblate?

weblate.org service

### Weblate versions

_No response_

### Weblate deploy checks

_No response_

### Additional context

_No response_

gzm0:
## Proposed changes

Do not copy the notes (comments) from the template/source when adding a translation for a key (in monolingual formats).

This fixes #11133. Copying the notes is problematic because the are not synced when they are updated. As a result, they can get out of date.

The implicit assumption behind this change is that most users only care about comments in source files (but not in target files).

## Checklist

### DIFF
--- a/weblate/formats/tests/test_formats.py
+++ b/weblate/formats/tests/test_formats.py
@@ -271,8 +271,8 @@ class BaseFormatTest(FixtureTestCase, TempDirMixin):
             self.assertEqual(unit.unit.getnotes().strip(), "")
         else:
             # Assume this is a multi-unit. Will fail otherwise.
-            for unit in unit.units:
-                self.assertEqual(unit.unit.getnotes(), "")
+            for subunit in unit.units:
+                self.assertEqual(subunit.unit.getnotes(), "")
 
     def test_find(self) -> None:
         storage = self.parse_file(self.FILE)
--- a/weblate/formats/tests/test_formats.py
+++ b/weblate/formats/tests/test_formats.py
@@ -265,6 +265,17 @@ class BaseFormatTest(FixtureTestCase, TempDirMixin):
         self.maxDiff = None
         self.assertEqual(testdata.decode().strip(), newdata.decode().strip())
 
+    def assert_no_notes(self, unit) -> None:
+        """
+        Assert that the underlying unit(s) do not have any notes.
+        """
+        if unit.unit:
+            self.assertEqual(unit.unit.getnotes().strip(), "")
+        else:
+            # Assume this is a multi-unit. Will fail otherwise.
+            for unit in unit.units:
+                self.assertEqual(unit.unit.getnotes(), "")
+
     def test_find(self) -> None:
         storage = self.parse_file(self.FILE)
         unit, add = storage.find_unit(self.FIND_CONTEXT, self.FIND)
@@ -378,7 +389,7 @@ class BaseFormatTest(FixtureTestCase, TempDirMixin):
         template_storage.save()
 
         # Add a new language and translate the new string.
-        self.FORMAT.add_language(main_file, Language.objects.get(code="cs"), "")
+        self.FORMAT.add_language(main_file, Language.objects.get(code="cs"), self.BASE)
         target_storage = self.parse_file(main_file, template=template_file)
         target_unit, add = target_storage.find_unit(
             self.NEW_UNIT_KEY, self.NEW_UNIT_KEY
@@ -389,7 +400,7 @@ class BaseFormatTest(FixtureTestCase, TempDirMixin):
         # Eagerly check that the target unit does not have notes.
         # This is the point where checking that the template unit still has the
         # notes would potentially fail (depending on the exact format).
-        self.assertEqual(target_unit.unit.getnotes(), "")
+        self.assert_no_notes(target_unit)
 
         target_unit.set_target("Translated string (CS)")
         target_storage.save()
@@ -407,12 +418,7 @@ class BaseFormatTest(FixtureTestCase, TempDirMixin):
         self.assertEqual(target_unit.notes.strip(), self.NOTE_FOR_TEST)
 
         # Check there are no notes on the underlying unit.
-        if target_unit.unit:
-            self.assertEqual(target_unit.unit.getnotes().strip(), "")
-        else:
-            # Assume this is a multi-unit. Will fail otherwise.
-            for unit in target_unit.units:
-                self.assertEqual(unit.unit.getnotes(), "")
+        self.assert_no_notes(target_unit)
 
     def test_flags(self) -> None:
         """
--- a/weblate/formats/tests/test_formats.py
+++ b/weblate/formats/tests/test_formats.py
@@ -266,9 +266,7 @@ class BaseFormatTest(FixtureTestCase, TempDirMixin):
         self.assertEqual(testdata.decode().strip(), newdata.decode().strip())
 
     def assert_no_notes(self, unit) -> None:
-        """
-        Assert that the underlying unit(s) do not have any notes.
-        """
+        """Assert that the underlying unit(s) do not have any notes."""
         if unit.unit:
             self.assertEqual(unit.unit.getnotes().strip(), "")
         else:
--- a/weblate/formats/tests/test_formats.py
+++ b/weblate/formats/tests/test_formats.py
@@ -340,84 +340,6 @@ class BaseFormatTest(FixtureTestCase, TempDirMixin):
         else:
             self.assertIn(self.NEW_UNIT_MATCH, newdata)
 
-    def test_do_not_clone_notes(self) -> None:
-        """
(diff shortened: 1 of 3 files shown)


gzm0:
Still WIP, because I have not figured out how to:
- run unit tests
- log into the development weblate (I managed to start it).

Any help greatly appreciated.

weblate/formats/ttkit.py
```diff
@@ -176,6 +177,15 @@ def flags(self):
...
         return flags.format()
 
+    def clone_template(self) -> None:
+        new_unit = copy(self.template)
+
+        # do not copy notes from the template (#11133)
+        new_unit.removenotes()
+
+        self.mainunit = self.unit = new_unit
+        self._invalidate_target()
```
nijel, line 187: Please avoid duplicating logic from the super class:

```suggestion
        super().clone_template()

        # do not copy notes from the template (#11133)
        self.unit.removenotes()
```

gzm0: Fair :)

nijel:
I usually squash commits when merging, so I'm not really concerned about evolution of the commits.

nijel:
Ah, I didn't notice this:

>     * run unit tests

This should be documented at https://docs.weblate.org/en/latest/contributing/tests.html

>     * log into the development weblate (I managed to start it).

If you're using `rundev.sh`, the admin password was printed on the first start. You can create new admin or update the password using https://docs.weblate.org/en/latest/admin/management.html#createadmin

weblate/formats/tests/test_formats.py
```diff
@@ -327,6 +327,37 @@ def test_new_unit(self) -> None:
...
+        # Check we get the aggregated notes through the wrapper.
+        self.assertEqual(target_unit.notes, "template note for test\n")
+
+        # Check there are no notes on the underlying unit.
+        if target_unit.unit:
+            self.assertEqual(target_unit.unit.getnotes(), "")
+        else:
+            # Assume this is a multi-unit. Will fail otherwise.
+            for unit in target_unit.units:
+                self.assertEqual(unit.unit.getnotes(), "")
```
gzm0, line 359: Alternative to this is to add a flag to the test so it can declare if it is multi-unit rather than detecting.

gzm0:
This is ready for review (tests pass locally, so hopefully they will on CI, barring obscure interactions with different versions).

I needed to fix two additional things that I have put in separate comments. @nijel LMK if you'd like me to put them into separate PRs.

I plan to squash the last "Fix doc comment" commit (once CI passes).

gzm0:
Rebased and squashed.

weblate/formats/base.py
```diff
@@ -276,7 +276,7 @@ def has_unit(self) -> bool:
         return self.unit is not None
 
     def clone_template(self) -> None:
-        self.mainunit = self.unit = copy(self.template)
+        self.mainunit = self.unit = deepcopy(self.template)
```
nijel, line 279: We used to have this, but this scales horribly: when adding 100 units, you need to clone 100 times the underlying objects. See 8df491b98ed372aaddcc611234d0dd486737ef51

gzm0: I see :-/ I'm concerned that this can actually lead to bugs.

My plan:
- Extract PHP test fix (IMO worth merging on its own): #11999
- Figure out of the shallow copy causes end-user visible issues, then take it from there.

nijel: Yes, it can lead to problems (I think we had this in the past). I think the only problematic piece is `xmlelement` attribute for LISA-based units, what can be addressed in several ways. Maybe doing a copy of it before wiping the notes would be enough to address this issue?

gzm0: Hmmm... I'm a bit concerned that this might lead to a lot of special casing down the line.

I'm wondering if rather than copying the whole unit, creating a new unit and copying only what we care about could be a viable approach. Have you ever toyed with this idea? (I have not the slightest idea what the implications for various formats would be).

FWIW: I have not managed to find any (trivial) issues caused by a shallow copy so far (I have found #12000, but you already fixed that in record time :P).

nijel: I think shallow copy could cause issue in case both template and translation store are being updated. Depending on the ordering, it might wipe notes from the template. See #2144 / 0fa7a7ecf4eef66867be5bebf81b457738c7861c for what kind of issues this can cause.

Rebuilding new unit would be an option as well, but I'm afraid that the translate-toolkit API is very inconsistent in this regard. 

Looking at it, there is actually a `TranslationUnit.buildfromunit` method which might do what we need here. The question is whether it works reliably. Some formats even implement `copy` method where they deal with the xml, see https://github.com/translate/translate/blob/ae6abd4c8935c0675933eea49eaed7255dca7981/translate/storage/tmx.py#L105-L115. Polishing these in the translate-toolkit and using it from Weblate might be the cleanest approach.

gzm0: - I have opened a CI only PR (#12015) to check what happens if we use `buildFromUnit` in `clone_template`.
- It seems like the deepcopy issue we see in the unit tests does not actually affect usage (I've tested this in a local dev instance). I will try to adjust the test cases to roundtrip via actual files (which is probably closer to actual usage anyways).

nijel: deepcopy is problematic as it can copy huge data in the background several times. Imagine 50 MB XLIFF file where you add 20 strings. You would suddenly need 1 GB of memory just for the deep copies of the XML (you would actually need at least 2x more as that also copies the store with parsed units).

gzm0:
Ah, @nijel I see you have already picked this up again. I was waiting on #12051.

I hope the current form (with the a bit more complicated test) is acceptable. I was planning on doing a manual test run again to make sure this does not destroy template comments in formats affected by the shallow-copy issue.

gzm0:
I'm looking into the test failure. Looks like my local manual test selection wasn't wide enough.

weblate/formats/tests/test_formats.py
```diff
@@ -328,6 +339,84 @@ def test_new_unit(self) -> None:
...
+        self.assertFalse(add)
+
+        # Check we get the aggregated notes through the unit wrapper:
+        # We always (additionally) display the template notes to the user
+        # (if they are different from the target unit notes).
+        self.assertEqual(target_unit.notes.strip(), self.NOTE_FOR_TEST)
+
+        # Check there are no notes on the underlying unit.
+        self.assert_no_notes(target_unit)
+
```
nijel, line 419: Please include test for side effects of shallow copy here. I don't think we can guarantee that it will be never hit inside Weblate, so better to be safe. Following should work:

1. Save template store
2. Parse it again
3. Verify that the note is there

gzm0:
I have followed the following manual procedure for both Android XML (a format affected by the shallow copy) and Java Properties (a format not affected by the shallow copy).

1. Add a new string with description in VCS.
1. Update repository.
1. Check the comment shows under "Source string description" of the string.
1. Translate the string to another language.
   - Check the comment shows as comment on the base language (in dark pill)
   - Check the comment shows under "Source string description".
1. Click "Save and stay"
1. Go back to the source string and check "source string description" still has the comment.
1. Commit & push pending changes.
1. Check written file does not have the comment.

nijel:
I've added the tests for not changing template in https://github.com/WeblateOrg/weblate/pull/12074 (they pass locally, so it should be auto-merged soon). Including notes tests in that should be easy.

gzm0:
TY for the help @nijel . The unit tests in `weblate.formats.tests` now fully pass locally. Let's see what CI has to say>

nijel:

PR #11938, opened 2024-06-25
fix issue #11928 suggestions table width

JohnRDOrazio:
BEFORE the fix:
Weblate screenshot BEFORE table fix

AFTER the fix:
Weblate screenshot AFTER table fix


JohnRDOrazio:
fixes issue https://github.com/WeblateOrg/weblate/issues/11928

JohnRDOrazio:
new screenshot: 
image

weblate/static/editor/full.js
```diff
@@ -578,7 +578,7 @@
           .attr("dir", this.state.dir)
           .html(el.html),
       );
-      row.append($("<td>").html(el.diff));
+      row.append($("<td>").attr("class", "machinery-diff").html(el.diff));
```
JohnRDOrazio, line 581: another approach could be to simply add the already defined `machinery-text` class which uses the rule `word-wrap: anywhere`, and has pretty much the same effect

JohnRDOrazio:
this PR force-overwrites PR #11929 (which was therefore automatically closed)

weblate/static/editor/full.js
nijel: That sounds even better. There is no reason to apply different logic to diff than to the full text.

nijel, line 581:
```suggestion
      row.append($("<td>").attr("class", "machinery-text").html(el.diff));
```

weblate/static/style-bootstrap.css
```diff
@@ -1621,6 +1621,10 @@ input.color_edit:checked + label {
   word-wrap: anywhere;
 }
 
+.machinery-diff {
+  word-break: break-word;
+}
+
```
nijel, line 1627:
```suggestion
```

nijel:

PR #12009, opened 2024-07-04
Fix the initial values of period in ChangeForm when there is a user input

harriebird:
## Proposed changes

When using the date range search filter on the Changes page, the dates just go back to the current date instead of setting the initial date range to the dates that were submitted. To solve this, the form must be able to supply the submitted date range values and be set as initial start and end dates in the form.

## Checklist


nijel:

PR #12068, opened 2024-07-16
CI only: Repro write template store

gzm0:
Debugging PR for https://github.com/WeblateOrg/weblate/pull/11937#discussion_r1679520763

It looks like in the formats that fail, the unit gets *moved* from the template to the target store: After both stores are saved, the key only exists in the target store, but not the template anymore.


gzm0:
Closing. A working version of this was merged in #12074.

Issue #5459, opened 2021-02-16
Contributor stats in API


Duncan-Idaho:
**Is your feature request related to a problem? If so, please describe.**
We're exporting translation files to outside contractors and reimporting them later.
This process is automatic, and I need to have a way to extract their contributions for bill verification for example.

Many parts of the API expose statistics already, but: 

* On user statistics, we lack the amount of words for example.
* We have no way to filter statistics per period.

Trying to call the existing contribution stat page is error prone at best due to the authentification and CSRF mechanisms.

**Describe the solution you'd like**

Either:
* a new API endpoint that receives exactly 4 parameters (start_date, end_date, period and format) and returns exactly what the /counts/:project-slug returns, or
* an improvement on the api/users/:email/statistics endpoint to include wordcount and support two query string parameters (start_date and end_date)

**Describe alternatives you've considered**

See above

**Screenshots**

**Additional context**

comradekingu:
@Duncan-Idaho Are they working on it concurrently, and is it separated into exclusive parts?

By dividing work into greater deliveries (and also roles), I find the accountability of only having someone to blame once something is really wrong is only worth avoiding not getting paid to the average translator.

If you can't account for not being able to work faster than you should, which by all accounts stops with more specific timing, there is no way to average out being held up in review when doing all the work to begin with would be faster, and easier.

Somewhere the rubber has to meet the road, and you can either test people against each-other and see what the resulting numbers of errors amount to, or there could be shorter deadlines with fewer strings per batch.

With where that is going, translators are not about to take the hit on what it costs to average out ability. Nor should they, but that is usually company money.

To combat the idea of strings translated and/or time spent amounting to end value:

It would be nice if each translator had their hourly rate set, and then the time it takes to fix the strings of others would have their payout deducted by said rate (for the time it takes to do). Some portion of this could be awarded the translator that ends up with the most sane fixes.

Either by actual metrics string-for-string, or by average speed and/or estimated/calculated complexity of the source material for changed strings, there are many ways to arrive at a number.

So how do you decide on a review being good or bad? Asking people to document errors others make for free isn't going to happen, and what company is paying more to have their already lost time documented at the cost of what their better translators charge? Add to that it is a boring thing to do, and why would anyone unless push came to shove? If not by way of external contributions if a company gets clever, certainly on the platform itself.

[...]

PR #12084, opened 2024-07-18
API: contributor stats

vinceh121:
## Proposed changes
#5459

API routes that report contributor stats, similar to the JSON export of the Reports tab.

## Checklist

## Other information

Currently, the contributor stats format uses a "positional" way to represent contribution datums, perhaps the API may benefit from using a format that relies on objects?

```json
[
    {
        "French": [
            {
                "email": "contact@vinceh121.me",
                "username": "vincent",
                "contributions": 39
            }
        ]
    }
]
```

Instead of:

```json
[
    {
        "French": [
            [
                "contact@vinceh121.me",
                "vincent",
                39
            ]
        ]
    }
]
```

## Left to do in this PR

- [x] Project route
    - [x] Mandatory date filters
    - [x] Language filters
- [x] Component route
    - [x] Mandatory date filters
    - [x] Language filters


nijel (commented):
I think changing the JSON to be object-based is a good idea. It just needs to be documented in the changelog as a breaking change.

Besides what you've mentioned, following will be needed:

- [ ] tests covering new API
- [ ] documentation

weblate/api/views.py
```diff
@@ -726,8 +731,52 @@ def destroy(self, request, *args, **kwargs):
...
+        if isinstance(obj, Project):
+            kwargs = {"translation__component__project": obj}
+        elif isinstance(obj, Component):
+            kwargs = {"translation__component": obj}
+        else:
+            raise TypeError("Expected project or component")
+
+        data = generate_credits(
+            None
+            if request.user.has_perm("reports.view", obj) or request.user.is_anonymous
```
nijel, line 768: Anonymous user should not get any reports as it exposes e-mail addresses. That's why regular users get only own report and only users with `reports.view` permission get full one.

```suggestion
            if request.user.has_perm("reports.view", obj)
```

vinceh121: Got it. So what should be the behaviour for anonymous users? Should they get a 401 on this route or should the JSON not contain the email fields?

I'm leaning more on the idea of the JSON not having the email fields.

nijel: I would make it fail with 401.

weblate/trans/views/reports.py
```diff
@@ -136,7 +141,10 @@ def get_credits(request, path=None):
                 translators=format_html_or_plain_join(
                     "\n",
                     translator_format,
-                    ((t[0], t[1], t[2]) for t in translators),
+                    (
+                        (t["email"], t["full_name"], t["change_count"])
+                        for t in translators
+                    ),
```
nijel, line 147: When we already have dict with values, it makes sense to change the above format strigs to use named placeholders and format using the dict directly.

weblate/api/views.py
```diff
@@ -726,8 +731,53 @@ def destroy(self, request, *args, **kwargs):
...
+        if "lang" in request.query_params:
+            language = request.query_params["lang"]
+
+        kwargs: dict[str, Any]
+        if isinstance(obj, Project):
+            kwargs = {"translation__component__project": obj}
+        elif isinstance(obj, Component):
+            kwargs = {"translation__component": obj}
+        else:
+            raise TypeError("Expected project or component")
```
nijel, line 767: Maybe we can push this logic to `generate_credits` so that it doesn't have to be duplicated?

weblate/api/views.py
```diff
@@ -726,8 +731,53 @@ def destroy(self, request, *args, **kwargs):
         return super().destroy(request, *args, **kwargs)
 
 
+class CreditsMixin:
+    @action(detail=True, methods=["get"])
+    def credits(self, request, **kwargs):
+        if request.user.is_anonymous:
+            self.permission_denied(request, "Must be authenticated to get credits")
```
nijel, line 738: Can you please cover this with a test?

weblate/trans/views/reports.py
vinceh121: I'm not sure I can change the `format_plaintext_join` and `format_html_join` methods to accept dicts due to their use of generators

nijel: Ah, I didn't realize that `format_html_join` doesn't support kwargs. Without that my comment isn't really meaningful.

nijel:
Looks good now, scheduled for merge. Thanks!

PR #12113, opened 2024-07-23
chore: add sentry connection keep-alive

michael-smt:
To avoid spurious missed check-in issues with Sentry Crons for Celery Beat.

## Proposed changes

Since the introduction of Celery Beat monitoring with Sentry Cron in https://github.com/WeblateOrg/weblate/commit/1d1a92ba456646f99143ccc2a1c1074a2d6e0e9d we are getting a lot of spurious missed check-in alerts similar to these here: https://github.com/getsentry/sentry-python/issues/3179

Using [`keep_alive=True`](https://docs.sentry.io/platforms/python/configuration/options/#keep-alive) has fixed it for us.

## Checklist

## Other information

Using `SENTRY_EXTRA_ARGS` for this configuration is not possible with the docker container.

### DIFF
--- a/weblate/utils/errors.py
+++ b/weblate/utils/errors.py
@@ -122,6 +122,7 @@ def init_error_collection(celery=False) -> None:
             ],
             attach_stacktrace=True,
             _experiments={"max_spans": 2000},
+            keep_alive=True,
             **settings.SENTRY_EXTRA_ARGS,
         )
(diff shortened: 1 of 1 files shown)


nijel:
Thanks, I think it makes sense to turn it on by default.

nijel:
Let's keep it hardcoded for now, it can be made configurable if we see some issues with it. Thanks

PR #12118, opened 2024-07-24
7705 suggestion already exists bug

### LINKED ISSUE #7705: When trying to add a suggestion that is == current translation, it says "suggestion already exists", even if there are actually none
### Describe the issue

Sounds nitpicky, but it confused me when is was editing glossary since there's an "Explanation" text area which is in place of where the translation text area usually is on non-glossary items. Maybe this also needs to be changed.

### I already tried

_No response_

### Expected behavior

_No response_

### Screenshots

_No response_

### Exception traceback

_No response_

### How do you run Weblate?

weblate.org service

### Weblate versions

_No response_

### Weblate deploy checks

_No response_

### Additional context

Code:

https://github.com/WeblateOrg/weblate/blob/4b2e079b80103364292464b585186943b7172890/weblate/trans/views/edit.py#L288-L297

https://github.com/WeblateOrg/weblate/blob/d92daadd0e3cd8d150cb1386e9a7b3ed541f37c3/weblate/trans/models/suggestion.py#L47-L48

gersona:
Raise a specific exception when the submitted suggestion is the same as the current translation
Any code calling Suggestion.objects.add is responsible of handling that error

## Checklist

## Other information

fixes #7705

### DIFF
--- a/weblate/trans/autotranslate.py
+++ b/weblate/trans/autotranslate.py
@@ -63,7 +63,7 @@ class AutoTranslate:
             len(item) > unit.get_max_length() for item in target
         ):
             suggestion = Suggestion.objects.add(
-                unit, target, request=None, vote=False, user=user
+                unit, target, request=None, vote=False, user=user, raise_exception=False
             )
             if suggestion:
                 self.updated += 1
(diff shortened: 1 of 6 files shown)


nijel:

PR #12120, opened 2024-07-24
12081 project count fix

### LINKED ISSUE #12081: Incorrect projects count on hosted weblate homepage for anonymous user
### Describe the issue

The number of projects displayed at the bottom of the homepage doesn't match the actual number of projects an anonymous user can browse.
Currently, it says there are 1400+ projects , but when we click on the button, the paginator only displays 12 pages available (with a limit of 100 projects per page), which add up only to 1200 projects at most

### I already tried

1 - Go to https://hosted.weblate.org/ as an anonymous user and click the "Browse all [number] projects"
2 - You will be navigate to the https://hosted.weblate.org/projects/ url, and scroll to the bottom of the page where the paginator is located
3 - notice the discrepancy between the pages available and the initial number of projects

### Expected behavior

The number in the "Browse all [number] projects" should match the actual number of projects a user can browse through

### Screenshots

Capture d’écran 2024-07-18 à 08 51 21
Capture d’écran 2024-07-18 à 08 53 46

### Exception traceback

_No response_

### How do you run Weblate?

weblate.org service

### Weblate versions

_No response_

### Weblate deploy checks

_No response_

### Additional context

_No response_

gersona:
## Proposed changes
Add a new "public_projects" metric data, that is used to display project count for anonymous users

## Checklist

## Other information
fixes #12081


weblate/metrics/migrations/0002_new_public_projects_metric_data.py
```diff
@@ -0,0 +1,72 @@
...
+        "public_projects",
+    ]
+    Metric = apps.get_model("metrics", "Metric")
+    Project = apps.get_model("trans", "Project")
+    public_projects_count = Project.objects.filter(
+        access_control__in={access_public, access_protected}
+    ).count()
+    if latest_metric := Metric.objects.order_by(
+        "date"
+    ).last():  # tests will not have Metric objects yet
```
nijel, line 52: This should look for global metric only.

weblate/metrics/migrations/0002_new_public_projects_metric_data.py
```diff
@@ -0,0 +1,72 @@
...
+        dict_data = dict(
+            zip_longest(new_metric_order, latest_metric.data or [], fillvalue=0)
+        )
+        dict_data["public_projects"] = public_projects_count
+        latest_metric.data = [dict_data.pop(name, 0) for name in new_metric_order]
+        latest_metric.save()
```
nijel, line 58: I don't think all the zip logic is needed, we can look at the length of the list only. Something like this should be IMHO sufficient:

```suggestion
        if len(latest_metric.data) == 27:
            latest_metric.data.append(public_projects_count)
            latest_metric.save(update_fields=["data"])
```

Really, the only thing we need to handle is an unlikely option of metrics being collected with the new code and migrations executed afterward, and looking at length does that.

gersona: that makes sense, I'll bring that change in

gersona:
@nijel The patch gets a low coverage score because some conditions in the migration file are not met in test environment. I could definitely write a unit test that creates those conditions then applies the migration to test it, but the test will become irrelevant when the migrations will be squashed in the future. 
Is this something I need to do or how do you usually deal with this type of scenario ?

nijel:
The migrations are squashed at major version where we stop supporting upgrades from older major version (see https://docs.weblate.org/en/latest/admin/upgrade.html#upgrade-from-an-older-major-version).

The migration's testing is done in `ci/run-migrate` and covers really basic things right now. Adding a metric to migrate there should be easy (it needs to be added in the script, as the fixture is used from the old version). But generally, not all migrations have to be CI tested.

nijel:

PR #12122, opened 2024-07-24
Update last-changes-content.html template

### LINKED ISSUE #5464: Better digest and changed string e-mail layout
Papercut
Something ala
bilde
vs
bilde

Right now it isn't super clear which string "Show" actually will display, since it isn't inline.
Also, there isn't any separation between source and translation, so that part could be improved.

Having the user-selected alternative languages would be a bonus, since that is what I am used to looking at. More errors could be found then too.

meel-hd:
## Proposed changes
Refactor the breadcrumb element from an ordered list to an unordered list for better accessibility and styling consistency.

Closes #5464

## Checklist

### DIFF
--- a/weblate/templates/last-changes-content.html
+++ b/weblate/templates/last-changes-content.html
@@ -9,11 +9,11 @@
     <td class="history-event">
         <strong>{{ change.get_user_display }}</strong>
         <p class="text-muted">{{ change.get_action_display }}</p>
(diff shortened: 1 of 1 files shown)


nijel:
I'd like to have similar rendering in the email as we have on web. It is shown as project / component / language instead of the list.

Issue #8356, opened 2022-11-16
Quick action in suggestion box hide text in RTL languages


sr258:
### Describe the issue

If there are suggestion in a RTL language, the quick action ("accept", "accept and edit", "delete", ...) hide the text in the suggestion box: 
![grafik](https://user-images.githubusercontent.com/15268740/202223933-5f935ab6-40b4-4574-8152-27e9431f2e1f.png)

It's not possible to read what's below the actions.

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar issues in this repository.

### How do you run Weblate?

Docker container

nijel:
https://github.com/WeblateOrg/weblate/issues/8259 is somewhat related (it also questions placement of these buttons).

PR #12129, opened 2024-07-25
Update suggestions box

meel-hd:
### Fixed suggestions box
The actions box no longer covers the text field under it. Also fixed spacing between the buttons inside the box.
### Preview
image

Fixes: #8356

## Checklist

### DIFF
--- a/weblate/static/style-bootstrap.css
+++ b/weblate/static/style-bootstrap.css
@@ -624,6 +624,12 @@ form + .login-label {
   }
 }
 
+.btn-float .btn,
+.btn-padded .btn {
+  padding: 2px 4px !important;
+  line-height: 0;
+}
+
 kbd {
   border: 1px solid #bfc3c7;
   -webkit-box-shadow: none;
@@ -1105,6 +1111,10 @@ thead th {
 
 .btn-info {
   border: 1px solid #2eccaa;
+}
+
+.btn-link {
+  padding: 6px;
 }
 
 .btn-link.btn-xs {
--- a/weblate/static/style-bootstrap.css
+++ b/weblate/static/style-bootstrap.css
@@ -583,6 +583,12 @@ form + .login-label {
   padding: 0;
 }
 
+.container-gapped {
+  display: flex;
+  flex-direction: column;
+  gap: 25px;
+}
+
 .comment-content {
   border-radius: 0 5px 5px 5px;
   background-color: #e5f0f2;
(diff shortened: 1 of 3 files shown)


weblate/static/style-bootstrap.css
```diff
@@ -602,20 +602,26 @@ form + .login-label {
...
+  border-radius: 5px;
   position: relative;
-  top: -5px;
+  transform: translateY(-100%);
   z-index: 10;
-  padding: 0 2px;
 }
 
-.btn-float .btn,
-.btn-padded .btn {
```
nijel, line 616: Why `btn-padded` is removed?

weblate/static/style-bootstrap.css
```diff
@@ -1101,10 +1107,6 @@ thead th {
   border: 1px solid #2eccaa;
 }
 
-.btn-link {
```
nijel, line 1104: This class is used in many other locations, I'd expect this breaks rendering there.

weblate/static/style-bootstrap.css
meel-hd: It applies smaller padding to the buttons than the default, it causes users to miss click. When removed the buttons reset to their default padding which looks better.

meel-hd: It applies smaller padding to the buttons than the default, it causes users to miss the click. When removed the buttons reset to their default padding which looks better.

weblate/static/style-bootstrap.css
meel-hd: Indeed. it is used on many elements. Also in the css there are multiple definitions of this class. I assume the default padding applied by bootstrap is enough. But I will leave the class styles here and remove it from the element instead.

nijel: Removing the CSS makes these padding smaller, look at zen mode.

Before this PR:

obrazek

With this PR applied:

obrazek

meel-hd:
# Preview
image

nijel:
This needs more changes:

* The content of the suggestion is still not right aligned for RTL languages, but starts at the toolbar box:
  
  image
* This breaks comments layout which uses same buttons:
  
  obrazek

What about making it a toolbar for a full width of the comment/suggestion?

meel-hd:
**Comments tooltip now look like this:**
image
**And RTL text is fixed:**
image

meel-hd:
> What about making it a toolbar for a full width of the comment/suggestion?

I don't think it will look better. It will only consume more space than it needs.

nijel:

Issue #2136, opened 2018-07-22
Allow Machine Translation from secondary languages


adrmzz:
### Steps to reproduce
1. A user is working in a project from language A to language C. There are no available Machine Translation systems from A to C.
2. The same project has already been translated from language A to language B. The user can access translations under "Other translations" or secondary languages. Among Weblate's current MT providers, there are available Machine Translation systems from B to C (and also from X and Y to C).

### Actual behaviour
The user depends on external services to check Machine Translation proposals from its secondary source language to its main target language and copy-paste into Weblate.

### Expected behaviour
Integrate Machine Translation (from current providers) that can provide the user with machine translation proposals from secondary languages or even from any of the already translated languages (shown under "Other translations").

miloshavlicek:
My situation is actually that I want to translate Czech to Slovak rather than English to Slovak when using Google Translate API Machine translation. My source language is English, however, I have complete Czech translation for the project. So, maybe it could work that if I do not have completed translation to the secondary language (czech), that machine translation will use primary language (english) instead in that case.

nijel:
This would be especially useful for single purpose services like https://github.com/WeblateOrg/weblate/pull/12154.

nijel:
There can be several use cases here:

* Widening support for machine translation pairs by choosing a source language that works for the target language. This will also make specific services like cyrtranslit more usable by choosing a matching source language.
* Better machine translation by selecting close languages (for example, machine translating Czech to Slovak will be typically yield better results than translating English to Slovak because the languages share a lot of logic).

Both these probably could be automated. The first can be fully automatic based on a list of supported languages which we already know for each service.

The second one would need to define groups of “close languages” which would be preferred for machine translation. This might be tricky, but should improve user experience with automated translations.

Is there any use case to make this user configurable? Please share them if you see any.

* Do we want to skip that logic in some situations?
* Should user be able to manually select language to use as source?

severo:
also needed for our project where the source language is French (https://codeberg.org/cartes/web/issues/2241)

nijel:
You can already select an alternative source language when installing the machine translation service or in the component configuration. The only missing bit here is automatic fallback when the service doesn't support some language pair, but would support other.

severo:
[sorry if I'm highjacking this issue - if so, I can delete and create a new one]

[...]

PR #12154, opened 2024-07-29
11923 cyrtranslit suggestion service

### LINKED ISSUE #11923: Add automatic suggestion service for transliteration of Cyrillic script to Latin
### Describe the problem

Several languages can be written in Cyrillic and Latin. As these are just different script for the same language, it should be possible to automatically convert these.

There is [cyrtranslit](https://github.com/opendatakosovo/cyrillic-transliteration) module which can do transliteration of Cyrillic script to Latin. It can be used to automatically generate `_Latn` and `_Cyrl` translations.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

gersona:
## Proposed changes
New machine translation class `CyrTranslitTranslation` is used as interface to the cytranslit module

## Checklist

## Other information
fixes #11923


weblate/machinery/cyrtranslit.py
```diff
@@ -0,0 +1,107 @@
...
+        is_cyrillic = False
+        for key, values in CYRTRANSLIT_TO_WEBLATE_LANGS.items():
+            if source in values["latin"] and language in values["cyrillic"]:
+                cyrtransl_lang = key
+                is_cyrillic = True
+                break
+            if source in values["cyrillic"] and language in values["latin"]:
+                cyrtransl_lang = key
+                is_cyrillic = False
+                break
```
nijel, line 83: Using `map_language_code` and `is_supported_language` might be a better approach here to map and filter supported languages. It would also avoid the need to override `download_multiple_translations`.

pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies = [
   "cryptography>=42.0.2",
   "cssselect>=1.2,<1.3",
   "Cython>=3.0.0,<3.1",
+  "cyrtranslit>=1.1.0",
```
nijel, line 39: Please add upper bound as well to make sure new versions are CI tested before accepted.

```suggestion
  "cyrtranslit>=1.1.0,<2.0",
```

nijel (commented):
Can you please also add documentation?

* `./manage.py list_machinery` will output template to update `docs/admin/machine.rst`
* Add changelog entry referencing the new engine

We probably should then revisit https://github.com/WeblateOrg/weblate/issues/2136 as it would be useful to translate from another translation here instead of the source language (which is likely to be English).

docs/admin/machine.rst
```diff
@@ -112,6 +112,22 @@ This service uses an API and you need to obtain an ID and API key from Baidu to
 
     `Baidu Translate API <https://api.fanyi.baidu.com/api/trans/product/index>`_
 
+.. _mt-cyrtranslit:
+
+Cyrtranslit
+-----
```
nijel, line 118:
```suggestion
Cyrtranslit
-----------

.. versionadded:: 5.7
```

nijel:
I've added changelog entry in 3d28298bd0 and I think this is ready to merge.

gersona:
@nijel thanks for your help on the documentation, I was still trying to get the sphinx to build properly locally

PR #12161, opened 2024-07-30
feat(glossary): limit terminology creation

### LINKED ISSUE #8807: Users should not be able to add entries to other languages' glossaries by default
### Describe the problem

Currently, if a user has the rights to add a glossary entry, they can add it for every language by checking the "Terminology" check box. I don't think this is good default behaviour for Weblate. Yes, the unneeded entries can be deleted, but it's an unnecessary annoyance.

The "Terminology" check box should be disabled by default, except for managers (or possibly also reviewers, if deemed necessary).

_No response_

### Screenshots

_No response_

### Additional context

_No response_

nijel:
## Proposed changes

Only users having glosssary.terminology permission can now add terminology. The existing Glossary role is adjusted to include that.

Fixes #8807

## Checklist


nijel:
@orangesunny Please review docs and the new permission name.

orangesunny (approved):

Issue #6710, opened 2021-10-25
Upload screenshot from clipboard


StefanBrand:
**Is your feature request related to a problem? If so, please describe.**

I often find myself taking screenshots from our web app via the Firefox screenshot tool. By default I always copy to clipboard so that I can paste the screenshot into other tools later. Unfortunately Weblate does not allow this. I must go back to the webapp, take the screenshot again and save it to disk. Then I must upload it to Weblate via the "Browse" button (navigating to the screenshot folder on my disk).

**Describe the solution you'd like**

Allow pasting screenshots into the "Add new screenshot" dialogue.\*

\*Might have to be resized automatically to comply with 2000x2000 px limit.

**Describe alternatives you've considered**

Download & Upload screenshots.

**Screenshots**

Paste screenshot into this dialogue:
![image](https://user-images.githubusercontent.com/8948865/138648653-4388e812-7afc-4192-b746-38e64e5fe8ea.png)

ptse888:
@nijel i would like to work on this if possible! could you possibly direct me as to where i should be looking to make this modification?

nijel:
The screenshot upload form is here:

https://github.com/WeblateOrg/weblate/blob/f785813127b8d40b4f9d9e41d6d26f63e35ac47a/weblate/screenshots/forms.py#L40-L59

It is being rendered here:

https://github.com/WeblateOrg/weblate/blob/f785813127b8d40b4f9d9e41d6d26f63e35ac47a/weblate/templates/screenshots/screenshot_list.html#L19-L32

ParthS007:
@nijel Where we add javascript code for some actions on html, do we have any rules for that?

What the recommended way to test changes in html files, I am trying to see what data comes in `add_form` and implement the logic for pasting from clipboard?

nijel:
No rules so far. Historically, there has been a single JS file, but recently, we've started to split out some logic - we now have also `weblate/static/editor/`.

We're using django-crispy-forms, it should allow customizing the form rendering, see https://django-crispy-forms.readthedocs.io/en/latest/layouts.html. The screenshot forms are here: https://github.com/WeblateOrg/weblate/blob/main/weblate/screenshots/forms.py, and for now, I'd make this change only to allow pasting screenshots, not anything else.

PR #12181, opened 2024-08-02
Feat: Add clipboard paste functionality for screenshots

meel-hd:
## Proposed changes
**Added:** Paste screenshot from clipboard.

The user clicks paste. Then the clipboard is read if the permesion is given and the context is secure (as required by the clipboard API). Then the image is pasted into the select file input. Otherwise if no image is available we only inform the  user about that.

Closes: #6710

## Checklist

### DIFF
--- a/weblate/screenshots/forms.py
+++ b/weblate/screenshots/forms.py
@@ -3,11 +3,19 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 from django import forms
-from django.utils.translation import gettext_lazy
+from django.template.loader import render_to_string
+from django.utils.html import format_html
 
 from weblate.screenshots.models import Screenshot
 from weblate.trans.forms import QueryField
 from weblate.utils.forms import SortedSelect
+
+
+class ScreenshotInput(forms.FileInput):
+    def render(self, name, value, attrs=None, renderer=None, **kwargs):
+        rendered_input = super().render(name, value, attrs, renderer, **kwargs)
+        paste_button = render_to_string("screenshots/snippets/paste-button.html")
+        return format_html("{}{}", rendered_input, paste_button)
 
 
 class ScreenshotEditForm(forms.ModelForm):
@@ -16,6 +24,9 @@ class ScreenshotEditForm(forms.ModelForm):
     class Meta:
         model = Screenshot
         fields = ("name", "image", "repository_filename")
+        widgets = {
+            "image": ScreenshotInput,
+        }
 
 
 class LanguageChoiceField(forms.ModelChoiceField):
@@ -31,6 +42,7 @@ class ScreenshotForm(forms.ModelForm):
         fields = ("name", "repository_filename", "image", "translation")
         widgets = {
             "translation": SortedSelect,
+            "image": ScreenshotInput,
         }
         field_classes = {
             "translation": LanguageChoiceField,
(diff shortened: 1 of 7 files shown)


weblate/templates/translate.html
```diff
@@ -11,6 +11,7 @@
 {% compress js %}
 <script defer data-cfasync="false" src="{% static 'editor/base.js' %}{{ cache_param }}"></script>
 <script defer data-cfasync="false" src="{% static 'editor/full.js' %}{{ cache_param }}"></script>
+<script defer data-cfasync="false" src="{% static 'js/screenshots/clipboard-paste.js' %}{{ cache_param }}"></script>
```
nijel, line 14: This file seems missing.

meel-hd: This commit builds on the previous commit. The file was introduced there.

nijel: Ah, sorry I didn't notice that GitHub showed me just a commit and not a full diff.

weblate/static/js/screenshots/clipboard-paste.js
```diff
@@ -0,0 +1,70 @@
...
+  // The paste trigger button
+  const $pasteScreenshotBtn = $("#paste-screenshot-btn");
+  // The file input to store the screenshot file
+  const $screenshotFileInput = $("#screenshot-form-container input#id_image");
+
+  $pasteScreenshotBtn.on("click", async (e) => {
+    e.preventDefault();
+    try {
+      // Read clipboard content
+      const clipboardItems = await navigator.clipboard.read();
```
nijel, line 15: As I have slightly outdated Firefox, I ended up with:

```
TypeError: navigator.clipboard.read is not a function
```

It would be great to not show the button at all in such case.

weblate/templates/screenshots/screenshot_list.html
```diff
@@ -22,7 +30,13 @@
 <div class="panel panel-default">
 <div class="panel-heading"><h4 class="panel-title">{% trans "Add new screenshot" %}</h4></div>
   <div class="panel-body">
-  {{ add_form|crispy }}
+  <p id="paste-screenshot-info-label" class="text-center"></p>
+  <button id="paste-screenshot-btn" class="btn" type="button">
+    {% trans "Paste from clipboard" %}
+  </button>
```
nijel, line 36: Can this be placed in the form, just next to the screenshot field:

image

It would need introducing rendering helper here and adding `Div` element to render the additional content:

https://github.com/WeblateOrg/weblate/blob/c006da8c8954d56fcac444833407e42e12f014cc/weblate/screenshots/forms.py#L25-L52

An example can be found here:

https://github.com/WeblateOrg/weblate/blob/c006da8c8954d56fcac444833407e42e12f014cc/weblate/trans/forms.py#L2305-L2312

weblate/static/js/screenshots/clipboard-paste.js
meel-hd: I see. Okay I will work on it.

weblate/templates/screenshots/screenshot_list.html
meel-hd: Okay. Thanks for the examples. I found where the form was created but I thought it would be better if I didn't change it here but change the markup instead. However I think it's better if it's closer to the input.

nijel: Part of the button stayed, removed in d944d52623.

meel-hd: 😳 Thanks for the fix.

weblate/screenshots/forms.py
```diff
@@ -51,6 +52,29 @@ def __init__(
...
 class SearchForm(forms.Form):
     q = QueryField(required=False)
+
+
+class PasteButtonWidget(forms.Widget):
+    def render(self, name, value, attrs=None, renderer=None):
+        info_label = '<p id="paste-screenshot-info-label" class="text-center"></p>'
+        btn_text = gettext_lazy("Paste from clipboard")
+        btn_template = f'<button id="paste-screenshot-btn" class="btn" type="button">{btn_text}</button>'
+        return info_label + btn_template
```
nijel: I've rewritten this to avoid using dummy input in 4d957772160ed9743c45cdbfeec4d9eebde422a8

nijel (commented):
I've made some changes to this:

* d944d526236685ecdaf5f3970e0213cddc3c56f7 removes extra html after changes in the PR
* 4d957772160ed9743c45cdbfeec4d9eebde422a8 customizes input widget instead of adding a dummy one
* 9d74d1b7b332f5f98d610eadfd58197c50000cb6 adds pasting to the screenshot editing
* c7013bb9e5 adds changelog entry

PR #12198, opened 2024-08-05
feat(checks): detect extra curly braces in Python brace format

### LINKED ISSUE #12197: Detect to many curly brackets (Python)
### Describe the problem

Translated strings used in Python do have (sometimes) to many curly brackets. Weblate do not warn about that. It cause Exceptions in the running code.

Weblate could detect it somehow.

_No response_

### Screenshots

image

### Additional context

_No response_

nijel:
## Proposed changes

Fixes #12197

## Checklist

### DIFF
--- a/weblate/checks/format.py
+++ b/weblate/checks/format.py
@@ -7,7 +7,7 @@ from __future__ import annotations
 import re
 from collections import Counter, defaultdict
 from re import Pattern
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Literal
 
 from django.utils.functional import SimpleLazyObject
 from django.utils.html import format_html_join
@@ -17,6 +17,8 @@ from django.utils.translation import gettext, gettext_lazy
 from weblate.checks.base import SourceCheck, TargetCheck
 
 if TYPE_CHECKING:
+    from collections.abc import Iterable
+
     from weblate.trans.models import Unit
 
 PYTHON_PRINTF_MATCH = re.compile(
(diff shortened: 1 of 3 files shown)


buhtz:
Thank you for this quick response.
I am not sure but the code looks like that it wouldn't catch all situations. And I don't understand the unit tests. The test code is not self explaining and has no comments. It is not clear for me which one of the cases should be an error and which not.

Other cases:
- `"foo }bar{"` (seems not to fail in your code)

Maybe my [own solution](https://github.com/buhtz/backintime/blob/898ca22d97a7858582a549d423ccd1293e1c46d9/update_language_files.py#L192-L254) using a combination of regex can give you some inspiration.

I also encountered strings like this: `"{{foobar}}"`. I am aware that this is a valid Python string to escape curly brackets. But in real world in most cases this is a typo from the translator. So if you see a possibility it would be nice if Weblate could warn about that, too. Or maybe compare it to the original string and look if there are escaped brackets, too.

weblate/checks/format.py
```diff
@@ -340,7 +344,9 @@ def normalize(self, matches: list[str]) -> list[str]:
     def extract_matches(self, string: str) -> list[str]:
         return [self.cleanup_string(x[0]) for x in self.regexp.findall(string)]
 
-    def check_format(self, source: str, target: str, ignore_missing, unit: Unit):
+    def check_format(
```
buhtz, line 347: The doc strings has potential to be improved. Not clear what it does. Arguments and return values are unclear.

Because of that I am unable to value the new unit tests using this method.

nijel: Indeed, the documentation always has potential to get improved. Currently, I'm at least focusing on adding type annotations when touching the code.

nijel:
https://github.com/WeblateOrg/weblate/pull/12199 should improve this.

There is currently nobody to do a review, but I hope this to change soon as some new team members will get fully onboarded.

buhtz:
Are you and the planed reviewers employed by Weblate company?
You might contact the Codeberg.org community and people. The might have capacity and they depend on Weblate.

PR #12199, opened 2024-08-05
fix(checks): make python brace format more robust

nijel:
## Proposed changes

- look for escaped braces in the original string as well
- treat { } escapes same as with %
- remove bogus detection of % placeable
- detect any stray { } besides parsed format string

See #12198 and #12197

## Checklist

### DIFF
--- a/weblate/checks/format.py
+++ b/weblate/checks/format.py
@@ -104,6 +104,8 @@ PASCAL_FORMAT_MATCH = re.compile(
 
 PYTHON_BRACE_MATCH = re.compile(
     r"""
+    }(})|                                 # escaped {
+    {({)|                                 # escaped }
     {(                                  # initial {
         |                               # blank for position based
         (?P<field>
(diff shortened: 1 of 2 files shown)


buhtz:
Thanks for this. But valuable doc strings are still missing. You won't attract new contributors with that especially unpaid ones.

buhtz:
The bots polluting the "Files changed" view IMHO don't help improve the situation but make it more worse because it distracts people.

nijel:
The bot comments on missing code coverage in the patch, and that is useful. If you don't see a value in that, it doesn't mean it's not useful for somebody else.

We don't need guidance on where more work can be done, I'm aware of many areas where Weblate can improve. Weblate is now healthier than ever with growing community contributions, sustainable funding and a growing team which will be soon backed by a Weblate company. It indeed was a one-man show for a long time, but now it's not the time to complain about that.

But this is really off-topic here, please consider using discussions next time because we're polluting the pull request discussion and distracting people.

PR #12173, opened 2024-08-01
feat(auth): add two-factor authentication

nijel:
## Proposed changes

- Implemented using django-otp and django-otp-webauthn
- Support for TOTP, WebAuthn and recovery codes

Fixes #1681

## Checklist

Pending tasks:

- [x] Implement configuration views
- [x] Implement generating backup codes
- [x] Implement view backup codes with audit log tracking (compare to https://github.com/settings/auth/recovery-codes)
- [x] Implement naming of security tokens (see https://github.com/Stormbase/django-otp-webauthn/issues/16)
- [x] Implement TOTP registration
- [x] Implement authentication flow for login
- [x] Implement authentication flow for social auth
- [x] Do manual testing
- [x] Implement WebAuthn login flow
- [x] https://github.com/Stormbase/django-otp-webauthn/issues/14 needs to fixed (perhaps by https://github.com/Stormbase/django-otp-webauthn/pull/15) and fix released
- [x] https://github.com/Stormbase/django-otp-webauthn/issues/17 needs to be fixed and fix released

- [x] Lint and unit tests pass locally with my changes.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] I have added documentation to describe my feature.
- [x] I have squashed my commits into logic units.
- [x] I have described the changes in the commit messages.

PR #12200, opened 2024-08-05
feat(auth): add option to enforce second factor authentication

nijel:
## Proposed changes

Built on top of https://github.com/WeblateOrg/weblate/pull/12173
   
    
- projects can define this and it will block users from editing
- teams can define this and roles from the teams will not be applied unless user is 2fa verified
- middleware warns user about not meeting this requirement

## Checklist


orangesunny:
@nijel, I would wait with a review until 12173 is merged, as it touches the same files and I did some corrections there. I would rather not get lost or create conflicts.

nijel:
Ready for review, #12173 is merged and this one rebased on top of that.

orangesunny (approved):
Good to go; can’t wait to use it!

docs/admin/projects.rst
```diff
@@ -200,6 +200,16 @@ Configure per project access control, see :ref:`acl` for more details.
 
 The default value can be changed by :setting:`DEFAULT_ACCESS_CONTROL`.
 
+.. _project-enforced_2fa:
+
+Enforced two factor authentication
```
nijel, line 205:
```suggestion
Enforced two-factor authentication
```

Issue #12209, opened 2024-08-06
.stringdict plural forms incorrect


dwaynebailey:
### Describe the issue

Comparing Android to iOS plural forms for Afrikaans. They are different while both claiming to be using CLDR. I suspect this is only a visual anomaly but it is enough to confuse translators.

Android: 
- https://hosted.weblate.org/translate/organicmaps/android/af/?checksum=e8e3cb8cdbc14c55
- Plural form: `n != 1`
- Results in: 
```
Android
integer -> index -> cldr form
0 -> 0 -> other
1 -> 1 -> one
2 -> 0 -> other
3 -> 0 -> other
4 -> 0 -> other
5 -> 0 -> other
10 -> 0 -> other
  ```
iOS: 
- https://hosted.weblate.org/translate/organicmaps/wip-ios-plurals-please-dont-use/af/?checksum=efcf782307b0b6f3
- Plural form: `n==0 ? 0 : n!=1 ? 1 : 2`
- I am aware of the iOS `zero` injection for stringdict, so that is ignored
- Results on:
```
iOS
integer -> index -> cldr form
0 -> 0 -> zero
1 -> 2 -> other
2 -> 1 -> one
3 -> 1 -> one
4 -> 1 -> one
5 -> 1 -> one
10 -> 1 -> one
``` 

I've traced the problem here:
https://github.com/WeblateOrg/weblate/blob/stable/weblate/lang/data.py#L235
but I'm not sure:
1. How these where derived
2. If they are merely informational and for providing the CLDR examples, or
3. If they are actually what Android uses

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

1. Ensure you have iOS stringdict and Android projects for similar strings
2. Enable Afrikaans (af) which has two forms (one, other)
3. Compare the suggested numeric help between Android and iOS `one` and `other` are swapped

### Expected behavior

Honour the CLDR forms in iOS plurals https://www.unicode.org/cldr/charts/42/supplemental/language_plural_rules.html#af

The correct formula would be: `n==0 ? 0 : n==1 ? 1 : 2`

### How do you run Weblate?

weblate.org service

nijel:
See https://github.com/WeblateOrg/weblate/issues/6368

nijel:
Sorry, I didn't understand this from the initial post.

[...]

PR #12219, opened 2024-08-07
Fix plus zero plural formulae

dwaynebailey:
Partial fix for #12209 plural forms plus zero

## Checklist


nijel:
As you call the fix partial, what do you think it missing? I think these are the only ones wrong formulas.

dwaynebailey:
> As you call the fix partial, what do you think it missing? I think these are the only ones wrong formulas.

You are probably right, but I didn't have time to check through the other forms :(

nijel:
That's okay, I just wanted to ensure that I'm not missing anything.

PR #12227, opened 2024-08-08
Add webpack bundling for client side libraries.

meel-hd:
## Proposed changes
- Added webpack for bundling 3rd party libraries.
- Added Sentry using this workflow.
- Installs and updates are easy and familiar now.

## Checklist

## Other information
See `client/README.rst`.


client/README.rst
```diff
@@ -0,0 +1,75 @@
...
+    <script src="{% static 'js/vendor/<lib-name>.js' %}"></script>
+
+References
+----------
+
+For more information on Node.js, visit the official Node.js website: https://nodejs.org/
+
+To learn more about Yarn package manager, refer to the Yarn website: https://yarnpkg.com/
+
+For details on Webpack, check out the Webpack documentation: https://webpack.js.org/guides/getting-started/
```
nijel, line 75: I'd rather put this into docs/contributing/frontend.rst and have here just a short readme pointing to https://docs.weblate.org/en/latest/contributing/frontend.html.

nijel (commented):
Looks good! It still lacks some parts to replace the current workflow:

* Move documentation to a single place (mentioned in review comment)
* Adjust .github/workflows/yarn-update.yml to include the new workflow as an additional step, so that the the vendored libraries are automatically updated via GitHub actions
* Remove sentry from `scripts/yarn` as it is now handled via a different workflow

client/README.rst
meel-hd: Working on it

client/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "client-3rd-party",
+  "version": "0.0.1",
+  "description": "Used to manage 3rd party libraries for the  client side.",
+  "license": "GPL-3.0-or-later",
+  "scripts": {
+    "build": "webpack"
+  },
+  "dependencies": {
+    "@sentry/browser": "^8.24.0"
```
nijel, line 10: I think we should pin exact version here:

```suggestion
    "@sentry/browser": "8.24.0"
```

This way the webpack build will be reproducible. The update pull requests will be automatically done by renovate.

client/README.rst
meel-hd: should I remove the old section about dependency management in `frontend.rst`. And replace it with the new steps?

meel-hd:
> * Adjust `.github/workflows/yarn-update.yml` to include the new workflow as an additional step, so that the vendored libraries are automatically updated via GitHub actions

@nijel  I didn't add this in the previous commit [a17f8ce](https://github.com/WeblateOrg/weblate/pull/12227/commits/a17f8ce77ecebf6d9d9c54f3fd39ba6a01c27a0d) . I am not sure exactly what to do. I might break some things. If you may add this please.

`.github/workflows/yarn-update.yml` is used to update all client libraries. And the new instal/update workflow isn't used for all of them. I don't think we should update this Github Workflow, yet?

nijel:
I've pushed f8045595149bc569d66c372b79754fcabef98a00 which removes sentry and browserify from the existing package.json and lock file.

I think the GitHub action can use both workflows for now, so just after calling `script/yarn-update` it would have another step to build bundles with webpack. Once we migrate all libraries, the legacy code would be dropped.

meel-hd:
> so just after calling `script/yarn-update` it would have another step to build bundles with webpack.

Uh true. I will add it.

nijel:
I've merged this with fixes mentioned before. Let's see how first dependency update with the new stack will go in https://github.com/WeblateOrg/weblate/pull/12239

Issue #12293, opened 2024-08-19
IntegrityError occurs when deleting a component added from sratch


harriebird:
### Describe the issue

The IntegrityError occurs when adding an Android String Resource component from scratch and deleting it. The component and is deleted but it will lead to an error page same as shown below:

![Screenshot 2024-08-19 at 22-54-41 IntegrityError at _projects_blank-project_](https://github.com/user-attachments/assets/118eaf7f-0dcf-4da4-9e0e-eb686bd0d988)

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

1. Create a new Project.
2. In the newly created project, add a new component. In the Create component page, select `Start from scratch` tab. Provide a Component name, and set its File format to ` Android String Resource `. Click `Continue` to proceed with the creation of the component.
3. After a successful creation of the component, the Community localization checklist page is shown. Click `Return to the component`  to go to the Component page. In the Component page, click `English` in the list of languages.
4. In the English project language page, you will be a adding new strings via upload since it is blank. To add the new strings, click the `Files` dropdown and select `Upload translation`. This will show the Upload form for the translation.

5. Save the code below as XML file (.xml). This will be uploaded using the Upload form to add new strings. 
```
<resources>
    <string name="app_name">Github Visualizer</string>
    <string name="my_bio">"A material metaphor is the unifying theory of a rationalized space and a system of motion."</string>

    <!-- TODO: Remove or change this placeholder text -->
    <string name="hello_blank_fragment">Hello blank fragment</string>
    <string name="title_activity_profile">ProfileActivity</string>
    <string name="action_settings">Settings</string>
    <string name="title_activity_repository_info" translatable="false">RepositoryInfoActivity</string>
    <string name="title_activity_test_repo">TestRepoActivity</string>

    <string name="title_activity_file_viewer">FileViewerActivity</string>
    <string name="dummy_button">Dummy Button</string>
    <string name="dummy_content">DUMMY\nCONTENT</string>
</resources>
```

6. In the Upload form, set the saved XML file as file to be uploaded. Set the File upload mode to `Add new strings` and click `Upload` to begin the adding of new strings.
[...]

PR #12309, opened 2024-08-21
Add an exception to handle IntegrityError in the format_unit_source

harriebird:
## Proposed changes

This adds an exception to handle `IntegrityError` in the format_unit_source.
This solves issue #12293.

## Checklist


nijel:
This approach just hides the error and potentially others, so it's not a good way to deal with this. Let's find a better solution, see https://github.com/WeblateOrg/weblate/issues/12293#issuecomment-2306729809

PR #12320, opened 2024-08-23
10896 Duplicate component shortcut

### LINKED ISSUE #10896: Issues when duplicating a component

gersona:
## Proposed changes

- A New "Duplicate Component" was added to the `Manage` menu in component details page
- Additional attributes are copied from the source when duplicating : `agreement", "merge_style", "commit_message", "add_message", "delete_message", "merge_message", "addon_message", "pull_message",`

Capture d’écran 2024-08-23 à 09 42 01

## Checklist

## Other information
fixes #10896 

## Summary by CodeRabbit

- **New Features**
	- Introduced a "Duplicate Component" option in the menu for easier component replication.
	- Added functionality to duplicate components directly from the editing interface for users with appropriate permissions.

- **Improvements**
	- Enhanced commit messages for translation operations to provide clearer context and information.
	- Updated forms to allow for more flexible data entry by permitting certain fields to be left blank.

- **Tests**
	- Improved test coverage for component creation and category management, ensuring robustness in functionality.


nijel:
@coderabbitai review

weblate/trans/models/component.py
```diff
@@ -672,6 +677,7 @@ class Component(models.Model, PathMixin, CacheKeyMixin, ComponentCategoryMixin):
             "please consult the documentation for more details."
         ),
         validators=[validate_render_addon],
+        blank=True,
```
nijel, line 680: We really don't want these to be blank at model level. If these need to be optional in a particular form, let's override it there, but not in the model as that has an impact on all users (for example API).

weblate/trans/views/create.py
```diff
@@ -498,6 +540,10 @@ def get_form(self, form_class=None, empty=False):
             ).order_project()
             form.branch_data = self.branch_data
         elif isinstance(form, ComponentSelectForm):
+            if self.duplicate_existing_component:
+                self.components |= Component.objects.filter(
```
nijel, line 544: This should filter components user has access to avoid the possibility of cloning restricted component.

```suggestion
                self.components |= Component.objects..filter_access(self.request.user).filter(
```

weblate/trans/tests/test_create.py
```diff
@@ -180,22 +182,120 @@ def test_create_component_wizard(self) -> None:
...
+                    "name": "Create Component From Existing",
+                    "slug": "create-component-from-existing",
                     "component": self.component.pk,
+                    "is_glossary": self.component.is_glossary,
                 },
                 follow=True,
             )
+
         self.assertContains(response, self.component.get_repo_link_url())
+        parsed_query = urllib.parse.parse_qs(response.request["QUERY_STRING"])
```
nijel, line 222: Why not to use request.GET here?

PR #12345, opened 2024-08-27
Fix IntegrityError when redirected from a Component, Category, and Project delete action

harriebird:
## Proposed changes

Issue #12293 is caused by prefetching records that are related to the deleted Component, Category, or Project. To solve this, a boolean field `to_delete` is added to the `Component`, `Category`, and `Project` models. The `to_delete` field will help determine those records that are to be deleted and exclude them in the prefetch. 
This solves issue #12293.

## Checklist


weblate/trans/views/settings.py
```diff
@@ -198,16 +209,27 @@ def remove(request: AuthenticatedHttpRequest, path):
...
+        flag_categories_to_delete(categories)
         category_removal.delay(obj.pk, request.user.pk)
         messages.success(request, gettext("The category was scheduled for removal."))
     elif isinstance(obj, Project):
         parent = reverse("home")
+        obj.to_delete = True
+        obj.save()
+        for component in obj.component_set.all():
+            component.to_delete = True
+            component.save()
```
nijel, line 230: Avoid using look when database can do that more effectively:

```suggestion
        obj.component_set.update(to_delete=True)
```

weblate/trans/views/settings.py
```diff
@@ -175,6 +175,17 @@ def dismiss_alert(request: AuthenticatedHttpRequest, path):
...
 
+def flag_categories_to_delete(categories):
+    while len(categories) > 0:
+        category = categories.pop()
+        category.to_delete = True
+        category.save()
+        categories += category.category_set.all()
+        for component in category.component_set.all():
+            component.to_delete = True
+            component.save()
```
nijel, line 186: This can be handled more effectively by using `update()`, see below.

weblate/trans/views/settings.py
```diff
@@ -198,16 +209,27 @@ def remove(request: AuthenticatedHttpRequest, path):
         messages.success(request, gettext("The translation has been removed."))
     elif isinstance(obj, Component):
         parent = obj.category or obj.project
+        obj.to_delete = True
+        obj.save()
```
nijel, line 213: The `save()` here will trigger repository rescan what is not really desired. Better use:

```suggestion
        Component.objects.get(pk=obj.pk).update(to_delete=True)
```

nijel (changes requested):
What I miss is filtering on `to_delete` when browsing. It will probably also need a database index to make queries on it effective.

weblate/trans/views/settings.py
harriebird: Optimized this function using `update()` and removing the loop over components.

weblate/trans/views/settings.py
harriebird: Applied this change suggestion.

weblate/trans/views/settings.py
harriebird: Applied this change suggestion.

harriebird:
I added a filter to exclude in those Projects, Categories, and Components that has the `to_delete` value of `True` in the listing.

nijel:
@coderabbitai review

harriebird:
@coderabbitai review

weblate/trans/views/settings.py
```diff
@@ -199,16 +215,23 @@ def remove(request: AuthenticatedHttpRequest, path):
...
+        categories = [obj]
+        flag_categories_to_delete(categories)
         category_removal.delay(obj.pk, request.user.pk)
         messages.success(request, gettext("The category was scheduled for removal."))
     elif isinstance(obj, Project):
         parent = reverse("home")
+        Project.objects.filter(pk=obj.pk).update(to_delete=True)
+        obj.component_set.update(to_delete=True)
+        categories = obj.category_set.all()
+        flag_categories_to_delete(categories)
```
nijel, line 234: The very same would have to happen in the API calls as well.

weblate/trans/models/component.py
```diff
@@ -770,6 +783,7 @@ class Component(models.Model, PathMixin, CacheKeyMixin, ComponentCategoryMixin):
     remote_revision = models.CharField(max_length=200, default="", blank=True)
     local_revision = models.CharField(max_length=200, default="", blank=True)
     processed_revision = models.CharField(max_length=200, default="", blank=True)
+    to_delete = models.BooleanField(default=False)
```
nijel, line 786: This would need a database index. Otherwise it will make the things unnecessarily slow.

weblate/trans/models/change.py
```diff
@@ -114,14 +115,26 @@ def prefetch(self):
...
+                "component",
+                queryset=component.objects.exclude(category__to_delete=True),
+            ),
             "project",
             "component__source_language",
-            "unit",
+            Prefetch(
+                "unit",
+                queryset=unit.objects.exclude(translation__component__to_delete=True),
+            ),
```
nijel, line 137: This is unnecessarily complex, better to filter changes for objects not scheduled for removal.

nijel (commented):
I'm still not convinced this is a viable approach. We would have to fix *all* queries that could possibly run into this, what is a complex and error-prone task and will pollute the code.

Furthermore, in case of error in the removal (like https://github.com/WeblateOrg/weblate/issues/12627), this would lead to zombie components which are not visible and cannot be removed again.

PR #12383, opened 2024-09-01
feat(js): use local storage instad of cookies for tab persistence

nijel:
## Proposed changes

This was never needed server-side, so there is no need to use cookies for this. It also allows to drop one JavaScript library.

## Checklist


weblate/static/editor/full.js
```diff
@@ -199,16 +199,11 @@
   FullEditor.prototype.initTabs = function () {
     /* Store active tab in a cookie */
     $('.translation-tabs a[data-toggle="tab"]').on("shown.bs.tab", function () {
-      const current = Cookies.get("translate-tab");
+      const current = window.localStorage.getItem("translate-tab");
```
meel-hd, line 202: No need for calling `window.localStorage` from the window object. As `localStorage` is available on the global scoop by default.

Also in the `loader-bootstrap.js`.

weblate/static/loader-bootstrap.js
```diff
@@ -611,10 +611,12 @@ $(function () {
     }
   } else if (
     $(".translation-tabs").length > 0 &&
-    Cookies.get("translate-tab")
+    window.localStorage.getItem("translate-tab")
   ) {
     /* From cookie */
```
meel-hd, line 616: `/* From localStorage */`

meel-hd (commented):
Overall no major issues. Only some small notes

Issue #7085, opened 2022-01-10
Avoid scrolling when adding screenshots


luebbe:
### Describe the problem

When adding/editing screenshots for the translators, the OCR doesn't have a chance to find all strings that should be associated with this screenshot. The most obvious ones are hint text for buttons which have only an icon and no caption. Sometimes also strings that contain accelerator keys like "Save &all" are not recognized properly, because they have an underscore "Save _all" under the letter (at least on Windows). 
This means that I have to add all relevant strings from memory and this also means that I probably will miss a lot.

**Edit**:  The actual issue is described in https://github.com/WeblateOrg/weblate/issues/7085#issuecomment-1009950253

### Describe the solution you'd like

At least the .po format contains a very valuable information, the location of the string. If the search function in the screenshot dialogue would allow me to search for the location of a string (either optional or always), it would be much easier to find all strings associated to a certain form.

### Screenshots

![grafik](https://user-images.githubusercontent.com/3872576/148743317-fd3eedc5-5fa0-4800-9619-406a8d5153c6.png)
The form is called "ModeTemplates.dfm". I know that one of the button hints contains the string "create" so I can add it by searching for "Create". It would be nice to be able to add all strings for this form by simply searching for "ModeTemplates.dfm"

nijel:
The search is same as elsewhere, so use `location:search-term`. See https://docs.weblate.org/en/latest/user/search.html

luebbe:
Thanks! That's cool. The idea didn't occur to me...

luebbe:
That's true. Only while adding strings, the top list grows and pushes the bottom list and the items on which I may want to click next off screen.

meel-hd:
@nijel  this issue is related to OCR. Where is the problem with scrolling? Wouldn't it be better to open a new issue and close this one as it seems resolved?

nijel:
I think the remaining issue to address here is:

> while adding strings, the top list grows and pushes the bottom list and the items on which I may want to click next off screen.

So while adding, the viewport should not move because of strings added to the top list.

PR #12418, opened 2024-09-04
Avoid scroll in Add Strings to screenshot

meel-hd:
## Proposed changes
**Avoid scroll in Add Strings to screenshot**

- Added a height limit to the first table showing strings related to the screenshot. to prevent scroll and hiding the second table.
- Fixed buttons overallaping in the actions of the tables on mobile devices

Fixes: #7085

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced a new CSS file for improved styling of the sources listing component, enhancing usability and aesthetics, especially on mobile devices.
	- Added a flexible `extra_styles` block in the HTML templates to allow for custom CSS in derived templates, improving styling options for the screenshot detail page.

- **Style**
	- Enhanced layout and spacing for the footer and buttons in the sources listing component.
	- Updated the search input element for better styling consistency.

### DIFF
--- a/weblate/static/styles/screenshots/screenshot_detail.css
+++ b/weblate/static/styles/screenshots/screenshot_detail.css
@@ -1,0 +1,26 @@
+/*
+ * Copyright © Michal Čihař <michal@weblate.org>
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+#sources-listing {
+  max-height: 400px;
+  overflow-y: auto;
+}
+
+@media (max-width: 767px) {
+  #sources-listing {
+    max-height: 300px;
+  }
+
+  #sources-listing td > .btn {
+    margin-bottom: 5px;
+  }
+
+  .panel-footer {
+    gap: 20px;
+    display: flex;
+    flex-direction: column;
+  }
+}
--- a/weblate/static/styles/screenshots/screenshot_detail.css
+++ b/weblate/static/styles/screenshots/screenshot_detail.css
@@ -4,13 +4,13 @@
  * SPDX-License-Identifier: GPL-3.0-or-later
  */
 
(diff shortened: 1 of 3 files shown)


weblate/static/styles/screenshots/screenshot_detail.css
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright © Michal Čihař <michal@weblate.org>
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+#sources-listing {
```
nijel, line 7: I think it would look nicer if only the inner table was scrollable, not the full panel.

nijel (commented):
Looks better now. Can the margin below the table be removed? I think it was removed by bootstrap when the table was directly in the panel.

meel-hd:
> Can the margin below the table be removed?

Yes. But there will be no separation between it and the below element.
Is that what is desired? To hint the two tables are related. If otherwise I don't think it will look good.

nijel:
I mean this:

image

The space is there only if the table is scrolled up, so it really doesn't serve as a separator.

Before scrolling down, it is correct:

image

weblate/static/styles/screenshots/screenshot_detail.css
```diff
@@ -4,6 +4,10 @@
  * SPDX-License-Identifier: GPL-3.0-or-later
  */
 
+ #sources-listing {
+  margin-bottom: 0px;
+ }
```
nijel: This is not what I meant, see the screenshots I've just posted.

meel-hd: Uh I see. Fixing it.

nijel:

Issue #4106, opened 2020-06-29
Live preview of search queries


nijel:
**Is your feature request related to a problem? Please describe.**
When typing search queries in places like bulk edit or custom download, it would be useful to get a preview of matching strings.

**Describe the solution you'd like**
* Show number of matching string next to the search input. Should be live updated as typing, possibly with some delay to reduce server-side load.
* Add link to search showing matched strings.

GG-Sheridan:
Should this issue be labeled as Good First Issue?

We are doing research about how to help newcomers onboarding in an open source project and how to accelerate the issue resolving process.
We have developed a tool to automatically label the Good First Issue (GFI), i.e., the issue which is good for newcomers. 
We first extract 72 features from different dimensions of issue reports, such as readability, semantic meaning, associated code snippets, etc, and experimentally investigate to what extent these features correlate with GFI. Results turn out that a large portion of these features do have a medium to large correlation with GFI. 
We then build a machine learning model based on these features to predict GFI. The experiments on a dataset of 12k issues from Github show that 0.90 AUC can be achieved. 

Our tool shows that this issue has a very large possibility to be a GFI. 
Could you please tell us whether it should be labeled as a GFI?
And if you have any comments or suggestions with this tool, could you please tell us? 
Thank you so much.

meel-hd:
@nijel I started working on this.  first by working on `Search and Replace`. I will make the handler of the Post request of this tool check if a certain header is available indicating preview then it will return only the search result as JSON for the preview without performing the `Replace`. then consume the response and show it in a `popover` while the user is typing with a _certain interval._

nijel:
I think you can use API to perform the search server-side (see https://docs.weblate.org/en/latest/api.html#units). It might need additional predicate to support filtering by URL path to make things easier to implement.

meel-hd:
I have implemented it in [73a50ef](https://github.com/meel-hd/weblate/commit/73a50eff8e42379be16c522af64d16a335e125f1) for `Search and Replace` but not sure about the correctness of the results. meaning if the api returns the `units` related to the current project we are trying to edit, etc.
![image](https://github.com/user-attachments/assets/e5145613-7bc2-40a8-a26f-15de04217c10)

meel-hd:
@nijel from reading the docs about `api/search` I am not sure what is exactly needed to implement the search preview in `Tools`. I wonder what is missing to add, so this endpoint is complete for the search preview. I might jump on it and them!

[...]

PR #12470, opened 2024-09-11
feat(tools): add search preview functionality

meel-hd:
## Proposed changes
This commit adds the functionality to display a search preview when typing in the search input field. The search preview shows a list of search results that match the user's input. It also highlights the matching text in the search results. The search preview is displayed below the search input field and disappears after a delay when the input field loses focus.
- The search also uses filters if provided.
The commit includes the following changes:
- Added a new file `weblate/static/editor/tools/search.js` that contains the JavaScript code for the search preview functionality.
- Added CSS styles in `weblate/static/styles/main.css` to position and style the search preview element.

Related: #4106 

## Checklist

### Note
`GET api/units` with the query `q`  is case insensitive. In contrast to the underlying search in `Search and Replace`. If the preview consumed it solely, its results are inaccurate. This is avoided with client side filtering for accurate preview considering case sensitivity. 

I think we should consider this. as case sensitive search is important for some parts of the application.
### Preview
image

## Summary by CodeRabbit

- **New Features**
	- Introduced a dynamic search preview feature that provides real-time feedback on search results as users type.
	- Enhanced search experience with clickable results that open in new tabs and highlight relevant text.

- **Style Enhancements**
	- Added new CSS styles for the search preview component to improve aesthetics and user interaction, including hover effects for search results.

- **Documentation**
	- Updated HTML template to include the new search functionality script, enhancing the web application’s capabilities.

### DIFF
--- a/weblate/static/editor/tools/search.js
+++ b/weblate/static/editor/tools/search.js
@@ -144,7 +144,6 @@ $(document).ready(() => {
       .val();
     if (projectPath) {
       builtSearchQuery = `path:${projectPath}`;
-      console.log(builtSearchQuery);
     }
 
     // Add filters to the search query
--- a/weblate/static/editor/tools/search.js
+++ b/weblate/static/editor/tools/search.js
@@ -3,28 +3,22 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 
 $(document).ready(() => {
-  searchPreview("#replace", "#id_replace_search", "#id_replace_q");
-  // searchPreview('#bulk-edit', '#id_bulk_q');
+  searchPreview("#replace", "#id_replace_q");
 
   /**
    * Add preview to the search input search results.
    *
    * @param {string} searchForm The selector string of the parent element of the search input
    * @param {string} searchElment The selector string of the search input or textarea element
-   * @param {string} filtersElement The selector string of the filters input element
-   * @param {number} fetchMax The maximum number of preview results to fetch
    *
    */
-  function searchPreview(searchForm, searchElment, filtersElement, fetchMax) {
+  function searchPreview(searchForm, searchElment) {
     const $searchForm = $(searchForm);
     const $searchElment = $searchForm.find(searchElment);
-    const $filtersElement = $searchForm.find(filtersElement);
-
-    const includeFilters = !!filtersElement;
 
     // Create the preview element
     const $searchPreview = $('<div id="search-preview"></div>');
-    $searchElment.after($searchPreview);
+    $searchElment.parent().parent().after($searchPreview);
 
     let debounceTimeout = null;
 
@@ -32,7 +26,7 @@ $(document).ready(() => {
     $searchElment.on("input", () => {
       $searchPreview.show();
       const userSearchInput = $searchElment.val();
-      let searchQuery = `'${userSearchInput}'`;
+      const searchQuery = buildSearchQuery($searchElment);
 
       // Clear the previous timeout to prevent the previous
       // request since the user is still typing
@@ -41,28 +35,20 @@ $(document).ready(() => {
       // fetch search results but not too often
       debounceTimeout = setTimeout(() => {
         if (userSearchInput) {
-          searchQuery = buildSearchQuery(
-            searchQuery,
-            $filtersElement,
-            includeFilters,
-          );
           $.ajax({
             url: "/api/units/",
             method: "GET",
-            data: { q: searchQuery, page_size: fetchMax },
+            data: { q: searchQuery },
             success: (response) => {
               // Clear previous search results
               $searchPreview.html("");
-
+              $("#results-num").remove();
               const results = response.results;
               if (!results || results.length === 0) {
                 $searchPreview.text(gettext("No results found"));
               } else {
-                showResults(results, userSearchInput);
+                showResults(results, $searchElment);
               }
-            },
-            error: (error) => {
-              console.error("Error fetching search results: ", error);
             },
           });
         }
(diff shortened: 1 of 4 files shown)


weblate/static/editor/tools/search.js
```diff
@@ -0,0 +1,185 @@
...
+                for (const result of results) {
+                  const key = result.context;
+                  const source = result.source;
+                  const sourceHighlighted = highlightMatch(
+                    userSearchInput,
+                    source.toString(),
+                  );
+
+                  // Make the URL relative
+                  const url = result.web_url.replace(/^(?:\/\/|[^/]+)*\//, "/");
```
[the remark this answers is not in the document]
nijel: What is reason for this?

weblate/static/editor/tools/search.js
```diff
@@ -0,0 +1,185 @@
...
+   * */
+  function buildSearchQuery(searchQuery, $filtersElement, includeFilters) {
+    let builtSearchQuery = searchQuery;
+
+    // Add path lookup to the search query
+    const pathSplited = window.location.pathname.split("/");
+    const projectPath = `${pathSplited[2]}/${pathSplited[3]}`;
+    if (projectPath) {
+      builtSearchQuery = `${searchQuery} path:${projectPath}`;
+    }
```
nijel, line 154: Maybe it would be better to include this in the form instead of figuring this from the URL?

nijel (commented):
The preview really makes sense for the filter field above, not this one. The filter supports all queries, while the search is a simple string to replace in the translation.

weblate/static/editor/tools/search.js
meel-hd: I tried to get the `action` attr from the form but somehow I was getting `undefined` so I simply went for the pathname as it is always readily available.

meel-hd: Also  the part of url manipulation is done to remove the domain and the  protocol of the `web_url`returned by the api. In local development it was returning  the ip of the server which when I go to it was unresolved (server running in wsl was port forwared to localhost).  Also not sure how it would behave in production, I didn't want to make any assumptions. So in this case a relative path would pick up the domain and the protocol of the currently loaded page. Which is I guess better.

meel-hd:
Oh!? I think the main operation here is concerned with the **strings** to replace. And filters are additional. In the current solution the preview is mainly based on the search string  while considering the filters. I don't think adding search preview to filters is the best option. as users usually don't search for filters but for strings.

meel-hd:
@nijel this commit [e3bd4d](https://github.com/WeblateOrg/weblate/pull/12470/commits/e3bd4d444f8d3366176be877cef62a1cae6b05b2) moved the search preview to filters as you suggested.
## Preview

https://github.com/user-attachments/assets/5b8fcdac-cace-41c9-89d9-c8ef8acec709

weblate/static/editor/tools/search.js
```diff
@@ -0,0 +1,153 @@
...
+
+    /**
+     * Handles the search results and displays them in the preview element.
+     * @param {any} results fetched search results
+     * @param {jQuery} $searchElment The search input element
+     * @returns void
+     */
+    function showResults(results, $searchElment) {
+      // Show the number of results
+      if (results.length > 0) {
```
nijel, line 98: The actual number of matches is in reponse.count, response.results is paginated and contains at most 100 matches (what is okay for display). The count should be also used in the message shown below.

weblate/static/editor/tools/search.js
```diff
@@ -0,0 +1,153 @@
...
+        );
+      } else {
+        $("#results-num").remove();
+      }
+
+      for (const result of results) {
+        const key = result.context;
+        const source = result.source;
+
+        // Make the URL relative
```
nijel, line 110: What is the reason for  making URL relative?

weblate/static/editor/tools/search.js
```diff
@@ -0,0 +1,153 @@
...
+   * @param {jQuery} $searchElment - The user input.
+   * @returns {string} - The built search query string.
+   *
+   * */
+  function buildSearchQuery($searchElment) {
+    let builtSearchQuery = "";
+
+    // Add path lookup to the search query
+    const pathSplited = window.location.pathname.split("/");
+    const projectPath = `${pathSplited[2]}/${pathSplited[3]}`;
```
nijel, line 141: This is wrong - it strips language and it will break on categorized components. Instead of guessing it from location (what is easy to break in the future), let's push it from backend. I've done that in https://github.com/WeblateOrg/weblate/pull/12602, you can then get the path from the form.

weblate/static/editor/tools/search.js
meel-hd: the url returned contains different domain than the current one in the opened tab. Making it relative we insure it always open in the current domain not the one returned by the server. I could remove it. But I didn't want to assume in prod the url returned will be the same as the one of the page.

meel-hd: Because if we left the url as is. When the user clicks it opens in inaccurate
 url (in development at least, not sure in prod)

weblate/static/editor/tools/search.js
```diff
@@ -0,0 +1,156 @@
...
+     * @param {any} results fetched search results
+     * @param {number} count The number of search results
+     * @returns void
+     */
+    function showResults(results, count) {
+      // Show the number of results
+      if (count > 0) {
+        const t = ngettext("%s matching string", "%s matching strings", count);
+        $searchPreview.append(
+          `<h4 id="results-num">${interpolate(t, [count])}</h4>`,
```
nijel, line 101: This could be a link to `/search/?q=...` with the search used in the API.

nijel (commented):
Looks good. I've fixed merge conflicts and added a changelog entry.

weblate/static/editor/tools/search.js
meel-hd: Yeah! okay

orangesunny (approved):

Issue #7655, opened 2022-05-17
Rework UI & UX


6543:
### Describe the problem

If you are used to weblate it might not fall in your eyes ...
It's hard to warp around your head if you did got in touch with [pontoon](https://pontoon.mozilla.org/) or  [Crowdin](https://crowdin.com/)

*This issue is not a simple feature-request but try to start a discussion how make it better.*

I do really see a downside of using Weblate for a project that you only can expect contributions from developers, I could get normal people only translation on crowdin ...

### Describe the solution you'd like

Get some volunteer designeer ...  
https://opensourcedesign.net/jobs/ might be an idea

### Describe alternatives you've considered

use some other platform - I know this sounds harsh & that weblate is FOSS, so consider this not a personal criticism the features and what weblate is able to do and already achieved **is awesome** 

### Screenshots

especialy the translation editor and how you filter stuff:
![Screenshot from 2022-05-17 17-55-37](https://user-images.githubusercontent.com/24977596/168855821-7c8a5d94-e170-4463-9b13-d19cf5b86e74.png)
![Screenshot from 2022-05-17 17-56-09](https://user-images.githubusercontent.com/24977596/168855837-1a6f7193-d3dc-4c62-99ab-2ff8543947b2.png)
![Screenshot from 2022-05-17 17-57-07](https://user-images.githubusercontent.com/24977596/168855843-999068eb-0652-4d74-92af-9f5fddc9a4ae.png)

### Additional context

ref https://github.com/woodpecker-ci/woodpecker/issues/258

nijel:
There is always room for improvements (for example, some suggestions are in https://github.com/WeblateOrg/weblate/issues/7391), but I don't think copying Crowdin is a way to go. Actually, we've also received quite some feedback that users like Weblate UI over Crowdin, so it might also be a matter of personal preference...

6543:
hmm should I create a new discussion at https://opensourcedesign.net/, so designers could have a look at?

Aikatsui:
https://github.com/WeblateOrg/weblate/issues/6393

Aikatsui:
> @nijel but I don't think copying Crowdin is a way to go. Actually, we've also received quite some feedback that users like Weblate UI over Crowdin, so it might also be a matter of personal preference...

> @6543 or add some theme engine? so community can write one as they like?

I think we can improve **zen mode** for that and keep weblate default UI without touch (it can go for different path)

[...]

PR #12501, opened 2024-09-14
Make the UI modern and Minimal

meel-hd:
### Summary:
Improved UI to be more minimal by reducing the number of colors used and their intensity. Also made it Modern by using more rounded corners and subtle shadows. Plus started depending more on css variables than hard coded values.

### Modifications include changes in:
- Navbar and its items
- Dropdown menus
- Input fields
- Nav Pills
- Panels and their headings

These changes while being opinionated they introduce a beautiful and modern UI. More suitable and aligned to the current trends and the community.

Relalted: #7655

## Checklist

#### Main Page (note navbar, dropdown menu, and nav pills)
b
#### Statistics Page (note panels and their headings)
a

## Summary by CodeRabbit

- **New Features**
	- Introduced new CSS classes for improved navigation and UI components, including `.navbar-default`, `.panel`, and `.multi-wrapper`.
	- Enhanced styling for buttons, input fields, and progress bars with consistent border radius and color variables.

- **Improvements**
	- Updated CSS variables for better maintainability and responsiveness, including support for dark mode.
	- Refined hover states and active visuals for navigation elements.

- **Bug Fixes**
	- Adjusted the navigation bar's styling for a more cohesive design by removing the `navbar-inverse` class.

### DIFF
--- a/weblate/static/styles/main.css
+++ b/weblate/static/styles/main.css
@@ -58,8 +58,27 @@
 }
 
 .navbar {
-  margin-bottom: 0;
+  margin-bottom: 10px;
+  padding: 2px 0;
   border: none;
+}
+
+.navbar-default {
+  background-color: unset;
+  box-shadow: var(--box-shadow);
+}
+
+.navbar-default * {
+  color: var(--text-color) !important;
+}
+
+.navbar .dropdown .dropdown-toggle {
+  border: 1px solid transparent;
+}
+
+.navbar .dropdown.open .dropdown-toggle {
+  border-radius: var(--border-radius);
+  border: var(--border);
 }
 
 .navbar-toggle {
@@ -101,6 +120,10 @@ h1 {
   margin-top: 0;
 }
 
+input {
+  border-radius: var(--border-radius) !important;
+}
+
 .main-content {
   margin: 15px;
 }
@@ -120,6 +143,7 @@ h1 {
 
 .progress-bar {
   box-shadow: none;
+  border-radius: var(--border-radius);
 }
 
 .progress-bar-dark {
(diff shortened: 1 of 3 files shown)


nijel:
Thanks, but I don't think this is something we should do now. We will definitely work on UI in the future, but this needs to start with UI design and implementing desired changes then. Making random UI changes without prior decision is not the way to go.

Moreover, when making UI changes, there are some existing features to be considered:

* There is [appearance customization](https://docs.weblate.org/en/latest/admin/admin.html#manage-appearance) in the management interface and that one should not be broken (removing the background from navbar does that because its color is one of the things users can customize).
* The dark theme is also user selectable, thus using media query is not the way to introduce dark theme changes, they need to go into `weblate/static/style-dark.css`. Ideally, this file would only change variables which are used in the main theme, but we're not there.

And some changes look worse to me. The navbar menu is simple and aligned:

obrazek

But with button like selection and navbar shadow it looks more messy:

obrazek

meel-hd:
Oh! That's unfortunate. while effort was made on these changes I believe it could have been used somewhere else more beneficial and aligned with the project if the changes were discussed before. While the issues with the menu and dark-theme rules could be easily fixed, I think that's irrelevant to discuss anymore.

It was a fun  little experiment I thought it would produce beautiful results but beauty is subjective. I wonder what are some of the good things you see appropriate; maybe we can incorporate or modify a little bit so they are.

nijel:
Generally I'm not opposed to UI improvements, but if it is completely changes what users see, it needs to be carefully considered before.

I want you to focus on working on issues for now so that you can learn the code structure (which definitely would use some improvements like using CSS variables which were not widely supported back then when these CSS were initially written) and Weblate features (which also influence what you can do in the CSS to avoid breaking them). We will be definitely doing more UI changes in the future, but for the bigger changes I prefer to start with an UX design and then implementing them rather than doing that without having a consistent design guide.

meel-hd:
Okay. I understand

Issue #12362, opened 2024-08-29
Add Kabyle to Google Machinery


BoFFire:
### Describe the problem

Hi,

Google Translate added recently a list if new language amont which we can find `Tamazight`. A lot of people notices that this berber language was trained on a corpus language which is in majority `Kabyle` [I'm Kabyle speaker].

### Describe the solution you would like

This is why I suggest to add mapping `Tamazight` to `Kabyle` on Weblate by setting `ber-Latn` to point towards `kab` in the adequate machinery.

Please check the structure of the URL and if API responds well :
https://translate.google.com/?sl=en&tl=ber-Latn&text=Hello!&op=translate

On LibreTranslate, we use :

https://translate.fedilab.app/?source=en&target=kab&q=Hello%21

nijel:
Isn't this merely a bug on Google's side which will likely be addressed?

BoFFire:
Hi,
I'm working to introduce a PR for this.

PR #12504, opened 2024-09-14
Adding kabyle mapping to Google machinery

BoFFire:
Adding kabyle mapping to Google machinery by mapping the so called Tamazight Latin.

## Proposed changes

As described here : https://github.com/WeblateOrg/weblate/issues/12362

## Summary by CodeRabbit

- **New Features**
	- Introduced support for the Kabyle language (ber-Latn) in the translation functionality, expanding available language options for users.


nijel (commented):
This will map Weblate `bet-Latn` to `kab`, I thought it should be the other way. Additionally, by question from #12362 is still valid. Will this last, or is this a temporary issue on Google's side which will produce unexpected results in a future?

BoFFire:
Hi @nijel 
Actually what I want to achieve is that Weblate would send a translation query to Google Translate and when Google Translate will answer in `ber-Latin`, map the answer to offer it to `kab` language under suggestions to translators.

Now, Google doesn't know something called Kabyle language or per code `kab`. People noticed that Google is calling it Tamazight Latin. Yes Kabyle is « a Tamazight » or « Berber » language.

And in fact the translation in response is mostly Berber based on Kabyle corpus, because their Tamazight model was mainly trained on Kabyle corpus.

It is not a Google Translate issue, they can still calling Tamazight for « political » reasons.

But in other hand, I will not advise to map « Tamazight in Tifinagh script » to `kab-tfng` because the answer to the query is addressed more to Tamazight of Atlas in Morocco than to Kabyle in Algeria.

To go further, actually even what I'm requesting now will not be useful to translators as the translation generated by Google Translate is poor comparing with the first translation model made by LibreTranslate community.

And yes, I've asked LibreTranslate if it is possible to add the Kabyle model to the main list of models. My request is marked as `enhancement` : https://github.com/LibreTranslate/LibreTranslate/issues/652

So even if my actual suggestion to map the Tamazight Latin language to Kabyle, we will be able to use better TM via LibreTranslate because there is no reliable TM for Kabyle all over the web as it is considered as a « low resourced language ».

And if my suggestion for mapping makes sense, I suggest that we disable pre-auto-translation for kabyle based on Google Translate because of the poor suggestion generated, same for OpenAI :)

nijel:

PR #12542, opened 2024-09-20
chore(js): pin exact webpack versions

nijel:
## Proposed changes

There might be difference in output between minor versions and that would cause mess in our CI.

## Checklist


meel-hd (approved):
Looks legitimate

PR #12585, opened 2024-09-24
feat: add openapi spec for weblate rest api

### LINKED ISSUE #12584: Describe the Weblate REST API with OpenAPI
### Describe the problem

There is not a problem to solve per se. The [current API documentation is good](https://docs.weblate.org/en/latest/api.html). In my opinion, the proposed OpenAPI standard might be an improvement upon such documentation.

[OpenAPI](https://swagger.io/specification/) (formerly known as Swagger) is one of the most popular specifications to describe REST APIs formally. It allows developers to define the entire Weblate REST API with a root OpenAPI document.

Some advantages this standard might offer over the current API documentation are:

- Make it easier for developers to code their own third-party REST clients to connect to the Weblate REST API, by using tooling which generates code automatically based on the OpenAPI document.
- Make it easier to document the Weblate REST API:
  - The OpenAPI specification provides a standard way to document REST APIs. It should be easier to document the API when you can use a well-known syntax and keywords (structure), without having the API documentation written in free-form (`.rst` files).
  - Sphinx seems to have plugins to integrate with OpenApi. If they support version 3.1 of the specification, it should be possible to display the OpenAPI document directly within the Weblate documentation.
- It should facilitate automating testing of the Weblate REST API.

_No response_

### Screenshots

This is how the Weblate REST API looks in Swagger UI:

demo-weblate-api-swagger-ui

### Additional context

I already wrote the first draft of a OpenAPI document for the Weblate REST API, as a learning experience of the spec. I'm linking a PR here later.

walpox:
# Changelog

The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).

## Added

  - OpenAPI document describing the Weblate REST API.

---

## Proposed changes

It closes #12584 .

## Checklist

## Other information

I tried to use a tool called `drf-spectacular` to generate the OpenAPI document automatically from Weblate's source code, but it didn't work for me at all, so I decided to do it manually with information gathered from testing (with Weblate v5.7.2 instances) and from what is already written in [the API documentation](https://docs.weblate.org/en/weblate-5.7.2/api.html).

 The OpenAPI document shows fine in my local Swagger UI installation, as shown in the screenshot in the linked issue.

**I don't think the PR is ready yet. The OpenAPI document is missing:**

- Response bodies operations such as POST, PUT and PATCH. In the operations where the body is present, it usually has only the mandatory properties documented, excluding the optional properties you can set for a resource.
- A few `#TODO`s in the documents.
- Some response codes for operations are missing the appropriate response from the server and are using generic response template instead.

I'm happy to discuss further changes to do here.

### DIFF
--- a/docs/openapi/openapi.yaml
+++ b/docs/openapi/openapi.yaml
@@ -1,0 +1,541 @@
+# SPDX-FileCopyrightText: 2024 Javier Pérez <68171111+walpox@users.noreply.github.com>
+# Weblate version in test Docker container: 5.7.2
+
+openapi: 3.1.0
+info:
+  title: Weblate REST API
+  description: |
+    The API is based on [Django REST framework](https://www.django-rest-framework.org).
+    You can interact with the API using the [Weblate Client](https://docs.weblate.org/en/latest/wlc.html)
+    or a third-party REST client of your choice.
+
+    ## Authentication
+
+    Authentication works with tokens placed in the `Authorization` HTTP request header:
+
+    - Each user has a personal access token which can be obtained in their respective user profile. Newly generated user tokens have the `wlu_` prefix.
+    - It is possible to create project-scoped tokens to grant a project access to the API. These tokens can be identified by the `wlp_` prefix.
+
+    Although some of the API operations are available without authentication,
+    it is still recommended to authenticate your requests:
+
+    - Operations such as `GET /api/users/` return an incomplete representation of the
+    requested resources if the request has not been authenticated and authorized.
+    - Unauthenticated requests are heavily rate limited, by default, to 100
+    requests per day. On the other hand, authenticated requests are rate limited
+    to 5000 requests per hour by default.
+
+    ## API rate limiting
+
+    Rate limiting can be adjusted in the `settings.py` file; see [Throttling in Django REST framework documentation](https://www.django-rest-framework.org/api-guide/throttling/)
+    for more details on how to configure it.
+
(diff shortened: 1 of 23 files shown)


walpox:
Note: my local Swagger UI container takes between 2.5 and 7 minutes to load the entire Weblate specification, with around 270 XHR requests. This is probably caused by the high number of `$ref`s in the documents to keep things organized and DRY.

nijel:
Thanks!

The annoying part of API documentation is keeping it in sync with code. It is easy to miss updating the list of fields in the docs when adding one to the serializer. Having a way to validate DRF implementation against docs would be great. Do you think something like that would be possible with OpenAPI when we decide to maintain it manually? 

Maybe `drf-spectacular` would be a better approach in the end, so that all this can be generated (I guess some additional annotation in the code is needed for that)?

Also, have looked at options of integrating this into our existing documentation?

walpox:
I agree that using an automated tool such as `drf_spectacular` is better in the long term for maintainability of the API documentation. Unfortunately, I couldn't get this tool to work in my local environment with Weblate code: it kept throwing exceptions.

I have tested the OpenAPI document in this PR with two Sphinx plugins and uploaded the corresponding screenshots in the linked issue. They are both a worse experience that Swagger UI... Still, if we can get `drf_spectacular` to work, we could use it to generate an OpenAPI document and link it in the regular [API documentation](https://docs.weblate.org/en/latest/api.html). Like this:

download-weblate-openapi

That way, at least the OpenAPI document would be up to date with Weblate code, even if the API document page is not.

walpox:
I'm closing this PR until I have enough Python experience to fix the issue with `drf_spectacular` to generate OpenAPI documents from Weblate source code.

Issue #9372, opened 2023-06-09
Separate color settings for dark theme in apepearance customization


tomkolp:
### Describe the problem

Currently, the color settings are the same for light and dark theme.  There should be separate color settings for dark and light theme.

### Describe the solution you'd like

Ability to change the color of each theme.

mjp-riley:
@nijel I'm assuming this has been solved given your update a few weeks ago? I noticed the current solution is to load the dark css file unless the user has set the theme setting to light, nearly doubling the css load time for the default setting and 2/3 of the total options. The change also does therefore not take place until after settings page reload. Would a better solution not be to get rid of the dark css file and alter the original css file to use custom properties for different theme on the :root? - See [https://codepen.io/mjp-riley/pen/eYLwOPr] / sc: [https://gist.github.com/mjp-riley/23ea3d91152337d5bab932a48a74c452]

Similar implementation of the js on #switcher in the example could be implemented on clicking save in the settings menu so change takes place immediately rather than on page reload. Then on page reload you would use the same logic as you currently use but instead of the settings change deciding whether a stylesheet is included it would just load the html tag with a data-theme attribute for their choice. If they select auto on settings load it without the attribute and the @media prefers dark kicks in if this is their system setting. Alternatively store the setting in storage and negate the need for the if statement in the html file at all, if they switch browsers and the value cant be found then get it from the backend.

Let me know if this is still an open issue that needs contribution or if there's a reason you decided to go the route that you have rather than this approach I'd appreciate some insights as to why so I can continue to learn.

nijel:
@mjp-riley This issue is about supporting appearance customization in the dark theme. Rigth now there is a single customization option which applies to both dark and light themes.

> I noticed the current solution is to load the dark css file unless the user has set the theme setting to light

The dark theme is loaded only if the user agent prefers a dark theme using media query.

Using variables would make things definitely easier, but Bootstrap 3 (which we currently use) is not really ready for this. There is an issue for upgrading, but nobody is working on it: https://github.com/WeblateOrg/weblate/issues/2916

nijel:
@meel-hd this will need a backend changes as well, but I think we could start with using CSS variables for the existing customization and then build the dark theme customization on top of that?

meel-hd:
I don't think we have to change the backend for persisting each user's preferred theme.

By default, the first time a user loads the app we select either light, dark, or based on the system theme. Then later he can change that based on his preference and we store that in local storage.

[...]

PR #12613, opened 2024-09-27
Add separate dark theme color settings

meel-hd:
## Proposed changes

- Modifed the form to have two color fields (one for dark and one for light theme).
- The colors are stored  in this form"#ffffff,#000000". Where the first one is for light theme and the other for dark theme.
- Then the custom.css consumes the values and performs a split(defined in custom_filters.py) on the color string to get the two colors.
- Those colors then are defined in css variables and consumed by the elements that need them.
- The css variables change base on the current theme (for dark values to be applied it is expected that the body tag has `data-theme="dark"`)

Closes: #9372 

## Checklist

#### Dark theme selected + (manually added data-theme="dark" to the body tag, I will add it later in the second commit)
Screenshot 2024-09-27 125432
#### Light theme selected
Screenshot 2024-09-27 125537

### DIFF
--- a/weblate/templates/configuration/custom.css
+++ b/weblate/templates/configuration/custom.css
@@ -1,53 +1,80 @@
+{% load custom_filters %}
+
+{% with header_color_split=header_color|split:"," %}
+{% with header_text_color_split=header_text_color|split:"," %}
+{% with navi_color_split=navi_color|split:"," %}
+{% with nav_text_color_split=navi_text_color|split:"," %}
+{% with focus_color_split=focus_color|split:"," %}
+{% with hover_color_split=hover_color|split:"," %}
+
+body {
+  --header-color: {{ header_color_split.0 }};
+  --header-text-color: {{ header_text_color_split.0 }};
+  --navi-color: {{ navi_color_split.0 }};
+  --navi-text-color: {{ nav_text_color_split.0 }};
+  --focus-color: {{ focus_color_split.0 }};
+  --hover-color: {{ hover_color_split.0 }};
+}
+
+body[data-theme="dark"] {
+  --header-color: {{ header_color_split.1 }};
+  --header-text-color: {{ header_text_color_split.1 }};
+  --navi-color: {{ navi_color_split.1 }};
+  --navi-text-color: {{ nav_text_color_split.1 }};
+  --focus-color: {{ focus_color_split.1 }};
+  --hover-color: {{ hover_color_split.1 }};
+}
+
 {% if header_color %}
   .navbar-inverse {
-    background-color: {{ header_color }};
+    background-color: var(--header-color);
   }
 {% endif %}
 {% if header_text_color %}
 .navbar-inverse .navbar-brand,
 .navbar-inverse .navbar-nav > li > a {
-  color: {{ header_text_color }};
+  color: var(--header-text-color);
 }
 .navbar a path {
-  fill: {{ navi_text_color }};
+  fill: var(--navi-text-color);
 }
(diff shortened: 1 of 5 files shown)


meel-hd:
@nijel  I am still working on this. I should add the `data-theme` attr to the body tag based on the current theme. However, I would love some early feedback

weblate/wladmin/forms.py
```diff
@@ -59,29 +60,52 @@ def __init__(self, **kwargs) -> None:
...
+
+class ThemeColorField(forms.MultiValueField):
+    widget = ThemeColorWidget
+
+    def __init__(self, **kwargs):
+        fields = (forms.CharField(required=False), forms.CharField(required=False))
+        super().__init__(fields=fields, require_all_fields=False, **kwargs)
+
+    def compress(self, data_list):
+        return ",".join(data_list) if data_list else None
```
nijel, line 85: Can't we just store the list in the database field? That would avoid the need for all the split/join logic.

nijel (commented):
What is missing is a migration path. When user currently has customized theme, he will get all black dark theme:

obrazek

Probably the best approach would be to do a database migration on upgrade to fix the affected settings.

weblate/wladmin/forms.py
meel-hd: Tbh I'm not sure what is the best choice in this case. What you see is better we will go with it

nijel: I'd prefer to avoid conversion to string, it's backed by a JSON field in the end, so it should have issues with storing list.

weblate/templates/configuration/custom.css
```diff
@@ -1,53 +1,104 @@
...
+  {% endif %}
+  {% if hover_color_split.1 %}
+    --header-text-color: {{ hover_color_split.1 }};
+  {% endif %}
+}
+
 {% if header_color %}
   .navbar-inverse {
-    background-color: {{ header_color }};
+    background-color: var(--header-color);
```
nijel, line 54: What happens if variable is not set?

meel-hd: Nothing, css ignores the rule. And the previous style will still be applied

meel-hd:
@nijel I got overwhelmed by this previous [test fail](https://github.com/WeblateOrg/weblate/actions/runs/11252714774/job/31286445125?pr=12613) it says it is missing a `header_color` variable in a dozen of tests. I don't know why and what to do. Any guides or thoughts would be helpful. Thanks.

nijel:
I currently don't see what might be causing that.

Can we split this PR into two? One would introduce using variables in the custom CSS without doing any functional change and second would then add the dark theme customization? That way, we limit what might be causing the failures.

meel-hd:
Thanks for the suggestion. I'll go ahead and split the PR into two as you recommended. I'll first focus on the dark theme customization and then follow up with introducing the variables into the custom CSS in a separate PR.

meel-hd:
@nijel  the previous commit [ac1289b](https://github.com/WeblateOrg/weblate/pull/12613/commits/ac1289b5f86e3a8aabf0b0a8e19afd2065ba73ea) above, makes this PR concerned only with the dark theme customization. I'll continue in another PR that depends on this one where the custom CSS variables will be consumed and used.

meel-hd:
@nijel this PR and #12794 are now finished. Also note they depend on each other.

- This first PR does the custom theme setttings
- #12794 consumes those theme settings and applies them.

The issues with the failing tests are resolved now. lol finally 🎉

PR #12621, opened 2024-09-28
Fix docs

ib:


docs/workflows.rst
```diff
@@ -145,9 +145,9 @@ from additional member(s) before it is accepted as a translation.
 +---------------------------------+-------------+------------------------------------+
 | Enable suggestions              | on          |                                    |
 +---------------------------------+-------------+------------------------------------+
-| Suggestion voting               | off         |                                    |
+| Suggestion voting               | on          |                                    |
```
nijel, line 148: Good catch, this was broken since f5080b4d329.

nijel:

Issue #12601, opened 2024-09-26
Add pagination also to the top of the Changes page


ib:
### Describe the problem

When searching for something on the Changes page, it is inconvenient to always scroll to the bottom of the page to move forward or backward.

### Describe the solution you would like

A second pagination at the top of the page would be helpful.

nijel:
With the current implementation, there can be only one paginator on page, see f8160c48216bbc15f9d8ded95640dc23ce09392c and #6888. @meel-hd can you please look into that? Adding second one to the page is just matter of additional include in the page, but in the current implementation the editing of page does not work on second one:

```diff
diff --git weblate/templates/trans/change_list.html weblate/templates/trans/change_list.html
index 88147fed97..2fc41c67ce 100644
--- weblate/templates/trans/change_list.html
+++ weblate/templates/trans/change_list.html
@@ -35,6 +35,8 @@
 
 {% block content %}
 {% if form.is_valid %}
+  {% include "paginator.html" %}
+
   {% include "last-changes-content.html" with last_changes=object_list %}
 
   {% include "paginator.html" %}
```

Het4304:
Hi i am interested in solving the issue can you assign it to me?? @nijel @ib

nijel:
@Het4304 This issue is already assigned to @meel-hd

meel-hd:
@nijel  I'm on it

Issue #6888, opened 2021-11-30
Bottom pagination component in changes page is unclickable to “summon” a text field


Geeyun-JY3:
### Describe the issue

Bottom pagination component in changes page is unclickable to “summon” a text field where users can input number to jump to a page.

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar issues in this repository.

### Steps to reproduce the behavior

1. Go to any Changes page
2. Scroll down to bottom pagination component
3. Click on 'x/y' part of the component
4. No response

### Expected behavior

Click then the text field appears inside the pagination component

### Screenshots

![weblate_pagination](https://user-images.githubusercontent.com/49396421/143915374-486acc55-0f10-4ade-a46a-0421a61d4a0e.gif)

### How do you run Weblate?

weblate.org service

PR #12648, opened 2024-10-01
Add second paginator to the top of changes page

meel-hd:
## Proposed changes
- Add second paginator to the top of changes page
- Fixed bug of inactive input when a second paginator is added.

Closes: #12601 

## Checklist

### Breakdown
@nijel the main issue was the use of `id` selectors with more than one element (which is incorrect). Plus, ofc the logic for closing, hiding and focusing the input field in the paginator wasn't adapted for two paginators.
### Preview
There are two paginators (one at the top, and one at the bottom) + the bug discovered in #6888 is fixed.
image

### DIFF
--- a/weblate/static/loader-bootstrap.js
+++ b/weblate/static/loader-bootstrap.js
@@ -1009,32 +1009,41 @@ $(function () {
   });
 
   /* Click to edit position inline. Disable when clicked outside or pressed ESC */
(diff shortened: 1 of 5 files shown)


nijel:

Geeyun-JY3:
The paginator does not look good on the full translation editor page:
image
And the changes page:
image
image

PR #12661, opened 2024-10-02
fix: use only supported languages when validating Apertium settings. Fixes: #12579

### LINKED ISSUE #12579: Unable to add Apertium APy authomatic suggestions because "en>de" pair is missing
### Describe the issue

I use a private Apertium APy instance, without en>de lang pair. If I try to add Apertium APy authomatic suggestions to Weblate I get an error because my Apertium server answers a 400 error (because en>de pair isn't available).

You can reproduce the same effect trying to add Apertium APY authomatic suggestions with official Apertium APy URL: https://apertium.org/apy/, because it lacks en>de pair too.

`Could not fetch translation:: 400 Client Error: Bad Request for url: https://apertium.org/apy/translate?langpair=en%7Cde&q=test&markUnknown=no`

So, Apertium APY is unusable. I guess the issue is in [this line](https://github.com/WeblateOrg/weblate/blob/main/weblate/machinery/base.py#L150), because settings validation has a translation en>de test hardcoded.

### I already tried

1. Try to add an Apertium authomatic suggestions using an APy URL without en>de language pair support
2. Click on Save settings
3. An error occours

### Expected behavior

I expected a test using an avaliable language pair at APy server were used.
Alternatively, provide a field to indicate a custom language pair for testing and validate settings.

### Screenshots

_No response_

### Exception traceback

_No response_

### How do you run Weblate?

Docker container

### Weblate versions

_No response_

### Weblate deploy checks

_No response_

### Additional context

_No response_

SukiCZ:
## Checklist

## Other information

Hi,
This seems like a good project for a [ hacktoberfest](https://github.com/WeblateOrg/weblate/labels/hacktoberfest) may I contribute?
I've tried to fix an issue #12579 with an option to overwrite the validation arguments.

Any feedback is welcome :smiley: 

P.S. I really love the documentation of your project - and specially the custom `checks` messages. You really know the Django framework, that's awesome! :rocket: :muscle:

### DIFF
--- a/weblate/machinery/apertium.py
+++ b/weblate/machinery/apertium.py
@@ -4,6 +4,9 @@
 
 from functools import reduce
 
+from django.core.exceptions import ValidationError
+from django.utils.translation import gettext
+
 from .base import (
     DownloadTranslations,
     ResponseStatusMachineTranslation,
@@ -90,6 +93,28 @@ class ApertiumAPYTranslation(ResponseStatusMachineTranslation):
         if code not in self.all_langs and code in LANGUAGE_MAP:
             return LANGUAGE_MAP[code]
         return code
+
+    def validate_settings(self) -> None:
+        try:
+            languages = self.download_languages()
+        except Exception as error:
+            raise ValidationError(
+                gettext("Could not fetch supported languages: %s") % error
+            ) from error
+        try:
+            source, language = languages[0]
+        except IndexError as error:
+            raise ValidationError(
+                gettext("No supported languages found: %s") % error
+            ) from error
+        try:
+            self.download_multiple_translations(
+                source, language, [("test", None)], None, 75
+            )
+        except Exception as error:
+            raise ValidationError(
+                gettext("Could not fetch translation: %s") % error
+            ) from error
 
     def download_languages(self):
         """Download list of supported languages from a service."""
--- a/weblate/machinery/tests.py
+++ b/weblate/machinery/tests.py
@@ -544,6 +544,16 @@ class ApertiumAPYTranslationTest(BaseMachineTranslationTest):
         )
 
     @responses.activate
+    def test_validate_settings(self):
+        self.mock_response()
+        machine = self.get_machine()
+        machine.validate_settings()
+        self.assertEqual(len(responses.calls), 2)
+        _, call_2 = responses.calls
+        self.assertIn("langpair", call_2.request.params)
+        self.assertEqual("eng|spa", call_2.request.params["langpair"])
+
+    @responses.activate
     def test_translations_cache(self) -> None:
         self.mock_response()
         machine = self.MACHINE_CLS(self.CONFIGURATION)


nijel:
The problem described in #12579 is caused by the option to have any set of language pairs installed in the Apertium server. This makes hard-coding the language pair impractical. Instead, it should be dynamically generated based on the supported languages.

The validation anyway downloads a list of supported languages, so it could use that to select a good language pair. This needs to be backend-specific because `download_languages()` has different semantics per backend. So I'd stick with hard-coding the pair for all services and doing dynamic selection in Apertium subclass only.

SukiCZ:
Thanks for your swifty feedback. I've extended the Apertium to use supported languages when validating settings as you suggested :blush:

weblate/machinery/apertium.py
```diff
@@ -91,6 +94,27 @@ def map_language_code(self, code):
             return LANGUAGE_MAP[code]
         return code
 
+    def validate_settings(self) -> None:
+        try:
+            languages = self.download_languages()
```
nijel, line 99: This can raise any error, these should be handled similarly as in parent class (and outside the IndexError block).

SukiCZ: Yes. That part is on line 105 (`except Exception as error:`) same as parent class but
```py
# languages = []
source, language = languages[0]
```
will raise `IndexError`, if there are no supported languages at all.

nijel: In parent class all  exceptions from dowload_languages are catched and reported to user, here not. It should be like:

```
try:
   languages = download_languages()
except Exception:
    raise ....
try: 
    source, language = languages[0]
except IndexError:
    raise ...
```

SukiCZ: Right. That is more explicit. Sorry for confusion.

weblate/machinery/tests.py
```diff
@@ -546,8 +546,15 @@ def mock_response(self) -> None:
...
-        machine.delete_cache()
+        machine = self.get_machine(cache=True)
+
+        machine.validate_settings()
+        self.assertEqual(len(responses.calls), 2)
+        self.assertIn("langpair", responses.calls[1].request.params)
+        self.assertEqual("eng|spa", responses.calls[1].request.params["langpair"])
+        responses.reset()
+
+        self.mock_response()
```
nijel, line 557: This really should go into a separate test case and not mixed witch cache testing.

SukiCZ:
Rebased with main to have mypy action passing

SukiCZ:
Not sure what's wrong here and why the mypy check passes on [my fork](https://github.com/SukiCZ/weblate/actions/runs/11188811712/job/31108230593). I thought it was fixed in 9cd5a23 which I've included yesterday while rebasing.

Any help is welcome :pray:

nijel:
Merged, thanks for your contribution! The mypy error should be hopefully fixed by e0f680f5dab9f072c50b253dcad4a8bae0dd8813.

PR #12673, opened 2024-10-03
Cache key collision in render checks for suggestions

vinceh121:
## Proposed changes

There is a cache key collision between renders generated in checks for the unit and for its suggestions.

How to reproduce the collision:
  1. Create a component with the max-size check
  2. Create a unit in the source language and translate it in a target language
  3. The render check (whether valid or incorrect) will display the render for the target string
  4. Add a different suggestion to the same unit
  5. The render check will display the render for the suggestion instead of the target string

## Checklist

### DIFF
--- a/weblate/checks/base.py
+++ b/weblate/checks/base.py
@@ -165,12 +165,15 @@ class BaseCheck:
     def render(self, request: AuthenticatedHttpRequest, unit: Unit) -> StrOrPromise:
         raise Http404("Not supported")
 
-    def get_cache_key(self, unit: Unit, pos: int) -> str:
-        return "check:{}:{}:{}:{}".format(
+    def get_cache_key(
+        self, unit: Unit, pos: int, fake_discriminator: int | None = None
+    ) -> str:
+        return "check:{}:{}:{}:{}:{}".format(
             self.check_id,
             unit.pk,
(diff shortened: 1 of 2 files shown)


weblate/checks/base.py
```diff
@@ -165,12 +165,15 @@ def get_fixup(self, unit: Unit) -> Iterable[tuple[str, str, str]] | None:
     def render(self, request: AuthenticatedHttpRequest, unit: Unit) -> StrOrPromise:
         raise Http404("Not supported")
 
-    def get_cache_key(self, unit: Unit, pos: int) -> str:
-        return "check:{}:{}:{}:{}".format(
+    def get_cache_key(
+        self, unit: Unit, pos: int, fake_discriminator: int | None = None
+    ) -> str:
```
nijel, line 170: The new argument seems unused.

weblate/trans/models/suggestion.py
```diff
@@ -224,6 +224,7 @@ def get_checks(self):
         fake_unit = copy(self.unit)
         fake_unit.target = self.target
         fake_unit.state = STATE_TRANSLATED
+        fake_unit.suggestion_id = self.pk
```
nijel, line 227: This would be a simpler solution and it would not need ad-hoc attribute:

```suggestion
        fake_unit.id = -self.pk
```

Issue #3703, opened 2020-04-09
Introduce type checking to codebase


nijel:
Weblate now requires Python 3.6, so it's now possible to use type hints and introduce type checking to our CI.

My experience with this is exactly zero and our code base is huge, so getting some experienced help with this is welcome.

nijel:
This is now a work in progress and is expected to take some time.

* There is `mypy` extra dependency to install type checking depedencies
* We now have CI job running mypy (ignoring the failures for now)
* Parts of the code base have been annotated
* There are still 400+ errors reported by mypy, most of these come from lack of annotations in Weblate or third-party (things like https://github.com/typeddjango/django-stubs/issues/2224 or https://github.com/typeddjango/django-stubs/issues/1514).

Any contribution to this is welcome, just pick one file and enhance it!

PR #12690, opened 2024-10-05
mypy: fix mypy errors in wladmin module. Rel: #3703

SukiCZ:
## Proposed changes

Fixes (7) mypy errors in wladmin module.

## Checklist

## Other information

list of fixed errors
```
weblate/wladmin/models.py:333:9: error: Name "support_status" already defined on line 331  [no-redef]
weblate/wladmin/sites.py:41:28: error: Argument 1 of "each_context" is incompatible with supertype "AdminSite"; supertype defines the argument type as "HttpRequest"  [override]
weblate/wladmin/sites.py:41:28: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
weblate/wladmin/sites.py:41:28: note: This violates the Liskov substitution principle
weblate/wladmin/views.py:326:18: error: Argument "key_type" to "get_key_data_raw" has incompatible type "str"; expected "Literal['rsa', 'ed25519']"  [arg-type]
weblate/wladmin/views.py:348:44: error: Argument "key_type" to "generate_ssh_key" has incompatible type "str"; expected "Literal['rsa', 'ed25519']"  [arg-type]
weblate/wladmin/views.py:432:21: error: Incompatible types in assignment (expression has type "str | list[object]", target has type "str | bytes")  [assignment]
weblate/wladmin/views.py:438:52: error: "Field" has no attribute "parser"  [attr-defined]
weblate/wladmin/views.py:506:16: error: "BillingManager" has no attribute "prefetch"  [attr-defined]
```

### DIFF
--- a/weblate/wladmin/sites.py
+++ b/weblate/wladmin/sites.py
@@ -38,7 +38,7 @@ class WeblateAdminSite(AdminSite):
             return settings.URL_PREFIX
         return "/"
 
-    def each_context(self, request: AuthenticatedHttpRequest):
+    def each_context(self, request: AuthenticatedHttpRequest):  # type: ignore[override]
         from weblate.wladmin.models import ConfigurationError
 
         result = super().each_context(request)
--- a/weblate/wladmin/models.py
+++ b/weblate/wladmin/models.py
@@ -328,12 +328,12 @@ class SupportStatusDict(TypedDict):
 
 def get_support_status(request: AuthenticatedHttpRequest) -> SupportStatusDict:
     if hasattr(request, "weblate_support_status"):
-        support_status = request.weblate_support_status
+        support_status: SupportStatusDict = request.weblate_support_status
     else:
         support_status = cache.get(SUPPORT_STATUS_CACHE_KEY)
         if support_status is None:
             support_status_instance = SupportStatus.objects.get_current()
-            support_status: SupportStatusDict = {
+            support_status = {
                 "has_support": support_status_instance.name != "community",
                 "in_limits": support_status_instance.in_limits,
             }
--- a/weblate/wladmin/views.py
+++ b/weblate/wladmin/views.py
@@ -50,6 +50,7 @@ from weblate.utils.tasks import database_backup, settings_backup
 from weblate.utils.version import GIT_LINK, GIT_REVISION
 from weblate.utils.views import show_form_errors
 from weblate.vcs.ssh import (
+    KeyType,
     add_host_key,
     can_generate_key,
     generate_ssh_key,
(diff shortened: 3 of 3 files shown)


weblate/billing/models.py
```diff
@@ -193,7 +193,7 @@ class Billing(models.Model):
     # with payment processor
     payment = models.JSONField(editable=False, default=dict, encoder=DjangoJSONEncoder)
 
-    objects = BillingManager.from_queryset(BillingQuerySet)()
+    objects: BillingQuerySet = BillingManager.from_queryset(BillingQuerySet)()
```
SukiCZ, line 196: Oh, I'd overlooked this one. I've tried to investigate a bit but couldn't find a working solution, just this issue ticket typeddjango/django-stubs#1067 or `#type: ignore[assignment]` or using `cast` (which seems too brutal)

Should I remove this change?

nijel: Probably yes. I think this is best to be addressed at django-stubs or mypy, otherwise we end up doing that on every model.

weblate/wladmin/sites.py
```diff
@@ -12,7 +12,7 @@
 from django.utils.translation import gettext, gettext_lazy
 
 if TYPE_CHECKING:
-    from weblate.auth.models import AuthenticatedHttpRequest
+    from django.http import HttpRequest
```
nijel, line 15: The intention here is that Django's `HttpRequest` can have `AnonymousUser` what can't happen in Weblate. That's why `AuthenticatedHttpRequest` was introduced. Without that each access to `reuqest.user` has complains from mypy.

weblate/billing/models.py
nijel: Furthermore, this does address only accessing query set via `.objects` but not when using related manager.

weblate/wladmin/sites.py
nijel: Actually, this is an upstream recommended approach: https://github.com/typeddjango/django-stubs?tab=readme-ov-file#how-can-i-create-a-httprequest-thats-guaranteed-to-have-an-authenticated-user

SukiCZ: Yes, I've seen that answer in FAQ. The reason of this change is because of a method `each_context`, that has a signature with `HttpRequest`. The mypy error is:

error: Argument 1 of "each_context" is incompatible with supertype "AdminSite"; supertype defines the argument type as "HttpRequest"  [override]
note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
note: This violates the Liskov substitution principle

For now I have two solutions on my mind, `cast` and `ignore`.

* `cast("AuthenticatedHttpRequest", request)` doesn't make much sense, because the function doesn't actually need a `request.user`
* `# type: ignore[override]` doesn't seem that bad actually

weblate/billing/models.py
SukiCZ: That's true. Thanks :smirk:

weblate/wladmin/sites.py
nijel: I was using ignore for override so far, IMHO it is the best approach

nijel:

Issue #12694, opened 2024-10-06
List reviewers on the Information tab


tjhietala:
### Describe the problem

As a translator, there does not seem to be a way to see which users (if any) currently hold the reviewer role for a particular language in a project/component. I don't think this information should be secret.

### Describe the solution you would like

A good place for the list of reviewers would be the Information tab, e.g. https://hosted.weblate.org/projects/weblate/application/fi/#information

### Additional context

Hosted Weblate

nijel:
There is `User.objects.having_perm` to list such users, adding these to the page should be easy.

busch44:
Hello, I'm a beginner and I'd appreciate to have a shot at this issue as my first.

nijel:
You're welcome to give it a try. You can look how `all_admins` works on the `User` model, something similar can be done for reviewers.

PR #12781, opened 2024-10-15
feat: adds list of reviewers on information tab

flamiedev:
## Proposed changes

Issue #12694 asks about showing the information of the current reviewers in the project. To solve this problem a new item is added on the information tab, showing this information with the `having_perm` method

## Checklist

### DIFF
--- a/weblate/templates/snippets/info.html
+++ b/weblate/templates/snippets/info.html
@@ -44,6 +44,17 @@
         {% endif %}
         {% endif %}
 
+        {% if project.having_perm %}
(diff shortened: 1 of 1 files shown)


weblate/templates/snippets/info.html
```diff
@@ -44,6 +44,17 @@
         {% endif %}
         {% endif %}
 
+        {% if project.having_perm %}
```
nijel, line 47: This method is available on users and needs a parameter, this can't work.

PR #12806, opened 2024-10-17
Strip probited initial character in Glossary CSV upload 

### LINKED ISSUE #12792: Add glossary check for source stings problematic in CSV
### Describe the problem

Some strings can't be exposed in CSV to automatic translation backends (see https://github.com/WeblateOrg/weblate/issues/12791). Weblate should warn that these are problematic.

Add glossary only check to warn about using any of `=+@-` at the beginning of the string for both source and translation.

https://github.com/WeblateOrg/weblate/issues/12791 will deal with such strings, but these will be missing from the sync (or problematic characters stripped), so user should be notified of that.

### Screenshots

_No response_

### Additional context

Currently, no checks are triggered on glossary components:

https://github.com/WeblateOrg/weblate/blob/64aa009bb957309a6938724b2e493ff69354c49c/weblate/trans/models/unit.py#L1353-L1355

gersona:
## Proposed changes
Problematic initial characters are stripped from glossary terms before upload . Prohibited characters are:
```{"=", "+", "-", "@", "|", "%"}```

I  chose this option over excluding the terms altogether

## Checklist

## Other information

fixes #12792

### CODE REVIEW

File: weblate/machinery/base.py [hunk 1]
```diff
@@ -737,6 +738,8 @@ def get_glossary_id(
         if not glossary_tsv:
             return None
 
+        glossary_tsv = self.cleanup_glossary_tsv(glossary_tsv)
```
nijel, line 741: What was your motivation for doing the cleanup as a separate step instead of doing that while generating it?

gersona: Well, that was just to have the logic isolated to make it easier to test, but I can definitely just include that in the `get_glossary_tsv` method.

nijel:

PR #12812, opened 2024-10-18
fix: update login section - Adjusted sizes to fix text overflow

Leetfs:
## Proposed changes

update login section - Adjusted sizes to fix text overflow

## Checklist

## Other information

old:

image

new:

image


nijel (changes requested):
This is changing a vendored library what is not really desired and will likely break some other UI parts.

If we're going to change this, using `col-sm-6` for both might be a better approach.

Leetfs:
@nijel okay~ I changed the .col-sm-6 width from 50% to auto to resolve the text overflow issue.

Leetfs:
I think I understand now, I will open a new PR to resolve this issue.

PR #12872, opened 2024-10-27
12700 dummy fields

JezFirth:
## Proposed changes

Modified `__setattr__` in `weblate.auth.models.py` to take the first name and last name attributes and pass them as full_name in extra data. 

## Checklist

## Other information

I struggled to replicate this issue through weblate. It seems that for `User.objects.get_or_create` with the `dummy fields` in the `defaults` parameter via the django shell, it would require `first_name` and `last_name` to be properties in the User class as it seems this function bypasses the weblate framework. This seems redundant however as their already exists a 'full_name` property.


nijel (commented):
This still doesn't handle the create() code path. Also when doing functional changes, please cover that with a test case.

Issue #7563, opened 2022-04-29
Selection of individual strings for bulk edit


keunes:
### Describe the problem

The bulk edit feature allows to perform a search and apply an action to all strings returned from that search. However, the search seems to allow to filter only based on one characteristic and there is no way to for example apply labels to a fine-grained selection of strings.
Not sure if custom search would work, but if it would the search would have to be quite long/detailed anyway to catch the relevant ones.

### Describe the solution you'd like

In addition to 'search & apply', it would be helpful to have a way to select individual strings to which the bulk action would be applied. A simple checkbox for each string would allow to make this selection. I hate to make the reference because the tool really isn't that good, but it could look/work similar as in Transifex:
![image](https://user-images.githubusercontent.com/11229646/165929607-6b498336-d8d5-48b0-8300-5e0d02bbb12f.png)

### Describe alternatives you've considered

* The live preview as described in #4106 would already help a bit but still would not allow for a fine-grained & visual selection of strings. Maybe this request (selection of individual strings) could be an enhancement on top of [#4106](https://github.com/WeblateOrg/weblate/issues/4106).
* _Combining_ this manual selection with a filter would be nice to help narrow down the list of strings one would have to look at, so that's a nice addition. But I also understand that that could potentially make it more tricky, 'cause what would happen if you select a string via the checkbox and then change the search that makes the selected string invisible...

keunes:
Also related (particularly filet/search): #4108

Issue #4108, opened 2020-06-29
Bulk operations and search and replace


nijel:
**Is your feature request related to a problem? Please describe.**
Right now, "Search and replace" and "Bulk edit" are distinct functions, while they serve almost same purpose, what can be confusing for users.

**Describe the solution you'd like**

* Extend "Bulk edit" to support text replacements as well.
* This would cover https://github.com/WeblateOrg/weblate/issues/4022 as all search capabilities would be available there
* The question is what to do with the preview. Is https://github.com/WeblateOrg/weblate/issues/4106 good enough replacement?

**Describe alternatives you've considered**
Alternative approach might be to add search field to the "Search and replace".

GG-Sheridan:
Should this issue be labeled as Good First Issue?

We are doing research about how to help newcomers onboarding in an open source project and how to accelerate the issue resolving process.
We have developed a tool to automatically label the Good First Issue (GFI), i.e., the issue which is good for newcomers. 
We first extract 72 features from different dimensions of issue reports, such as readability, semantic meaning, associated code snippets, etc, and experimentally investigate to what extent these features correlate with GFI. Results turn out that a large portion of these features do have a medium to large correlation with GFI. 
We then build a machine learning model based on these features to predict GFI. The experiments on a dataset of 12k issues from Github show that 0.90 AUC can be achieved. 

Our tool shows that this issue has a very large possibility to be a GFI. 
Could you please tell us whether it should be labeled as a GFI?
And if you have any comments or suggestions with this tool, could you please tell us? 
Thank you so much.

nijel:
This involves quite many changes in the codebase, so I don't think it's good issue to start with...

comradekingu:
I would have use for bulk editing to apply Priority:0-100 flags.
Searching by "string placement" and/or some way to add string-number ranges would really prevent doing things one string at a time.

For example GnuCash has a whole lot of strings for "2.6 tour" which are offered first for translation of the website https://hosted.weblate.org/translate/gnucash/website/nb_NO/?checksum=7a8c28997bb3c6de

I haven't actually used the bulk search, and I don't know if clicking "use" just does whatever it is told without any "are you sure". I'd like for bulk editing to be in the regular search for admins.

PR #12957, opened 2024-11-07
feat(tools): add search preview to bulk edit tool

michael-smt:
## Proposed changes

Adds the same search preview from _Search and replace_ to the _Bulk edit_ tool.

## Checklist

## Other information

Related to #4106, #4108 and #7563.

Does not affect the _Bulk edit_ add-on.

### DIFF
--- a/weblate/trans/forms.py
+++ b/weblate/trans/forms.py
@@ -2626,7 +2626,8 @@ class BulkEditForm(forms.Form):
     ) -> None:
         project = kwargs.pop("project", None)
         kwargs["auto_id"] = "id_bulk_%s"
-        kwargs["initial"] = {"path": getattr(obj, "full_slug", None)}
+        if obj is not None:
+            kwargs["initial"] = {"path": obj.full_slug}
         super().__init__(*args, **kwargs)
         labels = Label.objects.all() if project is None else project.label_set.all()
         if labels:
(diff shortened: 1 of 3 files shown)


weblate/trans/forms.py
```diff
@@ -2623,6 +2624,7 @@ class BulkEditForm(forms.Form):
     def __init__(self, user: User | None, obj, *args, **kwargs) -> None:
         project = kwargs.pop("project", None)
         kwargs["auto_id"] = "id_bulk_%s"
+        kwargs["initial"] = {"path": obj.full_slug}
```
nijel, line 2627: `obj` can be `None` here (for site-wide) and it then fails: `AttributeError: 'NoneType' object has no attribute 'full_slug'`

nijel: Passing blank path and not using it in lookup should do the trick in that case.

weblate/trans/forms.py
```diff
@@ -2625,6 +2626,7 @@ def __init__(
     ) -> None:
         project = kwargs.pop("project", None)
         kwargs["auto_id"] = "id_bulk_%s"
+        kwargs["initial"] = {"path": getattr(obj, "full_slug", None)}
```
nijel, line 2629: I'd rather do the condition this way:

```suggestion
        if obj is not None:
            kwargs["initial"] = {"path": obj.full_slug}
```

Issue #3507, opened 2020-02-21
Clicking placeholder in Zen scrolls view to top of string


comradekingu:
If there are lots of secondary languages and the string is sufficiently long, and the resolution/scaling isn't there to cover it all, editing field focus is lost.

![bilde](https://user-images.githubusercontent.com/13802408/75032204-46f63180-54a8-11ea-9705-bef590c7b1cf.png)

https://hosted.weblate.org/translate/remmina/remmina/sv/?checksum=a40059ac9985ca75

Write up a whole string, click </big>, view skips, click </tt> view skips again.

comradekingu:
@nijel 
This loss of editing field also happens (on 4.15) when just clicking the editing field, or clicking tab from the previous string.

The expected behaviour is to have the editing field visible at the bottom of the screen, and then lose out on the topmost alternative translations. (Instead of starting at the top alternative translation and then running out of screen-estate somewhere before the editing field is shown.)

meel-hd:
Given #12981 has been merged, this issue is fixed now.

PR #12981, opened 2024-11-09
fix(editor): Add sticky to bottom editor when out of view

meel-hd:
## Proposed changes
The logic used to the calculate scroll offset based on the closest tbody tag which is fine unless the space between them is very large (like in issue #3507) which hides the translation input. As a fix we calculate based on a closer element that can't grow in height very much to hide the translation input.

## Checklist


nijel:
The main intention here is to show the source string while the user navigates using the tab key. Your change destroys this (and mostly reverts to browser default behavior).

The whole thing is problematic because the content we want to show doesn't have to fit on the screen.

Ideas which could make it behave nicer:

* Move the viewport in a way that the editor is at the bottom, showing as much content above as possible.
* Do not adjust the viewport when the current editor was already focused.

meel-hd:
Or the editor should stick in the bottom of the screen when it goes out of screen. It will insure the user can scroll freely to view anything and be able to edit anywhere
**Example**

https://github.com/user-attachments/assets/56695901-cfb4-4e50-a53a-3594899b781a

nijel:
Great idea, it definitely makes sense as long as the editor is focused.

weblate/templates/zen-units.html
```diff
@@ -88,7 +88,10 @@
 </td>
 {% endif %}
 <td {% if user.profile.zen_mode != user.profile.ZEN_HORIZONTAL %}colspan="2"{% endif %} class="translator">
-{% crispy item.form %}
+  <button id="hide-sticky" class="btn btn-light btn-sm" title="{% trans "Collapse" %}" >
```
nijel, line 91: This form is there for each string, so using `id` is probably not a good idea as there will be more of them in the page.

meel-hd: Yeah that's not a problem in this case, we select only the closest hide btn to the string editor we are in. So others are ignored

nijel: Isn't `id` supposed to be unique in the document?

meel-hd: Yes that's what is said. But to be honest I'm not sure if it is that important or enforced by any means.

PR #12993, opened 2024-11-12
feat: Additional format support for translation memory import

### LINKED ISSUE #2280: Possibility to import the translation memory in formats different from TMX and JSON
I have some translation files in .po format that I don't know how to convert in tmx or json (every time I tried it seems like it didn't work, for some reason I don't know) and I would like to import them in the translation memory.

It would be great to have the possibility to import files into the translation memory in formats different from TMX and JSON, and also to be able to export in other formats, like Xliff and xml, like it's possible to do for single translations.

gersona:
## Proposed changes
Extend memory translation upload support to formats other than JSON and TMX with automatic format detection and content parsing. Other files usually require that `source_language` and `target_language` are passed as arguments

## Checklist

## Other information

- fixes #2280


weblate/memory/forms.py
```diff
@@ -2,19 +2,58 @@
...
+
 
 class UploadForm(forms.Form):
     """Uploading file to a translation memory."""
 
     file = forms.FileField(
         label=gettext_lazy("File"),
-        validators=[FileExtensionValidator(allowed_extensions=["json", "tmx"])],
+        validators=[
+            FileExtensionValidator(allowed_extensions=["json", "tmx", "xliff"])
```
nijel, line 21: Maybe allow more formats here?

weblate/memory/forms.py
```diff
@@ -2,19 +2,58 @@
...
+        help_text=gettext_lazy(
+            "Target language of the document when not specified in file"
+        ),
+        queryset=Language.objects.all(),
+        required=False,
+    )
+
+    def clean(self):
+        data = self.cleaned_data
+        extension = Path(data["file"].name).suffix[1:].lower()
```
nijel, line 46: I think the file can be in memory, so using `PurePath` sounds more reasonable here:

```suggestion
        extension = PurePath(data["file"].name).suffix[1:].lower()
```

weblate/memory/models.py
```diff
@@ -280,6 +294,70 @@ def import_tmx(
...
+        except Exception as error:
+            report_error("Could not parse memory")
+            raise MemoryImportError(gettext("Unsupported file!")) from error
+
+        if storage.monolingual is True:
+            raise MemoryImportError(
+                gettext("Monolingual format not supported for memory upload")
+            )
+
+        def get_language(language: Language | str | None) -> Language:
```
nijel, line 328: In method signature this is `str | None`, but here `Language | str | None`. Shouldn't these match?

weblate/memory/models.py
```diff
@@ -280,6 +294,70 @@ def import_tmx(
...
+            except Language.DoesNotExist as error:
+                raise MemoryImportError(
+                    gettext("Could not find language %s!") % language
+                ) from error
+
+        source_language = get_language(storage.source_language or source_language)
+        target_language = get_language(storage.language_code or target_language)
+
+        count = 0
+        for unit in storage.all_units:
```
nijel, line 348: Filtering for translated units only should happen here as well. Maybe using `iterate_merge` would be a better method (it is used for translations import).

weblate/memory/forms.py
gersona: Is it fair to assume all bilingual formats listed in the [Supported file formats](https://docs.weblate.org/en/latest/formats.html) can be included, since some also support monolingual content ?
i.e `["json", "tmx", "xliff", "po", "ts", "properties", "strings", "csv"]`

nijel: I'd definitely include at least `po` and skip all monolinguals, let's start with this:

```suggestion
            FileExtensionValidator(allowed_extensions=["json", "tmx", "xliff", "po", "csv"])
```

PR #12998, opened 2024-11-12
feat: add altcha integration

### LINKED ISSUE #1462: Add support for additional captcha implementation
Current captcha is quite simple and prevents from random bots filling in registration page, but does not provide defense against targeted attack.

Therefore it might be good idea to have option to choose captcha implementation and to provide stronger protection.

Probably we could allow plugging [existing captcha implementations for Django](https://djangopackages.org/grids/g/captcha/)

nijel:
## Proposed changes

This adds altcha in addition to existing match captcha.

Fixes #1462

## Checklist


nijel:
Some issue discovered in the Python implementation:

* https://github.com/altcha-org/altcha-lib-py/pull/3
* https://github.com/altcha-org/altcha-lib-py/pull/4
* https://github.com/altcha-org/altcha-lib-py/pull/5

weblate/static/styles/main.css
```diff
@@ -6,6 +6,9 @@
 
 :root {
   color-scheme: only light;
+  --altcha-color-text: #2a3744;
+  --altcha-border-radius: 4px;
```
nijel, line 10: Like this?

```suggestion
  --altcha-border-radius: var(--border-radius);
```

meel-hd: yes

PR #13039, opened 2024-11-15
fix: has:variant search filter fix

### LINKED ISSUE #12825: `has:variant` should always search for string(s) that have variant(s) in the language in which the user is searching
### Describe the issue

Currently, `has:variant` would search for string(s) that have variant(s) in any one of languages, rather than in the language in which the user is searching.

And there is a redundant `Things to check` panel with nothing but a title in the full editor while the current string has no variant in the current language but has variants in other languages.

### I already tried

1. Go to the search page of a language
2. Enter `has:variant`
3. Find a string that has variants that are not in the current language

### Expected behavior

`has:variant` should always search for string(s) that have variant(s) in the language in which the user is searching, rather than in any one of languages.

### Screenshots

Example: 
a string that has no variant in Chinese (Simplified Han script): 
image
(And there is a redundant `Things to check` panel with nothing but a title.)
but has variants in English:
image
and Estonian:
image

### Exception traceback

_No response_

### How do you run Weblate?

weblate.org service

### Weblate versions

5.8.2-dev

### Weblate deploy checks

_No response_

### Additional context

_No response_

gersona:
## Proposed changes
- when using `has:variant` in a string search, only string(s) that have variant(s) in the translation language of the search are displayed
- in the Translate view, the "Things to check" box is correctly hidden when there is no variant and nothing else to check:
 
Capture d’écran 2024-11-15 à 15 16 20
Capture d’écran 2024-11-15 à 15 20 59

## Checklist

## Other information

- fixes #12825


weblate/utils/search.py
```diff
@@ -470,7 +470,7 @@ def has_field(self, text: str, context: dict) -> Q:  # noqa: C901
         if text == "translation":
             return Q(state__gte=STATE_TRANSLATED)
         if text in {"variant", "shaping"}:
-            return Q(variant__isnull=False)
+            return Q(defined_variants__isnull=False)
```
nijel, line 473: AFAIR this covers only manually created variants, not the regex based on the key. Can you please add tests for both cases to check that?

nijel:

PR #13056, opened 2024-11-18
feat: Machinery configuration via API

### LINKED ISSUE #11926: Per-project automatic suggestion configuration via API
### Describe the problem

Per-project automatic suggestion configuration is stored in a *Project* model:

https://github.com/WeblateOrg/weblate/blob/32462c2b5e18ba86b46253748b28e59ec8df509e/weblate/trans/models/project.py#L219

There is currently no way to configure this via the API.

Add `machinery_settings` field to the project API including validation of the settings.

_No response_

### Screenshots

_No response_

### Additional context

The per-service validation is already implemented in https://github.com/WeblateOrg/weblate/blob/main/weblate/machinery/management/commands/install_machinery.py and the implementation should be shared, so it needs to be factored out from there (perhaps to the *Project* model) and used for both command line and API.

gersona:
## Proposed changes
- new endpoint has been added for Machinery service installation `/api/projects/{string:project}/machinery_settings/`
- validation has been moved out to `weblate.machinery.models` to be used by both API based and command based configuration

## Checklist

## Other information

- fixes #11926


weblate/api/views.py
```diff
@@ -1010,8 +1012,43 @@
...
+                return Response(
+                    {"errors": ["Missing service name"]}, status=HTTP_400_BAD_REQUEST
+                )
+
+            service, configuration, errors = validate_service_configuration(
+                service_name, request.data.get("configuration", "{}")
+            )
+            if service is None:
+                return Response(
+                    {"errors": errors},
```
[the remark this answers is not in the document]
gersona: Those errors don't come from stack trace, but were manually defined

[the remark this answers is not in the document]
nijel: Yes, this alert seems too strict to me, there are several occurrences in existing code as well.

weblate/api/views.py
```diff
@@ -1010,6 +1012,41 @@ def file(self, request: Request, **kwargs):
             name=instance.slug,
         )
 
+    @action(detail=True, methods=["get", "post"])
+    def machinery_settings(self, request: Request, **kwargs):
+        """List or create/update machinery configuration for a project."""
+        project = self.get_object()
```
nijel, line 1018: GET should be restricted to `project.edit` as well as the configuration can contain credentials for the services.

weblate/api/views.py
```diff
@@ -1010,6 +1012,41 @@ def file(self, request: Request, **kwargs):
...
+            if errors:
+                return Response({"errors": errors}, status=HTTP_400_BAD_REQUEST)
+
+            project.machinery_settings[service_name] = configuration
+            project.save(update_fields=["machinery_settings"])
+            return Response(
+                {"message": f"Service installed: {service.name}"}, status=HTTP_200_OK
+            )
+
+        return Response(project.machinery_settings, status=HTTP_200_OK)
```
nijel, line 1048: Would it be possible to use a serializer here? That way, the endpoint would be properly documented in OpenAPI.

weblate/api/serializers.py
```diff
@@ -1513,3 +1525,66 @@ class MetricsSerializer(ReadOnlySerializer):
...
+            "Simple message serializer",
+            fields={
+                "message": serializers.CharField(),
+            },
+        ),
+        201: inline_serializer(
+            "Simple message serializer",
+            fields={
+                "message": serializers.CharField(),
+            },
```
nijel, line 1583: Error: The map key in schemas "Simple message serializer" does not match the regular expression "^[a-zA-Z0-9\.\-_]+$"

weblate/api/serializers.py
```diff
@@ -1513,3 +1525,66 @@ class MetricsSerializer(ReadOnlySerializer):
...
+        ),
+        400: inline_serializer(
+            "Simple error message serializer",
```
nijel, line 1586: Error: The map key in schemas "Simple error message serializer" does not match the regular expression "^[a-zA-Z0-9\.\-_]+$"

nijel (commented):
There are some errors in OpenAPI validation: https://github.com/WeblateOrg/weblate/pull/13056/checks (You can ignore the warnings, these were present before).

nijel:

Issue #13050, opened 2024-11-16
Confirmation Skip action, if a translation was written


gcb:
### Describe the problem

User can press Skip instead of save, after writing a long translation. (Also, the dark theme shows a bright green for all buttons *but* sava&continue, which shows dimmer!)

### Describe the solution you would like

Besides fixing button colors (maybe a red/orange for skip), showing a confirmation if user really wants to skip when there's text inputed in the translation field.

### Describe alternatives you have considered

none.

### Screenshots

![image](https://github.com/user-attachments/assets/651147ee-9b3a-43e6-a876-adf64b72761b)
(I click skip 1 out of 10 times. Guaranteed)

### Additional context

I'm a very casual user of weblate.

nijel:
The button colors seem really wrong in the dark theme. @meel-hd can you please:

* [ ] Invert the button colors in the dark theme? Presently, they match light one, but it should be the inverse to highlight the most desired action.
* [ ] Add warning if user presses the “Skip” button after doing some edits?

PR #13065, opened 2024-11-18
Add skip unsaved changes confirmation

meel-hd:
## Proposed changes
- Fixed buttons visual heirarchy in  translation editor.
- Added a confirmation popup when the user clicks skip if there are unsaved changes.

close: #13050

## Checklist

## Other information
image

### DIFF
--- a/weblate/static/editor/base.js
+++ b/weblate/static/editor/base.js
@@ -30,6 +30,7 @@ WLT.Utils = (() => ({
 
 WLT.Editor = (() => {
   let lastEditor = null;
+  let hasChanges = false;
 
   function EditorBase() {
     const translationAreaSelector = ".translation-editor";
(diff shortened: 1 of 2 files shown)


gcb:
Thanks. that was fast!

code for confirmation :+1: 

I'm not familiar with [bootstrap naming convention for button roles](https://getbootstrap.com/docs/4.0/components/buttons/)... they seem to be exclusive for color not meaning... but i assume the the default form action "save and continue" might be `btn-default`. (and then you could even leave the other buttons as btn-primary instead of btn-link, i guess)

nijel:
Let's split this PR into two:

* JavaScript confirmation, that one is clear and can be merged right now.
* UI changes. I think these really should stay as buttons, using links for saving/suggesting seems wrong. The skip button could be a link, but it might be better to have all of them as buttons for UI consistency. What might be a better approach is to switch `btn-primary` and `btn-warning` colors in the dark theme. Currently, they are the same as in light and that brings attention to different buttons as the background is inverse.

PS: The button classes really don't have a real meaning, for historical reasons, Weblate is using `btn-warning` as secondary level of buttons.

meel-hd:
@nijel okay, I'll make this PR concerned with the confirmation logic only.

nijel:
@meel-hd I think this is a good approach, please create a separate PR for that.

PR #13085, opened 2024-11-20
docs: clarify used regex engine for search

nijel:
## Proposed changes

See #7095 or #13022 or #5322

## Checklist


yarons:
I'm not sure it's relevant but maybe it's worth specifying in the UI which syntax is used?
Regarding Hosted Weblate, which one is it?

nijel:
It most cases, Weblate uses PostgreSQL as that is the recommended backend.

Issue #12172, opened 2024-08-01
JavaScript libraries bundling


nijel:
### Describe the problem

Right now, Weblate uses custom tooling to include vendored libraries (see https://docs.weblate.org/en/latest/contributing/frontend.html#dependency-management). This is tricky to deal with for new contributors and brings issues when updating some libraries (see for example https://github.com/WeblateOrg/weblate/actions/runs/10178733602/job/28152952153?pr=11956 in https://github.com/WeblateOrg/weblate/pull/11956).

### Describe the solution you would like

There should be some standard solution that works for us. Requirements (those can be discussed, though):

* No additional runtime or installation time requirements besides Python modules.
* Store only needed files in Git, not full vendored packages.

### Describe alternatives you have considered

Things like webpack could be a solution. It might also replace the currently used django-compressor.

### Additional context

List of libraries to migrate:

* [x] `@tarekraafat/autocomplete.js`
* [x] autosize
* [x] daterangepicker
* [x] mousetrap
* [x] mousetrap-global-bind
* [x] multi.js
* [x] prismjs
* [x] slugify
* [ ] source-code-pro
* [ ] source-sans
* [x] tributejs

Libraries which probably won't be migrated:

* bootstrap-rtl as we use a newer snapshot than the release available as a package
* boostrap is using customized build

Maybe we should focus on #2916 instead.

meel-hd:
@nijel  This will be a big improvement. 

To sum up, currently used 3d party code is included in the app by `weblate/templates/base.html`  by more than a dozen script tags. linking to large 3rd party library files. Which are installed by yarn and copied using custom tooling into `static/vendor`.

**So given the desired results here it is what we will have:**
- Ease of install and update (one command `yarn install library`).
- Familiar to new contributors (The same usual workflow used by the web dev community using npm/yarn, with no need for custom tooling.)
- Installs will be stored in the _gitignored_ folder `node_modules`.
- The use of the 3rd party code in `static/vendor` will be unnecessary and can be removed.
- There will be one source of truth for declaring the use of third party code.

**How:**
1- We will introduce a new directory `vendor/` for managing frontend 3rd party code.
2- It will contain the `index.js`, `package.json`, `yarn.lock`, `webpack.config.js` and the _gitignored_ `node_modules/`.
3- `index.js` Will be used to import the libraries necessary for the frontend. And remove them if needed too.
4- _Webpack_ will be used to build and bundle the libraries used and declared in `index.js` into a minified file.
[...]

PR #13093, opened 2024-11-21
Move source-code fonts to new build workflow

meel-hd:
## Proposed changes
Moved source-code fonts to new build workflow

### Changes Include:
- Removed the old `scripts/yarn/** `dir, as it is no longer needed.
- Moved `source-code-pro and source-sans` to `client/`
- Built and bundled the fonts.
- Updated templates to use the new built fonts and styles.
- Added new built bundles license info in `REUSE.toml`.

### Note:
- The old bundles did not get removed to keep this PR simple for review.

### TODO:
- Remove old fonts bundles.
- Update any dependents on them to use the new bundles.
- if `FONTS_CDN_URL` is configured it should be seted up with `static/js/vendor/fonts` instead of the old `font-source/`
- Remove the old `yarn-update` script and its configs and workflows

Related: #12172

## Checklist


weblate/static/js/vendor/fonts/fonts.css
nijel, line 1: This is automatically generated, right?

weblate/templates/snippets/meta-css.html
```diff
@@ -22,8 +21,7 @@
 {% endif %}
 {% endcompress %}
 {% if fonts_cdn_url %}
-  <link rel="stylesheet" href="{{ fonts_cdn_url }}font-source/source-sans-3.css" />
-  <link rel="stylesheet" href="{{ fonts_cdn_url }}font-source/source-code-pro.css" />
+  <link rel="stylesheet" href="{{ fonts_cdn_url }}fonts/fonts.css" />
 {% endif %}
```
nijel, line 25: This won't work with existing CDN, but I think it's time to ditch this option completely

scripts/yarn-update
nijel, line 1: Let's remove this file completely.

weblate/static/js/vendor/fonts/font-source/07e9568bc36099d13146.woff2
nijel, line 1: You add fonts in new location, but the old ones are not removed in the PR. Furthermore, `weblate/fonts/utils.py` needs to be adjusted to the new fonts location.

weblate/static/js/vendor/fonts/fonts.css
meel-hd: Yes

weblate/templates/snippets/meta-css.html
meel-hd: Alright

weblate/static/js/vendor/fonts/font-source/07e9568bc36099d13146.woff2
meel-hd: Exactly, I didn't want to complicate this Pr more. If okay I'll go ahead and do that.

scripts/yarn-update
meel-hd: Wouldn't that break some GitHub actions. I thought let's leave it for a separate Pull

nijel: Better to break the actions right now than making them silently do nothing.

weblate/static/js/vendor/fonts/font-source/07e9568bc36099d13146.woff2
nijel: I just don't want to end up with two copies of the fonts.

scripts/yarn-update
nijel: The action will deal with that graceully now: https://github.com/WeblateOrg/meta/commit/3c66f2f2ea51080bc5f36d3841cf7d991fa0b0d9

nijel (commented):
The `docs/conf.py` also needs to be changed to use the new location of the fonts. I've made the docs build fail when the fonts directory doesn't exist in 425131e944.

client/webpack.config.js
```diff
@@ -152,6 +153,13 @@ module.exports = {
         test: /\.css$/,
         use: [MiniCssExtractPlugin.loader, "css-loader"],
       },
+      {
+        test: /\.(woff|woff2|eot|ttf|otf)$/i,
+        type: "asset/resource",
+        generator: {
+          filename: "fonts/font-source/[hash][ext]",
```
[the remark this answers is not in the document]
meel-hd: true

meel-hd:
@nijel I rerun the test suite on `main` (without the new changes) and it seems they fail also in there, I'm not sure what to do next.

### summary
`weblate/trans/tests/test_selenium.py::SeleniumTests::test_add_component` missing element `#list-add-button`
`weblate/trans/tests/test_selenium.py::SeleniumTests::test_fonts` missing element `#upload_font_submit`
`weblate/trans/tests/test_selenium.py::SeleniumTests::test_login` missing element `#dashboard-return`
`weblate/trans/tests/test_selenium.py::SeleniumTests::test_screenshots`  missing element `#query-dropdown`
`weblate/trans/tests/test_selenium.py::SeleniumTests::test_weblate`

client/webpack.config.js
```diff
@@ -152,6 +153,13 @@ module.exports = {
         test: /\.css$/,
         use: [MiniCssExtractPlugin.loader, "css-loader"],
       },
+      {
+        test: /\.(woff|woff2|eot|ttf|otf)$/i,
+        type: "asset/resource",
+        generator: {
+          filename: "fonts/font-source/[name][ext]",
+        },
+      },
```
nijel, line 162: Can we filter the assets somehow? We don't use several variants of the fonts, so there is no need to distribute them. We essentially want the following logic here:

https://github.com/WeblateOrg/weblate/blob/0e419825dd7a6964684e43af55d73c827758c7e3/scripts/yarn-update#L18-L29

meel-hd: Yes, we can. But isn't the built  fonts.css  will have missing dependencies. Because if they weren't used webpack wouldn't include them in the final output

nijel: Yes, the fonts.css uses them, but we don't use all the font variants in UI, so there are parts of CSS unused.

meel-hd: @nijel  upon close inspection and trying a couple solutions, it seems it isn't possible with webpack and how it works atleast without heavy modification of the css we import from `source-code-pro` and `source-sans`  or using our own css fonts definitions or maybe we fork those libraries and customize them to the likings of weblate UI. Otherwise, I think we leave it as, the downside would be a couple Megabytes of server storage used on deployements. what do you think?

nijel: Okay, let's leave it this way.

weblate/trans/tests/test_selenium.py
```diff
@@ -63,10 +63,11 @@
 SOURCE_FONT = os.path.join(
     os.path.dirname(os.path.dirname(os.path.dirname(__file__))),
     "static",
+    "js",
     "vendor",
+    "fonts",
     "font-source",
-    "TTF",
-    "SourceSans3-Bold.ttf",
+    "SourceSans3-Bold.ttf.woff2",
```
nijel, line 70: I guess this is why the tests fails:

```suggestion
    "SourceSans3-Bold.ttf",
```

docs/conf.py
```diff
@@ -44,7 +44,13 @@ def setup(app) -> None:
...
     )
 
-    font_dirs: list[str] = []
+    font_dirs: list[str] = [
+        str(weblate_dir / font_dir)
+        for font_dir in (
+            "weblate/static/js/vendor/fonts/font-source/",
+            "weblate/static/vendor/font-kurinto/",
+        )
+    ]
```
nijel, line 53: Seems like a wrong merge, you should adjust `font_locations` above.

meel-hd: Uh true, it should be an empty list. Thanks!

weblate/trans/tests/test_selenium.py
meel-hd: I rerun the test with the original file (and without deleting the asset .ttf) but it seems not the reason

nijel: woff2 upload is directly rejected, so the test with it cannot pass.

meel-hd: > woff2 upload is directly rejected, so the test with it cannot pass.

The .ttf version isn't included in the bundle by webpack because it isn't included in the css we import. So .woff2 was used instead. 

Btw Why woff2 isn't supported?

weblate/static/vendor/font-source/TTF/SourceSans3-Bold.ttf
nijel, line 1: @meel-hd This is the problem. It no longer includes TTF fonts. These are not referenced from the CSS, but Weblate needs them internally to render PNG widgets, and it is also used in the test suite (and that's why we see the failure).

meel-hd: fixed

nijel: No, this is not fixed. The TTF fonts needs to be shipped because they are used by Pango (via `weblate/fonts/utils.py`) to render bitmap widgets. Using different font in tests does not address this issue at all.

meel-hd: Apologies for the misunderstanding. I'll look into it.

meel-hd: The ttf fonts now are included

nijel:
Merged, thanks! I've made a fix to the testsuite (and reverted unnecessary changes there).

nijel:
And I didn't notice messed up licensing info, fixed that now in 1781de26555be2ef65758bd4a6be441e97a1ac10.

nijel:
I've created a followup issue at https://github.com/WeblateOrg/weblate/issues/13255. Using webpack is better than what we had before, but we might need something different in the end.

PR #13120, opened 2024-11-25
fix: Improve performance of Change filtering by timestamp date

### LINKED ISSUE #12970: Changes date filtering performance
### Describe the issue

Changes browsing when filtering by date is slow. This is because PostgreSQL doesn't use database index on `timestamp` once it is coverted to date by using `timestamp__date` lookup.

### I already tried

1. Have many changes
2. See what happens on the database upon filtering for changes by date

### Expected behavior

_No response_

### Screenshots

_No response_

### Exception traceback

_No response_

### How do you run Weblate?

Git checkout

### Weblate versions

_No response_

### Weblate deploy checks

_No response_

### Additional context

606b60ef07076596e3656289a68481bf325178ca made the similar change for search.

A similar approach should be used in  `weblate/trans/views/changes.py`. I am not sure if the same issue affects `weblate/metrics/models.py` where date lookup is used, but it is there actually to look for date, but it should be investigated as well.

gersona:
## Proposed changes
New `ChangeQuerySet` methods were added to filter Changes without converting converting `timestamp` to date in order to use Database Indexing.

## Checklist

## Other information

- closes #12970

### DIFF
--- a/weblate/trans/models/change.py
+++ b/weblate/trans/models/change.py
@@ -4,7 +4,7 @@
 
 from __future__ import annotations
 
-from datetime import timedelta
+from datetime import date, datetime, timedelta
 from typing import TYPE_CHECKING
 
 import sentry_sdk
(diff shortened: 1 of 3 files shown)


weblate/trans/models/change.py
```diff
@@ -28,15 +28,28 @@
...
-    from datetime import datetime
-
     from weblate.auth.models import User
     from weblate.trans.models import Translation
 
 
 CHANGE_PROJECT_LOOKUP_KEY = "change:project-lookup"
 
 
+def dt_as_day_range(dt: datetime | date) -> tuple[datetime]:
```
nijel, line 38: The type annotation is wrong, it returns two datetimes:

```suggestion
def dt_as_day_range(dt: datetime | date) -> tuple[datetime, datetime]:
```

PR #13309, opened 2024-12-16
Allow right click to inspect project

### LINKED ISSUE #12434: Mid-clickable expanded language lists
Papercut issue

Following up in https://github.com/WeblateOrg/weblate/issues/4656 which I suppose details [what files are touched](https://github.com/WeblateOrg/weblate/pull/12426/files).

### Describe the problem

0. Open https://hosted.weblate.org/accounts/profile/#notifications
1. Notice it has expanded languge lists.
2.  Be happy
3. Notice elements are not mid-clickable

I want to visit projects on the followed part without following, to dip my toes in.
Do I want to follow, or do I not?

I thought about having a list of projects that are actively unfollowed.

### Screenshots

bilde

### Additional context

Some part of this functionality hopefully allows greater control in rooting out malicious/spam projects.

meel-hd:
## Proposed changes
Allow right click to inspect project in watched projects list in profile settings.

**Note:** it was suggest to use middle click, but it has weird behavior and cross browser compatibility issues. Simply I went with right click instead. As it is not used.

This allows right click in /accounts/profile/#notifications in the watched project list. So users can open a project and inspect it before follwing it or not, as was suggested in issue #12434

close #12434 

## Checklist


nijel:
@comradekingu wanted the list to behave as regular links to that mouse middle click would open the link in new tab (built-in browser behavior). This, instead, adds a custom behavior to right-click, what nobody will find and will be confusing (as it also replaces the current window).

meel-hd:
I think that will be confusing. The element isn't a link and the main action isn't opening it. So adding a behavior usually associated with links I think isn't a good idea. That being said, the right click isn't that bad of an option (if we added text informing users about it below). And about opening in a new tab I think yeah that's right. I'll do that.

meel-hd:
@nijel the previous commit [2e88565](https://github.com/WeblateOrg/weblate/pull/13309/commits/2e88565a0487d3e5d1bbc78aa2daa5c0de4b20b1) still uses right click. But adding middle click if it is the final choice should be fairly easy. We just need to make sure it is the right choice.

nijel:
This is not about implementing custom functionality. The goal should be to make these things behave as links so that generic browser behavior for links works (besides normal click which will toggle the option in the select as now).

meel-hd:
I understand. I'll go with that

nijel:

PR #13339, opened 2024-12-19
Fix verb to plural in changes.rst

rffontenelle:
## Proposed changes

Adjust the verb conjugation so that it matches the plural in "Android string resources" title.

Current rendered text in [Changes page](https://docs.weblate.org/en/latest/changes.html#weblate-5-8) is:
> [Android string resources](https://docs.weblate.org/en/latest/formats/android.html#aresource) now supports translatable attribute in its strings.

## Checklist


nijel:

PR #13429, opened 2025-01-04
docs(api): update api description

walpox:
## Proposed changes

- Move the API description string to a separate variable to improve code readability.
- Add content to the (Open)API description summarized from general sections of the REST API documentation.

## Checklist

- [ ] Lint and unit tests pass locally with my changes.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have added documentation to describe my feature.
- [ ] I have squashed my commits into logic units.
- [x] I have described the changes in the commit messages.

## Other information

The rationale behind this change is to help the OpenAPI document stand on its own.

Relates to #12584

PR #13430, opened 2025-01-05
Set EXTERNAL_DOCS setting for OpenAPI spec

walpox:
## Proposed changes

- Set the `EXTERNAL_DOCS` setting for the OpenAPI document.

## Checklist

## Other information

Relates to #12584


walpox:
The merge conflict has been resolved and the PR should be good to go now.

weblate/api/spectacular.py
```diff
@@ -9,6 +9,8 @@
 
 from django.utils.translation import gettext_lazy
 
+from weblate.utils.docs import get_doc_url
```
nijel, line 12: Unfortunately, this won't work because `get_doc_url` relies on settings and `get_spectacular_settings` is called from settings. In the end, Django fails to load the settings and fails with weird errors like you can see in the test suite.

walpox: I have changed it so that `EXTERNAL_DOCS` uses a hardcoded URL to the home page of documentation. I don't know how to fix this issue properly in a timely manner, and don't want to delay release for your milestone...

The changes to `get_doc_url` have also been reverted.

walpox: Also, note that `get_doc_url` is used in `api_description` in #13429 and it seems to work fine there. If you would be so kind to let me know what test doesn't pass with these changes...

Scutua (approved):

Scutua (approved):

weblate/api/spectacular.py
nijel: Using lazy evaluation might work, but it depends on when this setting is actually used.

PR #13472, opened 2025-01-10
fix: memory import KeyError when uploading unsupported files

### LINKED ISSUE #13416: KeyError: 'file' on translation memory upload
Sentry Issue: [WEBLATE-1WP1](https://weblate.sentry.io/issues/15711522/?referrer=github_integration)

```
KeyError: 'file'
(8 additional frame(s) were not displayed)
...
  File "weblate/memory/views.py", line 64, in dispatch
    return super().dispatch(request, *args, **kwargs)
  File "weblate/memory/forms.py", line 56, in clean
    extension = PurePath(data["file"].name).suffix[1:].lower()
```

gersona:
## Proposed changes
- Add extension check for in `import_file` method
- Only continue file validation if no previous errors found

## Checklist

## Other information

- closes #13416


weblate/memory/forms.py
```diff
@@ -17,19 +18,9 @@ class UploadForm(forms.Form):
...
-                    "xliff",
-                    "po",
-                    "csv",
-                ]
-            )
-        ],
+        validators=[FileExtensionValidator(allowed_extensions=SUPPORTED_FORMATS)],
         help_text=gettext_lazy(
-            "You can upload a file of following formats: TMX, JSON, XLIFF, PO, CSV."
+            f"You can upload a file of following formats: {', '.join(SUPPORTED_FORMATS)}."
```
nijel, line 23: This won't work well for localizing, you need to use format string instead to inject value into localized string returned by gettext.



```suggestion
            "You can upload a file of following formats: %s."
```

And then apend `% ', '.join(SUPPORTED_FORMATS)`.

weblate/memory/tests.py
```diff
@@ -205,9 +205,10 @@ def test_import_po(self) -> None:
         """Test the import of an GNU PO file."""
         self.import_file_with_languages_test("cs.po", "en", "cs", 1)
 
-    def test_import_tbx(self) -> None:
-        """Test the import of a TBX file."""
-        self.import_file_with_languages_test("cs.tbx", "en", "cs", 4)
```
nijel, line 210: Was this removal intentional?

PR #13512, opened 2025-01-15
feat: make file sync work on all components in a repository

nijel:
## Proposed changes

This is more expected behavior than having to do it one by one when the repository is shared.

## Checklist


garbear:
Thanks for fixing this!

Issue #8446, opened 2022-12-07
Add check for Sphinx/RST syntax


rffontenelle:
### Describe the issue

I've been translating Weblate Docs for sometime and one issue I've always came across was Sphinx roles incorrectly rendered because of translation mistakes. These mistakes are not reported for the translator (i.e. they were not catch by quality checks), so they might not be noticed by the translator when translation several strings.

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar issues in this repository.

### Steps to reproduce the behavior

2. Scroll down to '...'
3. Click on '...'

### Expected behavior

For the translator to be (automatically) notified of mistakes in Sphinx role translations that trigger WARNING: and ERROR: when building and that are incorrectly rendered in the translated documentation.

### Screenshots

Here is an example from [pt_BR translation of installation in Debian page](https://docs.weblate.org/pt_BR/latest/admin/install/venv-debian.html#configuring-weblate) of Sphinx role ``` :command:`weblate` ``` incorrectly translated as ``` :comando:`weblate` ``` and being rendered like that.

![Role incorrectly translated as "comando"](https://user-images.githubusercontent.com/1571783/206218016-79ee65e4-afe1-49de-abce-7d1f79807689.png)

### How do you run Weblate?

weblate.org service

### Additional context

How I found these translation mistakes?

I browsed the [build logs of weblate-pt-br at readthedocs](https://readthedocs.org/projects/weblate-pt-br/builds/) and found interesting information.

1. https://readthedocs.org/projects/weblate-pt-br/builds/
2. Entered the latest finished build ([#18844887](https://readthedocs.org/projects/weblate-pt-br/builds/18844887/), in my current case)
3. In the specific build page, clicked *See raw*
4. In the raw logs, rolled down until I found `reading sources... [100%]`
5. Now, Sphinx's `WARNINGS:` and `ERRORS:` start<br>
NOTE: The errors/warnings always shows `<filepath>:<linenumber>`,  e.g.  `home/docs/checkouts/readthedocs.org/user_builds/weblate-pt-br/checkouts/latest/docs/admin/auth.rst:676`. The useful part is `admin/auth.rst:676`.<br><br>
6. On the Weblate Documentation component, I found the string using the search filter `location:`  (e.g. `location:admin/auth.rst:676`)

Some examples of issues (removed the file fullpath, keeping only after docs/):

[...]

PR #13516, opened 2025-01-15
feat(ci): build localized documentation in CI

nijel:
## Proposed changes

This allows to catch errors in translations earlier.

## Checklist


nijel:
We should really introduce Weblate checks to catch these errors earlier, see https://github.com/WeblateOrg/weblate/issues/8446

Issue #3412, opened 2020-01-22
Support fallback to the "generic" language variant


jlsjonas:
**Is your feature request related to a problem? Please describe.**
We decided to use a fallback strategy to limit the amount of duplicate translations for similar languages in our application, this means that several country-specific language files are now almost empty (e.g. `nl-BE` contains only specific overrides on `nl`).
Weblate is showing a lot of untranslated strings as part of this (and also doesn't seem to have an easy method to _remove_ a duplicate)

**Describe the solution you'd like**
Allow an option/flag to take into account the _base language_ (i.e. `nl`) when providing insights for a country-specific variant (like `nl-BE`).
I believe that this should ideally mark translations as needing verification, and then marking it as translated internally (without adding the key to the country-specific file)

**Describe alternatives you've considered**
Using secondary language to at least see them, however the _secondary language_ will depend on which language is being translated, additionally this shows a lot of keys as not translated, which can be confusing.

**Additional context**

![image](https://user-images.githubusercontent.com/932193/72909876-fddf8000-3d37-11ea-8369-b6cd497ecb71.png)

jlsjonas:
I believe a lot of projects would work with such a use-case and are currently working around this limitation (there's also several older issues to back this thought up)

As for implementation requirements, I believe this is quite limited:

- ignore flag for "not translated" (needs editing would still show for new source-based entries, translation status would ignore not translated, and only take _needs editing_ into account)
- (ideally) ability to delete a non-source key (saving empty string could/should probably do this)

comradekingu:
I was going to file this too.
I think marking a translation as a secondary subset of a primary translation would be the first step.
(Either by default, or on a per-project/component basis).
Then strings in the secondary language could be unlocked on a per-string basis.

That way secondary translations only feature strings that actually differ, instead of maintaining two diverging stringbases. (Effort would be better spent improving the basis rather than duplicating it).

franco999:
> That way secondary translations only feature strings that actually differ, instead of maintaining two diverging string-bases. (Effort would be better spent improving the basis rather than duplicating it.)

@comradekingu I am interested in this way of translation, it would be easier to maintain.
In my case, I am currently maintaining the versions of translations in "**Spanish (Argentina)**" (my country with its own regionalisms) and "**Spanish (United States)**" (version as neutral as possible for Latin America), since the "Spanish" version has many strings with regionalisms specific to Spain. 
[...]

Issue #2320, opened 2018-10-15
Derivative or sub-translation


Kingdutch:
**Is your feature request related to a problem? Please describe.**
We have a project where we sometimes create derivative installations that require slightly different wordings for some translations (e.g. "Group" is known as "Community").

**Describe the solution you'd like**
I'd like to have some way, either a sub-project or a sub-language that can selectively translate some strings and use another translation/project for all other strings.

**Describe alternatives you've considered**
Found nothing that was easy so far. Re-importing another translation could work but it could overwrite customisations that have been made.

ali-avci:
Have the similar requirement, with a solution in mind. 

Connecting projects similar to class inheritance and the project that is inherited receives the change as suggestion from base project. (given that the contexts are same)

nijel:
What some projects are using is additional languages for this. You can use something like `en@community` or `en_community`. That will not address the non changed strings though...

nijel:
With a clever localization framework, the fallback from variant to base language happens automatically (for example Gettext) does that.

I think something like this has been implemented in https://github.com/vendasta/weblate, but it was never contributed back.

Issue #1585, opened 2017-08-04
Ability to mark translation string as not needed


KamilSowa:
The issue we currently experience is inability to tell Weblate to not require a translation in one of the supported languages.

Image we have two features (F1 and F2) in the application and two countries (C1 with language L1 and C2 with language L2). If feature F2 is only applicable to C1, then translating it to L2 does not make sense, right? Is there a way in Weblate to enable such behaviour? Otherwise all translations from F2 will be marked in L2 as untranslated (they are in fact not needed).

I would greatly appreciate if there's some sort of support for such a scenario.

nijel:
You can split the translation into several files (which would end up as separate components in Weblate). Each can be then translated into different languages. 

4. srpna 2017 13:33:52 GMT+01:00, Kamil Sowa <notifications@github.com> napsal:
>The issue we currently experience is inability to tell Weblate to not
>require a translation in one of the supported languages.
>
>Image we have two features (F1 and F2) in the application and two
>countries (C1 with language L1 and C2 with language L2). If feature F2
>is only applicable to C1, then translating it to L2 does not make
>sense, right? Is there a way in Weblate to enable such behaviour?
>Otherwise all translations from F2 will be marked in L2 as untranslated
>(they are in fact not needed).
>
>I would greatly appreciate if there's some sort of support for such a
>scenario.

Issue #12815, opened 2024-10-19
Untranslatable glossary should be per language


timothyqiu:
### Describe the issue

Marking a glossary term as untranslatable in one language affect all other languages.

One term is untranslatable in one language does not mean it is untranslatable in another. For example, 
McDonald's the brand name is almost always untranslatable except in Chinese. This is also true for many tech terms.

Users often set a term to read-only simply because it cannot be translated in their own language. Users of other languages who find that a term they need to translate is marked as read-only will in turn unmark it read-only. It goes back and forth.

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

1. Go to some glossary term.
2. Tools > Mark as untranslatable.
3. Check this term in another language.

### Expected behavior

Untranslatable is a per language setting, not per glossary term.

nijel:
#12779 is related to this (also wants per-translation flags).

nijel:
This will need https://github.com/WeblateOrg/weblate/issues/14298

PR #13521, opened 2025-01-15
feat(workflow): add string filter for languages

michael-smt:
## Proposed changes

This adds a new setting to the language workflow of a project to exclude units from translation.

With regional languages it can be useful to only translate a subset of units. This avoid duplicating translations which have to be kept up to date, instead relying on a fallback in the localization framework (e.g. https://github.com/WeblateOrg/weblate/issues/2320#issuecomment-2200032529). By limiting translation to only the relevant units it is clear what is region specific and the progress can be properly reflected. See also #3412. It could also support cases such as #1585.

For example first you label all units that should be regionalized for British English with a label like `uk-regional`. Then configure the project's `en_GB` language workflow with the filter `label:uk-regional`. Only these units are available for translation to `en_GB`.

## Checklist

## Other information

This PR is far from complete and the filter will need to be added in a lot of places. I first wanted to get a feeling if this feature makes sense.

Also the naming might not be ideal...


nijel (commented):
I'm not really convinced this is a practical approach, as this opens up several consistency issues. Two ones I was able to figure out quickly:

* This will hide the strings from the translating UI while they still being counted for in the statistics, making it impossible to complete the translation.
* The filtering will apply only at certain levels (when language settings apply) and not other (project- or site-wide search). So you will find the string in the site-wide search, and clicking on it will tell you that the string does not exist.

nijel:
Right now, we're using `read-only` flag for this. It works on the source string, but there is no technical limitation why it would not work on the translations as well. The only tricky issue is how to reasonably present this to the users.

nijel:
I think any per language string filtering will add a lot of complexity. There are other use cases for flags per translation (see https://github.com/WeblateOrg/weblate/issues/12815), so this might be the best approach. But my comment https://github.com/WeblateOrg/weblate/issues/12815#issuecomment-2436966495 still applies.

nijel:
The implementation already supports per translation flags (`extra_flags` attribute on `Unit`), it's “only” the interface to edit these which is missing. The tricky part is to not introduce too much confusion for users with two levels of flags for a string.

PR #13534, opened 2025-01-16
Updated language aliases handling 

### LINKED ISSUE #12368: Updated language aliases handling
### Describe the problem

Changes in `weblate-language-data` can turn existing languages into aliases (for example, in https://github.com/WeblateOrg/language-data/pull/1427). This will have an immediate effect on new installations, but will do nothing on existing installations.

The existing installation language definitions should be updated if `UPDATE_LANGUAGES` is turned on.

1. Factor out logic from the `move_language` management command to a module.
2. Better handle integrity errors in that code (when both languages exist in a component), most likely by emitting a warning.
3. Migrate content for any existing aliased language to the alias target in `Language.objects.setup` using factored out code.
4. Remove blank aliased languages.
5. The legacy codes in the URLs will be automatically redirected by the existing middleware, so no change is needed for that.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

gersona:
## Proposed changes
- migrate content from alias language to alias target on `Language.objects.setup()` to affect existing installation
- blank alias languages are deleted

## Checklist

## Other information
- fixes #12368


weblate/lang/models.py
```diff
@@ -588,8 +588,63 @@ def setup(self, update, logger=lambda x: x) -> None:  # noqa: C901
...
+        for plural in source.plural_set.iterator():
+            formulas = target.plural_set.filter(formula=plural.formula)
+            try:
+                new_plural = formulas[0]
+            except IndexError:
+                plural.language = target
+                plural.save()
+            else:
+                plural.translation_set.update(plural=new_plural)
+
```
gersona, line 647: This is a bit unclear to me from the `move_language` command, why isn't the plural language changed to the new target every time ?
This seems to leave some plurals attached to a potentially stale language. Or what is the reason behind this ?

weblate/lang/models.py
```diff
@@ -588,8 +588,63 @@ def setup(self, update, logger=lambda x: x) -> None:  # noqa: C901
...
+                # delete alias language if blank
+                if not any(
+                    [
+                        language.translation_set.exists(),
+                        language.profile_set.exists(),
+                        language.secondary_profile_set.exists(),
+                        language.component_set.exists(),
+                        language.change_set.exists(),
+                        language.announcement_set.exists(),
```
gersona, line 606: Depending on the answer to my other comment, this may also include
`language.plural_set.exist()`

weblate/trans/tests/utils.py
```diff
@@ -254,10 +254,12 @@ def create_po_branch(self):
...
         return self.create_po(push_on_commit=True)
 
-    def create_po_empty(self):
-        return self._create_component(
-            "po", "po-empty/*.po", new_base="po-empty/hello.pot", new_lang="add"
-        )
+    def create_po_empty(self, project=None):
+        kwargs = {"new_base": "po-empty/hello.pot", "new_lang": "add"}
+        if project:
+            kwargs |= {"project": project}
```
nijel, line 260: Why not directly set it?

```suggestion
            kwargs["project"] = project
```

weblate/lang/models.py
nijel: This should migrate all translations to plurals of the new language, either by using existing or by moving ones. The stale plurals would be removed upon the language removal.

weblate/lang/models.py
nijel: These should not block removal, but there should be nothing using these plurals at this point, so something like this might be a reasonable check here:

```python
Translation.objects.filter(plural__in=language.plural_set.all()).exists()
```

weblate/lang/models.py
```diff
@@ -588,8 +590,67 @@ def setup(self, update, logger=lambda x: x) -> None:  # noqa: C901
...
+                        Translation.objects.filter(
+                            plural__in=language.plural_set.all()
+                        ).exists(),
+                    ]
+                ):
+                    language.delete()
```
nijel, line 614: This is bad performance wise, as all those queries will be evaluated in any case. On the other side, it is error-prone to future changes in the models.

The most likely situation here is that translation will exist, so use it as fast check and then use `django.db.models.delection.Collector` (or `django.contrib.admin.utils.NestedObjects` as it might provide a better structured data) to figure out if object removal is going to remove anything else besides Language and matching Plural objects.

```suggestion
                if not self.has_no_childs(language):
                    language.delete()
```

The `has_no_childs` method then could look like (not tested):

```py
    def has_no_childs(self, language: Language) -> bool:
        # Quickly look for existing translations, this is fast and the
        # most likely object existing for the language
        if language.translation_set.exists():
             return False

        # Collect all objects to be deleted. This can be expensive
        # as it fetches all the objects from the database.
        collector = NestedObjects("default") # Can we get db alias somehow here?
        collector.collect([language])
        current = None
        for current := collector.edges[current]:
            if isinstance(current, Language) and current == language:
                continue
            if isinstance(current, Plural) and current.language == language:
                continue
             return False
        return True
```

gersona: and how about `Memory` object attached to a language ? they are not migrated in the `move_language` ?
Shoud they be ignored as `Plurals` or migrated to the new language as well ?

nijel: Good catch! These should be migrated, it was apparently missed in the old code.

nijel:

PR #13586, opened 2025-01-20
Deduplicate strings in API page

rffontenelle:
These sentences have one more similar in the page, and translating with the same text causes Reused Translation to pop up. This PR basically edits to uniformize, which will deduplicate strings for translation.


nijel:

PR #13643, opened 2025-01-23
feat: improve language mapping for Azure AI Translator

nijel:
The service supports new languages and is now stricter in accepting language codes.


hicksjacobp:
@nijel do you have an idea of when this fix will be in a new released version?

nijel:
It should be released next week, see https://github.com/WeblateOrg/weblate/milestone/133

Issue #13304, opened 2024-12-16
Ability to download the statistics


PadraigDillon1:
### Describe the problem

We recently had a business review with our localization vendor. They reported that they would love to see a feature allowing them to download the wordcount statistics into a local log for each language.  They spend alot of time manually doing this task and would improve efficiency if it could be exported or downloaded. 

### Describe the solution you would like

This could be as simple as a exported .csv file that contains each of the categories and the wordcounts.

orangesunny:
@nijel, let’s talk about this and also probably about exposing it in the UI as well

nijel:
Okay, let's use API endpoints only for the statistics and include the links in the UI at all levels. https://github.com/WeblateOrg/weblate/pull/13666 is the first step in that direction.

PR #13666, opened 2025-01-27
feat(api): add CSV renderer

nijel:
This allows to fetch API responses in the CSV formats. This might be especially useful for statistics as outlined in #13304.


orangesunny:
@jfaltenbacher better and easier stats exports are in the progress

PR #13671, opened 2025-01-27
docs: consolidate autofixes docs

nijel:
It was in two places and one of them was incomplete.


AliceVisek (commented):
OK: Changes are matching, placed correctly, expressed clearly.

Issue #13525, opened 2025-01-15
Bad contrast for Human verification sign up


jermanuts:
### Describe the issue

As title says, here's how it looks currently on dark theme

![image](https://github.com/user-attachments/assets/1a2b4c2d-a729-4f60-b280-b709abdea32b)

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

Click on [Register](https://hosted.weblate.org/accounts/register/)
Scroll to see "Human verification"

### Expected behavior

contrast should be better

### Screenshots

![image](https://github.com/user-attachments/assets/1a2b4c2d-a729-4f60-b280-b709abdea32b)

### How do you run Weblate?

weblate.org service

nijel:
Reopening as https://github.com/WeblateOrg/weblate/pull/13686 was reverted due to unwated effects it cause.

meel-hd:
fixed now without affecting other places.

PR #13686, opened 2025-01-28
Fix signup human verification text color

meel-hd:
Fix signup human verification text color n dark theme.
also grouped scattered css varaiables into one place in `variables.css`

Close: #13525


nijel:
This broke rendering of all panels:

Image

Original rendering:

Image

I'm reverting this and reopening #13525.

Please test your changes visually before creating a pull request.

PR #13740, opened 2025-02-03
chore(ci): switch to girepository 2

nijel:


nijel:
Needed once pygobject 3.52.0 is out.

PR #13741, opened 2025-02-03
Fix signup human verification text color

meel-hd:
Fix signup human verification text color in dark theme.
Close: #13525


nijel:

PR #13786, opened 2025-02-07
feat: replace re module with regex

### LINKED ISSUE #13728: Support unicode categories in regex translation checks
### Describe the problem

I need to be able to force tighter constraints on some strings across languages that done use a strict a-z.
Specifically I want to restrict a string down to any letter, number, -, _, and 1-32 in length.
There's a shorthand regex that works for this:
```regex
^[-_\p{L}\p{N}\p{sc=Deva}\p{sc=Thai}]{1,32}$
```
This works using python regex (not re), but can't seem to be used with the existing translation flags (tested on webalte 5.9.2).

I would like to be able to apply the following translation flag
```regex
regex:'^[-_\p{L}\p{N}\p{sc=Deva}\p{sc=Thai}]{1,32}$'
```

DYI checks and rejecting invalid strings when we load the translation

### Screenshots

Image

### Additional context

_No response_

4-9:
This replaces all imports of the `re` regex module with [`regex`](https://pypi.org/project/regex/). This should be backward compatible with `re`, so just changing the import should be enough.

Some tests fail for as yet unknown reasons, as mentioned `regex `should be backward compatible with `re`, so some tweaks to the tests might be needed.

In the UI, all works as expected. You can use the following test regex check to validate.
```
regex:'^[-_\p{Ll}\p{Lo}\p{N}\p{sc=Deva}\p{sc=Thai}]{1,32}$'
```

Fixes #13728

### DIFF
--- a/weblate/checks/placeholders.py
+++ b/weblate/checks/placeholders.py
@@ -4,9 +4,9 @@
 
 from __future__ import annotations
 
-import re
 from typing import TYPE_CHECKING, Any, Literal
 
+import regex as re
 from django.utils.functional import SimpleLazyObject
 from django.utils.html import escape, format_html, format_html_join
 from django.utils.safestring import mark_safe
--- a/weblate/checks/placeholders.py
+++ b/weblate/checks/placeholders.py
@@ -6,7 +6,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Any, Literal
 
-import regex as re
+import regex
 from django.utils.functional import SimpleLazyObject
 from django.utils.html import escape, format_html, format_html_join
 from django.utils.safestring import mark_safe
@@ -21,7 +21,7 @@ if TYPE_CHECKING:
 
 def parse_regex(val):
     if isinstance(val, str):
-        return re.compile(val)
+        return regex.compile(val)
     return val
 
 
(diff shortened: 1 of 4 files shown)


nijel:
How about doing that in the regex check only? There is no use for this elsewhere and I'm worried about performance regressions.

nijel:
Apparently, it's not 100% API compatible, so let's introduced it really in the single place where it is wanted:

```
AttributeError: '_regex.Pattern' object has no attribute 'decode'
```

4-9:
Just updating the one file file needed now.

weblate/checks/placeholders.py
```diff
@@ -4,9 +4,9 @@
 
 from __future__ import annotations
 
-import re
 from typing import TYPE_CHECKING, Any, Literal
 
+import regex as re
```
nijel, line 9: I'd prefer to use the original name to make it clear that a different implementation is in use.

```suggestion
import regex
```

nijel (commented):
Would you be able to add the test for the expression you wanted to use? The existing tests are in weblate/checks/tests/test_placeholders.py.

nijel:
Can you also please add `types-regex==2024.11.6.20241221` to the `types` dependency group in `pyproject.toml` so that the type checker does not complain.

4-9:
I've added a test now, but not entirely sure what I'm doing. It seems to work, and would fail if `re` is used

nijel:

nijel:
I've made additional changes for this:

* https://github.com/WeblateOrg/weblate/pull/13846
* https://github.com/WeblateOrg/weblate/pull/13847

Issue #13399, opened 2024-12-27
Cannot change position input


iamshoXy:
### Describe the issue

![image](https://github.com/user-attachments/assets/bb17d8a0-4d3d-4feb-9b92-4520f4177715)
Once you click on the position box, you should be able to enter numbers to navigate between pages more easily.
It seems that a certain update broke it, because as soon as it comes into focus, it turns uneditable again.

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

1. Go to any page where the navigation box is visible (Browse, Translate)
2. Left click on the navigation box to become editable
3. Left click inside the input which will cause the position input to become uneditable again

### Expected behavior

Just like before, the numbers should be editable again.

### How do you run Weblate?

Docker container

### Weblate versions

5.9.2

nijel:
@meel-hd this might be a regression from #12648

meel-hd:
I'm on it

PR #13816, opened 2025-02-10
Fix the pagination input.

meel-hd:
Now it is editable in /browse and in /changes as well.

close: #13399


nijel:
It now doesn't disappear, but the input is still not focused after clicking like it used to be.

meel-hd:
Focus is fixed. It accounts, now, for more than 1 paginator in either page, be it /browse or /changes.

nijel:

Issue #13886, opened 2025-02-15
Events in the `History` tab of the full editor overlap with the `String information` panel


Geeyun-JY3:
### Describe the issue

Events in the `History` tab of the full editor overlap with the `String information` panel in some cases.
The length of the event body is too long in the horizontal direction.

### I already tried

- [x] I've read and searched [the documentation](https://docs.weblate.org/).
- [x] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

1. Go to https://hosted.weblate.org/translate/weblate/glossary/zh_Hans/?checksum=e63e103261004099#history
2. Switch to the `History` tab

### Expected behavior

Events in the `History` tab of the full editor do not overlap with the `String information` panel in some cases.

### Screenshots

![Image](https://github.com/user-attachments/assets/f1f348ac-252e-44d8-93c0-b9886c2c2f83)

### Exception traceback

```pytb

```

### How do you run Weblate?

weblate.org service

### Weblate versions

Weblate 5.10

### Weblate deploy checks

```shell

```

meel-hd:
Fixing it!

PR #13900, opened 2025-02-17
Fix history events overlay in the editor

meel-hd:
If it overlaps it will be scrollable in boundry.

### DIFF
--- a/weblate/static/styles/main.css
+++ b/weblate/static/styles/main.css
@@ -653,6 +653,10 @@ form + .login-label {
   z-index: 10;
 }
 
+#history {
+  overflow-x: auto;
+}
+
 @media only screen and (max-width: 600px) {
   .btn-float {
     box-shadow: none;


weblate/static/styles/main.css
```diff
@@ -675,13 +678,21 @@ form + .login-label {
     gap: 20px;
   }
 
-  #changes-item {
+  .breadcrumb {
```
nijel, line 681: Why change `.breadcrumb`? I don't see any breakage, but that doesn't seem to be related to this change at all and can cause issues.

weblate/static/styles/main.css
```diff
@@ -640,24 +640,27 @@ form + .login-label {
...
 .btn-float {
   display: flex;
   gap: 5px;
   padding: 4px;
   background-color: #fff;
-  box-shadow: 0 2px 10px 0 rgba(0, 0, 0, 0.1);
   border-radius: 5px;
   position: absolute;
   right: 15px;
-  transform: translateY(-100%);
```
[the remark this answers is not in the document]
meel-hd: Damn, true, I will fix it?

weblate/static/styles/main.css
meel-hd: yeah it is related. It causes that view to overflow under (in cases when the text is too big for the view holding it).
image

 Without this change the view breaks on mobile

nijel:
I think the rendering for wide screens shout stay as it is now:

image

This pull request changes it to:

image

meel-hd:
@nijel this layout structure is one of 2 reasons why the UI overlaps as mentioned in #13886. This updated version helps avoid this. I am not sure what a better solution could be?

nijel:
Yes, but this could be again done only on smaller resolutions, doing that on wide screens just puts a lot of blank space into the view.

meel-hd:
That's true, but I would argue whitespace is considered a good practice in UI design. It makes it feel less cramped or  cluttered.

nijel:
While it is sometimes a good practice, I doubt that it would be a good practice to fit half of he content to the screen:

image

versus

image

Another good practice is UI consistency, and you've changed rendering of changes, but the suggestions are still using the existing layout. We want CSS to be reusable and not that specific so that the same layout can be reused.

meel-hd:
I see, I'll bring that layout back

PR #13991, opened 2025-02-24
Update password validation

nijel:
While we haven't seen any issues with longer passwords, the 72 is the maximal length recommended by zxcvbn-python. Because of unhandled error longer passwords lead to crash.

Also add a length validator that shows this limit visibly to the user.

django-zxcvbn-password is no longer maitained, so better switch to a maintained alternative.

### DIFF
--- a/docs/admin/config.rst
+++ b/docs/admin/config.rst
@@ -1357,6 +1357,23 @@ List for credentials for Pagure servers.
    :ref:`vcs-pagure`,
    `Pagure API <https://pagure.io/api/0/>`_
 
+.. setting:: PASSWORD_MINIMAL_STRENGTH
+
+PASSWORD_MINIMAL_STRENGTH
+-------------------------
+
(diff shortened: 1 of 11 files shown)


docs/admin/config.rst
```diff
@@ -1357,6 +1357,22 @@ List for credentials for Pagure servers.
...
+.. setting:: PASSWORD_MINIMAL_STRENGTH
+
+PASSWORD_MINIMAL_STRENGTH
+-------------------------
+
+.. versionadded:: 5.10.2
+
+Minimal password score as evaluated by the `zxcvbn
+<https://github.com/dwolfhub/zxcvbn-python>`_ password strength estimator.
+Defaults to 0 to what disables the strength checking.
```
AliceVisek, line 1369: Suggestion - replace with:
Default setting to 0 disables the strength checking.

AliceVisek (approved):

Issue #2825, opened 2019-06-22
Support for PostgreSQL and MySQL only


nijel:
Currently Weblate supports any database backend supported by Django. That's nice, however that limits possibilities to use database specific features. 

For example switching to PostgreSQL only would allow us to use [it's full text search](https://docs.djangoproject.com/en/2.2/ref/contrib/postgres/search/) removing need for external full text engine. The Whoosh usage currently leads to some problems for example see https://github.com/WeblateOrg/weblate/issues/2549 and https://github.com/WeblateOrg/weblate/issues/2543. Having in database full text engine would also heavily improve it's performance, because currently it's first doing search in Whoosh to get fulltext matches and then filter the results in the database to perform filtering by translation and permission checks, what leads to issues as https://github.com/WeblateOrg/weblate/issues/2876.

nijel:
The migration guide for the database can be found here https://www.calazan.com/migrating-django-app-from-mysql-to-postgresql/

nijel:
Another nice benefit of switching to Postgres might be using Agagama instead of writing translation memory service on our own, see https://github.com/translate/amagama

nijel:
Things to improve based on this change:

- [ ] Migrate to [JSONField](https://docs.djangoproject.com/en/stable/ref/contrib/postgres/fields/#jsonfield)
- [ ] Use [fulltext search](https://docs.djangoproject.com/en/stable/ref/contrib/postgres/search/)

maxking:
While whoosh is pretty slow, we have seen a lot of performance improvements when using Django-haystack with Elasticsearch in Mailman project. It is definitely another service to support, but with containers it isn't a lot of work to setup. There are system packages for most operating system.

Xapian is written in C++ and also is one the supported backends in Django-haystack.

Just a thought, in case the only reason to drop support for other databases is slow searching.

nijel:
In case the PostgreSQL only solution will not perform well, Elasticsearch is certainly worth trying. However my current impression is that we can do all what we need with PostgreSQL and reasonable query parser. We might stick with Whoosh for query parsing as it was quite easy to adapt for our purpose, see https://github.com/WeblateOrg/weblate/blob/master/weblate/utils/search.py.

PS: The performance is not the biggest issue with Whoosh for me, it does search quite fast. The biggest issue is *huge* memory consumption (the updating process easily consumes gigabytes of memory) and lockups while updating (I've not yet found time to debug this).

unho:
[...]

PR #14009, opened 2025-02-25
feat: Webhooks

### LINKED ISSUE #7604: Web Hooks

gersona:
- fixes #7604


gersona:
@nijel before continuing, I'd like to have your input on a few points:

### icon
Is it beneficial to add a new icon just for the webhooks add-on ? If so, I thought of something like the following, but am not sure if there's a repository we can get them from
https://icons8.com/icons/set/webhook--nocorrect

### delivery failures and retries
To simplify things, I rely on the celery tasks to for the request delivery , which handles retries on failure.
Does it seem OK to your or do we need a way to track delivery status for each webhook event. 

### Payload
I more or less replicated what was included in the Fedora messaging payload. Is there something special to be added or removed ?

### Payload signature
the standard-webhooks guidelines encourage to sign the payload, do we also need to provide a way to verify that signature in the code (a `verify` method that users can use), or should it just be explained in the documentation ?

weblate/addons/models.py
```diff
@@ -534,6 +535,14 @@ def store_post_load_handler(sender, translation: Translation, store, **kwargs) -
...
 
 
+@receiver(post_save, sender=Change)
+def change_post_save_handler(sender, instance: Change, created, **kwargs) -> None:
+    from weblate.addons.tasks import addon_change
+
+    if created:  # ignore Change updates, they should not be updated anyway
+        addon_change(sender, instance, **kwargs)
+
+
```
nijel, line 545: There is also a bulk signal triggered on bulk change creation which needs to be handled:

https://github.com/WeblateOrg/fedora_messaging/blob/710a5b5156631eb24479e9f02e79d754208f3480/weblate_fedora_messaging/models.py#L36-L40

weblate/addons/tasks.py
```diff
@@ -163,3 +163,25 @@ def setup_periodic_tasks(sender, **kwargs) -> None:
         cleanup_addon_activity_log.s(),
         name="cleanup-addon-activity-log",
     )
+
+
+@app.task(trail=True)
+def addon_change(sender, change, **kwargs) -> None:
```
nijel, line 169: The task cannot receive `Change` object, as it's not something you can safely serialize. It needs to accept ID and fetch it from a database. Or actually a list of IDs (see previous comment) to fetch all changes from the database and then process them.

weblate/addons/webhooks.py
```diff
@@ -0,0 +1,76 @@
...
+        config = self.instance.configuration
+        if change.action in config["events"]:
+            payload = self.build_webhook_payload(change)
+            webhook_delivery_task.delay(
+                url=config["webhook_url"],
+                headers=self.build_headers(change, payload),
+                data=payload,
+            )
+
+    def build_webhook_payload(self, change) -> dict:
```
nijel, line 32: I'd really prefer to base this on the JSON schema so that the payload content is formally defined. Maybe reusing the Fedora Messaging schema will be the way to go, but I really didn't investigate.

weblate/api/tasks.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def send_message(
+        self, *, url: str = "", headers: dict | None = None, data: dict | None = None
+    ) -> requests.Response:
+        headers = headers or {}
+        data = data or {}
+        try:
+            return requests.post(url, json=data, headers=headers, timeout=self.timeout)
+        except requests.exceptions.ConnectionError as error:
+            raise MessageNotDeliveredError from error
```
nijel, line 48: The failure should be added to the add-on log here.

weblate/trans/models/change.py
```diff
@@ -679,6 +680,79 @@ class Change(models.Model, UserDisplayMixin):
...
+        ACTION_REMOVE_CATEGORY: "category.removed",
+        ACTION_RENAME_CATEGORY: "category.renamed",
+        ACTION_MOVE_CATEGORY: "category.moved",
+        ACTION_SAVE_FAILED: "string.save.failed",
+        ACTION_NEW_UNIT_REPO: "repository.string",
+        ACTION_STRING_UPLOAD_UPDATE: "string.upload.updated",
+        ACTION_NEW_UNIT_UPLOAD: "string.upload.added",
+        ACTION_SOURCE_UPLOAD: "translation.source.upload.updated",
+        ACTION_COMPLETED_COMPONENT: "component.translation.completed",
+    }
```
nijel, line 754: This will be difficult to maintain. The whole ACTION_* should be migrated to `django.db.models.IntegerChoices` and then use just `.name` instead here. Can you please do that in a separate pull request? I've done such a change for some types already (for example d09307fa7da307bf7cf16f856b71137a41c26c6b, df27acdd811b5c5e8223674066ce8c486f0602b9, or d5531de8decbdce987f6c990438616c3a9f76e6d).

weblate/trans/models/change.py
```diff
@@ -1038,6 +1112,14 @@ def show_unit_state(self):
             self.ACTION_SUGGESTION_CLEANUP,
         }
 
+    def get_uuid(self) -> uuid.UUID:
+        """Return uuid for this change."""
+        return uuid.uuid5(uuid.NAMESPACE_OID, f"{self.action}.{self.id}")
```
nijel, line 1117: This is really not a OID. I think it's better to define the own namespace than using a standard one in a non-standard way.

weblate/addons/webhooks.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class WebhookAddon(ChangeBaseAddon):
+    name = "weblate.webhook.webhook"
+    # TODO: improve description
+    verbose = gettext_lazy("Webhooks")
+    description = gettext_lazy("some desc")
+
+    settings_form = WebhooksAddonForm
+    # TODO: find a webhook icon
```
nijel, line 20: We use MDI, there is webhook icon: https://pictogrammers.com/library/mdi/icon/webhook/, you need to download SVG and include it in static files.

weblate/api/tasks.py
```diff
@@ -0,0 +1,56 @@
...
+
+from weblate.utils.celery import app
+
+
+class MessageNotDeliveredError(Exception):
+    """Exception raised when a message could not be delivered."""
+
+
+class MessageDeliveryBaseTask(Task):
+    retry_backoff = True
```
nijel, line 16: I'm not sure that Celery retries are the best approach here, as it retries one second after the failure. That might be too short for being useful. The question is how to define an useful retry policy.

I think the best approach for the initial implementation is do not attempt retrying at all and we can figure out something later. Neither GitHub nor GitLab do retries on webhooks either (GitLab has https://gitlab.com/gitlab-org/gitlab/-/issues/355721 for that).

nijel (commented):
I think I've addressed most of the questions inline, the only remaining bit is signature. We need an implementation for verification anyway for testing this in the testsuite. Maybe there is some standard function we can use here instead of reimplementing it?

weblate/trans/models/change.py
gersona: - addressed by #14025

weblate/api/tasks.py
gersona: OK, I'll just get rid of the retry feature for now

gersona:
- waiting for #14024 to be merged for final submission

gersona:
some Screenshots
Capture d’écran 2025-03-04 à 18 31 06
Capture d’écran 2025-03-05 à 07 30 31

weblate/addons/tests.py
```diff
@@ -1577,3 +1585,302 @@ def test_json(self) -> None:
...
+
+    @responses.activate
+    def test_bulk_changes(self):
+        """Test bulk change create via the propagate() method."""
+        # create another component in project with same units as self.component
+        self.create_po(
+            new_base="po/project.pot", project=self.project, name="Component B1"
+        )
+        # listen to propagate change event
+        self.addon_configuration["events"].append(ActionEvents.PROPAGATED_EDIT)
```
nijel, line 1752: You modify a class variable here what might have undesired effects on other tests.

weblate/addons/tests.py
```diff
@@ -1577,3 +1585,302 @@ def test_json(self) -> None:
 class TasksTest(TestCase):
     def test_cleanup_addon_activity_log(self) -> None:
         cleanup_addon_activity_log()
+
+
+class WebhookAddonsTest(ViewTestCase):
+    """Test for Webhook Addon."""
+
+    addon_configuration = {
```
nijel, line 1593: Add type hint might catch some issues, but I don't think it will catch editing it, it would have to be combined with `Final` which is a Python 3.13 feature:

```suggestion
    addon_configuration : ClassVar[dict] = {
```

weblate/addons/forms.py
```diff
@@ -652,3 +662,32 @@ class PropertiesSortAddonForm(BaseAddonForm):
...
     )
+
+
+class ChangeBaseAddonForm(BaseAddonForm):
+    """Base form for Change-based addons."""
+
+    events = forms.MultipleChoiceField(
+        label=gettext_lazy("Change events"),
+        required=False,
+        widget=SortedCheckboxSelectMultiple(),
```
nijel, line 673: Maybe use `SortedSelectMultiple` instead? It might be a better UX to select the events.

weblate/addons/models.py
```diff
@@ -546,6 +548,23 @@ def store_post_load_handler(sender, translation: Translation, store, **kwargs) -
...
+    if created:  # ignore Change updates, they should not be updated anyway
+        addon_change(sender, [instance.pk], **kwargs)
+
+
+@receiver(change_bulk_create)
+def bulk_change_create_handler(sender, instances: list[Change], **kwargs) -> None:
+    """Handle Change bulk create signal."""
+    from weblate.addons.tasks import addon_change
+
+    addon_change(sender, [change.pk for change in instances], **kwargs)
```
gersona, line 565: @nijel I have a final issue I'd like to have your comments on.
Your [earlier comment](https://github.com/WeblateOrg/weblate/pull/14009#discussion_r1969207275) suggested to use a list of IDs instead of instances when calling a task.
Here, `instances`  is the result of `bulk_create`.
https://github.com/WeblateOrg/weblate/blob/4a551cb8c09567744e4c1f8fb583cfb590e476b7/weblate/trans/models/change.py#L241-L247
However, when `connection.features.can_return_rows_from_bulk_insert == False`, `addon_change` will just receive `[None]` because `bulk_create` don't set the `pk` (which is the case when using `mysql` in our tests).
What do you would be the best solution:
- adding a secondary primary key to the model that we can use to fetch instances from the database later (e.g an UUID, or anything that doesn't rely on autoincrement).
- pass a list of serialized instances instead of list of IDs (downside is that it may require to change a lot code downstream)

BTW, the same problem happens inside `dispatch_changes_notifications`.
https://github.com/WeblateOrg/weblate/blob/4a551cb8c09567744e4c1f8fb583cfb590e476b7/weblate/accounts/notifications.py#L102-L107

gersona: Or 3rd option would be to use a regular `save()` and make sure the changes don't propagate back.
Something like
```python
if not connection.features.can_return_rows_from_bulk_insert:
    changes = []
    for change in args[0]:
        change.save()
        changes.append(change)
```

nijel: Ah, I didn't realize that MySQL doesn't support this. The good thing is that nobody complained so far (the notifications are clearly broken for it since 05157d66523100f0b53fb71bec58b96b0ebbdd55).

This feature is already relied on in project backups, and we decided to simply not support MySQL there (documented at https://docs.weblate.org/en/latest/admin/backup.html#project-level-backups), but that's probably not an option here.

The best solution most likely would be to use `save()` inside `ChangeQuerySet.bulk_create()` when `can_return_rows_from_bulk_insert` is not available. This way it works even in this case and the performance penalty is limited to this scope.

The other alternative I see is dropping MySQL support, but that turned out to be not wanted by the community last time I tried (see https://github.com/WeblateOrg/weblate/issues/2825).

weblate/addons/models.py
```diff
@@ -553,7 +553,16 @@ def change_post_save_handler(sender, instance: Change, created, **kwargs) -> Non
...
+    ignore_list = []
+    if not connection.features.can_return_rows_from_bulk_insert:
+        # when condition is falsy, ActionEvents.PROPAGATED_EDIT changes
+        # are inidividually created in the ChangeQuerySet.bulk_create handler,
+        # and generate an unwanted post_save signal
+        ignore_list.append(ActionEvents.PROPAGATED_EDIT)
+
+    if (
+        created and instance.action not in ignore_list
+    ):  # ignore Change updates, they should not be updated anyway
```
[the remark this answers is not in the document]
gersona: That makes sense. I guess I started form the assumption that `post_save` signals may have different behavior from `post_bulk_save` signal.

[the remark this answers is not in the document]
nijel: `change_bulk_create` is Weblate specific signal to optimize handling of bulk changes. It is okay to fall back to Django native `post_save` if the optimization is not possible.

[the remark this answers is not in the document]
gersona: That makes sense. I guess I started from the assumption that behavior of `post_save` and `bulk_create` signal receivers might be different and that both necessarily need to be sent

docs/admin/addons.rst
```diff
@@ -959,6 +966,168 @@ Unused strings are removed, and new ones added as copies of the source string.
...
+                +-----------------+---------------+-------------------------------------------------+
+                | ``events``      | Change events | Available choices:                              |
+                |                 |               |                                                 |
+                |                 |               | ``0`` -- Resource updated                       |
+                |                 |               |                                                 |
+                |                 |               | ``1`` -- Translation completed                  |
+                |                 |               |                                                 |
+                |                 |               | ``2`` -- Translation changed                    |
+                |                 |               |                                                 |
+                |                 |               | ``5`` -- Translation added                      |
```
AliceVisek, line 989: Is there a reason why number 5 precedes 3 and 4 here? Similarly, number 28 seems misplaced as well.

nijel (approved):

docs/admin/addons.rst
nijel: It seems to mostly match order these are defined in `weblate/trans/actions.py`, but 28 is correctly placed there. https://github.com/WeblateOrg/weblate/pull/14501 should fix the source.

weblate/addons/tasks.py
```diff
@@ -176,3 +176,38 @@ def setup_periodic_tasks(sender, **kwargs) -> None:
         cleanup_addon_activity_log.s(),
         name="cleanup-addon-activity-log",
     )
+
+
+@app.task(trail=True)
+def addon_change(sender, change_ids: list[int], **kwargs) -> None:
```
nijel, line 182: This is our interface, so it should be strictly defined:

```suggestion
def addon_change(change_ids: list[int]) -> None:
```

nijel:
I've updated the documentation is 0fdbaa3855.

Issue #14019, opened 2025-02-26
Locale aware sorting crashes on macOS 15 with OSError


Jin-Shikai:
### Describe the issue

I start weblate from source code and it throws an OSError. It may be related to that the character set does not support non-Latin characters. I have written scripts to reproduce this.

### I already tried

- [x] I've read and searched [the documentation](https://docs.weblate.org/).
- [x] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

1. I started weblate from the source code as described in the documentation
2. Error occurs

### Expected behavior

not to crash

### Screenshots

<img width="1366" alt="Image" src="https://github.com/user-attachments/assets/2ba0b229-5096-4f94-95c9-4ef20791ec53" />

### Exception traceback

```pytb
Environment:


Request Method: GET
Request URL: http://localhost:8000/

Django Version: 5.1.6
Python Version: 3.13.2
Installed Applications:
['weblate.addons',
 'weblate.auth',
 'weblate.checks',
 'weblate.formats',
 'weblate.glossary',
 'weblate.machinery',
 'weblate.trans',
 'weblate.lang',
 'weblate_language_data',
 'weblate.memory',
 'weblate.screenshots',
 'weblate.fonts',
 'weblate.accounts',
 'weblate.configuration',
 'weblate.utils',
 'weblate.vcs',
 'weblate.wladmin',
 'weblate.metrics',
 'weblate',
 'weblate.gitexport',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django.contrib.admin',
 'django.contrib.sitemaps',
 'django.contrib.humanize',
 'social_django',
 'crispy_forms',
 'crispy_bootstrap3',
 'compressor',
 'rest_framework',
 'rest_framework.authtoken',
 'django_filters',
 'django_celery_beat',
 'corsheaders',
 'django_otp',
 'django_otp.plugins.otp_static',
 'django_otp.plugins.otp_totp',
 'django_otp_webauthn',
 'drf_spectacular',
 'drf_spectacular_sidecar',
 'drf_standardized_errors']
Installed Middleware:
['weblate.middleware.RedirectMiddleware',
 'weblate.middleware.ProxyMiddleware',
 'corsheaders.middleware.CorsMiddleware',
 'django.middleware.security.SecurityMiddleware',
[...]

PR #14023, opened 2025-02-26
feat(ci): run tests on macos

nijel:
This would allow to catch issues like #14019 earlier.


nijel:
Even after https://github.com/WeblateOrg/weblate/pull/14032 there is one failure in the Selenium tets.

PR #14025, opened 2025-02-26
Migrate Change.ACTION_* to IntegerChoices

### LINKED ISSUE #14024: Migrate Change.ACTION_* to IntegerChoices
The whole ACTION_* should be migrated to `django.db.models.IntegerChoices` and then use just `.name` instead here. Can you please do that in a separate pull request? I've done such a change for some types already (for example d09307fa7da307bf7cf16f856b71137a41c26c6b, df27acdd811b5c5e8223674066ce8c486f0602b9, or d5531de8decbdce987f6c990438616c3a9f76e6d).

_Originally posted by @nijel in https://github.com/WeblateOrg/weblate/pull/14009#discussion_r1969243047_

gersona:
- fixes #14024


weblate/accounts/notifications.py
```diff
@@ -507,11 +507,11 @@ def notify_monthly(self) -> None:
 @register_notification
 class RepositoryNotification(Notification):
     actions = (
-        Change.ACTION_COMMIT,
-        Change.ACTION_PUSH,
-        Change.ACTION_RESET,
-        Change.ACTION_REBASE,
-        Change.ACTION_MERGE,
+        Change.ACTIONS.ACTION_COMMIT,
```
nijel, line 510: Maybe we can remove the `ACTION_` prefix? That would be easier to read:

```suggestion
        Change.ACTIONS.COMMIT,
```

Maybe even directly use `ActionEvents` instead:

```suggestion
        ActionEvents.COMMIT,
```

What do you think?

weblate/trans/actions.py
```diff
@@ -0,0 +1,166 @@
...
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+
+from django.db.models import IntegerChoices
+from django.utils.translation import gettext_lazy, pgettext_lazy
+
+
+class ActionEvents(IntegerChoices):
+    # Translators: Name of event in the history
```
nijel, line 11: This needs to be before every line so that it is included in every translation string.

nijel (approved):
Looks good, will merge it after the release of 5.10.2 tomorrow.

weblate/trans/tests/test_edit.py
nijel, line 1: There are some unconverted Change.ACTION_* usages in this file, see the test failures.

nijel:

PR #14064, opened 2025-02-28
fix(utils): use strcoll for sorting strings instead of strxfrm

nijel:
This one works more reliably across platforms and typically better deals with some corner cases which cannot be handled by strxfrm.


nijel:
Okay, while `strcoll` doesn't fail on macos 15, it also doesn't sort properly, so this change brings no value.

PR #14088, opened 2025-03-03
fix(api): various small fixes for schema generation

walpox:
# Changelog

The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).

## Fixed

- [GroupViewSet]: Unable to derive type of path parameter "language_code" because model "weblate.auth.models.Group" contained no such field.
- [ComponentViewSet]: Unable to derive type of path parameter "project_slug" because model "weblate.trans.models.component.Component" contained no such field.
- [ComponentListViewSet]: Unable to derive type of path parameter "component_slug" because model "weblate.trans.models.componentlist.ComponentList" contained no such field.
- Annotated non-existent method on TasksViewSet.
- [TasksViewSet]: Unable to guess serializer.
- [Search]: Unable to guess serializer.

---

The purpose of this PR is to fix most the remaining error and warning log messages which are issued when a user sends a GET request to `/api/docs/`.

Relates to #12584.

### DIFF
--- a/weblate/api/spectacular.py
+++ b/weblate/api/spectacular.py
@@ -157,7 +157,7 @@ The OpenAPI specification is available as feature preview, feedback welcome!
             },
             {
                 "name": "tasks",
-                "description": "Added in version 4.4.",
+                "description": "Added in version 4.4.\n\nListing of the tasks is currently not available.",
             },
             {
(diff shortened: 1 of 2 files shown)


nijel:

PR #14144, opened 2025-03-09
Support sorting in translation reports

### LINKED ISSUE #13745: Customizable ordering of reports
### Describe the problem

Our project has about 100 translators each year. We want to follow up who makes a lot of translations to give them a shout out (or even swag) and to monitor our translations. When a language reaches 95% translated, it gets in to production. If it drops beneath 90% we give it a beta label, and at 80% we give it a alpha label.
For the moment I run my own scripts to get stats and store them in some [markdown pages](https://github.com/ctlaltdieliet/mattermost-i18n-scoreboard/tree/main/pages).

I used to translate on Pootle. As a translator you see your current ranking (week/month?/overall?) and it is a kind of gamification that encourages you to be the best translator of the project for that time frame.

When you download a language file and upload it again and mark the strings as translated, the current stats take all the strings in account (even if you only made a few changes) and not the effective changes.

I'd like to get an overview of the top translators for certain time frames.
I want to follow up how translations for each language evolve during time

Running your own scripts 

### Screenshots

_No response_

### Additional context

Thank you for being awesome!

amCap1712:
Allow sorting reports by the author's date_joined field and edit count in ascending or descending order.

Fixes #13745.


weblate/trans/tests/test_reports.py
```diff
@@ -384,6 +415,248 @@ def test_counts_inverse_daterange(self) -> None:
...
+        user3 = User.objects.create(
+            username="customtestuser3",
+            email="custom-weblate-3@example.org",
+            password=TESTPASSWORD,
+            is_superuser=True,
+            full_name="Weblate Test 3",
+            date_joined=datetime(2025, 3, 1, tzinfo=UTC),
+        )
+
+        self.client.login(username=user1.username, password="testpassword")
```
nijel, line 456: Should use the same constant as when creating:

```suggestion
        self.client.login(username=user1.username, password=TESTPASSWORD)
```

… the same applies to other occurrences below.

nijel (commented):
Can you please also adjust documentation (mention this in changes and extend reports docs to cover this)?

weblate/trans/tests/test_reports.py
amCap1712: Hi! I tried this but the login fails in that case because `self.client.login` requires the plaintext password whereas `TESTPASSWORD` is already hashed. I also checked other uses of `self.client.login` in the tests to confirm and they are using the plaintext version as well.

amCap1712:
Added documentation and the changelog entry.

weblate/trans/views/reports.py
```diff
@@ -364,9 +364,10 @@ def get_counts(request: AuthenticatedHttpRequest, path=None):
     else:
         start = "{0}\n{1} {2}\n{0}".format(
             RST_HEADING,
-            " ".join(f"{h:40}" for h in headers[:2]),
-            " ".join(f"{h:24}" for h in headers[2:]),
+            " ".join(f"{h:40}" for h in headers[:3]),
+            " ".join(f"{h:24}" for h in headers[3:]),
         )
+        print(start)
```
nijel, line 370: Forgotten debugging code?

amCap1712: my bad, fixed.

AliceVisek (approved):

nijel:

PR #14168, opened 2025-03-11
chore: use binary wheels for lxml/xmlsec

nijel:
These should be again compatible.


nijel:
Okay, this is premature. https://github.com/xmlsec/python-xmlsec/releases/tag/1.3.15 is there, but it's not yet on PyPI.

nijel:
Still not yet fully there: https://github.com/xmlsec/python-xmlsec/issues/344

Issue #13896, opened 2025-02-17
Comma in rendered lists is not localized


nijel:
### Describe the problem

When several values are rendered and joined with `,`, it is not localized. Since https://github.com/WeblateOrg/weblate/commit/0b54da635d614770f53919f469cdfefb6a511abe there is a solution for `format_html_join`, but there are still dozen of places where `", ".join(...)` is used.

### Describe the solution you would like

Use `format_html_join_comma` whenever the content is shown in the UI.

nijel:
Several issues were addressed, but there are more in the code base. Following pull requests can be examples how to address this:

* https://github.com/WeblateOrg/weblate/pull/13949
* https://github.com/WeblateOrg/weblate/pull/13948
* https://github.com/WeblateOrg/weblate/pull/13905

Jin-Shikai:
Hi, I'm new to open source projects. May I give it a try? I'd appreciate if you could tell me where similar problems might occur.

nijel:
Just search for `", ".join(`. When that is used in something shown to the user in UI, it is a good candidate to replace.

PR #14191, opened 2025-03-12
Replace ", ".join(...) with format_html_join_comma for proper localization

KasukabeDefenceForce:
Replaced instances of `", ".join(...) `with `format_html_join_comma`, ensuring proper localization when rendering UI content.

Fixes https://github.com/WeblateOrg/weblate/issues/13896


weblate/checks/flags.py
```diff
@@ -283,7 +285,9 @@ def _format_values(self):
         return (self.format_flag(item) for item in self._items.values())
 
     def format(self) -> str:
-        return ", ".join(sorted(self._format_values()))
+        return format_html_join_comma(
+            "{}", ((mark_safe(flag),) for flag in sorted(self._format_values()))
```
KasukabeDefenceForce, line 289: Ruff wants to remove mark_safe() but the tests fail if I remove it because the quotes get html escaped. Is it safe to use mark_safe here?

weblate/billing/models.py
```diff
@@ -249,7 +252,7 @@ def all_projects(self):
 
     @cached_property
     def projects_display(self):
-        return ", ".join(str(x) for x in self.all_projects)
+        return format_html_join_comma("{}", ((str(x),) for x in self.all_projects))
```
nijel, line 255: `str` is not needed here, format does it anyway:

```suggestion
        return format_html_join_comma("{}", ((x,) for x in self.all_projects))
```

weblate/billing/models.py
```diff
@@ -624,7 +627,7 @@ def clean(self) -> None:
 
         if overlapping.exists():
             msg = "Overlapping invoices exist: {}".format(
-                ", ".join(str(x) for x in overlapping)
+                format_html_join_comma("{}", ((str(x),) for x in overlapping))
```
nijel, line 630:
```suggestion
                format_html_join_comma("{}", ((x,) for x in overlapping))
```

weblate/checks/flags.py
nijel: This one should not be localized at all as it used to serialize flags into the database.

weblate/accounts/views.py
```diff
@@ -1551,7 +1552,9 @@ def saml_metadata(request: AuthenticatedHttpRequest):
     if errors:
         add_breadcrumb(category="auth", message="SAML errors", errors=errors)
         report_error("SAML metadata", level="error")
-        return HttpResponseServerError(content=", ".join(errors))
+        return HttpResponseServerError(
+            content=format_html_join_comma("{}", ((err,) for err in errors))
+        )
```
nijel, line 1557: As this is a server error, localizing is not that useful.

weblate/lang/models.py
```diff
@@ -1098,7 +1099,7 @@ def get_plural_label(self, idx):
         return format_html(
             PLURAL_TITLE,
             name=self.get_plural_name(idx),
-            examples=", ".join(self.examples.get(idx, [])),
+            examples=format_html_join_comma("{}", self.examples.get(idx, [])),
```
nijel, line 1102: Conversion to tuple is missing here?

weblate/lang/models.py
```diff
@@ -1118,7 +1119,7 @@ def list_plurals(self):
             yield {
                 "index": i,
                 "name": self.get_plural_name(i),
-                "examples": ", ".join(self.examples.get(i, [])),
+                "examples": format_html_join_comma("{}", self.examples.get(i, [])),
```
nijel, line 1122: Conversion to tuple is missing here?

weblate/trans/admin.py
```diff
@@ -84,8 +85,8 @@ class ProjectAdmin(WeblateModelAdmin, RepoAdminMixin):
 
     @admin.display(description=gettext_lazy("Administrators"))
     def list_admins(self, obj):
-        return ", ".join(
-            User.objects.all_admins(obj).values_list("username", flat=True)
+        return format_html_join_comma(
+            "{}", (User.objects.all_admins(obj).values_list("username", flat=False))
```
nijel, line 89: `flat` defaults to false, make it simpler:

```suggestion
            "{}", User.objects.all_admins(obj).values_list("username")
```

weblate/trans/forms.py
```diff
@@ -2535,7 +2535,7 @@ def get_glossary_flags(self) -> str:
             result.append("forbidden")
         if self.cleaned_data.get("read_only"):
             result.append("read-only")
-        return ", ".join(result)
+        return format_html_join_comma("{}", ((flag,) for flag in result))
```
nijel, line 2538: This is used for serialization of flags, needs to stick with comma.

weblate/trans/models/component.py
```diff
@@ -2951,7 +2952,7 @@ def clean_new_lang(self) -> None:
         if errors:
             message = gettext(
                 "Could not parse base file for new translations: %s"
-            ) % ", ".join(str(error) for error in errors)
+            ) % format_html_join_comma("{}", ((str(error),) for error in errors))
```
nijel, line 2955: str is not needed:

```suggestion
            ) % format_html_join_comma("{}", ((error,) for error in errors))
```

weblate/trans/templatetags/translations.py
```diff
@@ -968,15 +969,18 @@ def unit_state_title(unit) -> str:
         state.append(
             "{} {}".format(
                 pgettext("String state", "Failing checks:"),
-                ", ".join(str(check) for check in checks),
+                format_html_join_comma(
+                    "{}",
+                    ((str(check),) for check in checks),
```
nijel, line 974:
```suggestion
                    ((check,) for check in checks),
```

weblate/trans/templatetags/translations.py
```diff
@@ -968,15 +969,18 @@ def unit_state_title(unit) -> str:
...
+                ),
             )
         )
     checks = unit.dismissed_checks
     if checks:
         state.append(
             "{} {}".format(
                 pgettext("String state", "Dismissed checks:"),
-                ", ".join(str(check) for check in checks),
+                format_html_join_comma("{}", ((str(check),) for check in checks)),
```
nijel, line 983:
```suggestion
                format_html_join_comma("{}", ((check,) for check in checks)),
```

weblate/trans/views/edit.py
```diff
@@ -78,7 +79,7 @@ def display_fixups(request: AuthenticatedHttpRequest, fixups) -> None:
     messages.info(
         request,
         gettext("Following fixups were applied to translation: %s")
-        % ", ".join(str(f) for f in fixups),
+        % format_html_join_comma("{}", ((str(f),) for f in fixups)),
```
nijel, line 82:
```suggestion
        % format_html_join_comma("{}", ((f,) for f in fixups)),
```

weblate/trans/views/edit.py
```diff
@@ -414,7 +415,11 @@ def perform_translation(unit, form, request: AuthenticatedHttpRequest) -> bool:
             gettext(
                 "The translation has been saved, however there "
                 "are some newly failing checks: {0}"
-            ).format(", ".join(str(CHECKS[check].name) for check in newchecks)),
+            ).format(
+                format_html_join_comma(
+                    "{}", ((str(CHECKS[check].name),) for check in newchecks)
```
nijel, line 420:
```suggestion
                    "{}", ((CHECKS[check].name,) for check in newchecks)
```

nijel (commented):
Thanks, it goes in the correct direction. I've made some comments on places I see problematic. 

Maybe introducing an additional function that would accept a list and convert it to a list of tuples would be useful as I see this pattern repeats frequently.

weblate/lang/models.py
KasukabeDefenceForce: The format_html_join function spreads the elements of the lists(as in this case). We need a tuple when there is a string.

KasukabeDefenceForce:
> Thanks, it goes in the correct direction. I've made some comments on places I see problematic.
> 
> Maybe introducing an additional function that would accept a list and convert it to a list of tuples would be useful as I see this pattern repeats frequently.

Should I add a check inside the format_html_join_comma to check if  the element is a string then convert it to a tuple?

weblate/billing/models.py
```diff
@@ -203,10 +205,13 @@ class Meta:
     def __str__(self) -> str:
         projects = self.projects_display
         owners = self.owners.order()
+        owners_visible_names_tuples = list_to_tuples([x.get_visible_name() for x in owners])
         if projects:
             base = projects
         elif owners:
-            base = ", ".join(x.get_visible_name() for x in owners)
+            base = format_html_join_comma(
+                "{}", (x for x in owners_visible_names_tuples)
```
KasukabeDefenceForce, line 215: I have implemented the util function i created here. We would have to create an extra variable for some cases. Is this ok with you? Currently, I have only implemented it in this file. If this looks good I would apply it elsewhere as well.

weblate/trans/util.py
```diff
@@ -395,3 +395,8 @@ def count_words(string: str, language: Language | None = None) -> int:
                 even = not even
         return count
     return sum(len(s.split()) for s in split_plural(string) if not is_unused_string(s))
+
+
```
KasukabeDefenceForce, line 399: Is this the correct file for this function?

weblate/billing/models.py
```diff
@@ -203,10 +205,15 @@ class Meta:
...
+        owners_visible_names_tuples = list_to_tuples(
+            [x.get_visible_name() for x in owners]
+        )
         if projects:
             base = projects
         elif owners:
-            base = ", ".join(x.get_visible_name() for x in owners)
+            base = format_html_join_comma(
+                "{}", (x for x in owners_visible_names_tuples)
+            )
```
nijel, line 216: The additional comprehension here is not needed at all:

```suggestion
            base = format_html_join_comma(
                "{}", owners_visible_names_tuples
            )
```


But keeping it fully inline might be even easier to understand:

```suggestion
            base = format_html_join_comma(
                "{}", list_to_tuples(x.get_visible_name() for x in owners)
            )
```

weblate/billing/models.py
```diff
@@ -249,7 +256,9 @@ def all_projects(self):
 
     @cached_property
     def projects_display(self):
-        return ", ".join(str(x) for x in self.all_projects)
+        return format_html_join_comma(
+            "{}", (x for x in list_to_tuples(self.all_projects))
```
nijel, line 260:
```suggestion
            "{}", list_to_tuples(self.all_projects)
```

...and pretty much anywhere else.

weblate/trans/models/component.py
```diff
@@ -3491,6 +3492,7 @@ def load_intermediate_store(self):
             source_language=self.source_language.code,
         )
         if self.pk:
+            print(store)
```
nijel, line 3495: Debugging statements?

weblate/trans/util.py
```diff
@@ -395,3 +395,8 @@ def count_words(string: str, language: Language | None = None) -> int:
...
+def list_to_tuples(strings: list[str]) -> list[tuple]:
```
nijel, line 400: Make the type annotations more generic on input and specific on output:

```suggestion
def list_to_tuples(strings: Iterable[str]) -> list[tuple[str]]:
```

weblate/trans/util.py
nijel: Yes, it could be placed here.

nijel: Looking again at it, placing it next to `format_html_join_comma` might be a better fit.

weblate/billing/models.py
```diff
@@ -256,9 +253,7 @@ def all_projects(self):
 
     @cached_property
     def projects_display(self):
-        return format_html_join_comma(
-            "{}", (x for x in list_to_tuples(self.all_projects))
-        )
+        return format_html_join_comma("{}", list_to_tuples(x for x in self.all_projects))
```
nijel: These no-op expressions are not really needed (more of these are present):

```suggestion
        return format_html_join_comma("{}", list_to_tuples(self.all_projects))
```

nijel:

Issue #13851, opened 2025-02-13
Dark theme button colors


nijel:
### Describe the issue

The dark theme currently uses the same colors for buttons as the light one. This is wrong, as the different background makes the default button (dark) less visible than the alternate ones (green). The current colors work well in the light theme.

### I already tried

- [x] I've read and searched [the documentation](https://docs.weblate.org/).
- [x] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

1. Go to translate page.
2. Toggle between dark and light mode.

### Expected behavior

The primary button should be more visible in the dark theme. Maybe we can just switch the two colors for buttons?

### Screenshots

![image](https://github.com/user-attachments/assets/651147ee-9b3a-43e6-a876-adf64b72761b)
(I click skip 1 out of 10 times. Guaranteed)

### Exception traceback

```pytb

```

### How do you run Weblate?

weblate.org service

### Weblate deploy checks

```shell

```

### Additional context

Previous discussions:

* https://github.com/WeblateOrg/weblate/pull/13065
* https://github.com/WeblateOrg/weblate/issues/13050
* https://github.com/WeblateOrg/weblate/pull/13065
* https://github.com/WeblateOrg/weblate/pull/13084

meel-hd:
I'm on it!

PR #14201, opened 2025-03-13
Switch button colors in dark theme

meel-hd:
close: #13851
## Preview
image


nijel:
Can't we modify the CSS for `btn-warning`/`btn-primary` in the dark theme instead of complicating the template?

meel-hd:
We can but that would change it globally in the app. Is that what we want?

meel-hd:
> in the dark theme instead of complicating the template?

@nijel  Done!

nijel:
This makes all buttons green, not switching them:

image

meel-hd:
@nijel no I just checked it seems working fine for me as in the screenshot above. Check if your instance have any customization to colors.

weblate/static/style-dark.css
```diff
@@ -678,7 +678,7 @@ fieldset[disabled] .btn-default.focus {
 }
 .btn-primary {
   color: #fffdfa;
-  background-color: #25303b;
+  background-color: #2eccaa;
```
nijel, line 681: Followup to the previous comment: I don't think it's related to caching. How does a single color change in CSS should change color of two buttons?

meel-hd: The primary button is changed simply with css. But the other button color can be customized by the user. Using the `Focus color` option in color customization.

So I changed the default `Focus color`  in `weblate/wladmin/forms.py:106`. 

It should update and rebuild the custom.css but it doens't.

meel-hd: > I don't think it's related to caching

It is. When I commented the code respnosible for cashing the custom colors, it worked as I intended. 

The main problem is the cache doesn't update if we updated the default custom colors.

nijel: The problem is unrelated to custom colors, the buttons have the same color in the dark theme even when CSS customization is not turned on:

image

meel-hd: @nijel  True, I misunderstood how everything works together. I only updated the default color in the code which  doesn't get updated if the db is already populated before , This is fixed now by a migration.

nijel: What I describe happens when there is *no* CSS customization turned on.

meel-hd: The color of the secondary buttons is changed by `focus_color`  from color customazation even if we have'nt configured it, it takes the default value. That's why I changed the default color.

meel-hd: Oh ok, I see what you mean. You are right. Thanks for you patience. It is fixed now.

nijel (commented):
`:focus` and `:hover` selectors should be IMHO updated as well.

meel-hd:
Yeah, true. I will do that

nijel:
Did you intentionally skip `:focus` selectors?

nijel:

PR #14207, opened 2025-03-13
setuptools: use find_namespace_packages

### LINKED ISSUE #12778: Package would be ignored warnings
### Describe the problem

During the package build there are several warnings like:

```
        ********************************************************************************
        ############################
        # Package would be ignored #
        ############################
        Python recognizes 'weblate.templates.memory' as an importable package[^1],
        but it is absent from setuptools' `packages` configuration.

        This leads to an ambiguous overall configuration. If you want to distribute this
        package, please make sure that 'weblate.templates.memory' is explicitly added
        to the `packages` configuration field.

        Alternatively, you can also rely on setuptools' discovery methods
        (for example by using `find_namespace_packages(...)`/`find_namespace:`
        instead of `find_packages(...)`/`find:`).

        You can read more about "package discovery" on setuptools documentation page:

        - https://setuptools.pypa.io/en/latest/userguide/package_discovery.html

        If you don't want 'weblate.templates.memory' to be distributed and are
        already explicitly excluding 'weblate.templates.memory' via
        `find_namespace_packages(...)/find_namespace` or `find_packages(...)/find`,
        you can try to use `exclude_package_data`, or `include-package-data=False` in
        combination with a more fine grained `package-data` configuration.

        You can read more about "package data files" on setuptools documentation page:

        - https://setuptools.pypa.io/en/latest/userguide/datafiles.html


        [^1]: For Python, any directory (with suitable naming) can be imported,
              even if it does not contain any `.py` files.
              On the other hand, currently there is no concept of package data
              directory, all directories are treated like packages.
        ********************************************************************************
```

The warnings should be addressed. I think in all cases these are data files.

_No response_

### Screenshots

_No response_

### Additional context

* https://setuptools.pypa.io/en/latest/userguide/package_discovery.html
* https://setuptools.pypa.io/en/latest/userguide/datafiles.html

amCap1712:
When building the weblate package, there are several "Package would be ignored" errors. See https://github.com/WeblateOrg/weblate/issues/12778 for details.

These errors stem from the fact that any directory even if it doesn't contain python files can be imported in a python file as a namespace package. The template and local directories included in the weblate build as package data qualify as namespace packages. However, since find_namespace_packages was set to false setuptools wouldn't consider it as importable package leading to an inconsistent configuration.

To fix this issue, set find_namespace_packages to true in pyproject.toml. Since weblate is a "flat layout", the include directive is also needed to exclude other top level directories from being included in the final package.

To test this change, I extracted the source dist and wheels produced before this change and compared them to their respective versions produced after the change. The wheels comparison produced no diff. For source dist, the diff was:
```
567,569c567,568
< [tool.setuptools.packages.find]
< namespaces = true
< include = ["weblate*"]
---
> [tool.setuptools.packages]
> find = {namespaces = false}
```
which is expected.

Fixes #12778

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -565,8 +565,8 @@ license-files = []
 zip-safe = false
 
 [tool.setuptools.packages.find]
-namespaces = true
 include = ["weblate*"]
+namespaces = true
 
 [tool.tomlsort]
 ignore_case = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -564,8 +564,9 @@ include-package-data = true
 license-files = []
 zip-safe = false
 
-[tool.setuptools.packages]
-find = {namespaces = false}
+[tool.setuptools.packages.find]
+namespaces = true
+include = ["weblate*"]
 
 [tool.tomlsort]
 ignore_case = true


nijel:

PR #14212, opened 2025-03-14
Add an api to get all string translations

### LINKED ISSUE #13186: API to get all string translations
### Describe the problem

There is no way to get all the translated units for a source unit in the API.

Introduce an additional endpoint for unit (tentatively `/api/units/(int: id)/translations/`) that would work for source units only and would include links to all the translation units.

I was also considering including this directly in the unit serializer, but I think this would ruin the unit listing performance, or we would have to use a different serializer for listing and details. That is certainly a viable approach as well.

### Screenshots

_No response_

### Additional context

https://github.com/WeblateOrg/weblate/discussions/13146

amCap1712:
Added a translations action to UnitViewSet to retrieve all target source units for a given source unit. A 400 error is returned if the specified unit is not a source unit.

Fixes #13186

### DIFF
--- a/weblate/api/views.py
+++ b/weblate/api/views.py
@@ -40,6 +40,7 @@ from rest_framework.status import (
     HTTP_204_NO_CONTENT,
     HTTP_423_LOCKED,
     HTTP_500_INTERNAL_SERVER_ERROR,
+    HTTP_400_BAD_REQUEST,
 )
 from rest_framework.utils import formatting
 from rest_framework.views import APIView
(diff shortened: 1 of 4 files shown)


weblate/api/views.py
```diff
@@ -1971,6 +1978,23 @@ def destroy(self, request: Request, *args, **kwargs):
             )
         return Response(status=HTTP_204_NO_CONTENT)
 
+    @action(detail=True, methods=["get"])
+    def translations(self, request: Request, *args, **kwargs):
+        unit = self.get_object()
+        user = request.user
+        user.check_access_component(unit.translation.component)
+
+        if unit.pk != unit.source_unit.pk:
```
nijel, line 1987: Avoids additional SQL query:

```suggestion
        if unit.pk != unit.source_unit_pk:
```

AliceVisek (approved):

weblate/api/views.py
nijel, line 1987: Use a generic interface for this (which also avoids one additional SQL query): 

```suggestion
        if unit.is_source:
```

Issue #14101, opened 2025-03-04
Highlight string page number on click


Cwpute:
### Describe the problem

When you're translating and want to hop to a string page you remember the number of, you've got click once on the page number, and then you have to manually delete the digits and replace them with the desired number. It's a small grievance, but it can add up pretty quickly and feels unnecessarily cumbersome.

### Describe the solution you would like

When the user clicks on the string page number, make it so the input area's number is automatically highlighted so that you can type right away without having to click again.

### Screenshots

My screenshot app unexplicably crashes these days so i'm not able to provide any.

nijel:
I believe this should be already addressed by https://github.com/WeblateOrg/weblate/pull/13816, what version do you use?

nijel:
Ah, #13816 fixed focus, but indeed does not select the input.

meel-hd:
I'm on it

PR #14271, opened 2025-03-19
Highlight string page number on click

meel-hd:
close: #14101


nijel:

Issue #14263, opened 2025-03-19
Althca border  radius does not match input


nijel:
### Describe the issue

The altcha border radius matches panel one, not the input one:

![Image](https://github.com/user-attachments/assets/e319424a-410f-4bde-b169-5cb964fe2ee2)

### I already tried

- [x] I've read and searched [the documentation](https://docs.weblate.org/).
- [x] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

1. Do not authenticate.
2. Go to https://hosted.weblate.org/contact/

### Expected behavior

The altcha input should look the same as other inputs.

### Exception traceback

```pytb

```

### How do you run Weblate?

weblate.org service

### Weblate deploy checks

```shell

```

meel-hd:
On it 👍

PR #14283, opened 2025-03-20
Fix altcha input border radius

meel-hd:
### The fix
image
close: #14263


nijel:

Issue #14299, opened 2025-03-21
Consecutive duplicate word check should be disabled for Toki Pona


Math-Bee:
### Describe the problem

Word repetition happens a lot in toki pona (as shown in examples below), and is very likely to be on purpose rather than on accident.
Examples where duplicate words happen:
- Numbers: toki pona only has 6 words for numbers (0, 1, 2, 5, 20, and 100), so in order to make larger numbers like 3 or 74, it just combines existing number words (3 is "tu wan", lit. "2 1"). Because of this, a lot of numbers end up having to duplicate words. Ex:
	-  4 is "tu tu" (2 2)
	- 11 is "luka luka wan" (5 5 1)
	- any number 40-99 is "mute mute [stuff]" (20 20 stuff)
- Double meanings: toki pona has some words that have multiple meanings. A good example is "tawa", which is means "to/towards" (preposition) and also "movement" (content word). So, a sentence like "o pana e sitelen tawa tawa mi", lit. "Give the 'moving image' (video, animation) to me", would trigger the check. Non-prepositions can have similar problems, although it's a bit less common
- Emphasis: You can reduplicate some words to strengthen their meaning. Ex:
	- mama = parent, mama mama = grandparent
	- mute = a lot, >3, or 20; mute mute = very very much, 40

With all of these, the check is way more likely to trigger when duplicate word use is intentional than when it's not. It's quite annoying to dismiss the check every time, and possibly nonsensical to put a separator (like a comma) in between the duplicated words

### Describe the solution you would like

Disabling the Consecutive Word Check for toki pona [tok]

### Describe alternatives you have considered

There could be a whitelist for some toki pona words, but this would exclude cases of emphasis, and would be harder than just disabling it for this language

nijel:
We can disable the check completely or add a list of allowed repeated words for this language. We already apply per-language word exceptions for several languages.

PR #14320, opened 2025-03-23
feat: Disable consecutive duplicate word check for Toki Pona (#14299)

imShubh03:
### Description  
Toki Pona frequently uses repeated words intentionally in various contexts, making the consecutive duplicate word check unnecessary. This PR disables the check specifically for Toki Pona (`tok`) to prevent false positives.  

### Problem  
The duplicate word check is not suitable for Toki Pona because:  
- **Numbers**: Larger numbers require repetition (e.g., `tu tu` = 4, `luka luka wan` = 11).  
- **Double Meanings**: Some words are used in different contexts, leading to valid repetition (e.g., `tawa tawa` = "moving image").  
- **Emphasis**: Words are intentionally repeated to strengthen meaning (e.g., `mute mute` = "very much").  

This results in unnecessary warnings and requires users to manually dismiss them frequently.  

### Solution  
- Disable the consecutive duplicate word check for Toki Pona by adding `tok` to the `IGNORES` dictionary.  

### Alternative Considered  
- A whitelist for specific repeated words was considered but would not cover all valid cases, making a full exclusion more effective.  

### Related Issue  
Fixes: #14299

### DIFF
--- a/weblate/checks/duplicate.py
+++ b/weblate/checks/duplicate.py
@@ -28,6 +28,7 @@ IGNORES = {
     "hi": {"कर"},
     "tr": {"tek", "adım", "gıcır", "sık"},
     "sq": {"të"},
+    "tok": set(),
 }
 
 


nijel (changes requested):
Your change will not change the behavior at all. It will use and empty set to ignore workds in Toki Pona.

The best way to do this would be `should_skip` method:

https://github.com/WeblateOrg/weblate/blob/a3e112faf896bbbd8c813023e2375c1ba9b1eca4/weblate/checks/placeholders.py#L151-L154

You can then check for the language using `unit.translation.language.is_base({"tok"})`.

Adding a testcase will help you show that the fix is actually effective.

Issue #14260, opened 2025-03-19
Describe priority of automatic translation sources


orangesunny:
### Describe the problem

The priority of multiple MT sources set in AT add-on is defined in the code, but is not described in the documentation.

### Describe the solution you would like

@nijel, please describe the current status here in the comment, and we will include it in the docs with @AliceVisek.

orangesunny:
@AliceVisek, it should be probably placed in https://docs.weblate.org/en/latest/admin/machine.html or in https://docs.weblate.org/en/latest/admin/addons.html#automatic-translation and mentioned in a hint/note respectively.

tomkolp:
What does the translation selection logic look like if the source is a translation memory and the source text has several 100% translations?

nijel:
If there are several 100% matches, there is no logic in the selection. So the first one returned by the database will be used.

AliceVisek:
> * Each engine has defined maximal score in can produce in the code for example: 
>     
>       
>         [weblate/weblate/machinery/deepl.py](https://github.com/WeblateOrg/weblate/blob/0de33809acd0fc6936c2629279def88bab8207e9/weblate/machinery/deepl.py#L34)
>       
>       
>            Line 34
>         in
>         [0de3380](/WeblateOrg/weblate/commit/0de33809acd0fc6936c2629279def88bab8207e9)
>       
>     
>     
>       
>   
>           
>             
>              max_score = 91

@nijel how can I extract the max_score for each translation service into the documentation? I' like to add the number into the description of each service, alternatively it can also be done in one table.

nijel:
https://github.com/WeblateOrg/weblate/pull/14342 adds that.

PR #14322, opened 2025-03-23
doc: describe priority of automatic translations

AliceVisek:
Added description of priority of automatic translation sources, addresses issue #14260


docs/admin/machine.rst
```diff
@@ -28,6 +28,14 @@ Per-project automatic suggestion can also be configured via the :ref:`api`.
 
    :ref:`machine-translation`
 
+.. _mt-translation-services-priority:
+
+Priority of machine translation and translation memory services
+----------------------------------------------------------------
+:ref:`_mt-weblate-translation-memory` matches with 100% score take priority over machine translation services. If 100% match in translation memory is found, no machine translation is performed. If several 100% matches occur, the first one returned by the database is used.
```
nijel, line 35: Fix the reference:

```suggestion

:ref:`mt-weblate-translation-memory` matches with 100% score take priority over machine translation services. If 100% match in translation memory is found, no machine translation is performed. If several 100% matches occur, the first one returned by the database is used.
```

The labels are defined as 

```
.. _LABEL:
``` 

and then used as 
```
:ref:`LABEL`
```

See https://www.sphinx-doc.org/en/master/usage/referencing.html#ref-role

docs/changes.rst
```diff
@@ -15,6 +15,7 @@ Weblate 5.11
 * Widgets show more precise stats.
 * :ref:`upload` is now tracked in history with details.
 * Changes in string flags are now tracked in history.
+* :doc:`machine-translation-setup` documentation extended.
```
nijel, line 18:
```suggestion
* :doc:`/admin/machine` documentation extended.
```

Or:
 
```
:ref:`machine-translation-setup`
```

Issue #13051, opened 2024-11-16
Introduce widget/badge indicating number of languages


keunes:
### Describe the problem

We already have a widget indicating the percentage of completion. That doesn't allow us to show off how many languages we support though.

### Describe the solution you would like

Introduce a widget that shows the number of languages activated on a given project/component.

It'd be even cooler if we could indicate a threshold, similar to #3436 (e.g. count number of languages with 40% completion or higher).

nijel:
What we also need is to restructure the widgets page and have a standard way to parametrize them.

PR #14326, opened 2025-03-24
Revamp widget parameterization

amCap1712:
Based on https://github.com/WeblateOrg/weblate/issues/13051#issuecomment-2481236851, this PR explores a new way to parameterize widgets. Here is the what the proposed UI looks like:

image

The widget, language, component, code, and color options are available for all widgets. All the extra options per widget are defined in the `extra_parameters` attribute of the widget class and passed through JSON where jQuery reads it and renders it. Using the options, jQuery generates the urls dynamically, and updates the live preview and the code to embed the images. 

Also, this is just a draft to start discussion.

### DIFF
--- a/weblate/middleware.py
+++ b/weblate/middleware.py
@@ -72,6 +72,7 @@ CSP_DIRECTIVES: CSP_TYPE = {
 INLINE_PATHS = {
     "social:begin",
     "djangosaml2idp:saml_login_process",
+    "widgets"
 }
 
 
--- a/weblate/static/js/widgets.js
+++ b/weblate/static/js/widgets.js
@@ -14,53 +14,60 @@ $(document).ready(function() {
   $("#component").val(jsonComponent);
 
   function updateLivePreviewAndEmbedCode() {
-    const widgetName = $("#widget_type").val();
-    const componentId = $("#component").val();
-    const component = widgetsData["components"].find(component => String(component.id) === componentId);
+    const widgetName = $('#widget_type').val();
+    const componentId = $('#component').val();
+    const component = widgetsData['components'].find(c => String(c.id) === componentId);
+    const language = $('#translation_language').val();
+    const widget = widgetsData['widgets'][widgetName];
+    const color = $('#color_select').val();
 
-    const language = $("#translation_language").val();
-
-    const widget = widgetsData["widgets"][widgetName];
-    const color = $("#color_select").val();
-
-    let engageUrl = widgetsData["engage_base_url"];
-
-    const widgetBaseUrl = widgetsData["widget_base_url"];
+    let engageUrl = widgetsData['engage_base_url'];
+    const widgetBaseUrl = widgetsData['widget_base_url'];
     let widgetUrl = `${widgetBaseUrl}`;
-    if (component !== undefined && language !== "") {
+    if (component !== undefined && language !== '') {
       widgetUrl = `${widgetUrl}/${component.slug}/${language}`;
-    } else if (component !== undefined && language === "") {
+    } else if (component !== undefined && language === '') {
       widgetUrl = `${widgetUrl}/${component.slug}`;
-    } else if (component === undefined && language !== "") {
+    } else if (component === undefined && language !== '') {
       widgetUrl = `${widgetUrl}/-/${language}`;
     }
 
-    const newUrl = `${widgetUrl}/${widgetName}-${color}.${widget.extension}`;
-    $("#widgetImage").attr("src", newUrl);
+    // Include extra parameters in the URL
+    const params = new URLSearchParams();
+    $('#extra-parameters input').each(function() {
+      const paramName = $(this).attr('name');
+      const paramValue = $(this).val();
+      if (paramValue) {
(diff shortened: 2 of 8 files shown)


weblate/middleware.py
```diff
@@ -69,10 +69,7 @@
 }
 
 # URLs requiring inline javascript
-INLINE_PATHS = {
-    "social:begin",
-    "djangosaml2idp:saml_login_process",
-}
+INLINE_PATHS = {"social:begin", "djangosaml2idp:saml_login_process", "widgets"}
```
nijel, line 72: Please avoid this, these are third-party apps where we can't avoid that, but our code should not use inline javascript.

weblate/trans/widgets.py
```diff
@@ -1,4 +1,4 @@
-# Copyright © Michal Čihař <michal@weblate.org>
+# Copyright Michal Čihař <michal@weblate.org>
```
nijel, line 1: Seems like unintended change.

weblate/templates/widgets.html
```diff
@@ -23,110 +24,76 @@ <h2>{% trans "Promoting translation projects" %}</h2>
...
+        </textarea>
       </div>
-    {% endfor %}
+    </div>
   </div>
 
 {% endblock %}
+
+{% block extra_script %}
+  <script type="application/json" id="widgets-data">{{ widgets_json|safe }}</script>
```
nijel, line 97: I don't think safe should be used here. The JSON can still contain markup and that would break. Also passing this as data attribute would be better approach as that would not need inline scripts in CSP and you would have no issues with escaping.

nijel (changes requested):
I think this goes into the correct direction. I'm not sure about the frontend code, I will add @meel-hd to review that.

weblate/static/styles/main.css
```diff
@@ -2358,3 +2358,29 @@ table.table-simple tbody#glossary-terms td {
   word-break: break-word;
   white-space: normal;
 }
+
+.widget-card {
+  border: 1px solid var(--altcha-color-border);
+  border-radius: var(--altcha-border-radius);
```
meel-hd, line 2364: These variables are specific to a 3rd party lib, use the more general alternatives defined in  `variables.css`

weblate/static/styles/main.css
```diff
@@ -2358,3 +2358,29 @@ table.table-simple tbody#glossary-terms td {
...
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.1);
+  margin-bottom: 20px;
+  padding: 15px;
+  background-color: #fff;
```
meel-hd, line 2368: Update this in `dark-styles.css` to be suitable for dark theme

weblate/templates/widgets.html
```diff
@@ -22,110 +22,76 @@ <h2>{% translate "Promoting translation projects" %}</h2>
...
+          <img id="widget-image"
+               src="data:,"
+               alt="{% translate "Translation status" %}"
+               style="max-width: 500px" />
         </div>
+      </div>
 
+      <div class="widget-card p-3">
+        <h4 id="embedCodeLabel">{% translate "Embed Code" %}</h4>
+        <textarea id="embedCode" aria-labelledby="embedCodeLabel" class="code-example form-control" rows="3" readonly>
```
nijel, line 87: This could also be a div with a copy button, something similar to what we do with SSH keys:

https://github.com/WeblateOrg/weblate/blob/19908677f12e01fd216855968886616f9ef196b0/weblate/templates/snippets/ssh-key.html#L12-L22

weblate/templates/widgets.html
```diff
@@ -22,110 +22,76 @@ <h2>{% translate "Promoting translation projects" %}</h2>
   <h3>{% translate "Promoting specific translations" %}</h3>
 
   <form class="autosubmit">
-    {% translate "Besides promoting the whole translation project, you can also choose a specific language or component to promote:" %} {{ form.lang }} {{ form.component }}
```
meel-hd, line 25: is the deletion of `{{ form.lang }} {{ form.component }}` intentional here?

weblate/static/js/widgets.js
```diff
@@ -0,0 +1,169 @@
+// Copyright © Michal Čihař <michal@weblate.org>
+//
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+$(document).ready(() => {
+  const widgetsData = $("#widgets-data").data("json");
```
meel-hd, line 6: **Suggestion**: consider simplifying the null check with 
`let jsonLanguage = widgetsData.language || "" ` , assuming` widgetsData.language `can't be 0, false, or another falsy-but-valid value.

weblate/templates/widgets.html
amCap1712: Yes, both of those options have been moved to the widget settings dialog.

weblate/templates/widgets.html
amCap1712: Added.

meel-hd (approved):
Looks good to me.

nijel:

PR #14328, opened 2025-03-24
Add `site.webmanifest` to `LOGIN_REQUIRED_URLS_EXCEPTIONS` list

Izumi-kun:
Just noticed that when Weblate runs with `REQUIRE_LOGIN`, requests for the manifest fail because they are made without credentials.


nijel:

Issue #14098, opened 2025-03-04
Enhancement for the `Sort By` menu on the `Search` tab at any level


Geeyun-JY3:
### Describe the problem

- The `Sort By` dropdown menu on the `Search` tab at any level always shows `Sort By` after an option is selected. Users will not know if the selection was successful if the `Sort By` dropdown menu does nothing after the selection is made.
- The sort order button also looks the same regardless of the sort order.

### Describe the solution you would like

- Show the last selected option of the `Sort By` dropdown menu after selecting it, consistent with the `Sort By` dropdown menu in the search page and editor, and the Advanced query builder options below.
- The sort order button should change after users click it.

### Screenshots

- `Sort By` dropdown menu:

  https://github.com/user-attachments/assets/41f1ecab-f88c-45a3-bd9a-725e31e8b7d4

- The sort order button:

  https://github.com/user-attachments/assets/728ab58c-ae6d-4bbf-bdec-5ad30a22090a

meel-hd:
I am on it :)

PR #14329, opened 2025-03-24
Fix Sort By search menu

meel-hd:
Show Sort By selection. And update the `search order` icon based on what is selected.
## Preview
_

https://github.com/user-attachments/assets/da495fb5-056b-47e2-bffc-c41e7065e0c7

_
close: #14098

### DIFF
--- a/weblate/static/loader-bootstrap.js
+++ b/weblate/static/loader-bootstrap.js
@@ -1015,6 +1015,20 @@ $(function () {
     }
   }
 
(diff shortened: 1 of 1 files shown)


amCap1712:
On pages that don't have the sort by search menu, it causes an error:
image
 
because of https://github.com/WeblateOrg/weblate/blob/cb02e469b9db3e72b72f40fa71b963faf4e5a76f/weblate/static/loader-bootstrap.js#L1030

PR #14360, opened 2025-03-26
fix(unit): show detailed propagation errors only if not too many

nijel:
Show summary if there is too much of errors to avoid filling up whole page.


nijel:
Already merged.

Issue #9541, opened 2023-07-11
Can you please add coloring if we made any changes in current text


yilmazdurmaz:
### Describe the problem

Proofreading is a bit tricky process when translating, especially in electronic form.

When using pen and paper, you can make any marks on the part that needs editing.

For electronic active editing, such as Weblate, this is not possible in the expected sense.

And the worst part is to click on the wrong button to go to the next string, without saving if we had any changes.

### Describe the solution you'd like

You can help us in two ways, one is simple to implement, the other needs some expertise

1. add some color to the translation's edit box if we have any changes in the text.
  - you have already something similar working on the page: "needs edit" checkbox is disabled when we change the text. but it is not enough.
  - adding some coloring (red, yellow, orange?) to the edit box will have our attention.
2. you may add a diff tools to show which bits have we changed.
  - this too is already in use in other parts of the app: diff in history, and diff in automatic/suggestions.
  - the hard part is that the text in the edit box is still under active editing.
  - it may become slower on some machines, maybe add a checkbox to activate it, if you implement this.

### Describe alternatives you've considered

Other than trying to memorize, I haven't tried anything else.

yilmazdurmaz:
> Please try to clarify the use case

This one is for the bot:

Mainly for proofreading purposes, add some color indicator to the edit box to minimize:
- focus loss for long texts if there are any changes made in the text
- stress about saving any accidental edits

yilmazdurmaz:
Hi again,

I have experimented with border coloring. I used the following code in "weblate/templates/translate.html" after line 170 (where the edit box is added), and it adds color to the input box border if the text has changed.

````html
     <style>
      .border {
        border: 1px solid #dee2e6 !important;
      }
      
      .border-warning {
        --bs-border-opacity: 1;
        border-color: rgba( 220, 53, 69, 1) !important;
      }
    </style>     
    <script>
      window.addEventListener("DOMContentLoaded", (event) => {
        let editor = $(".translation-editor")
        let original = editor.text()
        let changed
        let detect = function(){
          changed = original !== this.value
            if (changed){
            editor.addClass("border border-warning")
          }else{
            editor.removeClass("border border-warning")
          }
        }
        editor.keyup(detect)
      });
    </script>
````

The code is long because:
- somehow jQuery does not load before this code, "$ is not defined", so I had to add extra lines.
- the copy of bootstrap files does not have border classes, so I had to add the custom style tag.

Can someone solve these two issues and make the code more elegant?

[...]

Issue #12489, opened 2024-09-13
Improve biome checker configuration


meel-hd:
### Describe the issue

The `pre-commit`  command shows linting issues in files that are not modified by the current commit or branch.

The weird part is it raises this issues only when there is an issue with the files you actually modified. If the issues raised by the changes you made are fixed the `pre-commit`  doesn't show the issues with the other files you didn't modify.

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

1. Introduce a dummy and  unused variable (to raise an issue within the file you modified).
2. Stage the changes `git add .`
3. Run `pre-commit run -all` (this will say there is an issue with an unused var we introduced and step 1, plus a ton of other issues in other files'
4.  Remove the unused variable we introduced in step 1 and `git add  .`;
5. Then re run the `pre-commit` command. (now that the issue we introduced is fixed the command no longer shows the other issues in the other files'

### Expected behavior

- Show only the issues with the files that are actually modified.
- The other issues should be fixed separately to avoid this behavior

### Screenshots
I introduced the variable in another file. But it says there is issues with `zen.js` which I didn't modify.
![image](https://github.com/user-attachments/assets/227d3b95-a7bc-47e8-9929-27d89565d796)

### How do you run Weblate?

Other

nijel:
I think it shows warnings, but only it fails on error. That's why you see more issues than changed. Probably biome should be configured better (either to not report warnings or to fail on them and fix underlying issues). https://biomejs.dev/guides/configure-biome/ has the documentation.

PR #14389, opened 2025-03-29
feat(editor): indicate unsaved changes in translation editor

meel-hd:
Indicate unsaved changes in translation editor
And ask for confirmation if the user navigate/reload/close the page.
image
close: #9541


weblate/static/editor/base.js
```diff
@@ -26,6 +26,19 @@ WLT.Utils = (() => ({
...
     $el.find('input[name="review"][value="20"]').prop("checked", true);
   },
+
+  /**
+   * Indicate that the translation has changed
+   * by appending a warning before the editor.
+   * @returns {void}
+   */
+  indicateChanges: () => {
+    const $warning = $("<span class='text-warning m-1'/>");
```
nijel, line 36: What is `m-1` supposed to do?

nijel (commented):
There are some biome errors (see check statuses), but it's hard to see what is wrong because of existing warnings (see https://github.com/WeblateOrg/weblate/issues/12489).

weblate/static/editor/base.js
meel-hd: Gives margin

nijel: In Bootstrap 5 it does, not not in Bootstrap 3.4 which we're currently using.

weblate/static/editor/base.js
```diff
@@ -138,6 +156,14 @@ WLT.Editor = (() => {
 
     this.$translationArea[0].focus();
 
+    // Show confirmation dialog if changes have been madae
+    onbeforeunload = (e) => {
```
nijel, line 160: Triggers biome lint/suspicious/noGlobalAssign: A global variable should not be reassigned.

meel-hd: Fixed

PR #14420, opened 2025-04-01
bug: Strip whitespace after prohibited character in glossary string

### LINKED ISSUE #14159: "Invalid glossary entries provided" when requesting DeepL suggestion where glossary string contains "%"
### Describe the issue

Glossary terms containing "%" breaks DeepL automatic suggestions

### I already tried

1. Enable DeepL automatic translations
2. Create a glossary term that contains %
3. Go to the translation page for a string and open the "Automatic translation" tab.

You will get an error message from DeepL that states:

> The request for machine translation using DeepL has failed: Invalid glossary entries provided

### Expected behavior

I would expect it to work despite containing the special character.

### Screenshots

_No response_

### Exception traceback

```pytb

```

### How do you run Weblate?

_No response_

### Weblate versions

_No response_

### Weblate deploy checks

```shell

```

### Additional context

_No response_

gersona:
- fixes #14159
Leading whitespaces were not correctly stripped from a glossary string if it also contained a prohibited character. 
`e.g "% de acertar cuerpo a cuerpo" => " de acertar cuerpo a cuerpo"`.

This was problematic for some Automatic suggestions machinery (DeepL)


weblate/machinery/tests.py
```diff
@@ -499,11 +500,19 @@ def test_glossary_cleanup(self) -> None:
...
         self.assertEqual(render_glossary_units_tsv([unit]), "foo\tbar")
 
-        # # no character cleaned
+        # whitespace correctly stripped
+        unit = MockUnit(code="cs", source=" foo  ", target=" bar  ")
+        self.assertEqual(render_glossary_units_tsv([unit]), "foo\tbar")
+
+        # whitespaces after prohibited characters correctly stripp
+        unit = MockUnit(code="cs", source="% foo  ", target="% bar  ")
+        self.assertEqual(render_glossary_units_tsv([unit]), "foo\tbar")
```
nijel, line 513: I think the code will still fail for some strings. Can you please add a test for `% % foo`?

gersona: you're right, that will fail.
I'll correct that

weblate/glossary/models.py
```diff
@@ -204,7 +204,14 @@ def cleanup(text):
         """
         text = text.translate(CONTROLCHARS_TRANS).strip()
         if text and text[0] in PROHIBITED_INITIAL_CHARS:
-            text = text.lstrip("".join(PROHIBITED_INITIAL_CHARS))
+            text = text.lstrip(
+                "".join(
+                    PROHIBITED_INITIAL_CHARS
+                    | set(
+                        " ",
+                    )
```
nijel, line 212:
```suggestion
                    | {" "},

```

weblate/glossary/models.py
```diff
@@ -204,7 +204,11 @@ def cleanup(text):
         """
         text = text.translate(CONTROLCHARS_TRANS).strip()
         if text and text[0] in PROHIBITED_INITIAL_CHARS:
-            text = text.lstrip("".join(PROHIBITED_INITIAL_CHARS))
+            text = text.lstrip(
+                "".join(
+                    PROHIBITED_INITIAL_CHARS | {" "},
```
nijel, line 209: DeepL talks about "Unicode whitespace character", so that probably means any whitespace, not just the one you've added here.

AliceVisek (approved):

PR #14465, opened 2025-04-04
feat: add support for discarding flags while inheriting

### LINKED ISSUE #14298: Add way to disable flag when inheriting
### Describe the problem

When inheriting flags from other scopes (currently from component, but this will extend with features like https://github.com/WeblateOrg/weblate/issues/10897 or https://github.com/WeblateOrg/weblate/issues/12779) there is no way to remove a flag that was set at a previous level.

Add a flag like `remove:FLAG` that will remove the previously added flag if present. This would be processed when merging flags in `Flags` object.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

nijel:
This allows sting to discard flags defined at the component scope, component to override flags defined by the file format or project.

Fixes #14298

@orangesunny , @AliceVisek Is `discard` a good name for this? As the flags can be used by users in many places, we cannot change the name in the future.


AliceVisek:
> This allows sting to discard flags defined at the component scope, component to override flags defined by the file format or project.
> 
> Fixes #14298
> 
> @orangesunny , @AliceVisek Is `discard` a good name for this? As the flags can be used by users in many places, we cannot change the name in the future.

`Discard` is a correct word use. Personally, I'd prefer a bit more specific name, such as `discard-inherited` or even `discard-inherited-flags`. 

You may also consider to use the word dump instead (shorter, meaning shifted more towards _you'll never get those flags back_).

Ideal for me would be `dump-inherited-flags`. But that's just like my opinion ;)

nijel:
I wanted to keep this short, as this is what we do with other flags. Being long is nice when there is a single flag used, but once there are more of them, being too verbose would make the review of the flags harder.

AliceVisek:
After seeing all the mentioned options, here's my list according to preference:

1. `discard:*` most preferred. Easy to read, clearly has a different use then types of flags with `*-*` , such as `ignore-*`.
2. `-*` is a neat minimalistic solution, but prone to mistakes, as it is easy to overlook it while reading or typing.
3. `ignore-*` least preferred. It may become very confusing with other `ignore-CHECK` flags and mix-ups will be prone to happen.

Let's have a look at a case where I want to perform a check on some strings, that is ignored on a component level. Looking at example above, it may look as follows:

- Project flags: `rst-text`
- Component flags: `md-text, discard:rst-text, ignore-md-link`
- String flag:  (_all variants mentioned below_)
   - `discard:ignore-md-link`
   - `-ignore-md-link`
   - `-ignore-ignore-md-link`
   
Evaluated string flags: `md-text`.

Looking at this, the `discard:*` syntax looks the most transparent to me.

nijel:
Thanks for the feedback! Should we clarify something in the pull request documentation?

docs/admin/checks.rst
```diff
@@ -64,7 +64,10 @@ A failing check will be triggered if the string ``### Index`` is translated as `
 To ensure that internal links are not being translated (i.e. `[test](../checks)`
 does not become `[test](../chequeos)`.
 
-
+The flags defined in the higher scope can be discarded using the
+``discard:NAME`` syntax. For example if component is configured to
+``safe-html``, you can add ``discard:safe-html`` to the string flags to skip it
+for this particular string.
```
AliceVisek, line 70:
```suggestion
The flags defined on a higher level can be discarded using the
``discard:NAME`` syntax. For example, if a component is configured to
``safe-html``, you can add ``discard:safe-html`` to the string flags to skip it
for this particular string.
```

docs/admin/checks.rst
```diff
@@ -110,6 +113,8 @@ Here is a list of flags currently accepted:
     Mark this string as a variant of string with matching source. See :ref:`variants`.
 ``regex:REGEX``
     Regular expression to match translation, see :ref:`check-regex`.
+``discard:NAME``
+    Discards flag defined in the higher scope.
```
AliceVisek, line 117:
```suggestion
    Discards flag defined on a higher level.
```

docs/changes.rst
```diff
@@ -8,6 +8,7 @@ Weblate 5.11
 * Added :http:get:`/api/units/(int:id)/translations/` to retrieve a list of all target translation units for the given source translation unit.
 * Added :http:delete:`/api/groups/(int:id)/roles/(int:role_id)` to delete a role from a group.
 * :ref:`check-automattic-components-format` check to validate placeholders in Automattic components.
+* Flags can now be discarded while inheriting, see :ref:`custom-checks`.
```
AliceVisek, line 11:
```suggestion
* Inherited flags can now be discarded, see :ref:`custom-checks`.
```

AliceVisek (commented):
No need for bigger changes in describing the feature, the example mentioned makes it clear. I only changed some word choice and grammar.

Issue #14312, opened 2025-03-22
Notification emails with "String is available for translation" subject are rendered wrongly


ldmpub:
### Describe the issue

When receiving Weblate notification email on my Android devices under Gmail => the display layout is broken.

🐛**The footer is displayed as a "column".**

### I already tried

- [x] I've read and searched [the documentation](https://docs.weblate.org/).
- [x] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

1. Subscribe to activity notification for a Weblate Translation Project with a Gmail account
2. Open notification email in Gmail Android app

### Expected behavior

The email should display correctly and shouldn't prevent user to click on buttons to directly access Weblate translation string.

### Screenshots

![Image](https://github.com/user-attachments/assets/857e1518-a6a9-4a2e-9d84-b68d50edab9a)

### How do you run Weblate?

weblate.org service

### Additional context

1. In previous Weblate versions the email layout was OK in Gmail under Android.
2. This layout pb seems to occur only on particular types of notification emails (and not on all weblate emails received) => yet layout bug identified on:
    - emails with subject "String is available for translation"

ldmpub:
Changes seem to have be done 🤩.
First notification email received this morning looks good 👌.

I guess the change has been made on Weblate side rather than on Gmail Android app 😉.

ldmpub:
Notice: it may not be the exact same type of notification email 🙃, the one received this morning and displaying Ok is "Suggestions awaiting" and not "new strings to translate" (please refer to emails subjects: there are différent in the 2 sreenshots).

![Image](https://github.com/user-attachments/assets/d398c57c-3497-4a5a-af14-37aa429f68fa)

nijel:
We didn't make any change to this in the past months, so if things suddenly break and fixed, it's most likely caused by a third-party app. Or it can be that some kinds of notifications are rendered wrongly.

ldmpub:
@nijel your 2d hypothesis sounds good => Can you please have a look at the "email template" used for the notification with the following subject https://hosted.weblate.org/translate/weblate/application/fr/?checksum=f307af4252f39fd2 ? It's the one which looks broken (at least under GMail app under Android). 

Thanks,

nijel:
There is one issue in generated HTML which I've addressed in https://github.com/WeblateOrg/weblate/pull/14341, but it did affect all notifications, not just this one.

@meel-hd Do you have an idea how to debug this?

meel-hd:
[...]

PR #14473, opened 2025-04-05
fix(style): Fix String is available for translation email layout

meel-hd:
I tracked the email rendering issue to be related to wrong `flex-direction` default to `row `of `.container` instead of `column `that's why the footer is broken.

> [!NOTE]  
> This is the structure of the sent out email. The footer should be a child of `body` not nested as below.

```HTML
<body>
      <div class="content">
         <div class="container">
            <div class="box">...</div>
            <div class="footer">...</div>
            ...
         </div>
      </div>
   </body>
```

> [!IMPORTANT]  
> @nijel  I only tested the effect of this change on `String is available for translation` email. Not sure how it will effect the other emails.

How to setup weblate with an email server. I tried with [github.com/maildev/maildev](https://github.com/maildev/maildev) it works fine with `./manage.py sendtestemail` but when I enable the emails for an account and trigger them I don't receive any (I am running Weblate without docker).

Related: #14312

### DIFF
--- a/weblate/templates/mail/style.css
+++ b/weblate/templates/mail/style.css
@@ -207,6 +207,7 @@ pre {
   margin: 0 auto;
   width: 100%;
   padding: 0 20px;
+  flex-direction: column;
 }
 @media (min-width: 768px) {
   .container {


nijel:
@meel-hd The notifications are usually delivered in the background using celery, see https://docs.weblate.org/en/latest/admin/install.html#background-tasks-using-celery. The eager execution won't work for digest mails.

nijel:
PS: `<div class="footer">...</div>` should not be child of `<body>` it needs to be under container to not be full-width.

PR #14506, opened 2025-04-08
feat: add language_code field to unit api

### LINKED ISSUE #14364: Add language to unit resource in Rest API
### Describe the problem

https://docs.weblate.org/en/latest/api.html#get--api-units-(int-id)- does not have a field for the language code.

Add language to the unit resource.

Parsing out the language from the "translation" field.

### Screenshots

_No response_

### Additional context

_No response_

amCap1712:
Add language_code field to unit API. 

Fixes #14364.


weblate/api/serializers.py
```diff
@@ -1092,6 +1092,9 @@ class UnitSerializer(serializers.ModelSerializer[Unit]):
         ),
         strip_parts=1,
     )
+    language_code = serializers.CharField(
+        source="translation.language_code", read_only=True
```
nijel, line 1096: Using a standardized language code instead of the one used in the filename is probably more useful:

```suggestion
        source="translation.language.code", read_only=True
```

amCap1712: done.

nijel (approved):

PR #14513, opened 2025-04-09
notifications: truncate large digest emails

### LINKED ISSUE #14130: Do not split digest emails, summarize instead
### Describe the problem

This morning I woke up to 79 emails from Weblate in my inbox. All sent within the span of two minutes, all notifying about a single project (Organic Maps).

I checked my preferences. They correctly say that I want to receive digests, not one email per message.

However, it seems like even though I have set my preferences like this, if there are more than so many notifications of the same type (a 100?), the notification system will split them into multiple digests and send them as multiple emails. This seems rather suboptimal.

Instead of sending dozens of emails, send just one, but add a note mentioning that "there are more changes than we have included here". Getting dozens of emails about new strings/suggestions/etc. isn't any more helpful than getting just one email. It's counter-productive even.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

amCap1712:
If there are a large number a notification in a digest, the digest is chunked into multiple emails. This can result in a large number of emails for the user which is annoying. Hence, only send the first 100 (current digest chunk size) notifications to the user in the digest. If the number of notifications exceeds this, a note that the digest was truncated is displayed to the user in the digest email.

Fixes #14130.

Screenshot From 2025-04-09 02-37-00


weblate/accounts/notifications.py
```diff
@@ -441,9 +441,15 @@ def notify_immediate(self, change) -> None:
                 if current_subscription.onetime:
                     current_subscription.delete()
 
-    def send_digest(self, language, email, changes, subscription=None) -> None:
+    def send_digest(
+        self, language, email, changes, subscription=None, overlimit=False
```
nijel, line 445: Please add type annotations to the new code:

```suggestion
        self, language, email, changes, subscription=None, *, overlimit: bool = False
```

Making boolean kwargs only avoids mistakes in invocation.

PR #14515, opened 2025-04-09
feat(machinery): automatically choose source language for cyrtranslit

### LINKED ISSUE #13769: Defining source language in machinery backend
### Describe the problem

Some machinery backends support only certain language pairs and not all languages. [Cyrtranslit](https://docs.weblate.org/en/latest/admin/machine.html#cyrtranslit) could be a typical example.

These backends fail to perform the translation when the target language is supported, but the source is not. In such case, it could automatically change the source language to a supported one if it is available.

* The implementation depends on https://github.com/WeblateOrg/weblate/issues/13768.
* It can extend the language fallback mechanism to choose supported one instead of automatically choosing source.
* It could probably take precedence over user supplied language as well when it is not supported.
* The scope of this should be limited to simple engines such as Cyrtranslit - when there it is clear that the engine can't select anything else as a source language to work.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

nijel:
It automatically picks matching translation to use. This allows to use the engine even if the source language is English.

Fixes #13769

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -21,7 +21,7 @@ Weblate 5.11
 * :ref:`addon-weblate.discovery.discovery` better handles hundredths of matches.
 * Dismissing :ref:`checks` automatically updates propagated strings.
 * Improved rendering of :ref:`additional-flags` and :ref:`additional-explanation` changes in history.
-* :ref:`mt-cyrtranslit` now automatically transliterates from matching translation.
+* :ref:`mt-cyrtranslit` now automatically transliterates from a matching translation instead of the source strings.
 
 .. rubric:: Bug fixes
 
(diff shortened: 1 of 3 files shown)


docs/changes.rst
```diff
@@ -21,6 +21,7 @@ Weblate 5.11
 * :ref:`addon-weblate.discovery.discovery` better handles hundredths of matches.
 * Dismissing :ref:`checks` automatically updates propagated strings.
 * Improved rendering of :ref:`additional-flags` and :ref:`additional-explanation` changes in history.
+* :ref:`mt-cyrtranslit` now automatically transliterates from matching translation.
```
nijel, line 24:
```suggestion
* :ref:`mt-cyrtranslit` now automatically transliterates from a matching translation instead of the source strings.
```

PR #14518, opened 2025-04-10
feat(machinery): configurable source language

### LINKED ISSUE #6232: Automatic translation from manually chosen language
There is a very popular case that developers usually use mixed translation source strings especially for multi regional team or using libraries which is previously written not by English friendly developers. I think that's the reason why we have English (developer) here (*_devel is supported too). In such case, the machine translation feature of the automatic translation add-ons and the automatic suggestion does not work well because the translation source is not always in a locale but hard to detect.

**SUGGESTION**
Supporting using another verified translated translation as source string for automatic translation features. Let's say, people try their best effort to do manual translation from English (developer) to en_US and it's better if we can use the en_US as translation source for any further automatic one rather than the current source strings.

**BENEFIT**
This helps a lot because we move the standardizing tasks to translators in stead of developers, who are not prefer to do the translation carefully.

nijel:
Allow to customize the source language in the machine translation.

Fixes #6232


docs/workflows.rst
```diff
@@ -260,6 +260,22 @@ based on what is now a polished source language.
...
+String consolidation using secondary language
+---------------------------------------------
+
+When :ref:`component-intermediate` cannot be used to consolidate the strings,
+you might choose one of the translations to act as a secondary language. The
+language can configured in the project (:ref:`project-secondary_language`) or
+component (:ref:`component-secondary_language`). It is then shown to the
+translators while translating (see :ref:`secondary-languages`) and can be
+optionally used as a source for the machine translation (see
+:ref:`mt-sources`).
```
AliceVisek, line 274:
```suggestion
When :ref:`component-intermediate` cannot be used to consolidate the strings,
you can choose one of the translations to act as a secondary language. The
language can be configured in a project (:ref:`project-secondary_language`) or in a 
component (:ref:`component-secondary_language`). It is then shown to the
translators while translating (see :ref:`secondary-languages`) and can be
optionally used as a source for machine translations (see
:ref:`mt-sources`).
```

docs/workflows.rst
```diff
@@ -260,6 +260,22 @@ based on what is now a polished source language.
...
+
+This setup can be useful with mixed language source strings and consolidating
+them into a single translation which is then used as a baseline for other work.
```
AliceVisek, line 277:
```suggestion
them into a single translation, which is then used as a baseline for other work.
```

AliceVisek (approved):
Just minor grammar issues to be resolved.

Issue #14467, opened 2025-04-04
A label count function for strings


djt-dy:
### Describe the problem

We use labels to show which release a string is included in. We've seen an uptick in our copy teams re-visiting strings and altering them. We're trying to work out how many times that's happening. There's effort, time and cost attached to those changes, and we'd like to try and determine that and the value of it to our business.

### Describe the solution you would like

Something like 

`count(labels)>=2`

Would be perfect!

Jin-Shikai:
Hi, please review #14527 to see if it adds this support properly.

djt-dy:
Thanks, Michal,

Looking forward to using this feature.

Rgds

David

From: github-actions[bot] ***@***.***>
Date: Thursday, 26 June 2025 at 11:56
To: WeblateOrg/weblate ***@***.***>
Cc: David Tomlinson ***@***.***>, Author ***@***.***>
Subject: [EXTERNAL EMAIL]Re: [WeblateOrg/weblate] A label count function for strings in search (Issue #14467)
CAUTION: This email originated from outside of the organization. Do not click links or open attachments unless you recognize the sender and know the content is safe.
[Image removed by sender.]github-actions[bot] left a comment (WeblateOrg/weblate#14467)<https://github.com/WeblateOrg/weblate/issues/14467#issuecomment-3008070603>

Thank you for your report — the issue has now been resolved.

The fix has been merged into the repository and will be included in the upcoming Weblate 5.13 release.

  *   If you encounter any issues with the fix, please comment in this thread.
  *   If you experience a similar but separate issue, open a new report so we can track it properly.
  *   If you find Weblate helpful, consider supporting its development<https://weblate.org/donate/>.

PR #14527, opened 2025-04-11
Feat(search): add support for filtering by labels count

Jin-Shikai:
Add support for issue #14467 

image

The total number can be read by the number of positions.

### DIFF
--- a/weblate/trans/models/unit.py
+++ b/weblate/trans/models/unit.py
@@ -202,9 +202,7 @@ class UnitQuerySet(models.QuerySet):
 
         result = self
         if "labels_count" in query:
-            result = result.annotate(
-                labels_count=Count("source_unit__labels") + Count("labels")
-            )
+            result = result.annotate(labels_count=Count("source_unit__labels") + Count("labels"))
 
         result = result.filter(parse_query(query, **context))
         return result.distinct()
--- a/weblate/trans/models/unit.py
+++ b/weblate/trans/models/unit.py
@@ -200,13 +200,7 @@ class UnitQuerySet(models.QuerySet):
         """High level wrapper for searching."""
         from weblate.utils.search import parse_query
 
-        result = self
-        if "labels_count" in query:
-            result = result.annotate(
-                labels_count=Count("source_unit__labels") + Count("labels")
-            )
-
-        result = result.filter(parse_query(query, **context))
+        result = self.filter(parse_query(query, **context))
         return result.distinct()
 
     def same(self, unit: Unit, exclude: bool = True) -> UnitQuerySet:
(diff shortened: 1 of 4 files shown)


weblate/trans/models/unit.py
```diff
@@ -200,7 +200,13 @@ def search(self, query, **context) -> UnitQuerySet:
...
         from weblate.utils.search import parse_query
 
-        result = self.filter(parse_query(query, **context))
+        result = self
+        if "labels_count" in query:
+            result = result.annotate(
+                labels_count=Count("source_unit__labels") + Count("labels")
+            )
+
+        result = result.filter(parse_query(query, **context))
```
nijel, line 209: The whole filtering should be implemented in `weblate/utils/search.py`. It is used from other places as well.

nijel (changes requested):
Thanks for your pull request, but the `parse_query` should return a valid query object for Django, so the logic needs to be implemented there, not postprocessed later. Also, the new search query should be test covered in `weblate/utils/tests/test_search.py`.

Jin-Shikai:
@nijel Hi, thank you for your guidance. I've implemented the changes based on your suggestion, moving the filtering logic to weblate/utils/search.py.

weblate/utils/search.py
```diff
@@ -520,6 +521,30 @@ def component_field(self, text: str, context: dict) -> Q:
...
+        suffix = OPERATOR_MAP.get(self.operator)
+        if suffix == "substring":
+            suffix = "exact"
+        filter_kwargs = {f"labels_count__{suffix}": value}
+
+        return Q(
+            id__in=Unit.objects.annotate(
+                labels_count=Count("source_unit__labels", distinct=True)
+                + Count("labels", distinct=True)
+            )
```
nijel, line 543: This will have horrible performance wise on larger sites as it filters *all* units with more labels to then search within a selected scope.

Jin-Shikai:
@nijel Hi, I referred to the handling of glossary in has_field function, and use "translation__component__project=project" to improve performance.

I also tried filtering on smaller scopes like "component" and "translation", but the context doesn't have them. I found that the context is passed in from the function "search" in weblate/trans/views/edit.py via  
"allunits = unit_set.search(query_string, project=project)"
so there's only the "project" element.

nijel:
Turns out this is not that easy issue as I expected. Passing more info in the context is certainly possible, but it will anyway lead to slow queries because of using subquery.

Looking at this again, I think the best approach would be for parse_query to return both filter and annotations and these would be then executed. This will need more changes across the code-base, but will allow to make effective seaching using annotations.

Issue #14530, opened 2025-04-11
Each time you save a string, the browser asks you whether to leave the page


tomkolp:
### Describe the issue

After yesterday's bleeding edge update, every time I save a string, a browser message appears with two buttons: "Leave page" and "Stay on page". The window appears on all browsers.

### I already tried

- [x] I've read and searched [the documentation](https://docs.weblate.org/).
- [x] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

1) save the string
2) An additional browser selection window appears

### Exception traceback

```pytb

```

### How do you run Weblate?

Docker container

### Weblate versions

<html>
<body>

5.11-dev — 99a97d1b54cdf6dad2ec7ff5f8dabed9a38ce8f6
--

</body>
</html>5.11-dev — [99a97d1b54cdf6dad2ec7ff5f8dabed9a38ce8f6](https://github.com/WeblateOrg/weblate/commits/99a97d1b54cdf6dad2ec7ff5f8dabed9a38ce8f6)

### Weblate deploy checks

```shell

```

meel-hd:
I'm on it

PR #14544, opened 2025-04-12
fix(editor): improve unsaved changes indication and reset on submit

meel-hd:
### Fixes
- Prevent unsaved changes dialog on **submit** of changes.
- Add unsaved changes indicator only to translation editor (it showed also in `add term to glossary` inputs).

close: #14530


nijel:

PR #14547, opened 2025-04-12
fix: use ngettext for languages widget label

nijel:
This makes it look better when localized.


amCap1712 (approved):
Didn't test locally but LGTM.

Issue #14546, opened 2025-04-12
Adding term to glossary shows JSON response


nijel:
### Describe the issue

Adding glossary while translating is no longer handled by javascript and does direct form submission causing browser to show the JSON response.

### I already tried

- [x] I've read and searched [the documentation](https://docs.weblate.org/).
- [x] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

1. Go to translate view
2. Click on add glossary term
3. Enter source and translation
4. Click save

### Expected behavior

The submission should be handled by javascript and page updated.

### Exception traceback

```pytb

```

### How do you run Weblate?

weblate.org service

### Weblate deploy checks

```shell

```

### Additional context

Could it be a regression from https://github.com/WeblateOrg/weblate/pull/14544?

meel-hd:
Inspecting it!

PR #14562, opened 2025-04-13
fix(editor): fix glossary dialog initialization and event handling

meel-hd:
issue:
The  event listener for submit of the form, and the one for modal.bs.show weren't attached at all to the elements.

close: #14546


nijel:

PR #14575, opened 2025-04-14
Allow replacing existing fonts with the same family and style

### LINKED ISSUE #14484: Replacing a existing font with a updated version
### Describe the problem

[Managing fonts](https://docs.weblate.org/en/latest/admin/checks.html#fonts) could support replacing a existing font with a updated version of the same font.

I'd like to be able to replace a existing font with a updated font file, provided the font family and font style is the same.

1. Uploading as a new font and changing the default font of the font group: 
    not possible because the updated font has the same attributes (family, style), so it cannot be added as a new font.
3. Deleting the font and the font group and recreating both: 
    possible but quite a few clicks and error-prone.

### Screenshots

_No response_

### Additional context

This should also invalidate all existing cached font renderings.

Zouu-X:
### Summary

This PR is for "issue#14484"
It allows replacing an existing font (with the same family and style) when uploading via the font management form. The updated file replaces the existing one

### Details

- Checks for an existing font using `family`, `style`, and `project`
- If found, replaces the file
- Avoids duplication and improves UX
- Fully backward-compatible

### Related issue

Fixes #14484


nijel:
Wouldn't it be better to allow editing from the font detail page instead of magically doing that on upload? That way, the user would know what to do when updating a font.

Zouu-X:
> Wouldn't it be better to allow editing from the font detail page instead of magically doing that on upload? That way, the user would know what to do when updating a font.

Thanks for the feedback! I’ll move this feature to the font detail page with a button "upload" and update the PR shortly.

Zouu-X:
I've pushed the updated logic to enable users update the font in the font detail page. I also try to keep all pages' style aligned :)
Let me know if there’s anything else to adjust!

weblate/fonts/views.py
```diff
@@ -83,23 +86,67 @@ def post(self, request: AuthenticatedHttpRequest, **kwargs):
...
+            return self.get(request, **kwargs)
+
+        # Compare file content hashes
+        new_hash = hashlib.sha256(new_file.read()).hexdigest()
+        new_file.seek(0)
+
+        current_hash = hashlib.sha256(self.object.font.read()).hexdigest()
+        self.object.font.seek(0)
+
+        if new_hash == current_hash:
```
nijel, line 138: Isn't it better to directly compare the content? It is read into the memory anyway.

nijel (changes requested):
Looks good, but it would be great to have this covered by the tests.

weblate/fonts/views.py
nijel, line 138:
```suggestion
        new_content = new_file.read()
        new_file.seek(0)

        current_content = self.object.font.read()
        self.object.font.seek(0)

        if new_content == current_content:
```

weblate/fonts/views.py
```diff
@@ -83,23 +86,67 @@ def post(self, request: AuthenticatedHttpRequest, **kwargs):
...
+            messages.info(
+                request, "The uploaded font file is identical to the current one."
```
nijel, line 140: Localize the message:

```suggestion
                request, gettext("The uploaded font file is identical to the current one.")
```

weblate/fonts/views.py
```diff
@@ -83,23 +86,67 @@ def post(self, request: AuthenticatedHttpRequest, **kwargs):
...
+            )
+        else:
+            self.object.font = new_file
+            self.object.user = request.user
+            self.object.timestamp = timezone.now()
+            self.object.save()
+            messages.success(request, "Font updated successfully.")
```
nijel, line 147:
```suggestion
            messages.success(request, gettext("Font updated successfully."))
```

weblate/fonts/views.py
```diff
@@ -83,23 +86,67 @@ def post(self, request: AuthenticatedHttpRequest, **kwargs):
...
+
+        new_file = request.FILES["font"]
+
+        try:
+            uploaded_family, uploaded_style = get_font_name(new_file)
+            new_file.seek(0)
+        except Exception:
+            messages.error(
+                request,
+                "Could not extract font metadata. Ensure it’s a valid font file.",
```
nijel, line 118:
```suggestion
                gettext("Could not extract font metadata. Ensure it’s a valid font file."),
```

weblate/fonts/views.py
```diff
@@ -83,23 +86,67 @@ def post(self, request: AuthenticatedHttpRequest, **kwargs):
...
     def post(self, request: AuthenticatedHttpRequest, **kwargs):
         self.object = self.get_object()
+
         if not request.user.has_perm("project.edit", self.project):
             raise PermissionDenied
 
-        self.object.delete()
-        messages.error(request, gettext("Font deleted."))
-        return redirect("fonts", project=self.project.slug)
+        if "font" not in request.FILES:
```
nijel, line 106: This really should use form validation instead of doing all the checks manually.

nijel (approved):
I've made necessary fixes:

* c0f51e4e977bbea3c8c6acb6800b8f599b7b6b7e fixes font deletion (it was removed), adds tests, and uses form validation
* 72b12eeea958638bef4b55deae78011c06aac3a7 adds tracking of changes in history
* 50fd24812ef82cca7333ea7f5fa7a47f403d66b8 adds changelog entry

PR #14618, opened 2025-04-17
Kabyle character normalization check

BoFFire:
This introduces first check for Kabyle strings on Weblate.


weblate/checks/chars.py
```diff
@@ -464,6 +464,36 @@ class PunctuationSpacingCheck(TargetCheck):
...
+        if unit.translation.language.code != "kab":
+            return False
+        return any(char in target for char in self.confusable_to_standard)
+
+    def get_fixup(self, unit: Unit):
+        return [
+            (re.escape(confusable), standard, "gu")
+            for confusable, standard in self.confusable_to_standard.items()
+        ]
+
```
nijel, line 496: Seems like you've put the class in middle of another class instead of putting it outside.

weblate/checks/chars.py
```diff
@@ -464,6 +464,36 @@ class PunctuationSpacingCheck(TargetCheck):
...
+        "\u03b3": "\u0263",
+        "\u0393": "\u0194",
+        "\u03b5": "\u025b",
+        "\u0395": "\u0190",
+        "\u011f": "\u01e7",
+        "\u011e": "\u01e6",
+    }
+
+    def check_single(self, source: str, target: str, unit: Unit) -> bool:
+        if unit.translation.language.code != "kab":
```
nijel, line 487: `unit.translation.language.is_base({"kab"})` is usually a better way to check this to cover possible variations. Also, this is better to be placed in the `should_skip` method.

nijel (changes requested):
The code seems misformated (the class placed inside another one) and few bits are missing:

* The check needs to be included in the default configuration to be used
* Tests for the check
* Documentation

weblate/checks/chars.py
```diff
@@ -100,6 +100,39 @@ def get_fixup(self, unit: Unit):
...
+        "\u0393": "\u0194",
+        "\u03b5": "\u025b",
+        "\u0395": "\u0190",
+        "\u011f": "\u01e7",
+        "\u011e": "\u01e6",
+    }
+
+    def should_skip(self, unit: Unit) -> bool:
+        # Only run on Kabyle (covers 'kab' plus any variants)
+        return not unit.translation.language.is_base({"kab"})
```
nijel, line 123: Keep the parent class logic in place as well:

```suggestion
        if not unit.translation.language.is_base({"kab"}):
           return True
        return super().should_skip(unit)
```

nijel (commented):
Any luck in writing tests for this? (And other things I've mentioned before)

nijel:
https://github.com/BoFFire/weblate/pull/1 adds docs and tests, please review (especiall the docs) and merge.

nijel (approved):

nijel:

BoFFire:
Thank you <3! And very much ! @nijel

PR #14620, opened 2025-04-17
bugfix (gitlab): check status code 409 for existing pull requests

amCap1712:
The error message returned by Gitlab might be localized. The error message may also change in the future, hence check for the more reliable 409 status code that Gitlab returns in case of a resource conflict.


nijel (commented):
We should figure out when GitLab changed the behavior to determine whether it is still needed to keep compatibility with the previous logic.

BingYanchi:
Although merge requests can be submitted correctly, Weblate still treats them as errors and shows them in the alert.

nijel:
@BingYanchi Are you talking about situation with this patch applied or without?

nijel:
No, the pull request is not yet merged...

nijel:
> We should figure out when GitLab changed the behavior to determine whether it is still needed to keep compatibility with the previous logic.

The HTTP 409 is documented in OpenAPI for three years, so I think it's okay to ditch the old solution completely.

amCap1712:
Sorry for the delay, I had found similar documentation before opening the PR but forgot to mention it.

PR #14624, opened 2025-04-18
feat: deleting a unit goes to next entry in batch

### LINKED ISSUE #14558: Go to next entry in batch after deleting entry
### Describe the problem

I am cleaning up our glossary (with many entries added by contributors, which do not belong in a glossary). I want to delete a bunch of them, but each time I do so, I get sent back to the component overview page. Then I have again to click on Browse, find the string where I was/that I wanted to delete next (maybe jump a page or two) and open the string.

After deleting a string, open the next string in the batch (with batch I mean the list that I opened, as indicated by the ## out of ## indication just below the breadcrumb).

#7563 / #11887

### Screenshots

_No response_

### Additional context

_No response_

amCap1712:
When a unit is deleted, go to the next unit in the batch instead of going back to the component overview page.

fixes #14558

### DIFF
--- a/weblate/templates/translate.html
+++ b/weblate/templates/translate.html
@@ -208,7 +208,11 @@
                       <ul class="dropdown-menu">
                         {% if user_can_delete %}
                           <li>
-                            <a href="" class="link-post" data-href="{% url "delete-unit" unit_id=unit.pk %}">{% trans "Delete string" %}</a>
+                            <!-- use this_unit_url as the next unit will become the unit at current offset after deletion -->
+                            <a href=""
+                               class="link-post"
+                               data-href="{% url "delete-unit" unit_id=unit.pk %}"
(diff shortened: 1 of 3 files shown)


weblate/templates/translate.html
```diff
@@ -208,7 +208,11 @@ <h4 class="panel-title">
                       <ul class="dropdown-menu">
                         {% if user_can_delete %}
                           <li>
-                            <a href="" class="link-post" data-href="{% url "delete-unit" unit_id=unit.pk %}">{% trans "Delete string" %}</a>
+                            <!-- use this_unit_url as the next unit will become the unit at current offset after deletion -->
```
nijel, line 211: Are you sure this is the case? The search result should be cached in the sessions, so the offset IMHO should not change (but I didn't verify that).

nijel: Okay, looking at the code I see that the cache is discarded upon deletion.

PR #14695, opened 2025-04-27
feat: add comment api

### LINKED ISSUE #14232: Expose 'add comment' API endpoint
### Describe the problem

We're migrating from another platform to Weblate, and would like to take along our comments. It's currently not possible to do this easily/programmatically. Consequently, we're using valuable contributor feedback on our strings.

We'd like to have an API endpoint that allows to post a comment on a given string. Matching:
* If we provide the email address of the poster, and it exists as user on Weblate, the comment should be associated with that user.
* We'd like to provide the string key to identify the relevant string on Weblate.

I wondered whether to request the following, but concluded that the benefits don't outweigh the (technical) challenges: If we provide a timestamp (in the past), it is maintained so that when viewing the comment in context there is additional metadata to consider whether the comment is still relevant.

### Screenshots

_No response_

### Additional context

_No response_

amCap1712:
Add `/api/units/{unit_id}/comments/` endpoint to create comments.

Fixes #14232

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -2286,6 +2286,18 @@ and XLIFF.
 
    Returns a list of all target translation units for the given source translation unit.
 
+.. http:post:: /api/units/(int:id)/comments/
+
+    .. versionadded:: 5.12
+
+    Create a new comment on the given translation unit.
+
+    :param id: Unit ID
+    :type id: int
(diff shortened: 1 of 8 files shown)


nijel (approved):
Looks good, but will need more work in a separate pull request, see https://github.com/WeblateOrg/weblate/issues/14232#issuecomment-2834537750

AliceVisek (approved):

PR #14708, opened 2025-04-28
docs: clarify reverse proxy setup

nijel:
Split the section into three topics to make it clear what information is needed.


AliceVisek (approved):
Commas and word choice adjusted. I recommend to resolve those before merging, but the text will be readable well enough even without it.

Issue #14716, opened 2025-04-28
In Zen mode, all entries are marked as 'Changes Unsaved' once a single one has been edited.


mont29:
### Describe the issue

Discovered on translate.blend.org (we recently upgraded from 5.10.1.0 to 5.11.0.0), but also reproducible on e.g. https://hosted.weblate.org/zen/godot-engine/godot-docs/fr/ : 

![Image](https://github.com/user-attachments/assets/988484c3-ac03-461f-ae94-11037a567662)

### I already tried

- [x] I've read and searched [the documentation](https://docs.weblate.org/).
- [x] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

1. Start to translate in Zen mode
2. Edit a string
3. See all listed translations get outlined in orange, with 'Changes Unsaved!' warning message.

### Exception traceback

```pytb

```

### Weblate deploy checks

```shell

```

meel-hd:
I am working on  it

PR #14731, opened 2025-04-29
fix(editor): pass event to indicateChanges for accurate context

meel-hd:
Close: #14716


nijel:
The change indication should also go away when string is saved what happens automatically in the zen mode.

meel-hd:
Oh okay

nijel:

PR #14734, opened 2025-04-29
feat: import comment api

### LINKED ISSUE #14232: Expose 'add comment' API endpoint

amCap1712:
Update the /api/units/{unit_id}/comments/ endpoint to accept user and timestamp for a comment if the user making the request is a project admin. Return the comment id in the response as it is useful in testing and can also be useful in future resolve/delete comment APIs.

Fixes #14232.

### DIFF
--- a/weblate/api/views.py
+++ b/weblate/api/views.py
@@ -2053,8 +2053,9 @@ class UnitViewSet(viewsets.ReadOnlyModelViewSet, UpdateModelMixin, DestroyModelM
     def comments(self, request, *args, **kwargs):
         """Add a new comment to a unit."""
         unit = self.get_object()
+        user = request.user
 
-        if not request.user.has_perm("comment.add", unit.translation):
+        if not user.has_perm("comment.add", unit.translation):
             self.permission_denied(request)
 
         serializer = CommentSerializer(data=request.data, context={"unit": unit})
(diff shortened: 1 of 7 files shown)


weblate/trans/models/comment.py
```diff
@@ -17,25 +18,38 @@
...
         unit_scope = unit.source_unit if scope in {"global", "report"} else unit
 
-        user = request.user
-        new_comment = self.create(
-            user=user,
-            unit=unit_scope,
-            comment=text,
-            userdetails={
+        kwargs = {
+            "user": user,
```
nijel, line 41: It is no longer set from request?

weblate/trans/models/comment.py
```diff
@@ -17,25 +18,38 @@
...
 class CommentManager(models.Manager):
-    def add(self, request: AuthenticatedHttpRequest, unit, text, scope) -> None:
+    def add(
+        self,
+        request: AuthenticatedHttpRequest,
+        unit: Unit,
+        text: str,
+        scope: str,
+        user: User,
+        timestamp: datetime | None = None,
```
nijel, line 34: I'd prefer to keep the existing semantics to extract user from the request if not set. Also making new args kwargs only will make the code easier to read with many parameters:

```suggestion
        *,
        user: User | None = None,
        timestamp: datetime | None = None,
```

weblate/api/views.py
```diff
@@ -2053,17 +2053,31 @@ def translations(self, request: Request, *args, **kwargs):
...
+        ):
+            self.permission_denied(request)
+
+        if user_email is not None:
+            override_user = User.objects.filter(email=user_email).first()
+            if override_user is not None:
+                user = override_user
+
+        comment = Comment.objects.add(request, unit, text, scope, user, timestamp)
+        return Response(data={"id": comment.id}, status=HTTP_201_CREATED)
```
nijel, line 2080: I think having a proper model serializer and returning newly created serialized comment would be better.

nijel (changes requested):

AliceVisek (approved):

weblate/trans/models/comment.py
```diff
@@ -17,25 +18,39 @@
...
+        scope: str,
+        *,
+        user: User | None = None,
+        timestamp: datetime | None = None,
+    ) -> Comment:
         """Add comment to this unit."""
         # Is this source or target comment?
         unit_scope = unit.source_unit if scope in {"global", "report"} else unit
 
-        user = request.user
```
nijel, line 29: The user variable is also used later in the code, and keeping it `None` value causes crashes (as can be seen in the tests).

nijel (approved):

nijel:

Issue #14296, opened 2025-03-21
Sort by component when translating multiple components


nijel:
### Describe the problem

When translating at project or category scope, the string ordering still defaults to position and priority. So the user first translates strings of all components and then the second strings of all components. This makes unnecessary jumping between contexts and makes the translation process annoying.

### Describe the solution you would like

The strings would be first ordered based on components (and their priorities).

ByteHamster:
Additional sort priority: non-locked components first. Otherwise, if users see a string from a locked component first, they might think that the entire translation is locked

Cwpute:
This could be coupled with a warning for the user, something like "You were translating strings for component X, this string is part of component Y." so that the change is clearly visible when translating loads of strings at times from several different components.

PR #14781, opened 2025-05-04
feat: sort units by component when translating multiple components

amCap1712:
Add sort choice to sort by component and unit priority to avoid unnecessary jumping between contexts. The units are sorted by the component in the same order as components are sorted in a project/category.

Further, move the sort name and sort query to the SearchForm from the views to ensure consistent sort defaults on all pages.

Also fix a bug in zen mode when changing the sort in zen mode results in a blank page.

#14296

### DIFF
--- a/weblate/trans/forms.py
+++ b/weblate/trans/forms.py
@@ -763,7 +763,7 @@ class SearchForm(forms.Form):
         self,
         *,
         request: AuthenticatedHttpRequest,
-        language=None,
+        language : Language | None = None,
         show_builder=True,
         obj: type[Model | BaseURLMixin] | None = None,
         **kwargs,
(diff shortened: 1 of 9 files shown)


weblate/trans/forms.py
```diff
@@ -758,10 +759,21 @@ def get_initial(request: AuthenticatedHttpRequest):
             return {"q": request.GET["q"]}
         return None
 
-    def __init__(self, user: User, language=None, show_builder=True, **kwargs) -> None:
+    def __init__(
+        self,
+        *,
+        request: AuthenticatedHttpRequest,
+        language=None,
```
nijel, line 766: I think we can add type annotation here as well:

```suggestion
        language : Language | None = None,
```

nijel (commented):
Looks good!

PR #14799, opened 2025-05-06
fix: restructure footer in email template for better compatibility with email clients

meel-hd:
Close: #14312


nijel:

PR #14806, opened 2025-05-07
fix: improve tab naming consistency

### LINKED ISSUE #14171: Info / Information consistency
### Describe the issue

@Geeyun-JY3 at https://github.com/WeblateOrg/weblate/issues/14096:

> BTW, why is the `Information` tab of the site-wide specific language (I'm not sure if the [string related to the component list](https://github.com/WeblateOrg/weblate/blob/8fb30c17090aca4b6713738d4a1fb5edd5c19ece/weblate/templates/component-list.html#L23) also counts) is named “[Information](https://github.com/WeblateOrg/weblate/blob/8fb30c17090aca4b6713738d4a1fb5edd5c19ece/weblate/templates/language.html#L33)” while the `Info` tabs of others are all named “[Info](https://hosted.weblate.org/translate/weblate/application/en/?checksum=c543aec340e3d71c)? Was this intentional? If not, please make them consistent.

### I already tried

1. https://hosted.weblate.org/languages/cs/ uses "Information"
2. https://hosted.weblate.org/projects/weblate/ uses "Info"

### Expected behavior

It should be consistent. Need to pick what better fits.

### Screenshots

_No response_

### Exception traceback

```pytb

```

### How do you run Weblate?

weblate.org service

### Weblate versions

_No response_

### Weblate deploy checks

```shell

```

### Additional context

_No response_

nijel:
- use "Overview" instead of "Info"/"Information" as that better describes the tab
- replace "Overview" by "Summary" in the panels to avoid repeating the label
- the top level translation label is now "Translation", not "Overview" to make it consistent with other levels

Fixes #14171

### DIFF
--- a/weblate/templates/category-language.html
+++ b/weblate/templates/category-language.html
@@ -13,7 +13,7 @@
       <a href="#overview" data-toggle="tab">{% translate "Components" %}</a>
     </li>
     <li>
-      <a href="#information" data-toggle="tab">{% translate "Info" %}</a>
+      <a href="#information" data-toggle="tab">{% translate "Overview" %}</a>
     </li>
     <li>
       <a href="#search" data-toggle="tab">{% translate "Search" %}</a>
--- a/weblate/templates/category.html
+++ b/weblate/templates/category.html
@@ -29,7 +29,7 @@
       <a href="#languages" data-toggle="tab">{% translate "Languages" %}</a>
     </li>
     <li>
-      <a href="#information" data-toggle="tab">{% translate "Info" %}</a>
+      <a href="#information" data-toggle="tab">{% translate "Overview" %}</a>
     </li>
     <li>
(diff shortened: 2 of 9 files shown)


AliceVisek (approved):

Geeyun-JY3:
> * the top level translation label is now "Translation", not "Overview" to make it consistent with other levels

I don’t think this is consistent with other levels, because the similar tabs in other levels list their sub-level objects, while the label of these tabs just reads the name of their sub-level objects. If the first tab in the translation level is labelled "Translation", it will be read as the name of the current level.

The rest is fine.

nijel:
@Geeyun-JY3 Any other suggestion for naming [this](https://hosted.weblate.org/projects/weblate/application/cs/#overview)? *Overview* will be the new name of the *Info* tab, so something else needs to be chosen.

Issue #3410, opened 2020-01-21
Errors atop Zen editing field


comradekingu:
![bilde](https://user-images.githubusercontent.com/13802408/72826510-46843400-3c79-11ea-803b-8c06898ce0c1.png)

When trying to copy something from a string by clicking source with the mouse, it shifts focus, thereby the string is saved.
The arising issue is popups for all failing checks being posted all over the screen, sometimes obscuring the editing field.

It could be that the placeholder isn't picked up here, but I am pretty sure it happens even if the placeholder is there (and clicked).

nijel:
@meel-hd Could zen saving be delayed for such events so that clicking into the source to insert placeholder doesn't trigger save of incomplete string?

* Losing focus sets timer to save the string.
* If the editor gets focus quickly again, the timer is cancelled?

meel-hd:
We can do that? it would reduce save attempts of incomplete strings.

meel-hd:
@nijel also we may need to update the notifactions pop ups. They take the so much width and show on the top section of the screen and stack with so much space in between. I saw some implementations that show them in the right bottom corner with less space between them.

PR #14821, opened 2025-05-08
Refactor translation editor save logic to improve focus handling and save timing

meel-hd:
Fixes:
- Stop save triggered when clicking source text or inserting placeholders.
- Cancel save operation after (1000ms) quick refocus
- Reduce possibility of multiple validation popups showing in the UI because of unintended saves

Close: #3410

### DIFF
--- a/weblate/static/editor/zen.js
+++ b/weblate/static/editor/zen.js
@@ -157,15 +157,8 @@
     const $this = $(this);
     focusTimeout = setTimeout(() => {
       if (!editorHasFocus) {
-        const $row = $this.closest("tr");
-        const checksum = $row.find("[name=checksum]").val();
-        const statusdiv = $(`#status-${checksum}`);
-        const form = $row.find("form");
-        const payload = form.serialize();
-        const lastPayload = statusdiv.data("last-payload");
-
-        // Trigger save if payload has changed or if it's the first save
-        if (payload && (lastPayload === undefined || payload !== lastPayload)) {
+        // Editor lost focus and has changes
+        if ($this.hasClass("has-changes")) {
           handleTranslationChange.call($this[0]);
         }
       }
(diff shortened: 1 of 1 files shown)


raiderss:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

weblate/static/editor/zen.js
```diff
@@ -145,52 +141,105 @@
...
+  $document.on("focusin", ".translation-editor", () => {
+    editorHasFocus = true;
+    // Focus returned quickly; cancel pending save
+    if (focusTimeout) {
+      clearTimeout(focusTimeout);
+      focusTimeout = null;
+    }
+  });
+
+  $document.on("focusout", ".translation-editor", function () {
```
nijel, line 156: Doesn't this trigger save even in case when there are no changes?

weblate/static/editor/zen.js
```diff
@@ -152,14 +152,22 @@
...
     focusTimeout = setTimeout(() => {
       if (!editorHasFocus) {
-        // Focus lost for over 1000ms, triggering save
-        handleTranslationChange.call($this[0]);
+        const $row = $this.closest("tr");
+        const checksum = $row.find("[name=checksum]").val();
+        const statusdiv = $(`#status-${checksum}`);
+        const form = $row.find("form");
+        const payload = form.serialize();
+        const lastPayload = statusdiv.data("last-payload");
```
nijel, line 165: This would always save on first focus out right? So if user just navigates with TAB, *all* entries would be saved? Can't it use `has-changes` class instead?

weblate/static/editor/zen.js
```diff
@@ -145,52 +141,111 @@
...
-    if (payload === statusdiv.data("last-payload")) {
+    // First save
+    if (lastPayload === undefined) {
+      statusdiv.data("last-payload", payload);
+      // Check if the editor doesn't have changes
+      if (!$this.closest(".translation-editor").hasClass("has-changes")) {
+        return;
+      }
+    }
+    // Guard: skip if nothing has changed
```
nijel, line 205: Isn't has-changes already covering this?

KarenKonou (approved):
One thing that would make this better in my opinion is to give the user some subtle visual feedback that the save timeout has started. Right now it feels quite opaque, after focusout nothing happens for a while until the actual save action triggers.

nijel:
Good point, let's deal with that separately - https://github.com/WeblateOrg/weblate/issues/15261.

PR #14869, opened 2025-05-14
feat(users): make invitation subject more describing

nijel:
Include the username and project in the subject so that it provides more info.

The full subject now looks like (project and site scoped invitations):

* [Weblate] testuser has invited you to join the Test project
* [Weblate] testuser has invited you to join


orangesunny (approved):
texts good

Issue #14872, opened 2025-05-14
Improve unpaid support nagging


orangesunny:
### Describe the problem

We need to better inform the users about their unpaid dedicated instance/support package

### Solution brainstorm

- [x] if the backups are turned off as a result of unpaid service, inform about that in the instance: at both /manage and /performance sites
- [x] show red announcement about unpaid service to all superusers and admins
- [x] at /manage page, write _Unpaid subscription_ (with link to /user) instead of _Community support_
- [x] send regular e-mail notifications for unpaid instances forever instead of the current state (only for a period of time after unpaid happens)
   - [ ] in these forever notification, include a link where recipients can show they will to terminate the service
   - [ ] regularly (start one month after being unpaid) send these forever e-mail notification to all server superuser and adminsof such dedicated instance/extended with unpaid subscription

orangesunny:
related: https://github.com/WeblateOrg/internal/issues/363

jfaltenbacher:
If I may add a wish: send reminder one month prior to the date of payment

nijel:
https://github.com/WeblateOrg/website/pull/2753 improves notifications from weblate.org

nijel:
There are still some bits missing for the website (add a way to terminate subscription by user), but as for Weblate code, this is completed.

PR #14890, opened 2025-05-15
feat: improve reporting broken backups or unpaid subscriptions

nijel:
- make unpaid subscription and configuration error
- indicate which backup services have errors in the listing
- add explanation if subscribed backup no longer works with unpaid subscription
- add custom status when using community support but had a subscription before

Issue #14872


orangesunny (approved):
texts good to go

PR #14892, opened 2025-05-15
feat: add PWA support with caching and offline support

meel-hd:
Weblate now can be installed on users OS as a typical app.
## The app is installed without the browser shell

If the  user go offline he still can use the app(the cached views)

### DIFF
--- a/weblate/templates/js/service-worker.js
+++ b/weblate/templates/js/service-worker.js
@@ -1,0 +1,84 @@
+const APP_VERSION = "{{ version }}";
+const CACHE_NAME = `weblate-pwa-cache-${APP_VERSION}`;
+
+const urlsToCache = [
+  // Core pages
+  "/",
+  "/dashboard/",
+  "/projects/",
+  "/languages/",
+  "/changes/",
+
+  // User account pages
+  "/accounts/profile/",
+  "/accounts/login/",
+  "/accounts/register/",
+
+  // Common static assets
+  "/static/css/style.css",
+  "/static/js/main.js",
+  "/static/weblate.js",
+  "/static/bootstrap/css/bootstrap.min.css",
+  "/static/bootstrap/js/bootstrap.bundle.min.js",
+  "/static/font-awesome/css/font-awesome.min.css",
+  "/static/favicon.ico",
+  "/static/weblate-192.png",
+  "/css/custom.css",
+  "/js/i18n/",
+
+  "/site.webmanifest",
+  "/robots.txt",
+
+  // Common translation paths
+  "/browse/",
+  "/translate/",
+  "/zen/",
+
+  // The service worker itself
+  "/service-worker.js",
+];
+
+// Install event: Pre-cache static assets
+self.addEventListener("install", (event) => {
+  event.waitUntil(
+    caches.open(CACHE_NAME).then((cache) => {
+      return cache.addAll(urlsToCache);
+    }),
+  );
+});
+
+// Fetch event: Network-first strategy that falls back to cache
+self.addEventListener("fetch", (event) => {
+  event.respondWith(
+    fetch(event.request)
+      .then((response) => {
+        // Clone the response and store it in the cache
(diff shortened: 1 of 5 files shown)


meel-hd:
The `py3.11, postgresql, minimal deps` run fails, due to DB flush error and Selenium test (likely cascading).

```txt
FAILED weblate/trans/tests/test_selenium.py::SeleniumTests::test_fonts - django.core.management.base.CommandError: Database test_weblate_gw1 couldn't be flushed. Possible reasons:
  * The database isn't running or isn't configured correctly.
  * At least one of the expected database tables doesn't exist.
  * The SQL was invalid.
Hint: Look at the output of 'django-admin sqlflush'. That's the SQL this command wasn't able to run.
```
I am not sure if it is related to the changes made. it only appears on  `py3.11, postgresql, minimal deps`.

nijel (commented):
I still fail to see how this is useful for offline work.

meel-hd:
> I still fail to see how this is useful for offline work.

This isn't only about offline work:
- if users have unstable or lose connection they can still view the app instead of nothing. 
- The app now can be installed on devices without the browser tabs, navigation, etc. I think it offer greater experience. And all of this without that complex of technical implementation. It is worth it.

weblate/templates/js/service-worker.js
```diff
@@ -0,0 +1,84 @@
+const APP_VERSION = "{{ version }}";
+const CACHE_NAME = `weblate-pwa-cache-${APP_VERSION}`;
+
+const urlsToCache = [
+  // Core pages
+  "/",
+  "/dashboard/",
```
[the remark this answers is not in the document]
nijel: My comment still applies, this URL does not exist.

weblate/templates/js/service-worker.js
```diff
@@ -0,0 +1,77 @@
+const APP_VERSION = "{{ version }}";
+const CACHE_NAME = `weblate-pwa-cache-${APP_VERSION}`;
+
+const urlsToCache = [
+  // Core pages
+  "/",
+  "/projects/",
```
KarenKonou: Does this only cache the root `/projects/` path or all subpaths too? If it's the former it might not be particularly useful for offline work, but on the flipside the later case might be a lot of data to download upfront when installing.

KarenKonou (commented):
I think a proper offline experience would require a bit more work, for example registering [sync events](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Guides/Offline_and_background_operation#background_sync0) on forms submissions, using background sync or fetch to automatically update data when connection is regained, caching more content upfront or more gracefully handling cases where network is unavailable and cache misses. As it stands the implementation is quite barebones.

Also I'm not sure if the PWA user experience is that good for Weblate currently, being designed as a regular web application I found it a little awkward to use without a browser back button.

Additionally PWAs are still not supported by every browser, notably Firefox doesn't support them at the moment.

weblate/templates/js/service-worker.js
```diff
@@ -0,0 +1,77 @@
...
+  "/languages/",
```
nijel: Let's remove these as they are not frequently used:

```suggestion
```

nijel:
@KarenKonou I've removed some cached paths. Still, the main question would be: Is this something that will bring some value to users as it is? Offline work is not something we want to focus on for now.

weblate/templates/site.webmanifest
```diff
@@ -2,6 +2,7 @@
 {
     "name": "{{ site_title }}",
     "short_name": "{{ site_title }}",
+    "start_url": "/",
```
nijel:
```suggestion
    "start_url": "{% reverse 'home' %}",
```

nijel:
I've done some cleanups in d0eb30f3dc0f659590351e655de6ceb089781fec and I think this is good to go as a way to install Weblate. There is no offline support and I don't expect we will add it in the near future.

PR #14913, opened 2025-05-18
feat(api): enable project creation within billing limits

### LINKED ISSUE #14854: Create projects using API with billing
### Describe the problem

When billing is enabled, it grants user certain limits of projects to be created. This should also allow creating projects using API in such case even if he doesn't have `project.add` permission.

### Solution brainstorm

* Allow to create projects with billing specified (if they fit inside billing limits).
* Automatically choose a billing for user if it has access to one and none is specified (most users will have just a single billing).

This logic might be shared with `CreateProject` so that UI and API use the same.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

amCap1712:
Added support for non-superusers to create projects if they have billing enabled and the project limit is not exceeded. Currently, a within limit billing is automatically used for the project. Since most users have only one billing, it should work fine for most cases.

To let users with multiple billings choose the desired billing, the billing id or a similar identifying field needs to be exposed to the users. Currently, the user can only view the billing id from the url of the billing details page, which is not exactly ideal. Thoughts?

Fixes #14854.


orangesunny (approved):
fine in terms of language. @nijel, please check the code.

weblate/api/tests.py
```diff
@@ -1148,6 +1150,74 @@ def test_create(self) -> None:
         )
         self.assertEqual(Project.objects.count(), 2)
 
+    def test_create_with_billing(self) -> None:
+        installed_apps = settings.INSTALLED_APPS
+        if "weblate.billing" not in installed_apps:
+            installed_apps = ["weblate.billing", *settings.INSTALLED_APPS]
+        with override_settings(INSTALLED_APPS=installed_apps):
```
nijel, line 1157: Using `modify_settings` instead would make this code simpler.

weblate/api/views.py
```diff
@@ -1112,26 +1112,29 @@ def addons(self, request: Request, **kwargs):
...
+                try:
+                    billing = Billing.objects.for_user_within_limits(self.request.user)[
+                        0
+                    ]
+                except IndexError:
+                    self.permission_denied(
+                        request, "No valid billing found or limit exceeded."
+                    )
+            else:
+                self.permission_denied(request, "Can not create projects")
```
nijel, line 1129: Can you please add test to cover billing not installed and project creation denied?

nijel (changes requested):

weblate/api/tests.py
amCap1712: Done.

weblate/api/views.py
amCap1712: Done.

nijel:

nijel (approved):

PR #14914, opened 2025-05-19
feat(api): enable project team creation

### LINKED ISSUE #14853: Create teams using API
### Describe the problem

Currently, creating teams in API requires site-wide permissions. There should be a way to create project-bound teams for project admins.

### Solution brainstorm

`GroupViewSet.create` should allow creating new group with `defining_project` pointing to a project where user has `project.permissions` permission (probably `meta:team.edit` permission handler could be extended to support `Project` instances as well as the logic is already there).

_No response_

### Screenshots

_No response_

### Additional context

_No response_

amCap1712:
Added support for non-superusers to create project scope teams.

Fixes #14853.


orangesunny (approved):
language OK. @nijel, please check the code

weblate/api/views.py
```diff
@@ -607,8 +616,16 @@ def update(self, request: Request, *args, **kwargs):
...
-        self.perm_check(request)
-        return super().create(request, *args, **kwargs)
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        self.perm_check(
+            request, project=serializer.validated_data.get("defining_project")
+        )
+
+        self.perform_create(serializer)
```
nijel, line 626: Wouldn't it be simpler to move the permission check inside `perform_create` and avoid overriding `create`?

amCap1712: Ah yeah makes sense.

nijel:

Issue #14798, opened 2025-05-06
Add message when moving between components while translating


nijel:
### Describe the problem

When translating on category or project level, user might not be aware of the individual components.

### Solution brainstorm

> This could be coupled with a warning for the user, something like "You were translating strings for component X, this string is part of component Y." so that the change is clearly visible when translating loads of strings at times from several different components. 

 _Originally posted by @Cwpute in [#14296](https://github.com/WeblateOrg/weblate/issues/14296#issuecomment-2757356922)_

Issue #9953, opened 2023-09-14
Use unit absolute URLs for translating


nijel:
### Describe the problem

Users often share link to a current search, what is not really a good way to share link to a unit.

### Describe the solution you'd like

* The actual translation URLs will include unit ID (via `get_url_path`)
* Search will navigate through a (session cached) list of matches
* If the current unit is not part of the search, a warning will be shown, but the user will still stay on that unit

PR #14921, opened 2025-05-19
Show message when moving between components while translating 

KasukabeDefenceForce:
Fixes https://github.com/WeblateOrg/weblate/issues/14798

When translating on category or project level, make user aware if the component of unit changes.

### DIFF
--- a/weblate/trans/views/edit.py
+++ b/weblate/trans/views/edit.py
@@ -278,6 +278,7 @@ def search(
             "name": str(name),
             "ids": unit_ids,
             "ttl": now + SESSION_SEARCH_CACHE_TTL,
+            "last_viewed_unit_id": None,
         }
         if use_cache:
             request.session[session_key] = store_result
@@ -638,13 +639,9 @@ def translate(request: AuthenticatedHttpRequest, path):
             messages.error(request, gettext("Invalid search string!"))
             return redirect(obj)
 
-    previous_unit_id = (
-        int(request.session["previous_unit_id"])
-        if "previous_unit_id" in request.session
-        else None
-    )
-    if previous_unit_id:
-        previous_unit = unit_set.get(pk=previous_unit_id)
+    last_viewed_unit_id = search_result.get("last_viewed_unit_id")
+    if last_viewed_unit_id:
+        previous_unit = unit_set.get(pk=last_viewed_unit_id)
         if unit.translation.component != previous_unit.translation.component:
             messages.warning(
                 request,
@@ -654,7 +651,7 @@ def translate(request: AuthenticatedHttpRequest, path):
                     "current": unit.translation.full_slug,
                 },
             )
-    request.session["previous_unit_id"] = unit.id
+    search_result['last_viewed_unit_id'] = unit.id
 
     # Some URLs we will most likely use
     base_unit_url = "{}?{}&offset=".format(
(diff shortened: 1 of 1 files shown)


weblate/trans/views/edit.py
```diff
@@ -638,12 +638,27 @@ def translate(request: AuthenticatedHttpRequest, path):
             messages.error(request, gettext("Invalid search string!"))
             return redirect(obj)
 
+    from_offset = int(request.GET.get("from_offset", offset - 1))
```
nijel, line 641: This will crash with a blank value (`form_offset=`).

weblate/trans/views/edit.py
```diff
@@ -638,12 +638,27 @@ def translate(request: AuthenticatedHttpRequest, path):
...
+    if from_offset and 1 <= from_offset <= len(search_result["ids"]):
+        previous_unit = unit_set.get(pk=search_result["ids"][from_offset - 1])
+
+        if unit.translation.component != previous_unit.translation.component:
+            messages.warning(
+                request,
+                gettext("You have shifted from %(previous)s to %(current)s.")
+                % {
+                    "previous": previous_unit.translation.component.name,
+                    "current": unit.translation.component.name,
```
nijel, line 651: Name is not good enough here as it does not have to be unique, at least categories should be included. Perhaps using full component name including project would be better for consistency with other views.

nijel (commented):
Do you see a reasonable way to make it work without passing this info in request? The long-term plan is to use unit URLs while navigating (see https://github.com/WeblateOrg/weblate/issues/9953).

weblate/trans/views/edit.py
```diff
@@ -638,27 +638,26 @@ def translate(request: AuthenticatedHttpRequest, path):
             messages.error(request, gettext("Invalid search string!"))
             return redirect(obj)
 
-    from_offset = int(request.GET.get("from_offset", offset - 1))
-    if from_offset and 1 <= from_offset <= len(search_result["ids"]):
-        previous_unit = unit_set.get(pk=search_result["ids"][from_offset - 1])
-
+    previous_unit_id = int(request.session['previous_unit_id']) if 'previous_unit_id' in request.session else None
```
nijel: This should be pinned to the search result, otherwise it will have too many corner cases:

* Translating one project, then starting translating another project would yield the message.
* Opening multiple translation tabs they will fight for the same previous_unit_id.

It could stored inside the search data which we already have in the `search` function.

weblate/trans/views/edit.py
```diff
@@ -638,6 +638,24 @@ def translate(request: AuthenticatedHttpRequest, path):
             messages.error(request, gettext("Invalid search string!"))
             return redirect(obj)
 
+    previous_unit_id = (
+        int(request.session["previous_unit_id"])
+        if "previous_unit_id" in request.session
+        else None
+    )
```
nijel: This should be pinned to the search result; otherwise it will have too many corner cases:

* Translating one project, then starting translating another project would yield the message.
* Opening multiple translation tabs they will fight for the same previous_unit_id.

It could be stored inside the search data, which we already have in the `search` function above.

abhinavohri: okay

nijel:

Issue #14923, opened 2025-05-19
SAML integration: Use RelayState instead of authid for session restoration


javiboo:
Describe the problem:
Currently, the SAML integration appends an `authid` query parameter to the Assertion Consumer Service (ACS) URL to restore the session after SAML authentication. Many SAML Identity Providers (IdP) require the ACS URL to match exactly, and the extra parameter causes authentication to fail with errors such as:

```
Invalid request, ACS Url in request ... doesn't match configured ACS Url ...
```

This strict matching breaks SAML login for platforms that do not allow query parameters in the ACS URL.

Solution brainstorm:
The session restoration mechanism should use the SAML RelayState parameter for passing `authid` or similar session data instead of appending it to the ACS URL as a query parameter. The SAML RelayState is designed for this purpose and is returned unmodified by the IdP, making it suitable for state restoration without breaking ACS URL matching.

Describe alternatives you have considered:
- Making the inclusion of `authid` in the ACS URL optional/configurable (not robust, still breaks for most strict IdPs)
- Documenting the issue and asking users to relax ACS URL matching (not always possible)

Screenshots:

Additional context:
This problem is tracked in the following function in the codebase:
https://github.com/WeblateOrg/weblate/blob/main/weblate/accounts/views.py#L1343

Fixing this would improve compatibility with a wider range of SAML IdPs and make SAML authentication more robust.

nijel:
We need to be able to preserve certain data over login and sessions are currently used for this. Unfortunately, the login from SAML ends with POST request back and due to `SameSite` cookie policy for the session cookie the cookie is not sent, and thus the session cannot be restored. We used the same workaround as worked for OpenID, but apparently it has its issues.

This is also an upstream issue described at https://github.com/python-social-auth/social-app-django/issues/481.

nijel:
https://github.com/python-social-auth/social-core/pull/1159 and https://github.com/python-social-auth/social-app-django/pull/705 is my attempt to address this.

PR #14924, opened 2025-05-19
fix: do not send authid parameter in saml backend

javiboo:
Describe the problem:
Currently, the SAML integration appends an `authid` query parameter to the Assertion Consumer Service (ACS) URL to restore the session after SAML authentication. Many SAML Identity Providers (IdP) require the ACS URL to match exactly, and the extra parameter causes authentication to fail with errors such as:

```
Invalid request, ACS Url in request ... doesn't match configured ACS Url ...
```

This strict matching breaks SAML login for platforms that do not allow query parameters in the ACS URL.

Solution brainstorm:
The session restoration mechanism should use the SAML RelayState parameter for passing `authid` or similar session data instead of appending it to the ACS URL as a query parameter. The SAML RelayState is designed for this purpose and is returned unmodified by the IdP, making it suitable for state restoration without breaking ACS URL matching.

Describe alternatives you have considered:
- Making the inclusion of `authid` in the ACS URL optional/configurable (not robust, still breaks for most strict IdPs)
- Documenting the issue and asking users to relax ACS URL matching (not always possible)

Screenshots:

Additional context:
This problem is tracked in the following function in the codebase:
https://github.com/WeblateOrg/weblate/blob/main/weblate/accounts/views.py#L1343

Fixing this would improve compatibility with a wider range of SAML IdPs and make SAML authentication more robust.

Related to the issue: https://github.com/WeblateOrg/weblate/issues/14923

### DIFF
--- a/weblate/accounts/views.py
+++ b/weblate/accounts/views.py
@@ -1338,7 +1338,7 @@ def social_auth(request: AuthenticatedHttpRequest, backend: str):
         msg = "Backend not found"
         raise Http404(msg) from None
 
-    # Store session ID for OpenID or SAML based auth. The session cookies will
+    # Store session ID for OpenID based auth. The session cookies will
     # not be sent on returning POST request due to SameSite cookie policy
     if isinstance(request.backend, OpenIdAuth):
         request.backend.redirect_uri += "?authid={}".format(
--- a/weblate/accounts/views.py
+++ b/weblate/accounts/views.py
@@ -1340,7 +1340,7 @@ def social_auth(request: AuthenticatedHttpRequest, backend: str):
 
     # Store session ID for OpenID or SAML based auth. The session cookies will
     # not be sent on returning POST request due to SameSite cookie policy
-    if isinstance(request.backend, OpenIdAuth) or backend == "saml":
+    if isinstance(request.backend, OpenIdAuth):
         request.backend.redirect_uri += "?authid={}".format(
             dumps(
                 (request.session.session_key, get_ip_address(request)),


javiboo:
In my experience with setting up many other SAML integrations, the standard and compatible way to pass dynamic state, like a session ID or context (authid in this case), across the SAML authentication flow is by utilizing the RelayState parameter.

Based on my experience configuring SAML with various Identity Providers like Google Workspace, Microsoft Entra ID (Azure AD), and others, none of these platforms allow or expect a dynamic ACS URL with query parameters in their configuration or in the AuthnRequest for matching purposes. They strictly require a static, predictable endpoint.

However, the mechanism you are currently using is preventing SAML users from logging in. Don't you think it's better to remove it and at least keep the possibility of access rather than keeping something that prevents users from logging in?

nijel:
SAML now works for some users while preserving the URL over login. Removing this would be regression for those users and they would complain. Therefore, I prefer a solution that will keep existing behavior.

javiboo:
> SAML now works for some users while preserving the URL over login. Removing this would be regression for those users and they would complain. Therefore, I prefer a solution that will keep existing behavior.

Are you sure the SAML integration works? I don't know any IdP that allows dynamic configuration of the ACS URL. Can you give me an example of an IdP that allows this?

javiboo:
> Hmm, maybe it now works out of the box since [python-social-auth/social-core@b631796](https://github.com/python-social-auth/social-core/commit/b6317968afe5b1701cdec8fd1d00d03f1a16acbc). Can you test that `next` parameter is preserved over SAML login without `authid`?

As long as the authid is being sent as a parameter in the request URI, I can't test anything since authentication never occurs in Weblate. How do you suggest I test that?

nijel:
> Are you sure the SAML integration works?

I'm sure it currently works with https://www.okta.com/, and I think it also works with Azure SAML (it definitely worked in the past).

> As long as the authid is being sent as a parameter in the request URI, I can't test anything since authentication never occurs in Weblate. How do you suggest I test that?

You can apply the patch locally.

Looking at it again, I doubt that the `RelayState` solution in python-social-auth will fully work for Weblate as there are more session saved states than redirect URL (for example invitations).

javiboo:
> > Are you sure the SAML integration works?
> 
> I'm sure it currently works with https://www.okta.com/, and I think it also works with Azure SAML (it definitely worked in the past).
> 

I've checked the Okta documentation, and nowhere does it state that wildcards can be used in the ACS URL:
https://help.okta.com/en-us/content/topics/apps/aiw-saml-reference.htm

The only way to define multiple ACS URLs is through 'Other Requestable URLs,' but these don't support wildcards either:
https://support.okta.com/help/s/question/0D5KZ00000NfxSa0AJ/query-based-sp-acs-urls-are-not-allowed-in-other-requestable-url-section-of-okta-saml-app?language=en_US

> > As long as the authid is being sent as a parameter in the request URI, I can't test anything since authentication never occurs in Weblate. How do you suggest I test that?
> 
> You can apply the patch locally.
> 
> Looking at it again, I doubt that the `RelayState` solution in python-social-auth will fully work for Weblate as there are more session saved states than redirect URL (for example invitations).

I think you should reconsider using authid in SAML while we find a solution.

I'm pretty sure the SAML integration hasn't worked since version 5.11.X with any IdP. We downgraded to version 5.10.4.0 and the SAML integration started working again.

weblate/accounts/views.py
```diff
@@ -1340,7 +1340,7 @@ def social_auth(request: AuthenticatedHttpRequest, backend: str):
 
     # Store session ID for OpenID or SAML based auth. The session cookies will
```
nijel, line 1341:
```suggestion
    # Store session ID for OpenID based auth. The session cookies will
```

PR #14929, opened 2025-05-20
fix(fonts): pass font weight for size rendering

michael-smt:
Not sure if this actually fixes a issue, but for completeness' sake I think it should be passed to `_render_size()`.


nijel (approved):

nijel:

PR #14959, opened 2025-05-22
fix(unit): keep previous source when present

### LINKED ISSUE #14957: Source change only shows the latest iteration
### Describe the issue

A string was changed twice, instead of Weblate showing the difference since the last translation, it instead only shows the difference in the source string in the last iteration.

The string went: Storage restore -> File backup restore -> File restore. However, Weblate only shows as if the changes to make are to remove backup.

See https://hosted.weblate.org/translate/calyxos/seedvault_storage/ar/?checksum=9c96d1f37e1c7401

### I already tried

0. You seem to need to have a remote set up, not a local weblate component
1. Have translations to a string
2. Commit and push
3. Change a string twice, with Weblate updating between each change
4. !

### Expected behavior

It should show the difference since the translation was last modified.

### Screenshots

_No response_

### Exception traceback

```pytb

```

### How do you run Weblate?

weblate.org service

### Weblate versions

_No response_

### Weblate deploy checks

```shell

```

### Additional context

_No response_

nijel:
If the source string was changed twice, keep previous change.

Fixes #14957

### DIFF
--- a/weblate/trans/models/unit.py
+++ b/weblate/trans/models/unit.py
@@ -895,6 +895,10 @@ class Unit(models.Model, LoggerMixin):
                     # Store previous source and fuzzy flag for monolingual
                     if not previous_source:
                         source_change = previous_source = self.source
+                        # Keep prevoious source if already set in case source
+                        # changes multiple times
+                        if self.previous_source:
+                            previous_source = self.previous_source
                     state = STATE_FUZZY
                 pending = True
             elif (


lucasmz-dev:
quick question, would this apply to existing diffs, or only to new ones? useful to know so things can be fixed

nijel:
Only to new ones.

Issue #12795, opened 2024-10-16
Parse extended info from TermWeb TBX


nijel:
### Describe the problem

TermWeb is a tool for managing terminology and supports TBX export with a lot of additional metadata. Weblate should understand this as much as possible.

### Describe the solution you would like

* `<descrip type="definition">Explanation</descrip>` should be parsed as explanation
* `<descrip type="Translation needed">No</descrip>` should convert to `read-only` flag (untranslatable term)
* `<termNote type="administrativeStatus">forbidden</termNote>` (or `obsolete`) should convert to `forbidden` flag
* `<descrip type="Usage note">Note</descrip>` should be parsed a note

### Additional context

Example files: https://github.com/openSUSE/suse-terminology/tree/main/TBX/multilingual/for_Weblate, especially https://github.com/openSUSE/suse-terminology/blob/main/TBX/multilingual/for_Weblate/SUSE_for_Weblate_all_lang_req_fields.xml

nijel:
@jfaltenbacher Do you see something missing in this specification? I tried to look up useful info in the files you've shared, but I might have missed something.

jfaltenbacher:
@nijel, just these 4 fields:

- Translation needed
- Definition
- Usage status
- Usage note

You caught them all :)

amCap1712:
The `definition` is currently parsed as `source_explanation`. IIUC, it should be removed from `source_explanation` and be parsed as `explanation` ?

nijel:
I think it's fine to keep it that way. We definitely do not want to change behavior for existing glossaries.

snoerenberg:
Hi, I still struggle to get TBX files imported.

When adding this testfile:
https://github.com/openSUSE/suse-terminology/blob/main/TBX/multilingual/for_Weblate/Export%20for%20Weblate.xml

To a repository I need to rename it to "Export for Weblate_en.xml" that it is found during scan.
Also the manual upload need to have a zip with the same file within.

BUT: in both ways nothing is parsed.

<img width="1585" height="596" alt="Image" src="https://github.com/user-attachments/assets/7c57b665-de7f-48a4-8551-325d2df51423" />

<img width="953" height="358" alt="Image" src="https://github.com/user-attachments/assets/db8a3f42-45d2-4c6f-b687-4b89507a75a5" />

The file by Suse is "multilingual" because it contains all languages within one.
On the help page https://docs.weblate.org/en/latest/formats.html it is written as "biligual".

[...]

PR #14998, opened 2025-05-24
feat(tbx): handle extended metadata in tbx format

amCap1712:
Handle TermWeb extended metadata: 'definition', 'Translation needed', 'administrativeStatus' and 'Usage note' in accordance with https://github.com/WeblateOrg/weblate/issues/12795.

The `_get_notes_list` method is reimplemented in `TBXUnit` based on translate toolkit's `tbxunit.getnotes` to appropriately filter the extended metadata terms based on node tag and attributes.


nijel:
Indeed pushing that to translate-toolkit seems reasonable. For the various comments, it could be also implemented there, the `getnotes` has `origin` parmeter to distinguish different kinds.

amCap1712:
https://github.com/translate/translate/pull/5601. For comments, explanation does not seem to be a unit field in translate-tooklit.

nijel:
https://github.com/translate/translate/releases/tag/3.15.3 is out with the changes, you can now base on that after adjusting the minimal required version.

nijel:

PR #15005, opened 2025-05-25
feat(announcement): Add permission for posting announcements and new role

### LINKED ISSUE #8974: Add separate 'post announcements' permission
### Describe the problem

For AntennaPod, we have a volunteer dedicated to translations coordination. This person engages with translators and representing them in the core team.

We have defined a Translations Coordinator 'team' and assigned it different roles that allow them to do their work, without introducing permissions that could potentially break things. For example, the Admin role is not suitable because it allows to manage components.
afbeelding

In order to support their work, they need a way to reach out to existing translators. They might, for example, announce that a new app version is upcoming, and that we would like to get all languages up to 100%. We currently cannot make that happen.

Introduce a new 'post announcements' permission. This permission should allow the user to post announcements, limited by the permissions they have on component and/or language level (see also #8973).

This permission should be assigned to a user via a (default) role. I would suggest introducing a new 'Translation coordinator' default role with the permissions that we have defined (as you can see above) as well as this new permission to send announcements.

I hesitated to request the addition of a new default role, because I think it implies extra work. The alternative I considered is assigning this permission to the 'Power user' role. However, the permission to send announcements is quite extensive, as it allows someone to represent/reach out on behalf of the project. That is no longer a (power) user role, but rather a management role.

### Screenshots

_No response_

### Additional context

_No response_

amCap1712:
Introduce a new `announcement.add` permission to control access to announcement posting. Also, add a new role "Translation coordinator" to include this permission. Fixes #8974.

Questions:

1. I haven't updated the documentation because the name and permissions for the to be introduced "Translation coordinator" are not yet finalized. Currently, I am using the same as mentioned in the linked issue.
2. There is a dummy `announcement.delete` permission. https://github.com/WeblateOrg/weblate/blob/6a5a25c34cd8a1f8a49e29f3d7f3de696b4ce8af/weblate/auth/permissions.py#L530-L531 I think that permission should either be merged with the `announcement.add` permission or converted to separate real permission.
3. There is an existing permission called `announcement.edit` on the global level. Should this new permission (and the delete permission) be renamed to something different to avoid confusion?

Reminder: add changelog note to run `setupgroups` to update role.


amCap1712:
Updated the docs. I'll open a ticket for making `announcement.delete` a real permission and adding it to the translation coordinator role.

PR #15011, opened 2025-05-26
feat(memory): add option to disable personal and project TM contribution

amCap1712:
Introduce `contribute_personal_tm` for user profiles and `contribute_project_tm` for components, to enable selective contribution to translation memory. If `contribute_personal_tm` is disabled, the user's translation memory is not updated. If `contribute_project_tm` is disabled for a component, the associated project's translation memory is not updated.

Question: Should the shared instance memory also be not updated when the `contribute_project_tm` is disabled for a component?


weblate/memory/tasks.py
```diff
@@ -45,6 +46,72 @@ def import_memory(project_id: int, component_id: int | None = None) -> None:
                 handle_unit_translation_change(unit, None, component, project)
 
 
+@app.task(trail=False)
+def import_user_memory(user_id: int) -> None:
+    """
+    Import all translations from a user into their personal translation memory.
+
+    This is used when a user enables personal translation memory after having it disabled.
+    """
```
nijel, line 55: I'm not sure if we want this at all.

weblate/memory/tasks.py
```diff
@@ -45,6 +46,72 @@ def import_memory(project_id: int, component_id: int | None = None) -> None:
...
+    if not user.profile.contribute_personal_tm:
+        return
+
+    with transaction.atomic():
+        units = (
+            Unit.objects.filter(
+                Q(change__author=user) | Q(change__user=user),
+                state__gte=STATE_TRANSLATED,
+                translation__component__is_glossary=False,
+            )
```
nijel, line 69: This will return all units where the user has ever contributed, not just the ones where the translation he made is currently used. If we want this functionality, it would have to be done based on the `Change` model.

weblate/trans/models/component.py
```diff
@@ -1006,6 +1020,9 @@ def save(self, *args, **kwargs) -> None:
         for project in self.links.all():
             project.invalidate_source_language_cache()
 
+        if update_tm:
+            import_memory.delay_on_commit(self.project.id, self.pk)
```
nijel, line 1024: Can we reasonably limit this to component only?

nijel (commented):
>  Should the shared instance memory also be not updated when the contribute_project_tm is disabled for a component?

Yes, it makes sense to make it as opt out from the translation memory.

On the other side, I think user setting should influence only the user translation memory (what your correctly did).

weblate/memory/tasks.py
nijel: Any feedback on the review comments?

amCap1712: Hi! Sorry I missed these comments. Checking now.

weblate/memory/tasks.py
amCap1712: I see, in that I'll remove it for now and add later if there is demand for it.

weblate/memory/tasks.py
amCap1712: IIUC, that would mirror the current behavior for user translation memory no? If a user makes a translation its added to their translation memory but it is not removed from their translation memory if another user changes the translation?

Either way, removing it for now so shouldn't be an issue.

weblate/trans/models/component.py
amCap1712: Sure, I'll try to refactor the code accordingly.

amCap1712: Checking the code again, it already filters on the component:
https://github.com/WeblateOrg/weblate/blob/66bfefae9ae7348b568156d6288e49026bb6a7c4/weblate/memory/tasks.py#L30

nijel:

PR #15042, opened 2025-05-30
Change webhook signature verification implementation

### LINKED ISSUE #15035: Use reference implementation for webhooks signing
### Describe the issue

Currently, we have copy of standard webhooks signing and verification. Instead it should use library https://pypi.org/project/standardwebhooks/

### I already tried

1. Look at:
  
   https://github.com/WeblateOrg/weblate/blob/1e809421eb83e83fc3d9d0259004f6f8fbb11d1d/weblate/addons/webhooks.py#L143-L218

### Expected behavior

_No response_

### Screenshots

_No response_

### Exception traceback

```pytb

```

### How do you run Weblate?

weblate.org service

### Weblate versions

_No response_

### Weblate deploy checks

```shell

```

### Additional context

_No response_

gersona:
- Use [standardwebhooks](https://pypi.org/project/standardwebhooks/) library for Webhooks request signature (and validation)
- `Secret` string is now required for Webhooks add-on configuration (because of library's implementation)
- fixes #15035


pyproject.toml
```diff
@@ -157,6 +157,7 @@ dependencies = [
   "siphashc>=2.1,<3.0",
   "social-auth-app-django>=5.4.2,<6.0.0",
   "social-auth-core>=4.5.4,<5.0.0",
+  "standardwebhooks>=1.0.0",
```
nijel, line 160: Let's add upper bound so that we can notice API breakages.

```suggestion
  "standardwebhooks>=1.0.0,<1.1",
```

Issue #10304, opened 2023-10-30
Allow editing the key of a translation unit/source string


tipa:
### Describe the problem

I wish to change the key of a source string. This is easily done in my source files ("Find & replace"), but how can I reflect that change in Weblate? I can add and remove source strings in Weblate, but I cannot rename/change the key of a source string

### Describe the solution you'd like

A new menu item "change key":
![image](https://github.com/WeblateOrg/weblate/assets/2863198/ba484643-34dd-4a9c-8d60-34c5703d68de)

or make the "Key" field of the source string editable:
![image](https://github.com/WeblateOrg/weblate/assets/2863198/d2aa720e-516b-4082-9052-f289561869c4)

### Describe alternatives you've considered

I could create a new string in the Weblate UI, but I would then have to update all translations manually and then remove the old string manually.
I could also re-upload all source files after renaming the key, but this would be a lot of work for many languages and it could mess with the History or Labels of the translations

### Additional context

I created my components by manually uploading the translation source files. I am not using GitHub or similar integrations.

tipa:
Any reason why this won't be considered?

tipa:
Happy to give some more context on why I had this situation a few times already when working with translations for my apps:
- Sometimes I find a typo in the source string key that I want to fix. The string value as well as its translations stay the same.
- I name the keys of my source strings like the source string value. E.g. something like `key: "Dialog_Cancel" - value: "Cancel"`. Occasionally I find a better naming for those texts and I might change "Cancel" to "Abort". I would then also change the key `Dialog_Cancel` or `Dialog_Abort`. I would not want to add a new translation unit as "Cancel" and "Abort" is too similar and the translations for the string would likely still be correct (e.g. both "Abort" and "Cancel" translate to "Abbrechen" in German)

nijel:
I understand the motivation, but in most cases this needs a corresponding change in the source code as well. Therefore, this would be an opt-in feature and only few users would actually use that.

Ionys320:
You're right @nijel, but even adding a translation key often requires a change in the source code. I'm currently renaming a lot of keys by removing them and readding them (because the keys of my project is based on the localisation of the translation in the project (page/component/etc)). 

By the way, if the changes are not too complex, I may work on it.

nijel:
Yes, this would be a useful feature for some use-cases, but we're not going to push it on our own.

What I believe would work:

[...]

Issue #9729, opened 2023-08-11
Editing source string in bilingual formats


yilmazdurmaz:
### Describe the problem

I am trying to clear up the source glossary of the project I work on. there are thousands of items added with an enormous amount not related to glossary (even punctuations). I am mostly deleting them, but others need editing on their explanations; the topic of this request.

There is a problem (bug?) in adding terminology items, such that the "explanation" I added in a language does not go into the "Source string explanation" of the string in the project's source language. It can still be seen in "adder" language, but not in the source.

And once a string is added to source language, it is not possible to edit/update it anymore. there are only 2 choices left after that: delete the string, add a variant. (save and suggest button are disabled)

the first takes long as I also need to add it again with correct explanation.

the latter makes thing more complex, because I don't want to add a variant, but if I do then I will have another copy with a "key" I cannot edit(!) later.

### Describe the solution you'd like

Make source strings in the "glossary" editable/updatable, at the least for their "explanations".

### Describe alternatives you've considered

1. go into the glossary component in the source language (mostly English)
1. find delete the current string
2. add the same string but now with correct explanation

yilmazdurmaz:
> There is a problem (bug?) in adding terminology items, such that the "explanation" I added in a language does not go into the "Source string explanation" of the string in the project's source language. It can still be seen in "adder" language, but not in the source.

Should I open another ticket about this?

yilmazdurmaz:
> You can edit explanation at both source and translation

Unfortunately, this is not true. Once an item is added as terminology from a non-source language, it becomes "read-only" in the source language, thus its explanation cannot be edited without deleting that one and adding anew. Hence, this request came.

https://hosted.weblate.org/translate/godot-engine/glossary/en/?checksum=d3a9ffbb9206cb04&q=has%3Acomment#others

save button says "it is read-only" and suggestion button says "insufficient privilege"

![readonly](https://github.com/WeblateOrg/weblate/assets/1812105/7ccb64a5-0a1d-429b-a9cc-eff106bace92)

yilmazdurmaz:
And, by the way, this is the other part.

> if you add a translation, it's a translation explanation, if you add the term to the source language, it is the source explanation.

A checkbox to add the string along with the explanation to the source would be nice. Currently, I have to either switch back to source first to add that string or do the whole add-find-remove-add-again cycle, which would be add-find-edit if you implement the main request of this issue. And would be check-add-maybe-find-edit if we had both.

[...]

PR #15070, opened 2025-06-02
 chore(pending changes): deepen the stack of changes

amCap1712:
Currently, there are multiple situations under which translating a unit requires that the previously pending changes to the unit be flushed to the repository. One such example is two different users editing the same unit, and the second user's changes cannot be saved till the first user's changes to the unit have been committed. This is a performance issue.

To fix this issue, introduce a new PendingUnitChange model that keeps track of the pending changes to be flushed to the repository. When a unit is translated or saved otherwise, a PendingUnitChange record is created for the change instead of flushing the changes to the repository.

The logic to group changes by author has changed as the previous code depended on the changes to a unit by a different author to have been already committed. The new logic handles these cases of changes to the same unit by multiple authors correctly.

### DIFF
--- a/weblate/utils/search.py
+++ b/weblate/utils/search.py
@@ -507,8 +507,6 @@ class UnitTermExpr(BaseTermExpr):
                     "|".join(re_escape(term) for term in terms)
                 )
             )
-        if text == "pending":
-            return Q(pending_changes__isnull=False)
 
         return super().has_field(text, context)
 


weblate/trans/models/translation.py
```diff
@@ -773,35 +841,23 @@ def git_commit(
...
-
-            details = unit.details
-
-            # Remove pending flag
-            unit.pending = False
+        for pending_change in pending_changes:
+            unit = pending_change.unit
+            details = pending_change.details
 
             if details.get("add_unit"):
```
amCap1712: I am unsure about accessing unit here. For instance, consider there are three pending changes for a unit. Applying them in the chronological order, the oldest change might use the latest value which would reflect wrongly in commits. To my understanding, get_source_plurals and get_target_plurals would not be affected by unit level changes so should be fine to use. What about context and source_unit.explanation though? Should these attributes be copied to PendingUnitChange when its created?

Should the unit.context be updated here or in save or somewhere else?

amCap1712:
(doesn't depend on #15011 but includes changes from it due to local migrations and testing)

weblate/trans/models/translation.py
nijel: The `context` currently cannot change (it would be a different unit), so it is safe to use from unit. At least until https://github.com/WeblateOrg/weblate/issues/10304 is implemented.

The source string is here only used for bilingual formats, and it cannot change there (at least until https://github.com/WeblateOrg/weblate/issues/9729 is implemented). So using `unit.get_source_plurals` should be safe for now. 

To make it future-proof, tracking these in `PendingUnitChange` might be useful. But it can be as well added later to support the identifier editing setup as described in the referenced issues (it will consist of two pending changes - removal of the old identifier and adding new one).

`unit.get_target_plurals()` would return the current unit target, not the one from pending change, so that needs to come from `PendingUnitChange`.

One option is to modify the Unit instance in memory based on the `pending_change` so that unit methods return correct values. This way the rest of the code does not have to be changed.

weblate/trans/models/translation.py
```diff
@@ -962,7 +1004,7 @@ def show_upload_not_found(
         count = len(not_found_log)
 
         strings = format_html_join_comma(
-            gettext("“{}”"), ((string,) for string in not_found_log[:string_limit])
+            gettext('"{}"'), ((string,) for string in not_found_log[:string_limit])
```
nijel: Why changing the quotes?

weblate/trans/models/translation.py
```diff
@@ -671,30 +672,33 @@ def _commit_pending(self, reason: str, user: User | None) -> bool:
...
             return False
 
-        units = list(
-            self.unit_set.filter(pending=True)
-            .prefetch_recent_content_changes()
-            .select_for_update()
+        pending_changes = list(
+            PendingUnitChange.objects.for_translation(self)
+            .select_related("unit", "author")
+            .order_by("timestamp")
```
nijel, line 678: This should probably use `select_for_update` (with similar logic to the `Unit` model to avoid locking primary key).

weblate/trans/models/pending.py
```diff
@@ -0,0 +1,105 @@
...
+        db_index=True,
+    )
+    target = models.TextField(default="", blank=True)
+    explanation = models.TextField(default="", blank=True)
+    state = models.IntegerField(
+        default=0,
+        choices=StringState.choices,
+    )
+    timestamp = models.DateTimeField(auto_now_add=True, db_index=True)
+    details = models.JSONField(default=dict, blank=True)
```
nijel: As this is currently only used for `add_unit` and changing `PendingUnitChange` will always be quite cheap, we might add `add_unit` directly here.

nijel (commented):
Additional changes needed:

* `weblate/templates/translate.html` will have to be changed as it indicates if unit is pending
* `UnitSerializer` exposes this as well, not sure if we want to synthetise it there as might be expensive when listing units...

weblate/trans/models/translation.py
amCap1712: erroneous, will revert.

weblate/trans/models/translation.py
amCap1712: Understood, that covers it all except the `unit.source_unit.explanation` used later in the same function. What about that?

nijel:
The test failures should be addressed by https://github.com/WeblateOrg/weblate/pull/15161.

Remaining changes needed:

* `weblate/templates/translate.html` will have to be changed as it indicates if unit is pending
* `UnitSerializer` exposes this as well, not sure if we want to synthesize it there as might be expensive when listing units.

weblate/trans/models/translation.py
```diff
@@ -1510,7 +1556,6 @@ def get_store_change_translations(self) -> list[Translation]:
 
     @overload
     def add_unit(
-        self,
```
nijel: Probably error from merge?

nijel:
Now I noticed `trans_unit_pending` index in `Unit` model which is no longer needed.

amCap1712:
Indeed but `DROP INDEX` requires the `ACCESS EXCLUSIVE` lock in Postgres so best to remove at same time as `pending` column.

nijel:
Looking at it, removing the column should also be fast, as it keeps data in place (see https://stackoverflow.com/a/15700213/225718). So let's make this complete and remove the pending field and index. That way, it will be clear that nobody mistakenly uses a no longer updated field.

When speaking of migrations, there should be a migration that would take all pending units and create matching `PendingUnitChange` objects for them, so that strings pending on upgrade are not lost…

weblate/trans/migrations/0040_auto_20250613_1813.py
```diff
@@ -0,0 +1,68 @@
...
+def migrate_changes(apps, schema_editor) -> None:
+    PendingUnitChange = apps.get_model("trans", "PendingUnitChange")
+    Unit = apps.get_model("trans", "Unit")
+
+    User = apps.get_model(*settings.AUTH_USER_MODEL.split("."))  # noqa: N806
+    try:
+        anonymous_user = User.objects.get(username=settings.ANONYMOUS_USER_NAME)
+    except User.DoesNotExist:
+        # should only happen during tests
+        anonymous_user = None
```
amCap1712: IIUC, the `anonymous_user` always exists in production database because its created just after the migrations are run. But it won't exist in a new test db, as there won't be any pending units in a new db we should be fine.

weblate/trans/models/translation.py
amCap1712: @nijel Sorry for the double ping but I think you missed the comment above.

amCap1712:
The pending field in backups also needs to be handled, should I update backups to remove pending field from Unit and include PendingUnitChange in dumps as well?

nijel:
Good point, backups need to handle this as well. For restoring existing backups, the pending field should be restored as pending unit change model. New backups should include it directly.

weblate/trans/models/translation.py
nijel: Yep, missed that one. To be safe, it should be tracked as well.

weblate/trans/backups.py
```diff
@@ -545,6 +563,49 @@ def restore_teams(self, data: list[dict]):
...
+        source_units: list[Unit],
+        units: list[Unit],
+    ) -> None:
+        if "pending_unit_changes" in data:
+            pending_unit_changes = [
+                PendingUnitChange(
+                    unit=Unit.objects.get(
+                        translation=translation_lookup[item["translation_id"]],
+                        id_hash=checksum_to_hash(item["unit_id_hash"]),
+                    ),
```
nijel: It might be more effective to lookup in `units` instead of hitting the database.

weblate/trans/models/pending.py
```diff
@@ -57,11 +57,12 @@ class PendingUnitChange(models.Model):
     )
     target = models.TextField(default="", blank=True)
     explanation = models.TextField(default="", blank=True)
+    source_unit_explanation = models.TextField(default="", blank=True)
     state = models.IntegerField(
         default=0,
         choices=StringState.choices,
     )
-    timestamp = models.DateTimeField(auto_now_add=True, db_index=True)
+    timestamp = models.DateTimeField(auto_now=True, db_index=True)
```
nijel: Does this preserve the inserted timestamp? I think it always overwrites it.

amCap1712: My intent is to override the timestamp, thinking of the case where:

1. User 1 edits a unit - creates a pending change.
2. User 2 edits the same unit - creates a second pending change.
3. User 1 edits that unit again - existing pending change updated since there can be only one pending change per unit and author.

Now in 3, the pending change of user 1 will be ordered before the pending change of user 2 if the timestamp is not updated. And as a result user 1's final change would be applied to the VCS before user 2's change and the user 2's change would be the final persisted one in the VCS which would be incorrect.

Hence, when updating an existing pending change the timestamp needs to be updated too.

nijel: I didn't consider this, I was rather thinking about backup restore, which should preserve the original timestamp.

As for the edits, shouldn't the 3. just create an additional pending change to keep full history in Git?

amCap1712: 1. Ah okay, makes sense.
2. Sure will update.

amCap1712: Fixed.

weblate/trans/backups.py
amCap1712: Updated.

weblate/trans/models/translation.py
```diff
@@ -861,28 +926,17 @@ def update_units(  # noqa: C901
                         action=ActionEvents.SAVE_FAILED,
                         target=self.component.get_parse_error_message(error),
                     )
-                    clear_pending.append(unit.pk)
                     continue
 
+                changes_applied.append(pending_change)
```
amCap1712: Thinking about handling failures here, `PendingUnitChange` rows are not deleted if the change application failed. Is it possible to have a case like this? A unit has two changes, applying the older change fails but the newer one works. The older change is kept in the db and applied successfully on next try, overwriting the final change.

weblate/trans/migrations/0040_auto_20250613_1813.py
nijel: Another option would be getting the user from the database lazily once it is needed.

weblate/trans/models/translation.py
nijel: Actually, this is a real situation, see 918315842dd and #9775. There could be syntactically invalid edit which cannot be saved. Another user then comes and fixes the string. We don't want the invalid edit to stick around infinitely. These are the two `TODO` comments just above this block.

Such an edit can be discarded if there was a newer one successfully applied to the same string. The discarded edit won't appear in Git (but there is no way to save it). This is the “Could not update unit” block.

For disappeared strings (“String disappeared” block) there should be no newer edit successfully applied, so these should stick around (until the files are parsed again and such a unit is removed from the database).

amCap1712: Makes sense, updated.

nijel:
I think this is good to merge once 5.12.2 is released from the `main` branch.

I've just pushed https://github.com/WeblateOrg/weblate_schemas/releases/tag/2025.4, so the dependency here can be updated to not use git.

PS: Maybe just let's merge this after https://github.com/WeblateOrg/weblate/pull/15181 is merged so that the changes are separate and we don't have to deal with merge conficlits of sbom and lock files here...

amCap1712:
Sounds good to me, let me know if any other changes are needed meanwhile. Or if you feel there are missing test cases that should be added to the test suite.

nijel:
I was considering adding tests for the migration, but I'm not sure if it is worth of the effort...

amCap1712:
I'll look into it.

nijel:
There are some existing migration tests in `ci/run-migrate`.

nijel:
@amCap1712 Can you please rebase on top of the current `main` branch which already has (conflicting) update of the schema dependency?

amCap1712:
@nijel Updated.

ci/run-migrate
```diff
@@ -83,6 +83,27 @@ check
 ./manage.py shell -c 'from weblate.memory.models import Memory; from weblate.lang.models import Language; Memory.objects.create(source_language=Language.objects.get(code="en"), target_language=Language.objects.get(code="cs"), source="source"*1000, target="target"*1000, origin="origin"*1000)'
 check
 
+# Add a pending unit
+semver_compare "$TAG" "<=" "5.12"
```
nijel: The TAG is something like `weblate-5.12', so this gets never triggered. Better to use $1 directly. Also, we want to test migrations from version older than 5.12:

```suggestion
semver_compare "$1" "<" "5.12"
```

ci/run-migrate
```diff
@@ -83,6 +83,27 @@ check
 ./manage.py shell -c 'from weblate.memory.models import Memory; from weblate.lang.models import Language; Memory.objects.create(source_language=Language.objects.get(code="en"), target_language=Language.objects.get(code="cs"), source="source"*1000, target="target"*1000, origin="origin"*1000)'
 check
 
+# Add a pending unit
+semver_compare "$1" "<" "5.12"
+pending_migration_test=$?
+if [ "$pending_migration_test" -eq 0 ]; then
+    ./manage.py shell << EOF
+from django.contrib.auth import get_user_model
+from weblate.trans.models import Unit, Change, STATE_TRANSLATED
```
nijel:
```suggestion
from weblate.trans.models import Unit, Change
from weblate.utils.state import STATE_TRANSLATED
```

ci/run-migrate
```diff
@@ -83,6 +83,28 @@ check
...
+# Add a pending unit
+semver_compare "$1" "<" "5.12"
+pending_migration_test=$?
+if [ "$pending_migration_test" -eq 0 ]; then
+    ./manage.py shell << EOF
+from django.contrib.auth import get_user_model
+from weblate.trans.models import Unit, Change
+from weblate.utils.state import STATE_TRANSLATED
+User = get_user_model()
+user = User.objects.get(username="admin")
```
nijel:
```suggestion
user = User.objects.get_or_create(username="admin")[0]
```

ci/run-migrate
```diff
@@ -95,3 +117,29 @@ check
 # Check migrated vote exists
 uv run --all-extras ./manage.py shell -c 'from weblate.trans.models import Vote; Vote.objects.get(value=1)'
 check
+
+# Check migrated pending unit
+if [ "$pending_migration_test" -eq 0 ]; then
+    uv run --all-extras ./manage.py shell << EOF
+from weblate.changes.models import PendingUnitChange
```
nijel:
```suggestion
from weblate.trans.models import PendingUnitChange
```

amCap1712: :facepalm:

ci/run-migrate
```diff
@@ -83,6 +83,28 @@ check
...
+# Add a pending unit
+semver_compare "$1" "<" "5.12"
+pending_migration_test=$?
+if [ "$pending_migration_test" -eq 0 ]; then
+    ./manage.py shell << EOF
+from django.contrib.auth import get_user_model
+from weblate.trans.models import Unit, Change
+from weblate.utils.state import STATE_TRANSLATED
+User = get_user_model()
+user = User.objects.get_or_create(username="admin")[0]
```
nijel:
```suggestion
user = User.objects.create(username="migratetest")
```

weblate/utils/search.py
```diff
@@ -507,6 +507,8 @@ def has_field(self, text: str, context: dict) -> Q:  # noqa: C901
                     "|".join(re_escape(term) for term in terms)
                 )
             )
+        if text == "pending":
+            return Q(pending_changes__isnull=False)
```
nijel: This seems extra, it changes the `has` lookup which does not have `pending` option.

```suggestion
```

nijel:
Scheduled for merge, thanks!

Issue #6094, opened 2021-05-27
Adding new language from project-level language tab (rather than component level)


shasheene:
HI, Weblate is a great program. But the user-interface is very confusing for new for users to add a new language to the project. Whether or not you're logged in, there is a clear "Start new translation" button at the _component level_ (see screenshot). This is good:

![not logged in component level language](https://user-images.githubusercontent.com/1414379/119770173-03844180-befb-11eb-82f1-cf3430410503.png)

But viewing the languages tab at the _project level_ does **not** display a "Start new translation" button (see second screenshot). This is bad:

![not logged in project level language](https://user-images.githubusercontent.com/1414379/119770175-04b56e80-befb-11eb-91c5-42a7f1cb01c4.png)

### Solution

**Please add a "Start new translation" button on the project level's languages tab.**

It's confusing and unnecessary for translators to have to understand the component-level structure of my project. The screenshot context images I have associated with each of my source strings already provides enough context.

This change would allow me to simply share the this link: https://hosted.weblate.org/projects/rescuezilla/#languages and any translator (even one who's language is not yet represented) will be able to immediately start translating my entire program (after they immediately realize they need to register to click the 'Start new translation' button). 

By the way, I can see the Weblate program has a "view" that can encompass every language within the installed Weblate database (https://hosted.weblate.org/languages/). But I am just asking for this feature at the project level's language tab. My feature request is definitely not required at that absolute _topmost_ level of Weblate.

nijel:
Related to https://github.com/WeblateOrg/weblate/issues/5082

CiaraG98:
Hi, I can look at this if it is still open?

keunes:
@nijel Thanks for the exchange today at FOSDEM! There was indeed already this ticket for the 'Add language' button at component level :-)

PR #15099, opened 2025-06-07
feat: start new translation for project

amCap1712:
Add a "Start new translation" at project level.

Questions:
1. What level of permissions should using this button require?
2. What to do when the user has access to only some components in the project?
3. What to do when components are configured as pointing to instructions url, contact maintainers and disabled for new translations? How to surface this information to the user?
4. How many messages to flash to the user? One per component or one in total?

1 and 2 - I think a user should be to view the button if the user has permissions to start or request translating any components in a project. When a user uses this feature, translations should be created in all components that allow add new language. 

3 - For the components configured as contact maintainers, the maintainers should be notified. For none and point to instructions url, nothing should be done. At the component level, its possible to surface the exact action that would be taken for a component. Since that's not possible at project level, I think we can mention all actions possible for various configurations.

Not sure what to do for 4.

Thoughts?

TODO:

Add + button for the same functionality.

#6094

### DIFF
--- a/weblate/trans/tests/test_views.py
+++ b/weblate/trans/tests/test_views.py
@@ -382,6 +382,49 @@ class ProjectLanguageAdditionTest(ViewTestCase):
     def create_component(self):
         return self.create_po_new_base()
 
+    def test_exising_language_excluded(self):
+        self.user.is_superuser = True
+        self.user.save()
+
+        response = self.client.post(
+            reverse(
+                "new-language", kwargs={"path": self.components["add"].get_url_path()}
+            ),
+            {"lang": "af"},
+            follow=True,
+        )
+        self.assertTrue(
+            self.components["add"].translation_set.filter(language_code="af").exists()
+        )
+        response = self.client.post(
+            reverse(
+                "new-language",
+                kwargs={"path": self.components["contact"].get_url_path()},
+            ),
+            {"lang": "af"},
+            follow=True,
+        )
+        self.assertTrue(
+            self.components["contact"]
+            .translation_set.filter(language_code="af")
+            .exists()
+        )
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 200)
+        self.assertIn("form", response.context)
+        codes = {c[0] for c in response.context["form"]["lang"].field.choices}
+        self.assertNotIn("af", codes)
+
+        response = self.client.post(self.url, {"lang": "af"}, follow=True)
+        self.assertEqual(response.status_code, 200)
+        messages = [m.message for m in response.context["messages"]]
+        self.assertCountEqual(
+            messages,
+            [
+                "Please fix errors in the form.",
+            ],
+        )
+
     def test_view_add_language(self) -> None:
         self.assertTrue(
             all(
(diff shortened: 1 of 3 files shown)


nijel:
1 and 2: I agree with your proposal here.

3: The configuration will often be mixed because of glossary components where users cannot explicitly add languages, but it will automatically happen once the language exists in the project. On the other side, this will be often consistent for all non glossary components. 

I think the best we can do is to not show the button if there is no way to add language in any component (all components have that disabled or configured to link).

4: It should show just the summary. Imagine projects with hundreds of components. The summary should exclude glossary components because these have special handling anyway. I think there could be at most two messages shown, something like:

Success: The language was added to {{ count }} components.

Warning: The language could not be added to {{ count }} components. Please look into individual components' configuration for more info.

Both should be shown only if count > 0 and using pluralized localization.

nijel:
Can you please add tests for the new functionality?

weblate/trans/views/basic.py
```diff
@@ -794,11 +754,168 @@ def new_language(request: AuthenticatedHttpRequest, path):
...
+        "component": component,
+        "details": {},
+    }
+    counts = Counter()
+    with component.repository.lock:
+        component.commit_pending("add language", None)
+        for language in Language.objects.filter(code__in=languages):
+            kwargs["details"]["language"] = language.code
+            if component.can_add_new_language(user):
+                # TODO: handle language already exists error specially?
```
amCap1712: Should the case of language existing in one of the components be handled separately? Could add a boolean parameter to return the existing translation in that case but the add_new_language method is already very complex.

weblate/trans/views/basic.py
```diff
@@ -794,11 +754,168 @@ def new_language(request: AuthenticatedHttpRequest, path):
...
+                    form.cleaned_data["lang"],
+                    component,
+                    show_messages=False,
+                )
+                counter.update(component_count)
+
+            if counter["added"] > 0:
+                messages.success(
+                    request,
+                    ngettext(
```
amCap1712: Just realised, these messages need to be pluralized with respect to languages too.

So in that case, how should summaries work? One summary alert per language. or One across all languages and components, but no proper way to counts components then.

nijel: In most cases, just one language is being added, so optimize for that use-case. So having one summary per language is okay.

weblate/trans/views/basic.py
nijel: I would also avoid extending the complexity there for now.

weblate/trans/tests/test_views.py
```diff
@@ -366,6 +366,102 @@ def test_remove(self) -> None:
...
+                name=f"test-{new_lang}",
+                project=self.project,
+            )
+        self.url = reverse("new-language", kwargs={"path": self.project.get_url_path()})
+
+    def create_component(self):
+        return self.create_po_new_base()
+
+    def test_view_add_language(self) -> None:
+        self.assertTrue(
```
nijel, line 429: Please test the GET method on the view as well (just to be certain it renders okay).

amCap1712: Done.

nijel:

PR #15102, opened 2025-06-09
feat: anonymize IP address in notificatin mails

nijel:
Do not include exact IP address in the notification mails.


orangesunny (approved):

nijel:
Good point, I've just added such assertion.

PR #15113, opened 2025-06-10
feat: start migrating to Bootstrap 5

KarenKonou:
Weblate currently uses Bootstrap 3 which has long been deprecated. Updating to a newer version is recommended for security reasons and will also improve the user and developer experience.

Due to the very large scope of this change the migration will be done in parts, several templates at a time. Despite this there is still a large amount of upfront work necessary

- [x] Vendor the Bootstrap v5 JS and CSS files
- [x] Allow conditionally loading the Bootstrap v5 stylesheets by providing an argument in a view
- [x] Migrate customizations from the deprecated v3 online customizer to CSS variables
- [x] Create an updated version of the base template to use with migrated templates
- [x] Update our custom CSS if necessary to keep the user experience consistent
- [x] Update a couple of templates to kick-off the migration proper
- [x] Use `crispy_bootstrap5` template pack for crispy forms where applicable

Part 1 of #2916


KarenKonou:
## Progress report

First few templates are starting to look ready, still finishing up some minor details, mobile view and dark mode. There are some differences in style mostly due to the Bootstrap 5 crispy template. I think it looks quite good but I could also style everything to look as close as it used to on Bootstrap 3 (@nijel we can also discuss this in person next week)

**Before**:
image

**After**: 
image


There is a **required settings change** to add the `crispy_bootstrap5` module in `INSTALLED_APPS`:

```
INSTALLED_APPS = (
    ...
    "crispy_forms",
    "crispy_bootstrap3",
    "crispy_bootstrap5",
    ...
)
```

and enable it in `CRISPY_ALLOWED_TEMPLATE_PACKS` while still keeping the Bootstrap 3 template pack as default:

```
CRISPY_ALLOWED_TEMPLATE_PACKS = ["bootstrap3", "bootstrap5"]
CRISPY_TEMPLATE_PACK = "bootstrap3"
```

nijel:
I don't see need to keep exactly same look. We will do changes in that anyway so let's not spend much time on this.

nijel:
I did some testing today. Overall it looks good, but there are a few issues I could find:

#### Button for removing custom notification settings for a component

image

Related code:

https://github.com/WeblateOrg/weblate/blob/25ffbb731874e0ae7508b09bb46980a66ee10840/weblate/templates/accounts/profile.html#L131-L135

#### Weird border on bottom of user identities

image

#### Extra border around avatar image

image

#### Top green border on tables

Teams listing:

image

Audit log:

image

KarenKonou:
Still need to fix a few things like the redirect to notifications from the watched dropdown and check that all of the js is working correctly.

KarenKonou:
I think this is pretty much ready for testing and review, probably need to disable pre-commit spellcheck on the CSS map file though, I'm not entirely sure how to do this😅

 One more question I have is, how do you notify admins of breaking settings changes (the crispy forms bootstrap 5 template pack in this case)? I wouldn't want someone's instance to break because of this.

nijel:
https://github.com/WeblateOrg/weblate/pull/15655 should address the pre-commit failure.

Did some testing, and this is what I found.

* The notification removal button is invisible in the dark theme (black color on dark background) (maybe it was the case even with the old code, I didn't really check):
  
  obrazek


* Weird border on the bottom of the user identities is still there:
 
  obrazek
  
  

Besides that I think this is ready to get merged.

nijel:
> One more question I have is, how do you notify admins of breaking settings changes (the crispy forms bootstrap 5 template pack in this case)? I wouldn't want someone's instance to break because of this.

Mention this in the **Upgrading** section in the changelog.

nijel:
And one more thing that I've noticed because of the lint error: Please follow https://docs.weblate.org/en/latest/contributing/frontend.html#dependency-management for vendoring the library; that way we have a webpack-based approach to track the dependencies.

KarenKonou:
One more thing I remembered is I forgot to add the changes from https://github.com/WeblateOrg/weblate/pull/15621 to the `base5` template so I'll do that as well to keep it consistent

nijel:
I think this is good to go now. That way we can get an additional feedback from our community. Missing bits:

* Add changelog entry with upgrading instructions.
* Use webpack-based dependency on Bootstrap so that we can easily follow upstream and automate updates.

If you need help with any of those, just let me know (we can also leave webpack stuff for a separate pull request).

KarenKonou:
I'd merge this now then and do the webpack bit separately, could you add the changelog entry yourself just so I know what it should look like for next time?

Thank you :)

nijel:
Changelog added in ad6ef20692 and scheduled for merge. Thanks for getting this moving!

PR #15127, opened 2025-06-11
feat: Category field in Webhook payload

### LINKED ISSUE #14599: Include category in webhook payload
### Describe the problem

The webhook payload is based on Fedora messaging stuff which was introduced before categories were. This causes the categories to not be exposed in the webhook.

### Solution brainstorm

Add `category` field which would include full slug of component categories. This is similar approach we are about to use in categories backups in https://github.com/WeblateOrg/weblate/pull/14531.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

gersona:
- depends on https://github.com/WeblateOrg/weblate_schemas/pull/448
- fixes #14599 

When available, a category field is included in the webhook payload, as an array of category slugs


docs/changes.rst
```diff
@@ -24,6 +24,7 @@ Weblate 5.12
 * :http:get:`/api/users/` now includes ``languages``.
 * :ref:`addon-weblate.webhook.webhook` documentation improved.
 * Improved support for extended metadata in :ref:`tbx`.
+* :ref:`addon-weblate.webhook.webhook` now includes category field when available.
```
nijel:
```suggestion
* :ref:`addon-weblate.webhook.webhook` now includes a category field when available.
```

weblate/addons/webhooks.py
```diff
@@ -104,7 +104,13 @@ def build_webhook_payload(self, change: Change) -> dict[str, int | str | list[st
             data["component"] = change.component.slug
         if change.translation:
             data["translation"] = change.translation.language.code
-
+        if change.path_object and (category := change.path_object.category):
```
nijel: `path_object` is not a good choice here, as it fails with `'Translation' object has no attribute 'category'`.
```suggestion
        if change.component and (category := change.component.category):
```

nijel:

PR #15135, opened 2025-06-11
feat(api): add user contributions

michael-smt:
Adds a API endpoint for the `user/<name:user>/contributions/` view.

~~Compared to the `GET /api/components/(string: project)/(string: component)/translations/` endpoint there is no pagination.~~ Pagination added, see below.

Also the retrieval/permission logic is duplicated from `weblate.accounts.views.user_contributions`.

### DIFF
--- a/weblate/api/tests.py
+++ b/weblate/api/tests.py
@@ -373,6 +373,15 @@ class UserAPITest(APIBaseTest):
         )
         self.assertEqual(request.data["commented"], user.profile.commented)
 
+    def test_contributions(self) -> None:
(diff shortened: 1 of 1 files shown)


nijel:
It should be paginated, there are users like https://hosted.weblate.org/user/kingu/contributions/.

michael-smt:
@nijel pagination is added. I did not add tests because I'm not sure if the test setup is worth it. Maybe better move the duplicated logic into a single place - but where should it go?

nijel:
Having a test that at least verifies that the endpoint works would be a good start.

nijel:

PR #15190, opened 2025-06-17
feat(editor): add accept and approve button when review workflow is enabled

### LINKED ISSUE #15115: Accept and approve directly in the automatic suggestions tab
### Describe the problem

It is not possible to accept and approve a automatic suggestion with a single click, slowing down translators who want to directly have a good automatic suggestion added and reviewed.

The same is possible with human suggestions.

### Solution brainstorm

The three small green buttons available for human suggestions could be used instead of the current "Clone to translation" and "Accept" buttons. This would further unify the look & feel of the "Suggestions" and "Automatic suggestions" tab, maybe even allowing them to be merged in the future.

Add another "Accept and approve" button, but this might become too many wide buttons for narrow screens.

### Screenshots

Image

Image

### Additional context

_No response_

michael-smt:
This adds a button "Accept and approve" to the machinery table when the review workflow is enabled.

image

Fixes #15115

### DIFF
--- a/weblate/static/editor/base.js
+++ b/weblate/static/editor/base.js
@@ -7,6 +7,8 @@ var WLT = WLT || {};
 
 WLT.Config = (() => ({
   IS_MAC: /Mac|iPod|iPhone|iPad/.test(navigator.platform),
+  HAS_REVIEW_WORKFLOW:
+    $('.translation-form input[name="review"][value="30"]').length > 0,
 }))();
 
 WLT.Utils = (() => ({
@@ -24,6 +26,13 @@ WLT.Utils = (() => ({
     $el.find('input[name="fuzzy"]').prop("checked", false);
     /* Review workflow */
     $el.find('input[name="review"][value="20"]').prop("checked", true);
+  },
+
+  markApproved: ($el) => {
+    /* Standard workflow */
+    $el.find('input[name="fuzzy"]').prop("checked", false);
+    /* Review workflow */
+    $el.find('input[name="review"][value="30"]').prop("checked", true);
   },
 
   /**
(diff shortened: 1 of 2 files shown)


weblate/static/editor/base.js
```diff
@@ -10,6 +10,9 @@ WLT.Config = (() => ({
   // biome-ignore lint/performance/useTopLevelRegex: TODO: factor out
   // biome-ignore lint/style/useNamingConvention: TODO: fix naming
   IS_MAC: /Mac|iPod|iPhone|iPad/.test(navigator.platform),
+  // biome-ignore lint/style/useNamingConvention: TODO: fix naming
```
KarenKonou: This lint suppression can be removed since it's no longer in the recommended ruleset in Biome v2.

weblate/static/editor/full.js
```diff
@@ -46,6 +46,20 @@
       submitForm({ target: this.$translationArea });
     });
 
+    /* Copy, approve and save machinery results */
+    this.$editor.on("click", ".js-copy-approve-save-machinery", (e) => {
+      const $el = $(e.target);
+      const raw = $el.parent().parent().data("raw");
+
+      // biome-ignore lint/complexity/noForEach: TODO
```
KarenKonou: This lint suppression can also be removed.

KarenKonou (approved):

weblate/static/editor/base.js
michael-smt: Thanks, done.

weblate/static/editor/full.js
michael-smt: Thanks, done.

nijel:

PR #15230, opened 2025-06-20
feat: use native locking in python-redis instead of third-party package

nijel:
The package is no longer maintained and there is now built-in implementation of locking which should work well for our use case.


docs/changes.rst
```diff
@@ -15,6 +15,8 @@ Weblate 5.13
 
 Please follow :ref:`generic-upgrade-instructions` in order to perform update.
 
+* The distributed locking now uses a different what introduced several changes in :file:`settings_example.py`.
```
amCap1712: now uses a different *implementation and that* maybe?

nijel:
```suggestion
* The distributed locking now uses a different implementation and that introduced several changes in :file:`settings_example.py`.
```

Issue #13623, opened 2025-01-22
Redesign component level locking


nijel:
### Describe the problem

The component level locking might be too broad for some situations (see https://github.com/WeblateOrg/weblate/issues/13345). It was introduced to avoid concurrency isues like https://github.com/WeblateOrg/weblate/issues/5686.

The problematic operations being:

* Changing strings, these should be handled by `select_for_update` and should not need component level lock. This needs to be verified.
* Creating new strings (add, parsing files). As these are newly created strings or translations, the row level locking will not help. This might be the remaining needed for component locking.
* Batch updating checks. This happens asynchronously in the Celery task, which holds component lock.
* Updating source checks. This is now synchronous while creating translations. Moving it to a Celery task might be better for both observed performance and data consistence. With it in place, there might be a separate lock for updating checks.

### Describe the solution you would like

Analysis of the locking and what is actually needed to lock to avoid consistency issues.

nijel:
Once https://github.com/WeblateOrg/weblate/issues/8770 is implemented, the lock for changing strings could be probably removed as that will no longer trigger commit (and the string updates are using SELECT FOR UPDATE).

nijel:
https://github.com/WeblateOrg/weblate/pull/15250 removes component locking for bulk edit and automatic translation.

nijel:
https://github.com/WeblateOrg/weblate/pull/18759 offloads checks processing.

PR #15248, opened 2025-06-24
fix(bulk-edit): remove no longer needed commit pending

nijel:
The bulk edit operation should no longer need to commit pending changes since #15070 allows to stack changes.

Issue #13623

### DIFF
--- a/weblate/trans/bulk.py
+++ b/weblate/trans/bulk.py
@@ -46,7 +46,6 @@ def bulk_perform(  # noqa: C901
         prev_updated = updated
         component.batch_checks = True
         with transaction.atomic(), component.lock:
-            component.commit_pending("bulk edit", user)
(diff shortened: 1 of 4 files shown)


weblate/trans/bulk.py
```diff
@@ -46,7 +46,6 @@ def bulk_perform(  # noqa: C901
         prev_updated = updated
         component.batch_checks = True
         with transaction.atomic(), component.lock:
-            component.commit_pending("bulk edit", user)
```
amCap1712, line 49: Since we are not committing here now, is the component.lock above still needed?

amCap1712: Ah I see, it was removed in a separate PR.

nijel: Yes, I've separated that.

Issue #15212, opened 2025-06-18
Improve automatic translation and bulk edit performance


nijel:
### Describe the problem

Translating many strings using automatic translation can be slow because of repeating actions:

- generating and saving change objects
- adding automatically translated label, will be replaced in https://github.com/WeblateOrg/weblate/issues/15762
- generating pending unit change (upcoming in #15070)

### Solution brainstorm

These edits are already marked by `is_batch_update` flags, it could be utilized for more things:

- store changes in `Translation.update_changes` and save them to database in bulk via `Translation.store_update_changes`
- similar approach could be used for pending unit changes
- I'm not sure if we can do something with the labels
- Other callers with `is_batch_update` should be investigated if they need to be modified as well

nijel:
A similar approach might be useful for bulk editing...

nijel:
...and for file upload in `Translation.merge_translations`.

PR #15249, opened 2025-06-24
fix(bulk edit): update unit state before saving PendingUnitChange

nijel:
The in memory state was updated after PendingUnitChange would could cause wrong data in PendingUnitChange.


nijel:
@amCap1712 I was more thinking into using regular `Unit.translate` to avoid need to test the differences here. But it would have to be extended to support bulk updates (not saving immediately). That would be useful for automatic translation as well (see https://github.com/WeblateOrg/weblate/issues/15212).

PR #15256, opened 2025-06-24
bugfix(pending-migration): handle missing author

### LINKED ISSUE #15254: Migration to pending model fails
### Describe the issue

It seems like the migration from https://github.com/WeblateOrg/weblate/pull/15070 doesn't work well in some situations:

```
django.db.utils.IntegrityError: null value in column "author_id" of relation "trans_pendingunitchange" violates not-null constraint
DETAIL:  Failing row contains (518, Slide X, , , 10, 2025-06-24 13:13:47.576142+00, f, null, 131403890).
```

### I already tried

Reproduced on the snapshot of Hosted Weblate, will figure out details later.

### Expected behavior

_No response_

### Screenshots

_No response_

### Exception traceback

```pytb

```

### How do you run Weblate?

Git checkout

### Weblate versions

_No response_

### Weblate deploy checks

```shell

```

### Additional context

_No response_

amCap1712:
If an author is not available for a unit change, set it to the anonymous user.

fix #15254.

### DIFF
--- a/weblate/trans/migrations/0040_auto_20250613_1813.py
+++ b/weblate/trans/migrations/0040_auto_20250613_1813.py
@@ -21,6 +21,8 @@ def migrate_changes(apps, schema_editor) -> None:
     pending_changes = []
     for unit in pending_units:
         try:
+            # can be None in some cases, for instance unit updated in the
+            # repository outside of weblate
             author = (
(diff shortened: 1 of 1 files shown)


nijel:

PR #15279, opened 2025-06-26
refactor: Change details and content display refactor

### LINKED ISSUE #15056: Refactor change rendering
### Describe the problem

The changes rendering is currently split between (too complex)  `Change.get_details_display` and template `weblate/templates/last-changes-content.html`. This makes it unclear where to handle new change events and makes it hard to reuse the logic for example in https://github.com/WeblateOrg/weblate/issues/3113.

### Solution brainstorm

The logic should be refactored to have a separate logic for rendering each change type. Probably some class-based approach could be used (similar to what we do with alerts).

In any case, there should be smaller code blocks involved and all rendering should live in a single location.

### Screenshots

_No response_

### Additional context

_No response_

gersona:
- fixes #15056


weblate/templates/snippets/last-changes-content.html
```diff
@@ -0,0 +1,67 @@
...
+            {% if item.ip_address %}<p>{{ item.ip_address }}</p>{% endif %}
+          </ul>
+        </td>
+        <td class="history-data full-cell">
+          <div class="comment-content">
+            {% if item.history_data.description %}{{ item.history_data.description }}{% endif %}
+            {% for history_field in item.history_data.change_details_fields %}
+              <label class="{{ history_field.label_class }}">
+                {{ history_field.label }}
+                {% for tag in history_field.tags %}<span class="badge">{{ tag }}</span>{% endfor %}
```
amCap1712, line 47:
```suggestion
                {% for tag in history_field.tags %}
                  <span class="badge">{{ tag }}</span>
                {% endfor %}
```

To ensure, the badges are rendered with a little space between them. Alternatively, add a small margin in CSS to `badge` or `tags-list` class.

weblate/trans/templatetags/translations.py
```diff
@@ -1735,3 +1736,53 @@ def format_json(value: dict) -> str:
...
+        user = AnonymousUser()
+
+    processed_changes = []
+    for change in last_changes:
+        # Permissions
+        can_revert = change.can_revert() and user.has_perm("unit.edit", change.unit)
+        can_block_user = (
+            change.user
+            and not change.user.is_anonymous
+            and change.project
```
amCap1712, line 1767: This matches current behavior but also shows the block button to a user on their own changes. Can change later separately if you prefer.

```suggestion
            and change.project
            and change.user != user
```

gersona: That makes sense. Thanks

weblate/templates/snippets/last-changes-content.html
gersona: The `badge` class already has a padding defined in the CSS file, so it should be covered
https://github.com/WeblateOrg/weblate/blob/ab83a3ed2ecb0cb7bdc61684c50799016c5d5d1a/weblate/static/vendor/bootstrap/css/bootstrap.css#L3980-L3984

amCap1712: `padding` is part of the html tag as in it uses the background making the current html appears like this.
image

The current main looks like the below because of the new lines add some whitespace in the html.
image

Adding margin to the `badge` class could do the same or the new lines here in the html as they are in main.

Issue #9947, opened 2023-09-14
To have context in Translation Memory


shrihari-k:
### Describe the issue

Below is a sample PO file I used. But the msgctxt is not there in the TMX when we download it. For example

msgid "Friend"
msgstr "Ami(e)"

msgctxt "FEMALE"
msgid "Friend"
msgstr "Amie"

msgctxt "MALE"
msgid "Friend"
msgstr "Ami"

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar issues in this repository.

### Steps to reproduce the behavior

Uploaded a PO file with context (as bilingual) - Works well
View and update the translation - Works well
Download the translation memory - Works Well
TMX file does not contain the context information.

### Expected behavior

 <tu>
  <prop type="context">FEMALE</prop>
    <tuv xml:lang="en">
        <seg>Friend</seg>
    </tuv>
    <tuv xml:lang="fr">
        <seg>Amie</seg>
    </tuv>
  </tu>

### Screenshots

Not applicable

### Exception traceback

```pytb
Not applicable
```

### How do you run Weblate?

Other

nijel:
Also discussed at https://github.com/WeblateOrg/weblate/discussions/9930

Weblate translation memory currently doesn't store context. The problem with this is that it is typically specific to the use case. In most cases, you want to get a translation memory match even if the context would be different (this is typical for monolingual translations where context is actually a string identifier).

shrihari-k:
The following are features I am thinking of.
1. TMX data should have an option to be used with context / without context during translation
2. To have an option for TMX to be downloaded with context / without context
Following are the use cases I envision
1. When there is an incremental update we know that contextually things have not changed then we can look at an option to possibly avoid manual review and save time.
2. To have the TMX updated with context so that the information in the PO / XLIFF file when imported is not lost for future use if someone wants to use it (since PO is a temporary file). We will have the information in the repo.

tomkolp:
- Adding context to the translation memory would be a precursor to: https://github.com/WeblateOrg/weblate/issues/5308

- Automatic suggestions could also display strings with the same context, source and translation first

I think this problem is related, but it probably requires a lot of work to implement. This is not necessary for me at the moment, but it is useful when migrating from other translation platforms.
Additionally, automatic translation based on context matching would save a lot of time (currently, each translator has to check all strings from automatic translation).

gersona:
[...]

PR #15316, opened 2025-06-30
TM management improvement

### LINKED ISSUE #7346: Improve translation memory management
## Situation

In our weblate self-hosted approach, any translation components are premanently onboarded. In order to benefit cross component / project suggestions e.g. with "automatic suggestions", both default machineries are activated (`Weblate`: live component look-up, `Weblate Translation Memory`: TM look-up).

Working on tanslation, one essential aspect is to harmonize terminology in use across various components or even across various projects. For example, we may start with a first translation then later on it is needed to revise it for the sake of harmonized terminology.
This results in having old, obsoleted strings enriched in TM, as well as the latest, harmonized and approved string.

Over the time, this leads to a polluted TM, where the machinery `Weblate Translation Memory` provides outdated, obsolete or (meanwhile) even forbidden strings. In e.g. "automatic suggestion" tab, potential suggestions meanwhile became a mix of old TM results, latest TM results and live results from active components.

Meanwhile, this is heavily puzzling translators and even leads to mistakes in a way, that translators pick outdated terms or even forbidden ones from TM.

## Goal

As a translator, I don't want to see a history of text memory. More specifically, I only want to see auto suggestions based on texts which are currently valid and approved translations.

## Problem

Today, weblate provides no means to manually or automatically clean up TMs in order to get rid of "old" stuff and hence avoiding translation mistakes if translators base on TM results. Therefore, with translations continuously happening and todays automatic enrichement of TM, the pollution of TM continuously grows.

This issue is collecting options in order to improve TM management. In order to make a specific option implementable, this is/will be carved out to a specific, individual issue.

## Option 1: improve global TM management - delete and recreate

This option is described with https://github.com/WeblateOrg/weblate/issues/7347

This would be very helpful in order to counteract TM pollution in a manual way, but with a kind of "mass operation", where clean-up does not happen on individual string base but on full TM scopes.

## Option 2: individual deletion of TM entries

This option is described with https://github.com/WeblateOrg/weblate/issues/6440

In selective situations this is helpful. Currently, for my situation, Option 1 would be sufficient.

## Option 3: automatic TM maintenance based on review state

This option is not yet put to a individual issue, since this requires discussion first.

- Only approved strings shall go to TM
- If string enjoys changes and is approved again, this shall replace the old string in TM (automatic clean-up). Old string to be removed from TM, so that NOT both are proposed for automatic suggestion
- This shall be an automated solution, no manual intervention required.
- Provide a configuration option in order to `enable`/`disable` this automatic clean-up.
  - allow to configure it for shared, project and personal TM
  - Default value to be set by env variable (docker use case)
  - If env var not set, put default = `disable` for backwards compatibility

Excluded:

- Manually uploaded TMs either on project level or personal level shall not be affected. This still shall work as-is.

Affects:
- Shared TM
- Project TM
- Personal TM

Pros
- Automated solution, no manual interaction

Cons
- Algorithm for automatic action might be difficult to define in order to not miss a use case or even in order to not destroy a use case.
- Automatic actions on TM management are globally executed. Selectiveness depends on algorithm implemented.

## Option 4: provide option to switch off automatic TM enrichement

This optoin is described with https://github.com/WeblateOrg/weblate/issues/7348

In our use case, we primarily build on the `Weblate` machinery providing a live look-up to all connected components. Once this option is available, we would switch off automatic TM enrichement.

## Remark to any options

All this only affects the automatic enrichement of TMs.
What still shall be possible (as-is today):
- Upload manually an external TM
- TM machinery to offer results from any existing TM (e.g. in "Automatic suggestion" tab)

## Preferred solution approach

For our use case we are requiring the following options as a best fit for a next improvement step:
- Option 1 and Option 4

gersona:
- fixes #7346 
- Documentation probably needs to be reviewed

**Addtional question**
- What would be a reasonable value for the penalty factor ?


docs/admin/config.rst
```diff
@@ -644,6 +644,13 @@ DEFAULT_SHARED_TM
 
 Configures the default value of :ref:`project-use_shared_tm` and :ref:`project-contribute_shared_tm`.
 
+AUTOCLEAN_TM
+------------
+
+.. versionchanged:: 5.13
```
nijel:
```suggestion
.. versionadded:: 5.13
```

docs/admin/config.rst
```diff
@@ -644,6 +644,13 @@ DEFAULT_SHARED_TM
 
 Configures the default value of :ref:`project-use_shared_tm` and :ref:`project-contribute_shared_tm`.
 
+AUTOCLEAN_TM
```
nijel:
```suggestion
.. setting:: AUTOCLEAN_TM


AUTOCLEAN_TM
```

weblate/memory/machine.py
```diff
@@ -7,6 +7,8 @@
 from weblate.machinery.base import DownloadTranslations, InternalMachineTranslation
 from weblate.memory.models import Memory
 
+PENDING_MEMORY_PENALTY_FACTOR = 0.7
```
nijel, line 10: I think this is a good value for pending penalty.

weblate/memory/migrations/0005_set_memory_status_active.py
```diff
@@ -0,0 +1,32 @@
...
+    dependencies = [
+        ("memory", "0004_memory_status"),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            set_status_to_active,
+            reverse_set_status_to_active,
+        ),
+    ]
```
nijel: Wouldn't it be better to handle this in the migration adding the field via default value?

weblate/memory/tasks.py
```diff
@@ -100,38 +101,65 @@ def update_memory(
...
+    ):
+        memory_status = Memory.STATUS_ACTIVE
+        if project.autoclean_tm:
+            Memory.objects.filter(
+                from_file=False,
+                source=source,
+                origin=origin,
+                source_language_id=source_language_id,
+                target_language_id=target_language_id,
+            ).delete()  # delete old entries, including those with different targets
```
nijel: This has risk of deleting unrelated entries as it matches only based on the source:

- There can be different entries from different components.
- There can be multiple same strings in one component, differing in key/context only.

weblate/memory/migrations/0005_set_memory_status_active.py
gersona: You mean having `active` as default value then ? Because at the moment I set the default to `pending`, but wanted any past memories to be set as `active`

nijel: If we want to keep `pending` as default in the model, it can still be done purely in Django migrations. Instead of having only `AddField(...default=0)`, use `AddField(...default=1), AlterField(...default=0)`.

This way, the migrated rows will directly get the `active` state and the `AlterField` then just modifies the table structure, what is cheap. Updating all the rows in the second migration would double the migration time because all rows are updated when adding the column, and it is then again rewritten.

gersona: thanks for the tip

weblate/memory/tasks.py
gersona: Doesn't filtering by `origin` exclude entries from different components ?

gersona: As for the same string but different key/context, we can introduce a context variable to the Memory.
But going this route raises another question:
- Should we suggest a memory if it has a different key/context than the unit being translated ?
- If so, should we also apply another penalty factor for those ?

nijel: You're right with `origin`, sorry for not realizing that.

Either context or a foreign key to Unit is needed to not delete entries for other strings. https://github.com/WeblateOrg/weblate/issues/9947 also describes a feature request to include context, so that is probably a better approach. Therefore, I've assigned it to you. Using the penalty approach for context mismatch is probably a good idea. I think 0.95 would work well in this case. We still want a 100% source match with context mismatch to rank above any machine translation.

gersona:
- Requires https://github.com/WeblateOrg/weblate_schemas/pull/469 and weblate_schemas `2025.5` release
- The features related to TMX  described in [this comment](https://github.com/WeblateOrg/weblate/issues/9947#issuecomment-1718988513) were not included as I think they're worth getting extra attention and discussion
- Context display in the suggestion list was not included (yet?) as it will impact all code using `TranslationResultDict`

weblate/trans/models/_conf.py
```diff
@@ -96,6 +96,7 @@ class WeblateConf(AppConf):
     DEFAULT_ACCESS_CONTROL = 0
     DEFAULT_RESTRICTED_COMPONENT = False
     DEFAULT_SHARED_TM = True
+    AUTOCLEAN_TM = False
```
nijel: Shouldn't it have `DEFAULT_` prefix as others?

weblate/trans/migrations/0042_project_autoclean_tm.py
```diff
@@ -0,0 +1,25 @@
...
+    dependencies = [
+        ("trans", "0041_remove_unit_trans_unit_pending_remove_unit_pending"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="project",
+            name="autoclean_tm",
+            field=models.BooleanField(
+                default=False,
```
nijel: This should use the current value from the setting rather than hardcoding it. Without that users with non-default settings will get a warning about missing migration.

weblate/trans/models/unit.py
```diff
@@ -2076,15 +2076,18 @@ def update_translation_memory(
...
+            )
         ):
             return
 
         translation = self.translation
         component = translation.component
         if (
             (not translation.is_source or component.intermediate)
-            and self.state >= STATE_TRANSLATED
+            and (self.state >= STATE_TRANSLATED or self.state != self.old_unit["state"])
```
nijel, line 2090: This will push units with STATE_FUZZY when changed from STATE_TRANSLATED, is that intentional?

weblate/memory/tasks.py
```diff
@@ -93,45 +95,75 @@ def update_memory(
...
     source: str,
+    context: str,
     target: str,
     origin: str,
     add_shared: bool,
     add_project: bool,
     add_user: bool,
     user_id: int | None,
     project_id: int,
+    state: int,
```
nijel: Maybe call it `unit_state` to clarify that this is different from the `state` in the translation memory? Originally I wrote that it should be called `status`, but then I realized that this is a unit state.

weblate/trans/models/unit.py
gersona: It is intentional. If the unit state is downgraded, this allows this change to be reflected in the suggestion quality. 
> If a translation's status is changed (e.g. to "needs editing"), the associated memory's status is reverted to pending.

Do you think it's still relevant ?

weblate/trans/models/_conf.py
gersona: That's right, I'll fix that

weblate/trans/models/unit.py
gersona: One of the tests displays this behavior
https://github.com/WeblateOrg/weblate/blob/f0ba39cdab4242fc4daa070c6b8e1be98f084314/weblate/memory/tests.py#L312-L323

nijel: Ah, that makes sense. This will also pass STATE_TRANSLATED -> STATE_EMPTY transitions, but that will be discarded by `is_valid_memory_entry` below.

docs/admin/management.rst
```diff
@@ -219,6 +219,17 @@ manually, as the cleanups happen automatically in the background.
 
    :ref:`production-cron`
 
+
+cleanup_memory
+----------------
```
nijel:
```suggestion
cleanup_memory
--------------
```

docs/admin/memory.rst
```diff
@@ -64,6 +64,32 @@ installations, as it can have severe implications:
...
+.. _autoclean-tm:
+
+Autoclean translation memory
+-----------------------------
+
+.. versionadded:: 5.13
+
+The translation memory is automatically cleaned up by removing obsolete and outdated entries.
+
+In the Docker container this can be configured using :envvar:`WEBLATE_AUTOCLEAN_TM`.
```
nijel:
```suggestion
In the Docker container this can be configured using :envvar:`WEBLATE_DEFAULT_AUTOCLEAN_TM`.
```

PR #15319, opened 2025-06-30
fix(zen editor): give better visual feedback on pending save

### LINKED ISSUE #15261: Improve visual feedback on editor changes
### Describe the problem

Since #14389 the editor indicates pending changes. Since #14821 the saving in the Zen mode is delayed. There are few issues with this:

* The “Unsaved changes” text caused layout changes, what makes it annoying.
* There is no visual indication that the pending change will be saved.

### Solution brainstorm

I think “Unsaved changes” text could go away, sticking with just border change. Or it has to be added in a way that will not cause layout changes.

Perhaps `unit-state-saving` style could be applied once the timer is started to indicate that the string will be saved?

_No response_

### Screenshots

_No response_

### Additional context

_No response_

KarenKonou:
* Adds visual feedback when there's a save operation pending (using the `unit-state-saving` animation).
* Fixes some logic issues caused by shared global variables between handlers resulting in a save cancelling when tabbing from one zen unit to another and possibly other bugs.
* Moves the "unsaved changes" indicator next to the toolbar, preventing a layout change. (thought maybe it might be better next to the character counter or somewhere else?)
image

#### Pending question:

Currently, there is a timeout before the `handleTranslationChange` function is called an and then another timeout before the actual save operation occurs. I'm not sure if this is necessary, maybe a timeout before `handleTranslationChange` is enough and then the save operation can continue without a further delay?

Fixes #15261

### DIFF
--- a/weblate/static/editor/zen.js
+++ b/weblate/static/editor/zen.js
@@ -225,9 +225,7 @@
           addAlert(val.text, val.kind);
         });
 
-        $row
-          .removeClass("translation-modified")
-          .addClass("translation-saved");
+        $row.removeClass("translation-modified").addClass("translation-saved");
         $row.find("#unsaved-label").remove();
         $row.find(".translation-editor").removeClass("has-changes");
 


nijel:
* Placing the message next to the character counter is probably more intutive as that is also state indicator, while to toolbar is place where actions are.
* As for the nested timeouts, having one would be probably enough. The primary motivation is to avoid saving incomplete strings when clicking on the placeholders, what causes short focus loss on the editor.

PR #15324, opened 2025-06-30
feat(api): add user contact field

michael-smt:
Adds the new contact field from #14930 to the API.

### DIFF
--- a/weblate/api/serializers.py
+++ b/weblate/api/serializers.py
@@ -222,7 +222,9 @@ class LanguageSerializer(serializers.ModelSerializer[Language]):
 
 class FullUserSerializer(serializers.ModelSerializer[User]):
     contact = serializers.CharField(
-        source="profile.contact", allow_blank=True, read_only=True,
+        source="profile.contact",
+        allow_blank=True,
+        read_only=True,
     )
     groups = serializers.HyperlinkedIdentityField(
         view_name="api:group-detail",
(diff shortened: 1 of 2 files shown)


weblate/api/serializers.py
```diff
@@ -221,6 +221,11 @@ def get_value(self, dictionary):
 
 
 class FullUserSerializer(serializers.ModelSerializer[User]):
+    contact = serializers.CharField(
+        source="profile.contact",
+        allow_blank=True,
+        read_only=True,
+    )
```
nijel, line 228: There shouldn't be need for this, the default serializer should. work just fine.

```suggestion
```

michael-smt: Thanks for the review. When I remove this section and use `profile.contact` directly in the `fields` I get:
> django.core.exceptions.ImproperlyConfigured: Field name `profile.contact` is not valid for model `User` in `weblate.api.serializers.FullUserSerializer`

Also tried providing the source in the `extra_kwargs` but no luck.

Is there some cleaner way to access this field via the Profile model?

nijel: Ah, we don't have a serializer for the `Profile` model, I thought you're just adding a field to it, but it's an `User` serializer. We really should have a full support for all `Profile` fields, not introduce support just for one of them.

michael-smt: Ok yes makes sense. Closing this for a full `Profile` endpoint instead.

PR #15326, opened 2025-06-30
Use Flags object on file format flags

### LINKED ISSUE #13598: Use Flags object on file format flags
### Describe the problem

The current `flags` interface in file formats (`TranslationUnit`) returns strings. The only reason for this is that this API predates the introduction of the `Flags` object. This makes inheritance unnecessary complex and brings in repeated `Flags` parsing and serializing.

* Change `TranslationUnit.flags` to return `Flags` objects.
* Change the overriden methods to extend super() return value to ease implementation of features like #13597.
* Add explicit flags serialization to string in `Unit.update_source_unit`.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

gersona:
- fixes #13598

### DIFF
--- a/weblate/checks/flags.py
+++ b/weblate/checks/flags.py
@@ -7,7 +7,6 @@ from __future__ import annotations
 
 import re
 from functools import lru_cache
-from types import NoneType
 from typing import TYPE_CHECKING, Any, ClassVar, cast
 
 from django.core.exceptions import ValidationError
@@ -289,9 +288,9 @@ class Flags:
 
     def __eq__(self, other: object) -> bool:
         """Check if current flags are equal to other flags."""
-        if type(other) not in {Flags, str, etree._Element, NoneType}:
+        if not isinstance(other, Flags):
             return False
-        return self.items() == set(self.get_items(other))
+        return self.items() == other.items()
 
     def __hash__(self) -> int:
         return hash(self.items())
--- a/weblate/checks/flags.py
+++ b/weblate/checks/flags.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 
 import re
 from functools import lru_cache
+from types import NoneType
 from typing import TYPE_CHECKING, Any, ClassVar, cast
 
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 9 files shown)


weblate/formats/convert.py
```diff
@@ -102,7 +102,7 @@ def is_translated(self):
     def flags(self):
         flags = Flags(super().flags)
```
nijel: Should not need to convert it:

```suggestion
        flags = super().flags
```

weblate/checks/flags.py
```diff
@@ -286,6 +286,10 @@ def __contains__(self, key: str) -> bool:
     def __bool__(self) -> bool:
         return bool(self._items)
 
+    def equals(self, other: object) -> bool:
```
gersona: I'm not entirely sure whether this is better than implementing the `__eq__` method, but the logic would remain the same

nijel: Having standard ``__eq__`` is probably better.

weblate/checks/flags.py
```diff
@@ -286,6 +286,10 @@ def __contains__(self, key: str) -> bool:
...
+        """Check if current flags are equal to other flags."""
+        return self.items() == set(self.get_items(other))
```
nijel: There should be type checking here to avoid crash in `get_items` later:

```suggestion
        if not isinstance(other, Flags):
             return False
        return self.items() == set(self.get_items(other))
```

weblate/checks/flags.py
```diff
@@ -288,6 +289,8 @@ def __bool__(self) -> bool:
 
     def __eq__(self, other: object) -> bool:
         """Check if current flags are equal to other flags."""
+        if type(other) not in {Flags, str, etree._Element, NoneType}:
```
nijel: I believe using `type` is wrong here because it would not match subclasses, `isinstance` would work better.

nijel: Furthermore, I'm hesitating whether to allow so wide matching. I think comparing to `Flags` only would work just fine, make the code simpler and more consistent with how other Python classes behave.

gersona: Sure, I'll fix that

nijel:

PR #15352, opened 2025-07-03
bugfix(pending-unit): source string explanation commit to file

### LINKED ISSUE #15344: Lots of email about "Attempt to parse store without a filename"
### Describe the issue

Starting from about June 30th, I receive "Parsing failed" emails every hour about "Attempt to parse store without a filename" for the [Godot Glossary](https://hosted.weblate.org/projects/godot-engine/glossary/en/) repository.

### I already tried

I believe we changed nothing recently.

### Expected behavior

_No response_

### Screenshots

_No response_

### Exception traceback

```pytb

```

### How do you run Weblate?

weblate.org service

### Weblate versions

_No response_

### Weblate deploy checks

```shell

```

### Additional context

_No response_

amCap1712:
No translation file exists for source strings in bilingual formats, hence we should not try to commit explanation changes for the same to a file. Do not create a PendingUnitChange entry is a source unit and the file format is bilingual.

Fixes #15344.

### DIFF
--- a/weblate/trans/models/unit.py
+++ b/weblate/trans/models/unit.py
@@ -2008,10 +2008,16 @@ class Unit(models.Model, LoggerMixin):
                     unit=unit,
                     author=user,
                 )
-            PendingUnitChange.store_unit_change(
-                unit=self,
-                author=user,
-            )
+
+            # translation file does not exist for source strings in bilingual formats
+            if (
+                not self.is_source
+                or self.translation.component.file_format_cls.monolingual
+            ):
+                PendingUnitChange.store_unit_change(
+                    unit=self,
+                    author=user,
+                )
 
         if save:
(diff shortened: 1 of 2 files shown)


weblate/trans/models/unit.py
```diff
@@ -2008,10 +2008,16 @@ def update_explanation(
...
+
+            # translation file does not exist for source strings in bilingual formats
+            if (
+                not self.is_source
+                or self.translation.component.file_format_cls.monolingual
+            ):
+                PendingUnitChange.store_unit_change(
+                    unit=self,
+                    author=user,
+                )
```
nijel, line 2020: Thanks, this is more future-proof than the original logic before a74d58071ea5e632c46de6bd3c370a3606e6cba8 which simply didn't set the pending flag on the source string at all.

Issue #2084, opened 2018-06-13
Discovery addon: component creation based on the monolingual template or new base


benwillig:
Currently, the discovery addon creates automatically components if the specified regular expression matches translation files.

In our implementation, the .pot files are generated automatically. But in some cases, there isn't any translations file. It means that a developer must push a translation file for the component to be created.

The idea would be to create the component if the .pot file is found. Then, any translators (even not a programmer) could start a new translation.

Thanks

dsnoeck:
Hi @nijel, how complex it is to implement this enhancement ? Where should I start if I want to have a try ?

nijel:
The discovery code is in [weblate/trans/discovery.py](https://github.com/WeblateOrg/weblate/blob/master/weblate/trans/discovery.py). The problem is that currently it's based on assumption it detects translated files, so it will need some adjustments to work in this case as well.

dsnoeck:
With the limited knowledge of Python/Django I have, I gave up ! But found a workaround, while I'm adding a template.pot file for new component, I'm also creating a file test.po which will be matched by the regex.

Jibec:
Same request here, i create a fake fr.po file by copying the pot file, and do it for thousands of files (the fedora documentation).

m-aciek:
I'm also interested in this feature; I wonder if potentially [msgmerge](https://docs.weblate.org/en/latest/admin/addons.html#update-po-files-to-match-pot-msgmerge) could be a place implementing that behavior (although this way it would be more difficult to configure it probably).

surli:
Any work planned on that feature? I realized only yesterday this behaviour for component discovery: I was counting on it to create automatically new components when a new module was created with a source translation without any translation, and we discovered that quite a few modules were missing in our weblate instances since we put that in place.

carohadad:
Just commenting here that this will be helpful for our projects at [@Horizontal-org](https://github.com/Horizontal-org) . We use Docusaurus sites for websites and documentation of our projects [Tella](https://tella-app.org/) and [Shira](https://www.shira.app/). T[he English version of the site is not added as a translation, it's in the root of the project ](https://docusaurus.io/docs/i18n/introduction). So in order for the add-on to automatically discover new pages we need to manually add those to the translation files.  We wanted to avoid the manual part when we added the automatic discovery add-on.

whenever this is coded, it will simplify our process a lot.

[...]

PR #15356, opened 2025-07-04
docs: Note component discovery requirement for at least one translation

nikclayton:
Component discovery doesn't work if a base file exists but there are no translations (https://github.com/WeblateOrg/weblate/issues/2084).

Document this requirement as a warning.

While I'm here, add a hint that recommends pairing with the "Add missing languages" add-on to ensure that newly discovered components converge on a consistent set of languages.

### DIFF
--- a/docs/admin/addons.rst
+++ b/docs/admin/addons.rst
@@ -446,7 +446,7 @@ Matching files:
 
    To be discovered a new component must contain a file that matches ``base_file_template``
    **and** at least one file with a name that matches ``match`` -- i.e., the component must
-   contain a base language file and at least one existing translation. Otherwise it will be
+   contain a base language file and at least one existing translation. Otherwise, it will be
    ignored.
 
 .. seealso::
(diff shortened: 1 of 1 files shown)


docs/admin/addons.rst
```diff
@@ -437,6 +437,18 @@ Matching files:
...
    This saves time with configuration and system resources too.
 
+.. hint::
+
+   Ensure the new component contains the full set of translatable languages with
+   :ref:`addon-weblate.consistency.languages`.
+
+.. warning::
+
+   To be discovered a new component must contain a file that matches ``base_file_template``
```
nijel:
```suggestion
   To be discovered, a new component must contain a file that matches ``base_file_template``
```

docs/admin/addons.rst
```diff
@@ -437,6 +437,18 @@ Matching files:
...
+   **and** at least one file with a name that matches ``match`` -- i.e., the component must
+   contain a base language file and at least one existing translation. Otherwise it will be
```
nijel:
```suggestion
   contain a base language file and at least one existing translation. Otherwise, it will be
```

nijel:

PR #15366, opened 2025-07-06
feat: introduce project level commit quality policy

amCap1712:
For each unit, finds the last pending change that makes it eligible for committing based on the project's commit policy, and return all changes up to and including that change. Currently two policies are supported, include only approved changes and skip needs-editing changes.


amCap1712:
Introducing a commit policy to exclude units with failing checks is harder. The `unit.has_failing_check` status can be recorded in the `PendingUnitChange` table. But currently, the checks run after the `PendingUnitChange` object is created in `Unit.save_backend`, if we postpone the `PendingUnitChange` creation to after the checks have run, it would work. However, in certain cases can be skipped by passing `run_checks=False` to `save_backend` and will have to look more into it to understand what further corner cases are present. 

Also, IIUC checks are not immediately run for bulk translations etc. and in that case the current logic would still not work as the failing check status would not be reflected in the `PendingUnitChange`.

One possible way would to update existing `PendingUnitChange` from `run_checks` based on `target` but that would not be atomic and in cases when checks run with sufficient delay after a change, the failing check changes would get committed.

nijel:
Not covering failing checks is okay, there are enforced checks which can achieve this. The strings with failing enforced check will be marked as needing editing, and this is a policy you've implemented.

weblate/trans/models/pending.py
```diff
@@ -30,8 +30,32 @@ def for_component(self, component: Component):
         return self.filter(unit__translation__component=component)
 
     def for_translation(self, translation: Translation):
-        """Return pending changes for a specific translation."""
-        return self.filter(unit__translation=translation)
+        """Return pending changes for a specific translation based on commit policy."""
+        from django.db.models import OuterRef, Q, Subquery
```
nijel: This import can be on top of the file.

weblate/trans/models/pending.py
```diff
@@ -30,8 +30,32 @@ def for_component(self, component: Component):
...
+
+        # For each unit, finds the last change that makes it eligible for committing
+        # based on the project's commit policy, and returns all changes up to and
+        # including that change.
+        latest_eligible_changes = (
+            PendingUnitChange.objects.filter(*filters, unit_id=OuterRef("unit_id"))
+            .order_by("-timestamp")
+            .values("timestamp")[:1]
+        )
+        return qs.filter(timestamp__lte=Subquery(latest_eligible_changes))
```
nijel, line 57: How does the subquery scale with many pending changes? Unfortunately, I don't have much experience with this.

I was originally thinking about filtering by the current unit state instead of investigating all the pending changes, but your approach is better because it would commit changes which approved the string even if the string is no longer approved.

amCap1712: I'd expect overall query in case of say 'without needs-editing' policy to be,
```
SELECT puc.*
  FROM trans_pendingunitchange AS puc
  JOIN trans_unit              AS u 
    ON u.id = puc.unit_id
 WHERE u.translation_id = :translation_id
   AND puc.timestamp
       <= (
          SELECT MAX(puc2.timestamp)
            FROM  trans_pendingunitchange AS puc2
           WHERE puc2.unit_id = puc.unit_id
             AND puc2.state <> 10
       );
```
I can see one improvement based on the model, the state field does not have an index so the database would have to do a seq scan. We could add an index on state field and I think it should scale fine. 

Alternatively, we could also change the existing index on unit field to a multi-field covering index `(unit, state) including timestamp` which would ideally lead to index-only scans for the subquery (no table row hits so faster than checking two indexes and reading data from table/index for timestamp). 
The downside is that queries filtering only on state and not on unit_id would not be able to use this index. I'll check the codebase again to be sure but off the top of my head I don't think that's something we are or will be doing. Further, given the limited kind of queries performed against the PendingUnitChange table, I think such targeted optimization might be fine in this case.

nijel: I don't think the Django ORM will optimize into `MAX()` here, it will be instead `ORDER BY timestamp DESC LIMIT 1`, but it doesn't matter much in this case.

Having the additional index is probably a good way to go. On the other side, I don't expect there will be typically a massive amount of pending changes stacked per unit, so we're optimizing for a corner case here.

nijel:

nijel (approved):

PR #15400, opened 2025-07-10
feat: Automatic flags based on string location

### LINKED ISSUE #13597: Automatic flags based on string location
### Describe the problem

Users have to manually add flags like `rst-text` or `md-text` for translation files extracted from reStructuredText (for example using Sphinx) or Markdown documents (using po4a). This should not be necessary when there is information that this string was extracted from `.rst` or `.md` file.

The `flags` property of `TranslationUnit` could inspect `locations` and add flags based on the filename.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

gersona:
- fixes #13597

### DIFF
--- a/docs/user/checks.rst
+++ b/docs/user/checks.rst
@@ -2033,3 +2033,15 @@ For example with Gettext in Python it could be:
 
 
 .. _reStructuredText Primer: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html
+
+
+.. _default-location-flags:
+
+Default location flags
+----------------------
+
+Some flags are added to strings by default based on their locations.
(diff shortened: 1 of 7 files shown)


docs/user/checks.rst
```diff
@@ -2033,3 +2033,15 @@ For example with Gettext in Python it could be:
 
 
 .. _reStructuredText Primer: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html
+
+
+.. _default-location-flags:
+
+Default location flags
```
nijel: I'm not sure that “default” is the right word here. How about:

```suggestion
Location-based automatic flags
```

gersona: That's definitely clearer

PR #15404, opened 2025-07-10
Move sort icons to the left for right-aligned table columns

softworkz:
## Before

image

## After

image

### Rationale

When the headings are not properly right-aliged, the table views are very confusing (the percentage bars add to that visual confusion btw).
I'm not aware of any CSS trick to move the spans to the other side (except float, but then it fully moves to the other edge), the order can only be changed at the source.

Hope this makes sense to you.


KarenKonou (approved):

nijel:

Issue #14645, opened 2025-04-22
Python segmentation fault on MacOS 15.4


nijel:
### Describe the issue

Since the update to macOS 15.4 the Python installed from brew crashes while running the test suite with segmentation fault. This can be seen here: https://github.com/WeblateOrg/weblate/actions/runs/14592037763/job/40929306650

### I already tried

- [x] I've read and searched [the documentation](https://docs.weblate.org/).
- [x] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

CI job does that.

### Exception traceback

```pytb

```

### Weblate deploy checks

```shell

```

### Additional context

I've asked at Homebrew forum, but that gave no hints: https://github.com/orgs/Homebrew/discussions/6094

nijel:
For now I've marked macos CI as not required when merging pull requests.

gersona:
Issue related to the version of the LDC compiler, the

nijel:
Let's see if that helps https://github.com/WeblateOrg/weblate/pull/14942

nijel:
Okay, I think that's a different issue. I don't think there is anything compiled with ldc in our stack and ldc is not even installed on the CI system.

nijel:
I've saved crash reports from the CI, but I can't find anything useful there: https://github.com/nijel/weblate/actions/runs/15155810723

[Python-2025-05-21-072126.ips.txt](https://github.com/user-attachments/files/20362908/Python-2025-05-21-072126.ips.txt)

gersona:
It might be another library available by default on macos 15.4 that has been compiled with LDC. 
At this point, it's too tricky to go down the rabbit hole of chasing the faulty one. I don't know if it's just more reasonable to wait for an OS update , but this may also take weeks before it's realeased

nijel:
The CI environment is not critical, so we can wait there.

PR #15406, opened 2025-07-10
Introduce template block for nav_pills

softworkz:
Hi,

first of all thanks for your hard work on this software! It excels in many aspects: it's well organized, clean and transparent and very well documented.  Setting up a production instance (Docker-less) with direct dev-to-production deployment was still an enormous task. I noticed too late that settings_docker is still what I need even without Docker. But well - that's fair - you cannot expect to get everything served for free on a silver plate.. 😉 

There's just a single big issue I have and that's about the layout. From top to bottom, there's a menu with some items, then there's a breadcrumps bar and then - there's (often) once another menu bar - right in the middle of the page where you normally expect content. Sometimes there are messages shown above it, which moves that bar even further down. 
This is somewhat odd - at least you don't see that very often. More typically it's Logo/Header > Menubar 1 > Menubar 2 > Breadcrumps > Notifications > Content > Footer 

That's what I wanted to achieve at least - unfortunately it cannot be reasonably done with CSS only, because the secondary menu bar ("nav-pills") is inside the main content container and you cannot reasonably get it out of that and re-order.
(it's possible to set body to display:flex, flex-dir:column, set display:contents for main-content and then order:1,2,3,4 for the elements, but that's ugly and then you don't have a main content container anymore for border/background differentiation)

Looking at the templates it appeared straightforward to introduce a new block nav_pills for the 2nd menu-bar, I did so and it worked well. With this PR it will be possible to make a single change in base.html in order to move the nav-pills bar above the breadcrumps (needs appropriate CSS of course).

I hope it's okay the way I did it (no experience with Django) - the one case I wasn't able to figure out in limited time is the project settings view (`/settings/<project-name>/`) - is this creating the pills dynamically?

For illustration, here's the final result:

image

(the site is for translation of our own software, not a commercial service, the breakdown of OneSky forced us into it..)

Best wishes,
sw


softworkz:
Looking at the [history of the macOS runs](https://github.com/WeblateOrg/weblate/actions/workflows/macos.yml), it's nothing I need to be concerned about, right?

nijel:
Yes, that's a known issue - https://github.com/WeblateOrg/weblate/issues/14645

nijel:

softworkz:
For reference, here's how I'm using it to show the secondary menu bar ("nav pills") above the breadcrumps bar - like shown in the screenshot above: https://github.com/softworkz/weblate/commit/fd4f368bb387a6a88270ac406d4545e9d2eaefb1

PR #15411, opened 2025-07-10
feat: indicate that language edits might be reverted

### LINKED ISSUE #15407: Edits to Language Names are getting reverted
### Describe the issue

When I the names of languages in the UI and then I update the deployment (which always runs migrate), the names are getting reset back to their original values. 
Now - I've read about the UPDATE_LANGUAGES setting - but I still don't quite understand this behavior. 
It makes sense to me that the language database is kept up-to-date, but is it actually intended that the language names are getting overwritten?

I mean - how can the software offer to edit something when it's clear that those edits will get lost with the next update?

### I already tried

Edit a language name
Update software

### Expected behavior

Edited name persists

### Screenshots

_No response_

### Exception traceback

```pytb

```

### How do you run Weblate?

Other

### Weblate versions

_No response_

### Weblate deploy checks

```shell

```

### Additional context

_No response_

nijel:
- motivate users to contribute to the Weblate language data package
- add warning that edits might be reverted upon update

Fixes #15407

### DIFF
--- a/weblate/templates/lang/language_edit_warning.html
+++ b/weblate/templates/lang/language_edit_warning.html
@@ -1,0 +1,15 @@
+{% load i18n translations %}
+
+<div class="alert alert-info">
+
+  {% blocktranslate with url="https://github.com/WeblateOrg/language-data/" %}Please consider submitting your edit to the <a href="{{ url }}">Weblate language data repository</a> so that it can be included in the future Weblate releases.{% endblocktranslate %}
+  <div class="clearfix"></div>
+</div>
+
+{% if update_languages %}
+  <div class="alert alert-warning">
+    {% documentation "admin/languages" "included-languages" as url %}
+    {% blocktranslate %}The language definitions are automatically updated by Weblate upon database migration. This might revert your edits. <a href="{{ url }}">Read more in the documentation.</a>{% endblocktranslate %}
+    <div class="clearfix"></div>
+  </div>
+{% endif %}
--- a/weblate/templates/lang/language_edit_warning.html
+++ b/weblate/templates/lang/language_edit_warning.html
@@ -9,7 +9,7 @@
 {% if update_languages %}
   <div class="alert alert-warning">
     {% documentation "admin/languages" "included-languages" as url %}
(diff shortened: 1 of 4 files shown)


weblate/templates/lang/language_edit_warning.html
```diff
@@ -0,0 +1,15 @@
+{% load i18n translations %}
+
+<div class="alert alert-info">
+
+  {% blocktranslate with url="https://github.com/WeblateOrg/language-data/" %}Please consider submitting your edit to the <a href="{{ url }}">Weblate language data repository</a> so that it can be included in the future Weblate releases.{% endblocktranslate %}
+  <div class="clearfix"></div>
+</div>
```
softworkz, line 7: Do you think that makes sense? At least from what I believe it is that 99% of edits that users are making here are not the kind of changes that you would want for your reference database.

Things like:

- Rename `Chinese (Simplified Han script)` to `Chinese (Simplified)` - simply due to column widths
- Rename `Norwegian Bokmål` to `nb-NO (Norwegian Bokmål)` because none of us even knows what "Bokmål" is

At least,  I think the banner is a bit too much. Maybe a small text line below the Save button?

weblate/templates/lang/language_edit_warning.html
```diff
@@ -0,0 +1,15 @@
...
+
+{% if update_languages %}
+  <div class="alert alert-warning">
+    {% documentation "admin/languages" "included-languages" as url %}
+    {% blocktranslate %}The language definitions are automatically updated by Weblate upon database migration. This might revert your edits. <a href="{{ url }}">Read more in the documentation.</a>{% endblocktranslate %}
```
softworkz: "Language definitions are overwritten by Weblate during database migration. This will likely revert your edits. Read the documentation about how to prevent this."

orangesunny: Thanks, this is useful.

weblate/templates/lang/language_edit_warning.html
orangesunny: @nijel ? :)

orangesunny (approved):

weblate/templates/lang/language_edit_warning.html
nijel: I have absolutely no clue what are the major edits here, so I assumed that in most cases these could be contributed back. Still, the message says "Please consider".

Issue #3149, opened 2019-10-25
Flat XML - configure default tag names


kgronowski:
Hello,

I have a question about the flatxml format.
Is it possible to configure/customize default xml tag names that weblate should use when reading / creating new translation files as is possible using the appropriate options (root, value, key) in [Translate Toolkit](http://docs.translatehouse.org/projects/translate-toolkit/en/latest/commands/flatxml2po.html)?

Basically I would like to change FlatXML format definition from 
```
<root>
  <str key="hello_world">Hello World!</str>
  <str key="resource_key">Translated value.</str>
</root>
```
to e.g. 
```
<dictionary xmlns="urn:translate-toolkit:flat-xml-dictionary">
  <entry name="hello_world">Hello World!</entry>
  <entry name="resource_key">Translated value.</entry>
</dictionary>
```

nijel:
You currently need to define new file format in Weblate which would pass needed params. It's not possible to specify this on the UI, see https://github.com/WeblateOrg/weblate/issues/1897

kgronowski:
Thank you for your answer.
Unfortunately I'm not really familiar with Python and Django and I'm not sure how I am supposed to add new format with needed params. I assume that I should add/create something similar to what can be found in [ttkit.py](https://github.com/WeblateOrg/weblate/blob/011dd3d7201b6f13b09d65f32539d4a0e66adcbc/weblate/formats/ttkit.py) but where/how should I set my params for converter.

Btw, I'm using Docker container for Weblate.

nijel:
The easiest approach seems to be to subclass `FlatXMLFile` overriding its `__init__` and changing parameters to desired values. You can then use the newly defined class as `loader` in the Weblate format class:

https://github.com/WeblateOrg/weblate/blob/011dd3d7201b6f13b09d65f32539d4a0e66adcbc/weblate/formats/ttkit.py#L1233-L1241

nijel:
Reopening as this can now be built on top of https://github.com/WeblateOrg/weblate/pull/15416.

PR #15416, opened 2025-07-11
feat: Parameters for File formats

gersona:
@nijel I have a couple of questions from this initial work:
- [x] Should the file format addons be removed in this PR as well, or are there some steps to respect before getting rid of Addon classes, may be to allow users to remove the class from setting ?
- [x] Should the file format params fields be displayed on the discovery forms ? Or is it enough to have them on the component creation and setting forms ?
- [x] The issue mentions that "the parse method will accept these params arg". From what I understand, those params are used in the `Translation.handle_source` method and in the translate toolkit store classes (via the `dump_args` attribute). Is that correct or there are other places where they are expected ?

---------
### Screenshots
file_format_settings

### DIFF
--- a/weblate/trans/file_format_params.py
+++ b/weblate/trans/file_format_params.py
@@ -186,7 +186,7 @@ class GettextPoLineWrap(BaseFileFormatParam):
     ]
     default = 77
     help_text = gettext_lazy(
-        "By default gettext wraps lines at 77 characters and at newlines."
+        "By default gettext wraps lines at 77 characters and at newlines. "
         "With the --no-wrap parameter, wrapping is only done at newlines."
     )
 


nijel:
The current add-ons should be replaced by the new parameters. So there should be migration that would turn all existing add-ons to appropriate configuration (keep in mind project- and site-wide add-ons, which affect only components with matching file format).

I'd skip parameters in the discovery for now. The discovery should copy the parameters from the main component if the file format is the same. We might add parameters later in case demand appears.

The current add-ons only configure the serializing part, so it is okay to pass them as class attributes (for example, `dump_args` for JSON, but this differs per format).

With the new approach, we want to support parameters that alter parsing as well. To have some examples:

* Configuring custom tags in flat XML format (see https://github.com/WeblateOrg/weblate/issues/3149)
* Configuring encoding for Properties or CSV formats (this is currently done by having more formats differing just with an encoding)

To achieve this, the translate-toolkit class needs to be customized before parsing, so these parameters need to be passed to `TranslationFormat.load`.

The `Translation.handle_source` is influenced as well, because it consumes these parameters and passes them to `TranslationFormat.update_bilingual`. This is a bit of a special case because `MsgmergeAddon` will stay while `GettextCustomizeAddon` shall be migrated into file format parameters. Maybe the msgmerge setting could be exposed in the parameters while adding a note that automatic triggering is only done when the add-on is installed?

weblate/trans/format_params.py
```diff
@@ -0,0 +1,223 @@
...
+        "yaml",
+        "ryml",
+        "yml",
+        "ruby-yaml",
+    )
+
+    def get_widget_attrs(self) -> dict:
+        attrs = super().get_widget_attrs()
+        attrs["data-file-formats"] = " ".join(self.file_formats)
+        return attrs
```
nijel: Doesn't this duplicate the `fileformats` attribute?

nijel (commented):
This will need to review all `store_post_load.send` invocations, because these are used to inject add-on customization to loading. I think this trigger/signal should no longer be needed once customizing add-ons are migrated to the settings and could be removed.

weblate/trans/format_params.py
gersona: You're right, it's a duplicate

nijel:
While updating the docs for gettext PO (https://github.com/WeblateOrg/weblate/pull/15542) I've realized another thing that will need updating: describe the parameters in the file format documentation.

gersona:
@nijel Some final questions:
- should the stale addons be actually deleted as it will also cascade into deleting some AddonActivityLogs and Events ?

weblate/trans/management/commands/list_file_format_params.py
```diff
@@ -0,0 +1,74 @@
...
+            key=operator.itemgetter(0),
+        )
+
+        def max_string_length(list_of_lists):
+            return max(len(s) for sublist in list_of_lists for s in sublist)
+
+        name_width = max_string_length([row[0] for row in table])
+        formats_width = max(len(s) for r in table for s in r[1])
+        label_width = max_string_length([row[2] for row in table])
+        help_width = max(len(s) for r in table for s in r[3])
```
nijel: We're already repeating this logic three times, this adds fourth copy. https://github.com/WeblateOrg/weblate/pull/15639 factors out the table rendering into standalone function.

nijel:
Sorry I missed this question yesterday:

> should the stale addons be actually deleted as it will also cascade into deleting some AddonActivityLogs and Events ?

Yes, these should be deleted. Actually, activity was not really logged for these (`AddonEvent.EVENT_STORE_POST_LOAD` is excluded from the log because it happens too often).

docs/admin/addons.rst
```diff
@@ -1605,6 +1615,10 @@ Allows adjusting XML output behavior, for example closing tags.
 Customize YAML output
 ---------------------
 
+.. deprecated:: 5.13
+
+  :guilabel: Replaced by :ref:`file_format_params`
+
```
nijel: It was actually removed, not deprecated:

```suggestion
.. versionremoved:: 5.13 Replaced by :ref:`file_format_params`.

```

docs/formats.rst
```diff
@@ -506,3 +506,18 @@ support for it in the `translate-toolkit`_.
...
 
 .. _translate-toolkit: https://toolkit.translatehouse.org/
+
+.. _file_format_params:
+
+File format parameters
+----------------------
+
+File format parameters provide a way to configure settings related to the file format.
+They are configured at component level
```
nijel: Incomplete sentence

weblate/trans/management/commands/list_file_format_params.py
```diff
@@ -0,0 +1,74 @@
...
+                    (
+                        "",
+                        f"``{value}``".replace("\\", "\\\\"),
+                        f"  {description}".replace("\\", "\\\\"),
+                    )
+                )
+        return result
+
+    def handle(self, *args, **options) -> None:
+        """List file format parameters."""
```
nijel, line 42: Please add note about generated content here (and include it in the rst file), so that it is clear that the rst file should not be edited manually. Something like:

https://github.com/WeblateOrg/weblate/blob/bae7bfd437602955b690a53da9285f4a561e1d09/weblate/addons/management/commands/list_addons.py#L69-L71

weblate/trans/migrations/0045_component_file_format_params.py
```diff
@@ -0,0 +1,178 @@
...
+    for file_formats, params in configs["component_wide"].get(component.id, []):
+        if component.file_format in file_formats:
+            file_format_params.update(params)
+
+    return file_format_params
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("trans", "0044_merge_20250721_1211"),
```
nijel: As this migration is operating on add-ons, it should have a dependency on that.

weblate/trans/migrations/0045_component_file_format_params.py
```diff
@@ -0,0 +1,178 @@
...
+
+    operations = [
+        migrations.AddField(
+            model_name="component",
+            name="file_format_params",
+            field=models.JSONField(
+                blank=True, default=dict, verbose_name="File format parameters"
+            ),
+        ),
+        migrations.RunPython(migrate_addons_config, migrations.RunPython.noop),
```
nijel: Make the migration elidable when squashing them:

```suggestion
        migrations.RunPython(migrate_addons_config, migrations.RunPython.noop, elidable=True),
```

weblate/trans/file_format_params.py
```diff
@@ -0,0 +1,309 @@
...
+        (-1, gettext_lazy("No line wrapping")),
+    ]
+    default = 77
+    help_text = gettext_lazy(
+        "By default gettext wraps lines at 77 characters and at newlines."
+        "With the --no-wrap parameter, wrapping is only done at newlines."
+    )
+
+    def setup_store(self, store: TranslationFormat, **file_format_params) -> None:
+        if wrapper := cast("BasePoFormat", store).store.wrapper:
```
nijel: This mypy error should be addressed by https://github.com/WeblateOrg/weblate/pull/15763.

weblate/trans/file_format_params.py
```diff
@@ -0,0 +1,309 @@
...
+
+@register_file_format_param
+class XMLClosingTags(BaseFileFormatParam):
+    name = "xml_closing_tags"
+    label = gettext_lazy("Include closing tag for blank XML tags")
+    field_class = forms.BooleanField
+    default = True
+
+    @classproperty
+    def file_formats(self) -> list[str]:
```
nijel: Perhaps:

```suggestion
    def file_formats(self) -> Sequence[str]:
```

amCap1712:
https://github.com/WeblateOrg/weblate/blob/783a829f1aa620f0391af75ca9194155e6d688fc/weblate/addons/events.py#L15

Can be removed?

weblate/trans/file_format_params.py
gersona: oh yes, I was wondering why it failed, I didn't realize it was drawing from a different path

gersona:
> https://github.com/WeblateOrg/weblate/blob/783a829f1aa620f0391af75ca9194155e6d688fc/weblate/addons/events.py#L15
> 
> Can be removed?

or should it rather be commented out as in ?
https://github.com/WeblateOrg/weblate/blob/e6e5ee5138fb2b687004d3cd127778ca06f7bd25/weblate/trans/actions.py#L37

nijel:
Probably better to leave the comment there so that it is clear why a value is missing.

weblate/trans/file_format_params.py
nijel: Well the intention in the past was that the C implementation would eventually become the default one in translate-toolkit, and we would migrate to it. But since then we started to rely on the Python implementation in several areas, and it should be clearly declared.

weblate/static/loader-bootstrap.js
```diff
@@ -1643,4 +1643,48 @@ $(function () {
     "font-size: 20px; font-family: sans-serif",
     gettext("See https://en.wikipedia.org/wiki/Self-XSS for more information."),
   );
+
+  /* Display relevant file_format_params field in Component forms */
+  const form_auto_ids = ["id", "id_scratchcreate"];
```
amCap1712, line 1648: When you try to create a new component from scratch, the file format options fields are not displayed on the first time. The options are only displayed after you select another format.

image

(after selecting another format and then the default again)

image

amCap1712: Because `displayRelevantFileFormatParams` is only triggered on the `change` event, it should also be called after initial form setup.

amCap1712: Also, I cannot see the options to modify the params in the component settings but they are present in the html so seems like its the same issue.

weblate/formats/base.py
```diff
@@ -385,6 +387,14 @@ def __init__(
             is_template=is_template,
         )
 
+        self.post_store_load_setup(file_format_params)
```
amCap1712: A bit confused. Since the `file_format_params` is passed to `self.load` already, why is this step needed?

weblate/trans/file_format_params.py
```diff
@@ -0,0 +1,309 @@
...
+
+@register_file_format_param
+class JSONOutputSortKeys(JSONOutputCustomizationBaseParam):
+    name = "json_sort_keys"
+    label = gettext_lazy("Sort JSON keys")
+    field_class = forms.BooleanField
+    default = False
+
+    def setup_store(self, store: TranslationFormat, **file_format_params) -> None:
+        dump_args = getattr(store.store, "dump_args", {})
```
amCap1712: The default `{}` is not useful here because it is not set on `store.store` instance later. For clarity, better to default to `None` and return early if attribute is absent. Same for other params below.

nijel: It should not be absent here (the parameter should be applied only to matching formats). This was probably done to avoid type errors. The typing on file formats is a bit messy for now...

amCap1712: I see, how about using manual `cast` like in https://github.com/WeblateOrg/weblate/blob/c105c32e3ceed78c13259479e8c905ca2cf8dcf3/weblate/trans/file_format_params.py#L307

nijel: That's probably the least annoying workaround for now.

The proper solution would be to extend type annotations for `TranslationFormat` (and `TranslationUnit`) to correctly expose types of underlying objects. But that is clearly out of scope in this pull request :-).

weblate/formats/base.py
gersona: Correct, that could also go in the `load` method of `TTKitFormat` as it's mainly to configure the output behavior. Initially, the `file_format_params` was passed to `load` to provide customization for file parsing.

nijel:
Does anybody see a blocker for merging this?

@gersona Anything you still have in your mind?

@KarenKonou Can you please review the frontend code?

@amCap1712 Any additional feedback?

@orangesunny Any feedback on the documentation changes? It is rendered at https://weblate--15416.org.readthedocs.build/en/15416/, the params are documented at https://weblate--15416.org.readthedocs.build/en/15416/formats.html#file-format-params

KarenKonou (approved):
Frontend looks good to me

gersona:
@nijel I'm good, just waiting for any additional comments from reviews

nijel (approved):

orangesunny (approved):

orangesunny:
Looks fine, nice work! I only added a few spaces. :)

weblate/trans/file_format_params.py
```diff
@@ -0,0 +1,307 @@
...
+        ),
+        (
+            65535,
+            gettext_lazy("Only wrap lines at newlines (like 'xgettext --no-wrap')"),
+        ),
+        (-1, gettext_lazy("No line wrapping")),
+    ]
+    default = 77
+    help_text = gettext_lazy(
+        "By default gettext wraps lines at 77 characters and at newlines."
```
nijel:
```suggestion
        "By default gettext wraps lines at 77 characters and at newlines. "
```

nijel:
Scheduled for merge, thanks!

PR #15422, opened 2025-07-12
Unify logic for license requirement when changing project access

softworkz:
This applies the same logic like in alert.py: MissingLicense(BaseAlert)

### Specific problem: 

Vars set according to docs:

```
DEFAULT_ACCESS_CONTROL=200
LICENSE_REQUIRED=False
LICENSE_FILTER = set()
```
This way, you can never make any project public.

### Simplified scenario:

You only set: 

```
LICENSE_FILTER = set()
```

Now, when you change a project to private, you cannot change it back to public anymore.

### DIFF
--- a/weblate/trans/forms.py
+++ b/weblate/trans/forms.py
@@ -2193,12 +2193,14 @@ class ProjectSettingsForm(SettingsBaseForm, ProjectDocsMixin, ProjectAntispamMix
                     )
                 }
             )
-        if self.changed_access and access in {
-            Project.ACCESS_PUBLIC,
-            Project.ACCESS_PROTECTED,
-        }:
+        if self.changed_access and access == Project.ACCESS_PUBLIC:
             unlicensed = self.instance.component_set.filter(license="")
-            if unlicensed:
+            if (
+                unlicensed
+                and settings.LICENSE_REQUIRED
+                and not settings.LOGIN_REQUIRED_URLS
+                and (settings.LICENSE_FILTER is None or settings.LICENSE_FILTER)
+            ):
                 raise ValidationError(
                     {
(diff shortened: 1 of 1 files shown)


nijel (changes requested):
It would be better to actually share the logic by reusing the code, not by copying it.

softworkz:
> It would be better to actually share the logic by reusing the code, not by copying it.

I totally agree. I'm just not familiar enough with the whole project to know where it would need to be placed.

The PR is still an improvement because 

- A: Currently: Two places - different logic	
- B: This PR: Two places - same logic
- C: Desirable: Singple place - same logic

So it's still an incremental improvement.

Options:

- Let me know how you want it to be done and I'll update the PR
- You merge the PR and do C yourself
- We close the PR and you do it yourself

I'm good with either way - thanks!

nijel:
It is an improvement, but it still makes it likely to break in the future as the code is copied and can easily get out of sync again. So once this has been identified I prefer to fix it for good rather than keeping the original cause. It would be great, if you could update the PR to address this. Additionally, the tests need to be adjusted because they are now failing.

softworkz:
It surely makes more sense to use shared code for this. You probably know already pretty well how you want to have it done and before I even start finding how it is likely supposed to be done in ths project, you would be done with it already. So it would surely be more economic when you do it directly. Besides, I've gotten a little short on time (spent too much on [15469](https://github.com/WeblateOrg/weblate/issues/15469)) 😉 

Thanks

nijel:
https://github.com/WeblateOrg/weblate/pull/16382 makes the logic actually unified.

softworkz:
Great!

Thank you very much.

PR #15458, opened 2025-07-16
🩹 — Use startswith() for comparing deepl URLs

ldidry:
In order to catch deepl URL if people add a trailing slash to the URL, let’s use startswith() instead of `==`.


nijel:

ldidry:
Thx for the software :-)

PR #15494, opened 2025-07-18
docs: describe placing custom code in Docker container

nijel:
The feature is there for years, it just was never properly documented. It also covers use-case described in #15493.


FSeidinger-XI:
followed your example and put a python file unter /app/data/python/customize.

First of all, I had to create the directory /app/data/python/customize in my custom entry point, because it is not present in the container (using Weblate 5.12). I guess that this should not be necessary and that a django app should already exist there.

And my guess is also, that for this reason it gets not picked up.

Did I miss something here?

Hint: My colleague Markus Schmitt tried to get a support contract with Weblate but it is still not active because we need a CV from your employee.

nijel:
Answered at https://github.com/orgs/WeblateOrg/discussions/15493...

Issue #15427, opened 2025-07-11
Memory Leak in File Import leading to OOM kill of Celery Worker


softworkz:
### Describe the issue

When importing translation files (TBX), there appears to be a memory leak. 

The Celery worker process memory usage is constantly growing until it gets eventually stopped by the oom killer.

It might have gone unnoticed because it takes a larger amount of strings to get to those limits.

### I already tried

- [x] I've read and searched [the documentation](https://docs.weblate.org/).
- [x] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

- Go to [Microsoft Terminology](https://learn.microsoft.com/en-us/globalization/reference/microsoft-terminology#downloading-microsoft-terminology)
- Download [Microsoft Terminology](https://download.microsoft.com/download/b/2/d/b2db7a7c-8d33-47f3-b2c1-ee5e6445cf45/MicrosoftTermCollection.zip)
- Create a new Glossary Project in Weblate
- Create a new Component (Upload Translation Files)
- Choose the downloaded zip file to upload

Now wait for a long time and watch the Celery worker process memory.

At the end, you will have a partial import.  

You can then go to "Repository Maintenance" and choose "Rescan" -  then it will go through all files again.

> [!NOTE]
> Interestingly, the scanning of already imported files does not seem to leak. The growth starts only when importing strings that are not imported yet.

### Expected behavior

Memory does not leak and gets freed after use.

### Kernel Log

```
2025-07-11T20:54:47 kernel: python3 invoked oom-killer: gfp_mask=0x140dca(GFP_HIGHUSER_MOVABLE|__GFP_COMP|__GFP_ZERO), order=0, oom_score_adj=0
2025-07-11T20:54:47 kernel: CPU: 2 UID: 999 PID: 48282 Comm: python3 Not tainted 6.11.0-1015-azure WeblateOrg/docker#15~24.04.1-Ubuntu
2025-07-11T20:54:47 kernel: Hardware name: Microsoft Corporation Virtual Machine/Virtual Machine, BIOS Hyper-V UEFI Release v4.1 08/23/2024
2025-07-11T20:54:47 kernel: Call Trace:
2025-07-11T20:54:47 kernel:  <TASK>
2025-07-11T20:54:47 kernel:  dump_stack_lvl+0x5b/0x70
2025-07-11T20:54:47 kernel:  dump_stack+0x10/0x20
2025-07-11T20:54:47 kernel:  dump_header+0x48/0x1d0
[...]

PR #15502, opened 2025-07-18
feat(translation): bulk create source strings while parsing

nijel:
This significantly improves performance of parsing bilingual translations when a lot of new source strings are introduced.

Issue #15427


nijel:
Great idea, I've extended `ComponentTest` to verify the sanity of the source strings.

PR #15538, opened 2025-07-22
feat(ghost-translations): use add project language for ghost translations

amCap1712:
1. Refactor function to determine components eligible for language addition for reuse.
2. Fix and simplify logic to determine which languages are not available in all components. Original logic was double counting source languages in components.
3. Make + button for add new translation work.
4. Adding a ghost translation uses add new project language.

Questions/Clarifications:
1. If a language is present in a project already but not in all components, the current logic shows it twice. Once as a normal translation language stat and once as ghost. This is obviously confusing and not ideal so I think if a language is present in any component in the project, we should not show a ghost translation. Similarly, if a language is present in any component in a category, it should not be shown at the category level. Thoughts?
2. The values of the stats are incorrect for Ghost Translations in some cases, for instance ghost translations are shown as completed instead of empty. Should I compute the total stats from the source translation of the component in this case? What about project and category level, sum up counts from all source translations of consisting components?
image

### DIFF
--- a/weblate/trans/tests/test_dashboard.py
+++ b/weblate/trans/tests/test_dashboard.py
@@ -59,14 +59,22 @@ class DashboardTest(ViewTestCase):
         self.assertEqual(len(response.context["componentlists"]), 1)
 
     def test_component_list_ghost(self) -> None:
+        new_component = self.create_po_new_base(
+            project=self.component.project, name="New component", new_lang="add"
+        )
         clist = ComponentList.objects.create(name="TestCL", slug="testcl")
-        clist.components.add(self.component)
+        clist.components.add(new_component)
+
+        response = self.client.get(reverse("home"))
+        self.assertNotContains(response, "Spanish")
 
         self.user.profile.languages.add(Language.objects.get(code="es"))
 
         response = self.client.get(reverse("home"))
-
-        self.assertContains(response, "Spanish")
+        # testing for language name in text is insufficient as profile languages name
+        # is always present in the languages drop down in the top bar and hence, it would
+        # pass the test even if the ghost translation was not added.
+        self.assertContains(response, f"{new_component} — Spanish")
 
     def test_user_component_list(self) -> None:
         clist = ComponentList.objects.create(name="TestCL", slug="testcl")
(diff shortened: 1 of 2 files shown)


nijel:
> * If a language is present in a project already but not in all components, the current logic shows it twice. Once as a normal translation language stat and once as ghost. This is obviously confusing and not ideal so I think if a language is present in any component in the project, we should not show a ghost translation. Similarly, if a language is present in any component in a category, it should not be shown at the category level. Thoughts?

I agree, it would be better to not show any language twice.
 
> * The values of the stats are incorrect for Ghost Translations in some cases, for instance ghost translations are shown as completed instead of empty. Should I compute the total stats from the source translation of the component in this case? What about project and category level, sum up counts from all source translations of consisting components?

You should be able to use `source_*` stats directly, and yes, it makes more sense to display ghosts as untranslated.

amCap1712:
One more question, when the user tries to add a translation from a component page, should the language be added to all components or just that component? I think the latter makes more sense as it is consistent with adding a new translation in general from the component page.

Also, can you please explain what you mean by `source_*` stats? From my understanding and playing around, category and project langauge stats have all source stats set to 0.

nijel:
> One more question, when the user tries to add a translation from a component page, should the language be added to all components or just that component? I think the latter makes more sense as it is consistent with adding a new translation in general from the component page.

I think adding at component should add only to that component (what is the current behavior). 
 
> Also, can you please explain what you mean by `source_*` stats? From my understanding and playing around, category and project langauge stats have all source stats set to 0.

It shouldn't be zero, this is what is shown in Source row in https://hosted.weblate.org/projects/weblate/#information.

weblate/templates/snippets/list-objects.html
```diff
@@ -175,6 +175,7 @@
         <tr id="{{ row_id }}">
           <td class="actions" rowspan="2">
             {% if object.is_ghost %}
+              {# {% perm 'translation.add' object.component as user_can_add_translation %} #}
```
nijel: I guess this should have been removed instead of commented out?

amCap1712: This is needed for the case where object is GhostTranslation, because particular components that might have restrictive language addition rules. Had commented only for isolating a bug.

nijel:
Can you please take a look at the missing test coverage? To me it seems that category language is missing tests.

nijel:

PR #15542, opened 2025-07-23
doc: document gettext wrapping and add-ons

nijel:


johnzhou721:
Hey @nijel, thanks for the PR! I was the one asking on Ticket#2011999 -- however, is the wrapping behavior based on the current gettext versions installed on the servers? If so, could you please reveal the version of gettext / Translate Toolkit currently installed on the Hosted Weblate servers?  Thanks.

nijel:
The wrapping is done by translate-toolkit, we typically run the latest version. The imporant thing to note is that gettext 0.23 changed the behavior and translate-toolkit followed in https://github.com/translate/translate/pull/5437.

PR #15545, opened 2025-07-23
feat(ci): try rundev on different platforms

nijel:
It should be possible to use the same workflow on macos and windows, so test it. See #15536


nijel:
Okay, Windows can't use containers on GitHub Actions, so we can't really test it there. The same applies to macos-15, so let's stay with Linux only testing of this.

More info:

* https://github.com/orgs/community/discussions/25491 
* https://github.com/douglascamata/setup-docker-macos-action/issues/35

PR #15576, opened 2025-07-26
chore: improve error message for incompatible backups restore

### LINKED ISSUE #15372: Provide friendlier error messages on failed project restore
### Describe the issue

When project backup is not compatible with current version it fails with a horrible error message:

Image

The full JSON validation error message is useless to the user and too long to be useful to anybody.

### I already tried

1. Go to Hosted Weblate
2. Backup a project
3. Try to restore it in a stable Weblate version

### Expected behavior

We could probably safely assume that this is caused by newer backup and tell user to upgrade. The full validation message could be hidden.

### Screenshots

_No response_

### Exception traceback

```pytb

```

### How do you run Weblate?

weblate.org service

### Weblate versions

_No response_

### Weblate deploy checks

```shell

```

### Additional context

_No response_

amCap1712:
Trying to restore backups from a newer version of Weblate into an older version, for example hosted weblate to self-hosted stable release instance, can lead to errors. The full error message is very gnarly and not useful to users, hence replace such errors with a simpler message.

Fix #15372.

### DIFF
--- a/weblate/trans/forms.py
+++ b/weblate/trans/forms.py
@@ -2443,10 +2443,12 @@ class ProjectImportForm(BillingMixin, forms.Form):
         try:
             backup.validate()
         except jsonschema.exceptions.ValidationError as error:
+            version = backup.data.get("metadata", {}).get("version", "unknown")
             raise ValidationError(
                 gettext(
-                    "Could not load project backup: The backup is from an incompatible version. Please upgrade your Weblate instance."
+                    "Could not load project backup: The backup is from an incompatible version (%(version)s). Please upgrade your Weblate instance."
                 )
+                % {"version": version}
             ) from error
         except Exception as error:
             raise ValidationError(


weblate/trans/forms.py
```diff
@@ -2441,6 +2442,12 @@ def clean_zipfile(self):
         backup = ProjectBackup(fileio=zipfile)
         try:
             backup.validate()
+        except jsonschema.exceptions.ValidationError as error:
+            raise ValidationError(
+                gettext(
+                    "Could not load project backup: The backup is from an incompatible version. Please upgrade your Weblate instance."
+                )
```
nijel, line 2450: `backup.data["metadata"]["version"]` should contain Weblate version that did the backup; it might be useful to include it here.

amCap1712: done.

nijel:

PR #15585, opened 2025-07-27
feat: use search queries for automatic translation

### LINKED ISSUE #4107: Use search queries for autotranslate
Right now autotranslate is probably the only remaining user of old filters interface. It should be migrated to new queries to allow more complex filtering and bring consistency with other user interface.

* Change the autotranslate api to accept query string rather than filter.
* Use SearchField display in in the form (both manual and addon).
* Migrate existing addon configuration from filters to search query.

Current approach works, but is inconsistent and relies on code which is not used elsewhere.

**Additional context**
https://github.com/WeblateOrg/weblate/issues/4106 should apply here as well.

amCap1712:
The Search query UI/field/parser power searching everywhere else in Weblate. It makes sense to migrate the remaining filter_type code for to search queries as well.

Fixes #4107.


weblate/addons/migrations/0007_update_autotranslate_filter_type.py
```diff
@@ -0,0 +1,60 @@
...
+
+    for addon in addons:
+        config = addon.configuration
+        old_filter = config.pop("filter_type")
+        config["q"] = filter_mapping.get(old_filter, old_filter)
+        addon.save(update_fields=["configuration"])
+
+
+def reverse_update(apps, schema_editor):
+    """Reverting this migration is not possible as search queries support more versatile filters than filter types."""
```
amCap1712: Should the original_filter type be retained in the configuration to allow for a reverse migration?

nijel: Reverse migrations are nice to have if it's not too much of work. In this particular case, the `filter_type` can be kept in the configuration during the migration, and the reverse migration will be a no-op. The additional field should cause no harm and will disappear once somebody saves the configuration.

weblate/addons/migrations/0007_update_autotranslate_filter_type.py
```diff
@@ -0,0 +1,60 @@
...
+        "unapproved": "state:translated",
+        "todo": "state:<translated",
+        "nottranslated": "state:empty",
+        "translated": "state:>=translated",
+        "suggestions": "has:suggestion",
+        "nosuggestions": "-has:suggestion state:<translated",
+        "comments": "has:comment",
+        "allchecks": "has:check",
+        "all": "",
+    }
```
nijel: `weblate.trans.filter.FilterRegistry` already has this mapping, so it probably could be reused.

weblate/addons/migrations/0007_update_autotranslate_filter_type.py
```diff
@@ -0,0 +1,60 @@
...
+
+    addons = Addon.objects.filter(
+        name="weblate.autotranslate.autotranslate", configuration__has_key="filter_type"
+    )
+
+    for addon in addons:
+        config = addon.configuration
+        old_filter = config.pop("filter_type")
+        config["q"] = filter_mapping.get(old_filter, old_filter)
```
nijel: Keeping an unknown old filter will most likely break things; it might be better to fail during the migration.

amCap1712: This is not for unknown filter but filters of type `check:CHECK_NAME`, `label:LABEL_NAME` etc.

nijel: Makes sense, adding a comment it would make it easier to understand.

amCap1712: Just checked that AutoForm disables the use of such filters anyway hence simplifying.

weblate/addons/migrations/0007_update_autotranslate_filter_type.py
amCap1712: Updated.

weblate/addons/migrations/0007_update_autotranslate_filter_type.py
amCap1712: Fixed.

weblate/trans/forms.py
```diff
@@ -919,9 +919,9 @@ class AutoForm(forms.Form):
         label=gettext_lazy("Automatic translation mode"),
         initial="suggest",
     )
-    filter_type = FilterField(
+    q = QueryField(
```
nijel, line 906: This change adds element with duplicate id=`id_q` to `AutoForm`. The form `__init__` should probably override `auto_id` to avoid that, see for example `ReplaceForm`.

This is also the reason for failing test.

amCap1712: Thanks, should be fixed now.

weblate/trans/forms.py
```diff
@@ -920,9 +903,9 @@ class AutoForm(forms.Form):
...
         required=True,
```
nijel, line 907: This has the same issue as described in #15686, but let's address it there.

weblate/addons/migrations/0007_update_autotranslate_filter_type.py
```diff
@@ -0,0 +1,44 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("addons", "0006_alter_addonactivitylog_component"),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            update_autotranslate_filter_type,
+            migrations.RunPython.noop,
+        ),
```
nijel, line 44: Allow discarding this migration when squashing:

```suggestion
            elidable=True,
        ),
```

nijel:
Sheduled for merge, thanks for your contribution!

PR #15597, opened 2025-07-29
fix(docker): correctly set oidc title and image

Wouter0100:
Due to a copy-paste error, the OIDC title and images were set on the oauth0 OIDC title and image. This commit corrects it.


nijel:

PR #15618, opened 2025-07-30
Add  CSS classes to footer and support menu items

softworkz:
This makes it a lot easier to customize these via custom CSS.

### DIFF
--- a/weblate/templates/base.html
+++ b/weblate/templates/base.html
@@ -390,7 +390,7 @@
                       <a href="https://weblate.org/support/">{% translate "Get professional support" %}</a>
                     </li>
                     {% if get_help_url %}
-                    <li class="supportmenu-gethelp">
+                      <li class="supportmenu-gethelp">
                         <a href="{{ get_help_url }}">{% translate "Get help" %}</a>
                       </li>
                     {% endif %}
--- a/weblate/templates/base.html
+++ b/weblate/templates/base.html
@@ -374,49 +374,49 @@
                   </span>
                 </a>
                 <ul class="dropdown-menu">
-                  <li class="dropdown-header">{% translate "Get help" %}</li>
-                  <li>
+                  <li class="dropdown-header supportmenu-header-help">{% translate "Get help" %}</li>
+                  <li class="supportmenu-contact">
                     <a href="{% url 'contact' %}">{% translate "Contact server admins" %}</a>
                   </li>
-                  <li>
+                  <li class="supportmenu-community">
                     <a href="https://github.com/orgs/WeblateOrg/discussions">{% translate "Community discussions" %}</a>
                   </li>
                   {% if support_status.has_support %}
-                    <li>
+                    <li class="supportmenu-support">
                       <a href="https://care.weblate.org/">{% translate "Professional support" %}</a>
                     </li>
                   {% else %}
-                    <li>
+                    <li class="supportmenu-support">
                       <a href="https://weblate.org/support/">{% translate "Get professional support" %}</a>
                     </li>
                     {% if get_help_url %}
-                      <li>
+                    <li class="supportmenu-gethelp">
                         <a href="{{ get_help_url }}">{% translate "Get help" %}</a>
                       </li>
                     {% endif %}
                   {% endif %}
 
-                  <li class="dropdown-header">{% translate "Documentation" %}</li>
-                  <li>
+                  <li class="dropdown-header supportmenu-header-documentation">{% translate "Documentation" %}</li>
+                  <li class="supportmenu-documentation">
                     <a href="{% documentation 'index' %}">{% translate "Documentation" %}</a>
                   </li>
-                  <li>
+                  <li class="supportmenu-shortcuts">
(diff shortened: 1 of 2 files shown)


nijel:
What kind of customization are you aiming at? I don't feel that adding a CSS class to each element is the way to go.

softworkz:
> What kind of customization are you aiming at?

Hiding and re-ordering. It is simply about being able to select specific items via CSS.

softworkz:
Or what I did today - add some text:

```css
footer ul li.footer-privacy a::before {
    content: 'ACME '
}
```

softworkz:
If you prefer IDs instead of classes, I can change that. The problem with Ids though, is the uniquenes requirement. From long experience, it's the same repeating story: you use Ids because you think that uniqueness won't be a problem but at some point later, it bites you. Could be years or just an hour ;-)

nijel:
It really doesn't matter whether it is `id` or `class`, my main concern is whether this is reasonable for us:

* The HTML structure and attributes will never be considered a stable interface for third-party use when doing changes on our code. So these might go away or be renamed when needed for further development or when somebody doesn't realize that it might be used elsewhere during cleanup.
* On the other side, accepting such a pull request would look like a promise to maintain this, which is something we cannot guarantee.
* I'm not sure if we want to make removing branding that easy. We want to encourage users to purchase a support subscription, because that's what funds Weblate development.

softworkz:
> * The HTML structure and attributes will never be considered a stable interface for third-party use when doing changes on our code. So these might go away or be renamed when needed for further development or when somebody doesn't realize that it might be used elsewhere during cleanup.
> * On the other side, accepting such a pull request would look like a promise to maintain this, which is something we cannot guarantee.

I do not expect that. I also do not ultimately need it. 

Yet I try to contribute back anything that might be valuable for the project, that's why I'm aiming to minimize direct changes to the Weblate source and keep all customizations outside as much as possible.
Making direct changes is much easier though: It takes me 60s for getting a local source code change deployed to the production site. Creating a PR on a different base commit, submitting it to GitHub,, explaining, defending and discussing it, takes a whole lot more time than the change itself. 

> I'm not sure if we want to make removing branding that easy. 

This doesn't really make it easier to removing branding. Here's just one of many ways:

```css
ul.dropdown-menu li:has(a[href*='weblate' i]) {
    display: none;
}
```

which doesn't need classes or ids and removes all links to Weblate (same can be done for the footer).

Note that we do not remove all branding and links. It's just a bit too much and confuses users. "Powered by Weblate" with link is still there and documentation links are going to Weblate, so those who are really interested in the software that is driving the site, will easily find their way to you. 


> We want to encourage users to purchase a support subscription, because that's what funds Weblate development.

That was the original intention actually, but then it turned out that there's so much more needed that I had to get hands-on myself, digging deep into the subject down to the source code level. 

I understand when you reject it from your point of view, but anybody who is able to run a self-hosted Weblate (in a serious way) will be easily able to manipulate those links in any desired way, so I don't believe that it matters in reality whether those classes exist or not..

nijel:
> Yet I try to contribute back anything that might be valuable for the project, that's why I'm aiming to minimize direct changes to the Weblate source and keep all customizations outside as much as possible.

I understand your motivation, but I fail to see a motivation to merge *and maintain* such a change. I don't see any value in this for users or developers. For example, https://github.com/WeblateOrg/weblate/pull/15406 improved the template structure, so it was reasonable to merge even if we don't agree on the final navigation placement.

> Making direct changes is much easier though: It takes me 60s for getting a local source code change deployed to the production site. Creating a PR on a different base commit, submitting it to GitHub,, explaining, defending and discussing it, takes a whole lot more time than the change itself.

Yes, making direct changes is quick. Maintaining them long-term will be a different story.

We are most likely going to break your custom CSS with https://github.com/WeblateOrg/weblate/pull/15113 anyway.

> I understand when you reject it from your point of view, but anybody who is able to run a self-hosted Weblate (in a serious way) will be easily able to manipulate those links in any desired way, so I don't believe that it matters in reality whether those classes exist or not..

You focused too much on the last point I've posted, and it was last for a reason: I see the previous ones as more important :slightly_smiling_face:. I know there are many ways to debrand Weblate; I just don't want to encourage that.

softworkz:
> For example, #15406 improved the template structure, so it was reasonable to merge even if we don't agree on the final navigation placement.

Yea, for the few remaining cases which are not covered by this, I don't really dare to submit a PR, because the changes for these are not that elegant like the one I submitted..

> Yes, making direct changes is quick. Maintaining them long-term will be a different story.

That's exactly my point. The more of those "special changes" I make to the source, the more painful will it be to stay in-sync, and eventually this usually makes me stop caring about upstream development and decouple from it.

For example, creating an OAuth2 provider for enabling login via our user forums (Invision) could nicely be done as a component - but getting Weblate to show the user profile photos meant to patch avatar.py - directly.  This kind of building up and growing debt is what's making things painful in the future.
The Bootstrap update is a lot less worrying. The custom CSS is a lot less than usual, shouldn't take long to adapt (unless you combine it with some more fundamental changes...

Anyway, I don't mind you closing this. 
Thanks

PR #15621, opened 2025-07-30
feat: use headers in the support menu

nijel:
I think this better structures the menu items making it easier to navigate. Also hides the donate link for sites with purchased support.

After:

image

Before:

image


nijel:
@KarenKonou This might have an impact on #15113 because I don't think we've used this before.

KarenKonou:
Thanks for letting me know, I'll rebase my branch once this is merged and check

PR #15678, opened 2025-08-04
Multi-Level Rate Limiting for Notification E-Mails

softworkz:
Multi-Level Rate Limiting for Notification E-Mails

Yesterday, a user had received almost 1000 notification e-mails after enabling notifications on new strings.

I've seen that there's a hardcoded rate lmit of 1000 e-mails per user per day - which is not much of a limit of course.
The problem why it's often increased after introduction, is that users get cut off from receiving e-mails for  any hours.

For example: User recives 1000 e-mails at once within 1h - then they are blocked from receiving any further e-mails for 23h.

A multi-level rate-limt method allows to equalize this and avoid bursts.
You'll probably  want to change the default values. The number of levels is not fixed. You can also define just one level with the previous values and you get the same result like before (if somebody really wants that).

### DIFF
--- a/weblate/accounts/notifications.py
+++ b/weblate/accounts/notifications.py
@@ -44,7 +44,7 @@ from weblate.trans.models import (
 )
 from weblate.utils.errors import report_error
 from weblate.utils.markdown import get_mention_users
-from weblate.utils.ratelimit import rate_limit
+from weblate.utils.ratelimit import multi_level_rate_limit
 from weblate.utils.site import get_site_domain, get_site_url
 from weblate.utils.stats import prefetch_stats
 from weblate.utils.version import USER_AGENT
@@ -282,11 +282,14 @@ class Notification:
         self, address: str, subject: str, body: str, headers: dict[str, str]
     ) -> None:
         encoded_email = siphash("Weblate notifier", address)
-        if rate_limit(f"notify:rate:{encoded_email}", 1000, 86400):
+        is_blocked, reason = multi_level_rate_limit(encoded_email, settings.RATELIMIT_NOTIFICATION_LIMITS)
+        
+        if is_blocked:
             LOGGER.info(
-                "discarding notification %s to %s after sending too many",
+                "discarding notification %s to %s due to rate limit: %s",
                 self.get_name(),
                 address,
+                reason,
             )
         else:
             self.outgoing.append(
--- a/weblate/settings_example.py
+++ b/weblate/settings_example.py
@@ -950,6 +950,14 @@ DATABASE_BACKUP = "plain"
 # Enable auto updating
 AUTO_UPDATE = False
 
+# Multi-level rate limiting for email notifications
+# Each tuple contains (max_emails, time_window_seconds)
+RATELIMIT_NOTIFICATION_LIMITS = [
+    (3, 120),    # Prevent burst sends - 3 emails per 2 minutes
+    (10, 3600),  # Equalize to avoid getting blocked for too long - 10 emails per hour
+    (50, 86400), # Daily limit: 50 emails per day
+]
+
 # PGP commits signing
 WEBLATE_GPG_IDENTITY = None
 
--- a/weblate/utils/ratelimit.py
+++ b/weblate/utils/ratelimit.py
@@ -159,3 +159,57 @@ def session_ratelimit_post(scope: str, logout_user: bool = True):
         return wraps(function)(_rate_wrap)
 
     return _session_ratelimit_post_controller
+
+
+class RateLimitCache:
+    """Cache wrapper for rate limiting with atomic check/decrement operations."""
+    
+    def __init__(self, base_key: str, attempts: int, window: int):
+        self.cache_key = f"notify:rate:{base_key}:{attempts}:{window}"
+        self.attempts = attempts
+        self.window = window
+        
+        # Initialize the bucket (atomically on redis) if it doesn't exist
+        if not is_redis_cache():
+            if cache.get(self.cache_key) is None:
+                cache.set(self.cache_key, self.attempts, self.window)
+        else:
+            cast("RedisCache", cache).set(self.cache_key, self.attempts, self.window, nx=True)
+    
+    def would_block(self) -> bool:
+        """Check if rate limit would be exceeded without decrementing counter."""
+        
+        current = cache.get(self.cache_key, 0)
+        return current <= 0 
+    
+    def decrement(self) -> None:
+        """Decrement rate limit counter for an email that will be sent."""
+        
+        with suppress(ValueError):
+            cache.decr(self.cache_key)
+
+
+def multi_level_rate_limit(encoded_email: str, rate_limits) -> tuple[bool, str]:
+    """
+    Multi-level rate limiting for email notifications.
+    
+    Returns:
+        tuple: (is_blocked, reason)
+    """
+
+    level_caches = [
+        RateLimitCache(encoded_email, attempts, window)
+        for attempts, window in rate_limits
+    ]
+    
+    # Check all without decrementing
+    for level_cache in level_caches:
+        if level_cache.would_block():
+            return True, f"rate limit exceeded ({level_cache.attempts}/{level_cache.window}s)"
+    
+    # If we get here, we can send - so decrement ALL counters
+    for level_cache in level_caches:
+        level_cache.decrement()
+    
+    return False, ""
+


weblate/utils/ratelimit.py
```diff
@@ -159,3 +159,57 @@ def _rate_wrap(
...
+        """Check if rate limit would be exceeded without decrementing counter."""
+        
+        current = cache.get(self.cache_key, 0)
+        return current <= 0 
+    
+    def decrement(self) -> None:
+        """Decrement rate limit counter for an email that will be sent."""
+        
+        with suppress(ValueError):
+            cache.decr(self.cache_key)
```
nijel, line 222: This class duplicates code from the current rate limiter, there should be just implementation of that.

weblate/accounts/notifications.py
```diff
@@ -282,11 +282,14 @@ def send(
         self, address: str, subject: str, body: str, headers: dict[str, str]
     ) -> None:
         encoded_email = siphash("Weblate notifier", address)
-        if rate_limit(f"notify:rate:{encoded_email}", 1000, 86400):
+        is_blocked, reason = multi_level_rate_limit(encoded_email, settings.RATELIMIT_NOTIFICATION_LIMITS)
```
nijel: `encoded_email` is an integer here; see the mypy error.

nijel (changes requested):
Thanks for your contributions. Can you please:

* Address comments I've posted.
* Add documentation for this.
* Add tests for the new rate limiter.
* Adjust current notification tests to not be limited by the limiter?

Overall, it might be better to make this a generic way to configure limits. Add support for the `LIMITS` suffix, which would be used if present.

weblate/utils/ratelimit.py
softworkz: Totally re-worked, no more duplication

softworkz:
> Thanks for your contributions. Can you please:
> 
> * Address comments I've posted.

Done

> * Add documentation for this.

Not yet,

> * Add tests for the new rate limiter.

Done

> * Adjust current notification tests to not be limited by the limiter?

Done


The first version was a bit half-baked, admittedly. I didn't want to spend too much time without knowing whether you are interested in this, but now I have done it in a nice and clean and unified way.

Note that I've dropped the is_redis_cache distinction. From my research cache.add() does exactly what's needed (i`f get is none then set`) and is atomic with a Redis backend.

> Overall, it might be better to make this a generic way to configure limits. Add support for the `LIMITS` suffix, which would be used if present.

Yea, that makes sense, but I'm not sure what you mean by "LIMITS" suffix..

Do you mean:

- If there's a `RATELIMIT_XXXXX_LIMITS` variable
  - interpret it as a list of 2-tuples `(attempts, window)`
  - ignore any `RATELIMIT_XXXXX_ATTEMPTS` and `RATELIMIT_XXXXX_WINDOW` variables
- If there's no such variable but a pair of  `RATELIMIT_XXXXX_ATTEMPTS` and `RATELIMIT_XXXXX_WINDOW` variables
  - Use them to build a single-item `RATELIMIT_XXXXX_LIMITS` variable
- Treat `RATELIMIT_XXXXX_LOCKOUT` separately (like before)


I'm wondering whether the "lockout" is even still needed with multi-level limiting? It can also be avoided easily (assuming somebody is developing a bot to shave-off information or tries to overload the site. What I mean is this:

Defaults

```python
    RATELIMIT_ATTEMPTS = 5
    RATELIMIT_WINDOW = 300  # 5 minutes
    RATELIMIT_LOCKOUT = 600  # 10 minutes
```

When somebofy is blindly hammering, 

- they will run into the lock for 10 minutes after 5 successful requests
- after the 10min have elapsed, it will happen again and they're locked for anouther 10 minutes
- afte one hour, they'll have 30 successful requests

but somebody more clever will

- make exactly 5 requests and wait 5 minutes
  => never run into the lock
- after 5 minutes - another 5 requests
- after an hour they' have 60 successful requests


But when you add a second level - like: `[(5, 300). (3600, 30)]]`
Nobody will be able to make more than 30 per hour, no matter how they're doing it...

weblate/utils/ratelimit.py
```diff
@@ -159,3 +109,119 @@ def _rate_wrap(
...
+class CacheCounterItem:
+    cache_key: str
+    attempts: int
+    window: int
+
+    def __init__(self, base_key: str, attempts: int, window: int):
+        self.cache_key = f"{base_key}:{attempts}:{window}"
+        self.attempts = attempts
+        self.window = window
+        cache.add(self.cache_key, attempts, window)
```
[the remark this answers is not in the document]
softworkz: > Using cache.add() in **init** may fail silently if the key already exists, potentially causing inconsistent initialization.

There's no "silent failure". Read the docs.

> Set a value in the cache if the key does not already exist.

Which is exactly the desired behavior. With a Redis backend,, the  `add()` is also atomic,

softworkz:
Not stale

weblate/settings_example.py
```diff
@@ -947,6 +947,14 @@
 # Enable auto updating
 AUTO_UPDATE = False
 
+# Multi-level rate limiting for email notifications
+# Each tuple contains (max_emails, time_window_seconds)
+RATELIMIT_NOTIFICATION_LIMITS = [
+    (3, 120),  # Prevent burst sends - 3 emails per 2 minutes
+    (10, 3600),  # Equalize to avoid getting blocked for too long - 10 emails per hour
+    (50, 86400),  # Daily limit: 50 emails per day
+]
```
nijel, line 956: This should probably go to `weblate/accounts/models.py` as a default value.

weblate/utils/ratelimit.py
```diff
@@ -159,3 +109,119 @@ def _rate_wrap(
...
+    def __init__(
+        self,
+        scope: str,
+        request: AuthenticatedHttpRequest | None = None,
+        address: str | None = None,
+        user: User | None = None,
+    ):
+        if request is not None and request.user.is_authenticated and user is None:
+            user = request.user
+        if user is not None:
```
nijel, line 180: Add type annotation:

```suggestion
        key: int | str
        if user is not None:
```

weblate/settings_example.py
softworkz: This is a point that used to confuse me: I've seen many settings which only exist in the settings sample and docker files, without having any base definition somewhere.
I'm not sure about your intended way for dealing with this. From what I've figured from looking at similar cases, the same block would need to be copied into the docker settings file...?

softworkz:
> * The multi-level rate limiting can indeed deal better with smarter hammering, but in reality, most of the hammering is dumb :-). But figuring out reasonable hourly limit for operations like translating might be tricky. Some people can be really fast in this.

Let's not forget that the motivating use case for the change were e-mail notifications. There can be events which are causing many notifications to be sent within a small period of time. When you set a daily limit for example of 100 e-mails within 24h, and those get sent to you within a shorter amount of time - be it 10min or 2 hours - the outcome will always be that you would be cut off from all notifications for the remaining 22h - and that's not what anybody would want. 

> But figuring out reasonable hourly limit for operations like translating might be tricky. Some people can be really fast in this.

That's exactly where the benefit lies in having multiple levels: The non-linearity. Looking at the example from the code
(the numbers are very subjective and arguable)

    (3, 120),  # Prevent burst sends - 3 emails per 2 minutes
    (10, 3600),  # Equalize to avoid getting blocked for too long - 10 emails per hour
    (50, 86400),  # Daily limit: 50 emails per day

It allows 3 within 2 minutes. That alone would mean it could be 15 within 10 minutes - but it allows only 10.
And 10 per 10 minutes would mean 1440 per 24h - but it allows just 50.

> Some people can be really fast in this.

Yes, it also fits that case, because people may be fast - but not continuously fast at the same pace over 24h. For that, you can allow  a higher rate over smaller time spans but block requests when the hit rate becomes non-human - i.e. continuous rate without pause.

softworkz:
> * The change makes the tests fail because it hits the rate limiter. Can you please look into that?

Would you be able to take care of that? I'm deep down into different work and I broke/lost my WSL partition. It would take a while until I'll get back to it.

weblate/settings_example.py
nijel: The example file should only contain settings the user is likely to need to change, and this is not the case. The models file contains default values if the setting does not exist in `settings.py`.

weblate/utils/ratelimit.py
```diff
@@ -159,3 +109,120 @@ def _rate_wrap(
...
+
+        self.cache_items = [
+            CacheCounterItem(self.key, attempts, window)
+            for attempts, window in rate_limits
+        ]
+
+    def is_limit_exceeded(self) -> tuple[bool, str]:
+        # Check all without decrementing
+        for cache_item in self.cache_items:
+            if cache_item.count_remaining <= 0:
```
nijel, line 141: This is where the test with RATELIMIT_MESSAGE_ATTEMPTS=1, RATELIMIT_WINDOW=0 fails.

nijel (commented):
What I wanted to say is that multiple limits work great for e-mails; it might not be that easy to figure out such limits elsewhere. Still, this is a great improvement.

As for the tests, there are two issues. I've commented on one inline, and the second one is that the new limit applies to the notification tests, so a settings override will be needed there.

softworkz:
> What I wanted to say is that multiple limits work great for e-mails; it might not be that easy to figure out such limits elsewhere.

Ah, got it now. Yes, that's really not easy. With regards to dealing with AI bots that are hammering the server (often from a multitude of IP addresses), there's an another approach that might be worth to consider or try out:

### Testing Client Patience

For humans, it's not unusual when a browser request takes a little longer sometimes. 
But for bots which are scripted in a way to grab as much data as possible, waiting for a response is wasting valuable time. They must have a limit on outgoing TCP connections (it's not infinite), and waiting for a response is unproductive - it hurts their throughput, which means that the cannot afford waiting too long for a response. 

This allows a pattern like follows:

- When the frequency of requests is suspiciously high, start delaying responses 
- Then, record how long it takes until the client closes the connection
- Repeat for the next two requests
- When the client has cllosed the connection 3 times after the same duration, put it on the blacklist

nijel:
I've rebased this on current main and fixed some of the issues in https://github.com/WeblateOrg/weblate/pull/16393. Let's continue there.

softworkz:
Nice - thank you°

Issue #15674, opened 2025-08-04
Add REDIS_USER env variable


scare21410:
### Describe the problem

I cannot configure redis properly if my redis connection requires a user name.

settings_docker.py calculates `CACHES.default.LOCATION` without possibility for adding user into the url.
Similarly, `CELERY_BROKER_URL` is calculated similarly.

### Solution brainstorm

`REDIS_USER` env variable could be used to calculate both redis cache `LOCATION` and `CELERY_BROKER_URL` correctly.

Himanshu-Yadav-0:
Hi, i would love to work on this, can i get assigned?

nijel:
@Himanshu-Yadav-0 Meanwhile there is https://github.com/WeblateOrg/weblate/pull/15687 implementing this.

Himanshu-Yadav-0:
ohh okay, i'll check some other issue then.

PR #15687, opened 2025-08-04
Add REDIS_USER env variable

scare21410:
Adds support for REDIS_USER env variable as per #15674 

No automatic tests exist for `settings_docker.py`, so I tested this manually to ensure that keeping REDIS_USER blank will not change current behaviour.

### DIFF
--- a/docs/admin/install/docker.rst
+++ b/docs/admin/install/docker.rst
@@ -1633,6 +1633,10 @@ instance when running Weblate in Docker.
 
     The Redis database number, defaults to ``1``.
 
+.. envvar:: REDIS_USER
+
+    The Redis database user, not used by default.
+
 .. envvar:: REDIS_PASSWORD
 
     The Redis server password, not used by default.
--- a/docs/admin/install/docker.rst
+++ b/docs/admin/install/docker.rst
@@ -1635,6 +1635,8 @@ instance when running Weblate in Docker.
 
 .. envvar:: REDIS_USER
 
+   .. versionadded:: 5.13
+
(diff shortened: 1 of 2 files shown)


docs/admin/install/docker.rst
```diff
@@ -1633,6 +1633,10 @@ instance when running Weblate in Docker.
 
     The Redis database number, defaults to ``1``.
 
+.. envvar:: REDIS_USER
+
```
nijel, line 1637:
```suggestion

   .. versionadded:: 5.13

```

weblate/settings_docker.py
```diff
@@ -1239,14 +1239,23 @@
...
 # Extract redis password
 REDIS_PASSWORD = get_env_str("REDIS_PASSWORD")
+REDIS_USER = get_env_str("REDIS_USER")
+REDIS_USER_PASSWORD = (
+    f"{REDIS_USER}:{REDIS_PASSWORD}"
+    if REDIS_USER and REDIS_PASSWORD
+    else f":{REDIS_PASSWORD}"
+    if REDIS_PASSWORD
+    else REDIS_USER  # can be None
+)
```
nijel, line 1249: Looking at the code, it should be adjusted to generate the Redis URL just once and use it in both caches and Celery configuration.

The different URLs are there because django-redis did not support standard Redis URLs, but this is no longer true since 3.8.0.

But this is just an optional cleanup I've noticed.

nijel:
Scheduled for merge. Thanks for your contribution!

Issue #15691, opened 2025-08-04
Repeated "Could not find string in the translation file"


nijel:
### Describe the issue

Some of the projects see a flood of "Could not find string in the translation file" errors.

### I already tried

- [x] I've read and searched [the documentation](https://docs.weblate.org/).
- [x] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

I'm not sure about the actual reproducer, but it appears that the commit is tried too often when the string could not be found.

### Expected behavior

I think it should try writing the string only once and not retry until the file is changed.

### Screenshots

<img width="927" height="2312" alt="Image" src="https://github.com/user-attachments/assets/1d21e923-e6bf-4f89-8341-374d8f71c0f2" />

### Exception traceback

```pytb

```

### Weblate deploy checks

```shell

```

### Additional context

https://hosted.weblate.org/changes/browse/klipper/index/ro/

nijel:
https://github.com/WeblateOrg/weblate/pull/15693 should address the root cause of this being triggered. This happens too often, but still we should not retry failed commits too often to avoid such a spamming of the history.

Possible causes of this error:

* The underlying file was changed without Weblate being updated. This should not happen, but it seems to be the case here due to another issue.
* The bug in file format that made it parse the string, but could not find it while saving. We had such issues in the past.

I think we need to have some retry policy, this will probably need additional fields (or a JSON field to store any possible future metadata) on the `PendingUnitChange` model.

The change writing should be retried if any of these conditions are met:

* The content of the file changes (`Translation.revision` field)
* The Weblate version changes (`weblate.utils.version.GIT_VERSION` to correctly track git cloned installs).
* Some time passed from the last failed write. I'm not sure here, but something one week sounds reasonable and not too annoying.

nijel:
Another two examples of this:

* https://hosted.weblate.org/translate/portmod/portmod/ta/?checksum=1d0fef99e2b88ded

  This is string with a syntax error which will never be possible to save.

* https://hosted.weblate.org/translate/odoo/odoo/zh_Hant/?checksum=fe6206e2dbfbb77a
 
  Something broken in the repository and the string is no longer present in the file.

This is really annoying for several reasons:

* additional load on the server because the commit is triggered every hour via the task
* the history is full of this error
* Sentry is full of errors from this

It is too late to fix this for 5.13, but we need a solution ASAP.

amCap1712:
[...]

PR #15693, opened 2025-08-04
fix(component): avoid push if needed for not commited changes

nijel:
When a change cannot be committed, do not trigger optional push.

Otherwise we end up in an infinite recursion Component.do_update -> Component.commit_pending -> Component.push_if_needed -> Component.do_push -> Component.do_update

Issue #15691

I think this answers why an attempt to commit in #15691 happens so often. Still, we need a solution to make the failed commit less noisy.

### DIFF
--- a/weblate/trans/models/component.py
+++ b/weblate/trans/models/component.py
@@ -2168,6 +2168,7 @@ class Component(
         skipped = set()
         source_updated_components = []
         translation_pks = defaultdict(list)
+        was_changed: bool = False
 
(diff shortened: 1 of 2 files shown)


nijel:
@amCap1712 I'm merging this as this is something we need not fix fast, but still your review is welcome!

amCap1712:
Didn't approve earlier because wanted to reproduce locally and confirm the fix but otherwise LGTM. Thanks!

PR #15701, opened 2025-08-04
Feature: support for plurals suggestions using OpenAI

ditwoo:
## Problem

The current version of Weblate does not support proper OpenAI suggestions for plurals.
This PR tries to fix this issue and fetch translations from OpenAI for plurals.

The next step - adapting this fix to the suggestions of an "Automatic translation".

### DIFF
--- a/weblate/machinery/base.py
+++ b/weblate/machinery/base.py
@@ -509,14 +509,27 @@ class BatchMachineTranslation:
             alternate_units = plural_mapper.get_other_units([unit], source_language)
 
         plural_mapper.map_units([unit], alternate_units)
+        # collect translations
+        if unit.is_plural:
+            # NOTE: using source plurals in order to translate something,
+            #       because in most cases the unit.plural_map yields an empty string
+            sources = [(unit.get_source_plurals()[0], unit)]
+            translations_mapper = plural_mapper
+        else:
+            sources = [(text, unit) for text in unit.plural_map]
+            translations_mapper = None  # note: don't need to map singulars
+
         translations = self._translate(
             mapped_source_language,
             target_language,
-            [(text, unit) for text in unit.plural_map],
+            sources,
             user,
             threshold=threshold,
+            plural_mapper=translations_mapper,
         )
-        return [translations[text] for text in unit.plural_map]
+
+        result = [translations[text] for text, _ in sources]
+        return result
 
     def download_multiple_translations(
         self,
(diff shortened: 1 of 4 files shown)


nijel:
@Changaco might be interested in reviewing this as he is the original author of PluralMapper and machinery changes for that.

Changaco:
The proposed changes to the `PluralMapper` class are clearly wrong.

ditwoo:
Hi @Changaco,

Could you help me understand this a bit better?

I made the changes so that the list of plurals could be returned for translation. I’m aware that this workaround feels somewhat hacky. The issue arises because, if we keep the `result.append((None, None))`, every element in the list ends up as `(None, None)` except the last one, which is `(-1, None)`. Since the mapper logic expects an integer value, encountering a None causes an exception.

Because I’d like to fix both manual and automatic plural translations, could you suggest the best way to update `PluralMapper`?
Or let me rephrase, what was the original idea with translation of plurals, and how to use `PluralMapper.target_map` properly?

Thank you!

Changaco:
@ditwoo The `PluralMapper` class was created to get as many plural forms as possible from translation services that don't support generating all the plural forms directly from the source strings. Since you're trying to use an LLM to generate all the plural forms in a single request, you probably don't need the plural mapper at all.

ditwoo:
@Changaco Thank you for your helpful message.

You’re right - what I need is simply a unit object. The existing `Unit` already contains the singular/plural information, and when a plural form is requested, it can provide both the number of possible plurals and their corresponding numbers (examples).

Thank you again for your guidance; I'll update my PR accordingly.

weblate/machinery/base.py
```diff
@@ -657,21 +677,37 @@ def batch_translate(
             alternate_units = plural_mapper.get_other_units(units, source_language)
         plural_mapper.map_units(units, alternate_units)
 
-        # TODO: fetch source from other units
-        sources = [(text, unit) for unit in units for text in unit.plural_map]
-        translations = self._translate(source, language, sources, user, threshold)
-
+        # NOTE: Instead of translating all at once (the old approach),
+        #       a step-by-step translation is more efficient/robust.
```
nijel, line 681: This removes purpose of the `BatchMachineTranslation`. If it is better (for whatever reason) to not batch translation for a single engine, subclassing `MachineTranslation` is the way to go.

weblate/machinery/base.py
```diff
@@ -509,14 +522,21 @@ def translate(
             alternate_units = plural_mapper.get_other_units([unit], source_language)
 
         plural_mapper.map_units([unit], alternate_units)
+        # collect translations
+        if unit.is_plural:
+            sources = [(unit.get_source_plurals()[0], unit)]
```
nijel, line 527: This definitely breaks with the alternate sources feature.

PR #15708, opened 2025-08-05
Add setting to hide shared glossary components in projects they are s…

softworkz:
Glossary components are typically shared into other projects to make them available for translation translation work.
When these are visible in the compnoent list of projects which are using them, it can cause confusion or distract translators from
the actual components that are meant to be translated.

This PR adds a setting which hides those shared glossary components.


softworkz:
Not stale

nijel:
The new option should be documented in https://docs.weblate.org/en/latest/admin/projects.html#share-in-projects so that it is clear that glossary components might not be visible with it turned on.

PS: Please address the Copilot reviews, these are definitely valid points.

softworkz:
I've commited the AI suggestions.

Also enabled edits: 

image

nijel:
Scheduled for merge after additional fixes, thanks for your contribution!

PR #15713, opened 2025-08-05
chore: improve type annotations

nijel:
- extend mypy filter
- add correctly type some optional paramters
- properly pass outgoing list to notification test

Issue #3703


nijel:
@amCap1712, @gersona This will start enforcing type checking on parts of the code base. It now covers only some smaller modules, but I want to extend the scope over the time.

PR #15726, opened 2025-08-06
chore: use webpack to manage bootstrap5 dependency

KarenKonou:
As discussed, uses yarn/webpack to manage the bootstrap 5 dependency.
Also moves existing vendored css files to the styles directory.

There is an issue with an empty `bootstrap5_rtl.js` file being generated by webpack, this appears to be a longstanding webpack bug: https://github.com/webpack/webpack/issues/11671


nijel:
Looks good, scheduled for merge.

Issue #4017, opened 2020-06-11
Redesign configuration


nijel:
Weblate uses Django settings since the beginning. It might be time to revisit this decision as in many cases this causes troubles.

- The settings is Python code, what makes it powerful, but most people do not need this
- The settings live in the Python site packages, what makes it look like code, while it actually is not
- Most of the settings are not supposed be changed by users and forgetting doing that on upgrade can lead to subtle breakage
- We have additional layer on top of that to load settings from the environment in Docker container, maybe we could use something shared for both environments

nijel:
We might want to try something like https://pydantic-docs.helpmanual.io/ for loading from the environment.

PR #15729, opened 2025-08-06
Implement Dynamic OpenAPI Generation with. OAS 3.0 Support

softworkz:
Last week, I wanted to generate a .NET ApiClient library for weblate in order to automate a few things.
We're auto-buiilding REST client for a range of languages/platforms, so that they're always up-to-date, so I didn't expect any problems in doing so.
Until I realized that Weblate OpenAPI description are using spec 3.1. That's a problem, because almost nobody is ready for that yet, even some are claiming to be, but when you look closer, you'll realize that an OAD according to OAS 3.1 is largely useless at the time of writing. Most of the tools and tooling hasn't been updated yet, and those who really support 3.1 are entirely new developments - but for 3.1, and do not properly work for 3.0.
That little minor bump is not as little as it might appear to be. 

3.1 is an update with a small set of extremely incompatible changes.

It took me several hours and multiple tools to convert the 3.1 from Weblate to a 3.0 version which - most importantly  - is actually accepted by 3.0 tools.
Also, the generated document is highly bloated by tons of error response models, spec extensions and inefficiently build component schema. This week, I wanted to try leeting some AI interact with the Weblate API, but those 2MB are way outside of what they are able to ingest.
The latter is the reason for the comprehensive filtering capability: being able to get openapi descriptions creaed in a compact and efficient way.

3.0 doesn't support Webhook descriptions, so I kept 3.1 as the defauilt. The schema generation endpoint (`/api/schema/?params...`) is also included in its own doc itself.
While at it, I also added a Swagger view because ReDoc doesn't have a "Try out" featue. 
Finally, the base path was wrong - which had caused all tool and consumers to failing out-of-the-box behavior.
Other changes:

- Removed the single quote from API doc title 
- Fix a typo in a drf config function name
- The OAS version is now included in the title,
  which also determines the file name when downloading
- Also the title will include 'Light' when light param is set, or "Subset" in all other cases when some filtering params are set

This is hwo it looks with those changes applied:

image


softworkz:
I've taken the default 3.1 version, the 3.0 version and a 3.0 Light version and ran it through several OpenApui validation tools (swagger-cli,  Redocly CLI, Prance and openapi‑spec‑validator or similar)


npx ajv-cli validate   --spec=draft2020   -s openapi-3.1.0-schema.json   -d '/mnt/c/Users/admin/Downloads/Weblate_3.1_invalid.yaml'   --strict=false


All passed - Redocly CLI showed a number of warnings only.

nijel:
Spectacular should be able to create a 3.0 schema directly without needing to maintain the conversion code. Unfortunately, it seems to be settings only (`OAS_VERSION`). Have you considered changing the configuration only instead of writing the conversion code?

softworkz:
There is no conversion code. I'´m setting `OAS_VERSION` to 3.0 and spectacular creates a 3.0 description.

The code that is there has different purposes:

####  _filter_patterns()

That's the provided way by SpectacularAPIView for filtering out certain paths. I'm pretending to do this "by Tags", while in reality, paths are filtered by the first path segment. But as this is always the same as the tag, it's essentially that: "filter by tag"

#### _prune_unused_tags

While the above is filtering out the paths, it doesn't filter out the actual Tag elements.  `_prune_unused_tags()` does this and so it avoids that there are lots of empty sections  in the left column when certain paths are filtered out.

#### _prune_unused_components()

The next problem is that when we filter by using the patterns parameter, sperctacular omits the compoinent schemas from the directly related definitions, but it still leaves most of it in the document.
`_prune_unused_components()` recursively parses the components and follows all references in order to detect and remove everything that is not referenced.
Validations have shown that it is working pretty well like this (even though I would had preferred not needing to do this manually).

#### _collect_used_security_schemes()

The only purpose of this function is to identify the security_scheme items, because they would otherwise be cleaned up by  `_prune_unused_components()`.

#### _strip_examples()

Does what it says for data examples.

Note: The code for stripping out the errors (when chosen) is still inline, it should probably be moved out for consistency.

Both of the latter play together with _prune_unused_components(). They are removing the directly related elements and `_prune_unused_components()` does the deep/recursive cleanup.

softworkz:
This means in turn: 

For a URL like `.../api/schema(?oav=3.0`, none of this code would need to run. At the moment it _does_ run, but the only actual effect comes from `_prune_unused_tags()`, which means that those tags get remvoed which do not have any content either way (i.e. Glossary and Tasks).

nijel:
Thanks for the explanation. I really didn't look at the patch as it's really hard to review because of moving code to another file and then adding new code instead of it - the diff is really messy. Having the fixes separate from the new feature would make it much easier to review. I will get to reviewing the patch, but probably not before the 5.13 release. If you can split it up, the fixes will be easier to review and can be processed sooner.

softworkz:
> If you can split it up, the fixes will be easier to review and can be processed sooner.

Have you seen that there are 9 separate commits, each one doing a very specific change only?

softworkz:
The second commit (https://github.com/WeblateOrg/weblate/commit/a2c3e570aa861024883797720498a3a5a0f039ce) moves the configuration without making an actual change to the contents, so that content changes can be reviewed properly.

(you just need to review commit-by-commit)

softworkz:
If you prefer the drf settings code to remain in the spectacular.py file, I can reshape the PR.
I could also move out the helper code into a separate py file (like openapi_utils.py or similar. 

The original reason for moving it was that I had encountered once another circular referencing problem, but I think it resolved naturally, later. I still found it reasonable to keep it separate, though, but there are many valid way for structuring things and I'm good with whichever way you like.

softworkz:
NOT STALE

nijel:
I'm still unsure about this:

* I'm sure there exist third-party tools that can achieve conversion to older versions of the OpenAPI specification or filtering results. I would rather not maintain that.
* The module name `spectacular` is indeed confusing and no longer bound to just drf-spectacular. This is long outstanding issue of mixing user-configurable settings with the ones user is not really supposed to configure, see https://github.com/WeblateOrg/weblate/issues/4017.

> Also, the generated document is highly bloated by tons of error response models, spec extensions and inefficiently build component schema. 

Indeed, this is something that should be improved, but I see no improvement in this here. Rather a workaround to hide some content by some parameters.

PR #15752, opened 2025-08-07
bugfix: reject backups with duplicate files

### LINKED ISSUE #15558: Crash upon restoring old backup
### Describe the issue

Weblate before 5.11.1 didn't properly backup categories (implemented in ed849811c826631a1943e96e38f2839713b4e770). This can cause backups that cannot be restored as there might be components with duplicate `name` or `slug`.

### I already tried

1. Create a project with two categories each having a component with the same name.
2. Back this up on Weblate 5.10.
3. Try to restore this on Weblate 5.12

### Expected behavior

The backup would restore. But it is not that easy because it lacks categories, so it cannot be fully restored. I can see two possible approaches:

* Detect this situation and reject restoring such a backup with a clear message.
* Detect this situation and rename problematic components. Both `name` and `slug` need to be handled.

If renaming would not be much additional work, I'd use this approach so that the backup can be restored even in this case.

### Screenshots

_No response_

### Exception traceback

```pytb

```

### How do you run Weblate?

weblate.org service

### Weblate versions

_No response_

### Weblate deploy checks

```shell

```

### Additional context

Sentry Issue: [WEBLATE-30Q5](https://weblate.sentry.io/issues/50966842/?referrer=github_integration)

```
UniqueViolation: duplicate key value violates unique constraint "component_slug_unique"
DETAIL:  Key (project_id, category_id, slug)=(9, null, authorizationdb) already exists.
  File "django/db/backends/utils.py", line 105, in _execute
    return self.cursor.execute(sql, params)
  File "psycopg/cursor.py", line 97, in execute
    raise ex.with_traceback(None)

IntegrityError: duplicate key value violates unique constraint "component_slug_unique"
DETAIL:  Key (project_id, category_id, slug)=(9, null, authorizationdb) already exists.
(25 additional frame(s) were not displayed)
...
  File "contextlib.py", line 85, in inner
    return func(*args, **kwds)
```

amCap1712:
Weblate before 5.11.1 didn't properly backup categories, this can lead to zip files with duplicate components and files. Detect this early and return an appropriate error.

fixes #15558


nijel:

Issue #9077, opened 2023-04-13
Move file format features from table to format pages


nijel:
### Describe the problem

Right now, there is a huge table of file format capabilities at https://docs.weblate.org/en/latest/formats.html#translation-types-capabilities

This is hard to read, navigate, and find.

### Describe the solution you'd like

The documentation should be moved to file format specific pages. Each page should have a heading section (table or [field list](https://www.sphinx-doc.org/en/master/usage/restructuredtext/field-lists.html)) including data from the table.

Additional info should be included:

* Support for read-only strings (currently at https://docs.weblate.org/en/latest/formats.html#read-only-strings)
* List of file extensions/names common for this format
* API identifier - ID which is used when identifying this file format in API, can be found as `format_id` in the source, for example: https://github.com/WeblateOrg/weblate/blob/5a62dca81de57aa19b32d856b632f3b7905fa37d/weblate/formats/ttkit.py#L1175

### Additional context

As a preparation for this, file format documentation was split to individual pages in 1d322635cb55ca5a455856a2dff26e9b83679aa7.

niccokunzmann:
I would like to add that my process was the following:

I asked myself: Which file formats support Feature XY - in my case Descriptions. Then I had a look which are the ones I can use most easily.

It is good for me to have an overview of the features and know which file formats allow to use them.

This is what it can look like:

----

## Source String Descriptions

This allows you to write notes to translators in files or add flags and they will come up on the weblate website.

Supported formats:
- [PO](...) - with flags, description, ...
- ...

----

nijel:
Yes, adding more detailed description of the capabilities is the way to go, I've started that in 624891a5c6. 

Adding guidance for a format to each feature is definitely a good idea, I've tried to add that in e45206b0bb (for plurals only).

Geeyun-JY3:
Bug in https://docs.weblate.org/en/latest/formats.html#supported-file-formats after recent updates

![Screenshot_20230414104737](https://user-images.githubusercontent.com/49396421/231928945-e3c4e587-9edc-4841-afd8-e353391440e4.jpg)

nijel:
When adding a new feature to file formats, I've introduced "Supported features" table in TBX documentation in https://github.com/WeblateOrg/weblate/pull/9252, this might be the way to go here.

nijel:
I think automatically generating the basics of this documentation would be the way to go. It might be useful to add annotations for all the features in the code and automatically update the documentation. It could use approaches described in #14279.

PR #15761, opened 2025-08-08
Update formats.rst: TBX supports Context

softworkz:
.if I'm not totally mistaken...?


nijel:

PS: This table is waiting for a full rewrite, see https://github.com/WeblateOrg/weblate/issues/9077.

PR #15776, opened 2025-08-08
chore: turn zipfile warnings into errors

### LINKED ISSUE #15665: Validate generated backup zip file
### Describe the issue

From https://github.com/WeblateOrg/weblate/issues/15558#issuecomment-3145622983:

> I generated a zip file to reproduce the issue again and listed its contents.
>
> Image

### I already tried

1. Create backup of a project with categories.
2. See the zip file has duplicate entries.

### Expected behavior

The backup should not generate invalid zip file. This particular case has been addressed, but the zip file generating should crash in such situation, because that backup is not valid.

### Screenshots

_No response_

### Exception traceback

```pytb

```

### How do you run Weblate?

weblate.org service

### Weblate versions

_No response_

### Weblate deploy checks

```shell

```

### Additional context

_No response_

amCap1712:
Turn warnings from zipfile module into errors. Fixes #15665.

1. Tested manually by applying the same fix to Weblate 5.10 and the warning is turned into an error. I couldn't add a test for this to the tests because pytest's warning configuration seems to take precedence during tests.
2. I do see the errors in the logs but I don't think errors in the backup process are surfaced to the user (at least in Weblate 5.10).
3. I am not sure if the best place for the warnings filter is the backup file itself. Off the top of my head, I can't think of other places that create zip files. The best place would be `settings.py` I guess as it would then apply always but that requires existing installations to manually update their settings files which is not ideal.

### DIFF
--- a/weblate/trans/backups.py
+++ b/weblate/trans/backups.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 
 import json
 import os
+import warnings
 from collections import defaultdict
 from datetime import datetime
 from functools import partial
@@ -56,6 +57,8 @@ if TYPE_CHECKING:
     from django.db.models import Model
 
     from weblate.billing.models import Billing
+
+warnings.filterwarnings("error", module="zipfile")
 
 PROJECTBACKUP_PREFIX = "projectbackups"
 


nijel:
Zip files are also created when downloading, but these should trigger no warnings as well.

nijel:

> I couldn't add a test for this to the tests because pytest's warning configuration seems to take precedence during tests.

Maybe the pytest configuration could be adjusted to treat zipfile warnings as errors as well?

amCap1712:
Yes that can be done but it wouldn't actually test the codebasee so not sure if its worth it.

nijel:
I think it makes sense to have the same warnings behavior in the tests as in production

PR #15810, opened 2025-08-12
Add option to hide glossary matches

softworkz:
This PR adds an option to hide the glossary panel in the translation editor:

image

The option is enabled by default for glossary components.
(might be debatable)

The rationale behind this is this:

- A glossary component can be used for various purposes, like
  - Adding explanations for translators
  - Providing/enforcing consistent translations of specific terms
  - Indicating certain terms as non-translatable
- In all of those (and probably most other) cases, the glossary component is meant to be NORMATIVE and binding for translators to follow
- In this regard, a glossary is logically the SOURCE of information, and shouldn't depend on or take input from other glossary components

=> That's why I think that a glossary should not show matches form itself or other glossary components

BTW, I also think that "Contribute to project translation memory" and "Allow translation propagation" should be off by default for glossary components.

This PR also includes a related fix:

Also prevents showing self-referencing matches in glossary panel (regardless of the new option)

Currently, when editing a glossary entry, you see that exact entry in the glossary panel..

### DIFF
--- a/docs/admin/projects.rst
+++ b/docs/admin/projects.rst
@@ -629,6 +629,24 @@ notification about any source string comments made in Weblate.
 With the :ref:`gettext` format, this address is also saved by Weblate in the
 :mailheader:`Report-Msgid-Bugs-To` header of the file.
 
+.. _component-hide_glossary_matches:
+
+Do not show glossary matches
+++++++++++++++++++++++++++++
+
+Hides the glossary panel and its matches in the translation editor for this component.
+When enabled, glossary suggestions for this component are not computed and the Glossary
+panel (including its “Add term to glossary” action) is hidden in the editor.
+
+Notes:
+- Glossary self-references are always excluded from matches even when this option is disabled.
+- Newly created glossary components enable this by default.
+
+.. seealso::
+
+   * :ref:`glossary`
+   * :ref:`component-is_glossary`
+
 .. _component-allow_translation_propagation:
 
 Allow translation propagation
--- a/weblate/trans/models/component.py
+++ b/weblate/trans/models/component.py
@@ -540,6 +540,11 @@ class Component(
             "Locked component will not get any translation updates."
         ),
     )
+    hide_glossary_matches = models.BooleanField(
+        verbose_name=gettext_lazy("Do not show glossary matches"),
+        default=False,
+        help_text=gettext_lazy("Hides the glossary panel in the translation editor."),
+    )
     allow_translation_propagation = models.BooleanField(
         verbose_name=gettext_lazy("Allow translation propagation"),
         default=settings.DEFAULT_TRANSLATION_PROPAGATION,
@@ -1127,6 +1132,7 @@ class Component(
             is_glossary=True,
             has_template=False,
             allow_translation_propagation=False,
+            hide_glossary_matches=True,
             license=self.license,
         )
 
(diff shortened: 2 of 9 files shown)


nijel (commented):
This definitely makes sense for some glossary use cases. 

What is debatable is whether this behavior should be the default for glossaries and whether existing glossary components should be migrated.

There are many ways to use multiple glossaries and in many of them you want to see matches from other glossaries while editing the term.

The first public project with many glossaries I remembered is Eso Spolszczenie and I belive seeing other glossaries on the glossary edit is useful here (like https://weblate.eso-spolszczenie.eu/translate/eso-spolszczenie/-skille-/pl/?checksum=855835cd5fd8df34). @tomkolp can probably comment on this.

nijel:
I've extracted the self-referencing entries in https://github.com/WeblateOrg/weblate/pull/16084 to have that change separate.

tomkolp:
From my perspective, it makes no real difference whether showing glossary matches within glossaries is enabled or disabled by default, as long as there is a global option to change this behaviour, along with the ability to configure it individually for each glossary. Of course, glossaries must remain functional within regular components.

softworkz:
> From my perspective, it makes no real difference whether showing glossary matches within glossaries is enabled or disabled by default, as long as there is a global option to change this behaviour, along with the ability to configure it individually for each glossary. Of course, glossaries must remain functional within regular components.

I had come to the same conclusion. Users may want it this or that way, and the next decision criteria for me was to keep the default aligned with the current behavior:  Means when someone is creating a new glossary after the change, they will get a glossary with the same behavior as before.

That being said, I'm fine with the other way as well.

docs/admin/projects.rst
```diff
@@ -629,6 +629,24 @@ notification about any source string comments made in Weblate.
 With the :ref:`gettext` format, this address is also saved by Weblate in the
 :mailheader:`Report-Msgid-Bugs-To` header of the file.
 
+.. _component-hide_glossary_matches:
+
+Do not show glossary matches
+++++++++++++++++++++++++++++
+
+Hides the glossary panel and its matches in the translation editor for this component.
+When enabled, glossary suggestions for this component are not computed and the Glossary
```
nijel:
```suggestion
When enabled, glossary suggestions for this component are not computed, and the Glossary
```

docs/admin/projects.rst
```diff
@@ -629,6 +629,24 @@ notification about any source string comments made in Weblate.
...
+panel (including its “Add term to glossary” action) is hidden in the editor.
+
+Notes:
+- Glossary self-references are always excluded from matches even when this option is disabled.
+- Newly created glossary components enable this by default.
```
nijel:
```suggestion
.. note::

   Glossary self-references are always excluded from matches even when this option is disabled.
```

weblate/trans/models/component.py
```diff
@@ -1127,6 +1132,7 @@ def create_glossary(self) -> None:
             is_glossary=True,
             has_template=False,
             allow_translation_propagation=False,
+            hide_glossary_matches=True,
```
```suggestion
```

softworkz:
Thanks for looking into it.

The pushes reminded me that I have disabled edits to the PR branches in some other PRs. When you need to make changes to any of those, just let me know and I'll flip that setting. I just found those bot changes annoying...

nijel:
Let's keep the current behavior; I've adjusted the pull request accordingly. The changelog entry is missing, and besides that, I don't see any issue.

nijel:

PR #15836, opened 2025-08-13
feat (formats): Configure FlatXML parsing with File format parameters

### LINKED ISSUE #3149: Flat XML - configure default tag names

gersona:
- fixes #3149


nijel:
This looks safe, so let's include this in 5.13.

docs/formats/flatxml.rst
```diff
@@ -10,11 +10,14 @@ Flat XML files
 
    :doc:`tt:formats/flatxml`
 
-Example of a flat XML file:
+Example of a flat XML files:
```
nijel:
```suggestion

.. versionchanged:: 5.13

   The tag and attribute names can now be customized using :ref:`file_format_params`.

Example of a flat XML files:
```

PR #15862, opened 2025-08-15
feat(ui): migrate core components to bootstrap 5

KarenKonou:
Migrates core templates (accessible from the main navbar) to bootstrap 5

TODOs:

- [x]  Search page
- [x]  Progress boxes on translation pages
- [x]  Unwanted border at the top of some components in dark mode
- [x]  Badges on history pages
- [x] Search page dark mode
- [x] User profile page browse all contributions button
- [x] Fix tests
- [x] Appearance customization from the manage menu w/ BS5 (will be in a separate PR)

### DIFF
--- a/weblate/templates/dashboard/user.html
+++ b/weblate/templates/dashboard/user.html
@@ -23,15 +23,24 @@
     </li>
     {% if all_componentlists %}
       <li class="nav-item">
-        <a data-bs-target="#componentlists" data-bs-toggle="tab" href="#" {% active_link "componentlists" %}>{% translate "All component lists" %} <span class="badge text-bg-secondary">{{ all_componentlists|length }}</span></a>
+        <a data-bs-target="#componentlists"
+           data-bs-toggle="tab"
+           href="#"
+           {% active_link "componentlists" %}>{% translate "All component lists" %} <span class="badge text-bg-secondary">{{ all_componentlists|length }}</span></a>
       </li>
     {% endif %}
     <li class="nav-item">
-      <a data-bs-target="#suggestions" data-bs-toggle="tab" href="#" {% active_link "suggestions" %}>{% translate "Suggested translations" %} <span class="badge text-bg-secondary">{{ suggestions|length }}</span></a>
+      <a data-bs-target="#suggestions"
+         data-bs-toggle="tab"
+         href="#"
+         {% active_link "suggestions" %}>{% translate "Suggested translations" %} <span class="badge text-bg-secondary">{{ suggestions|length }}</span></a>
     </li>
     {% if owned_projects %}
       <li class="nav-item">
-        <a data-bs-target="#managed" data-bs-toggle="tab" href="#" {% active_link "managed" %}>{% translate "Managed projects" %} <span class="badge text-bg-secondary">{{ all_owned_projects|length }}</span></a>
+        <a data-bs-target="#managed"
+           data-bs-toggle="tab"
+           href="#"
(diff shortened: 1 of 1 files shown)


nijel:
Let's keep the customization for a separate PR, this one is already big enough :-).

I've done some testing, and this is what I found.

The hamburger menu focus looks weird (there was no focus indication here in the old UI):

image

Component lists on the dashboard are misaligned:

image

Active menu item is not themed:

image

Form fields are no longer bold:

image

Repository maintenance has too big labels and spacing:

image

Once the repository maintenance tab is loaded, it is always shown below the current tab:

image

Some menu items do not correctly switch tabs:

image

The copy button is not right-aligned:

image

The sorting dropdown is not the same height as the search input:

image

nijel:
I've still located some minor issues, and I think we can merge this once these are addressed.

### Blue hover background for buttons:

image

Old behavior:

image

### Plural labels when showing string

These are too big (for example, in history):

image

### History is a bit too wide

There is a horizontal scrollbar:

image

### Bugs when loading page with anchor

* The menu opens visibly when an anchor points to it (for example `#history`)
 
  image

* On the dashboard, when loading the component list, it includes another tab:

  image

### List group inside card

There is a round border in the bottom:

image

nijel:
@KarenKonou Maybe the blue color in the Bootstrap 5 theme should be replaced by something that better fits our color scheme, as it's the second time we're addressing it.

KarenKonou:
Good idea, I'll change the color scheme variables to match our visual identity.

KarenKonou:
Still working on the tabs issues. I wasn't able to reproduce the component list tab problem. As for the open dropdown menu, it seems to be default bootstrap behavior when toggling a tab with a trigger in a dropdown with JS, I'll see if i can find a workaround.

weblate/templates/dashboard/user.html
```diff
@@ -15,53 +15,56 @@
...
     {% endif %}
-    <li {% active_link "suggestions" %}>
-      <a href="#suggestions" data-toggle="tab">{% translate "Suggested translations" %} <span class="badge">{{ suggestions|length }}</span></a>
+    <li class="nav-item" {% active_link "suggestions" %}>
+      <a class="nav-link" data-bs-target="#suggestions" data-bs-toggle="tab" href="#">{% translate "Suggested translations" %} <span class="badge text-bg-secondary">{{ suggestions|length }}</span></a>
     </li>
     {% if owned_projects %}
-      <li {% active_link "managed" %}>
-        <a href="#managed" data-toggle="tab">{% translate "Managed projects" %} <span class="badge">{{ all_owned_projects|length }}</span></a>
+      <li class="nav-item" {% active_link "managed" %}>
```
nijel: I think this is the problem for the weird tabs behavior on the dashboard that I can observe. `active_link` can conditionally output `class="active"` making the element have two `class` attributes.

KarenKonou: Should be fixed now.
One more  bug that I've noticed that seems to occur on the main branch as well: if a user sets their default dashboard view to a component list the dynamic content doesn't load properly until the tab is switched.

nijel: That's possible the dynamic loading was added recently here...

KarenKonou: The problem is because the `show.bs.tab` event doesn't fire on initial page load so that's another thing to find a workaround for.

nijel:
I've decided to merge this. Let's address remaining issues separately.

PR #15869, opened 2025-08-16
chore(pending change): make retries smarter

### LINKED ISSUE #15691: Repeated "Could not find string in the translation file"

amCap1712:
Currently, all failing pending changes are retried on every commit, and that happens around every hour or so. This creates lots of noise in the unit's changelog and sentry as failing changes are unlikely to resolve themselves. Hence, add a metadata field to PendingUnitChange to store the translation revision at the time of the failure, the time of the failure and the weblate version. Failing changes are only retried when either the revision or weblate version changes or a week has passed.

Also, failures are of two types - blocking due to missing unit in the translation files and non-blocking due to issues possibly in TranslateToolkit. Whether a change is blocking or not is stored in the metadata field as well. If a failed pending change of blocking type exists, all newer changes to the unit are excluded from selection. A non-blocking failed pending change might itself not be selected for retrying if conditions are not met but it does not prevent newer changes to the unit from selection. If any such changes are successfully applied, the earlier failed changes are deleted from the table.

Fixes #15691.

### DIFF
--- a/weblate/trans/tests/test_models.py
+++ b/weblate/trans/tests/test_models.py
@@ -527,6 +527,24 @@ class TranslationTest(RepoTestCase):
         ttk_unit, _ = translation.store.find_unit(unit.context, "Hello, ${ name }!")
         self.assertEqual(ttk_unit.target, "Ahoj, ${ name }!")
 
+    def test_commit_successful_deletes_failed_changes(self):
+        """Test that failed changes are deleted when a subsequent successful change to the unit is applied."""
+        user = create_test_user()
+        component = self.create_ftl()
+        translation = component.translation_set.get(language_code="cs")
+        unit = translation.unit_set.get(source="Hello, ${ name }!")
+
+        unit.translate(user, "Ahoj, ${ jméno }!", STATE_TRANSLATED)
+        unit.translate(user, "Ahoj, ${ name }!", STATE_TRANSLATED)
+        self.assertEqual(translation.count_pending_units, 2)
+
+        component.commit_pending("test", None)
+
+        # second change applied and first deleted as a result
+        self.assertEqual(translation.count_pending_units, 0)
+        ttk_unit, _ = translation.store.find_unit(unit.context, "Hello, ${ name }!")
+        self.assertEqual(ttk_unit.target, "Ahoj, ${ name }!")
+
 
 class ComponentListTest(RepoTestCase):
     """Test(s) for ComponentList model."""


weblate/trans/tests/test_models.py
```diff
@@ -450,6 +441,111 @@ def test_commit_policy(self):
...
+        translation = component.translation_set.get(language_code="cs")
+        unit = translation.unit_set.get(source="Hello, world!\n")
+        unit.translate(user, "Ahoj Země!\n", STATE_APPROVED)
+
+        self.assertEqual(translation.count_pending_units, 1)
+
+        # mocking TranslateUnit in weblate or ttk seems impossible in the
+        # relevant code, hence the following hacky property mock to ensure
+        # that the exception is raised in the "Store Translations" block
+        count = 0
```
amCap1712: As the comment says, I spent a lot of effort in trying to find a way better way to test this but couldn't so open to ideas.

nijel: You can probably use Fluent to test this without actual mocking, as this is what we see at https://hosted.weblate.org/translate/portmod/portmod/ta/?checksum=1d0fef99e2b88ded.

- Source string contains variable: `Are you sure you want to destroy prefix { $prefix }?`
- When the variable name is translated, it triggers an error upon save: `முன்னொட்டு {$ முன்னொட்டு} ஐ அழிக்க விரும்புகிறீர்களா?`
- This should fix saving: `முன்னொட்டு {$prefix} ஐ அழிக்க விரும்புகிறீர்களா?`

`test.ftl` can look like:

```fluent
destroy-prompt = Are you sure you want to destroy prefix { $prefix }?
```

amCap1712:
Huh, weird. The tests were passing locally. I'll check.

nijel:
Maybe caching issue that causes the stats to be used from previous tests. Maybe invalidation is missing somewhere.

amCap1712:
It turned out to be a peculiar JSON behavior in Django for mariadb/mysql. Use postgres locally so tests were passing locally.

amCap1712:
Hmm, it fixed it partially only.

weblate/trans/models/pending.py
```diff
@@ -31,12 +32,55 @@ def for_component(self, component: Component):
...
-        qs = self.filter(unit__translation=translation)
+        one_week_ago = timezone.now() - timedelta(days=7)
+
+        # filter changes that are new or eligible for retry
+        eligible_for_attempt_filter = (
+            Q(metadata__last_failed__isnull=True)
+            | ~Q(failed_revision=translation.revision)
+            | ~Q(weblate_version=GIT_VERSION)
+            | Q(metadata__last_failed__lt=one_week_ago.isoformat())
+        )
```
nijel, line 46: Maybe these lookups need to be changed to use KT as well?

amCap1712: Just fixed it I think, needed to cast to string manually. The issue is that django was treating `translation.revision` as JSON values and trying to extract a string from it to compare. only happens on mariadb/mysql.

weblate/trans/tests/data/test-base-repo.git.tar
nijel, line 1: @amCap1712 I missed this in the review: This file is generated and the source needs to be updated. Created https://github.com/WeblateOrg/weblate/issues/15980 to track this and https://github.com/WeblateOrg/weblate/pull/15979 documents this.

PR #15872, opened 2025-08-17
Exclude terms from Glossary display where target equals source and state is fuzzy

softworkz:
When a Glossary component is based on a file format where it is expected that values for all keys exist (source language term if no translation) these items should not be shown in the glossary panel to avoid confusion.

### DIFF
--- a/weblate/glossary/models.py
+++ b/weblate/glossary/models.py
@@ -15,12 +15,12 @@ from typing import TYPE_CHECKING, cast
 import ahocorasick_rs
 import sentry_sdk
 from django.core.cache import cache
-from django.db.models import Prefetch, Q, Value
+from django.db.models import Prefetch, Q, Value, F
 from django.db.models.functions import MD5, Lower
 
 from weblate.trans.models.unit import Unit
 from weblate.utils.csv import PROHIBITED_INITIAL_CHARS
-from weblate.utils.state import STATE_TRANSLATED
+from weblate.utils.state import STATE_TRANSLATED, STATE_FUZZY
 
 if TYPE_CHECKING:
     from weblate.trans.models.translation import Translation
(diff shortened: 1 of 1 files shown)


weblate/glossary/models.py
```diff
@@ -154,6 +154,9 @@ def fetch_glossary_terms(  # noqa: C901
             # Variant is used for variant grouping below, source unit for flags
             base_units = base_units.select_related("source_unit", "variant")
 
+            # Exclude fuzzy entries whose target exactly equals source
+            base_units = base_units.exclude(Q(state=STATE_FUZZY) & Q(source=F("target")))
```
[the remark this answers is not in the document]
softworkz: BS!

`target` isn't even a variable

weblate/glossary/models.py
```diff
@@ -154,6 +154,11 @@ def fetch_glossary_terms(  # noqa: C901
             # Variant is used for variant grouping below, source unit for flags
             base_units = base_units.select_related("source_unit", "variant")
 
+            # Exclude fuzzy entries whose target exactly equals source
+            base_units = base_units.exclude(
+                Q(state=STATE_FUZZY) & Q(source=F("target"))
```
nijel, line 159: What is wrong on target matching source? It can happen in many situations and it is still useful to know that the translation is not supposed to be changed.

nijel:
I think it is desired to show these. Hiding some particular strings can be confusing to users.

weblate/glossary/models.py
softworkz: It's not about "target matching source".

It's about "target matching source" PLUS "target is not translated yet" (fuzzy)

softworkz:
> I think it is desired to show these. Hiding some particular strings can be confusing to users.

It's my fault. I should have explained the case in a better way. Even I had to think for a moment which case it was about.

The case is this:

- We have a glossary that is based on the RESX file format
- In case of RESX, all languages must have values for all strings
- So we have configured the plugin accordingly to fill all untranslated strings with the source string value  
- The plugin also marks these target strings as "FUZZY" (not translated or translation not confirmed)

A Glossary is expected to be NORMATIVE for translations. That's why untranslated/FUZZY strings must not be shown as Glossary entries.

PR #15874, opened 2025-08-17
contact.html: Add class name to alert element

softworkz:
This allows hiding or changing appearance via custom CSS

### DIFF
--- a/weblate/templates/accounts/contact.html
+++ b/weblate/templates/accounts/contact.html
@@ -12,7 +12,7 @@
 
   <div class="container">
 
-    <div class="alert alert-warning">
+    <div class="alert alert-warning alert-contact-form">
       <ul>
         <li>
           {% if site_title != "Weblate" %}


softworkz:
> Once again the caveat applies that we can't guarantee stability of attributes in HTML templates

That's totally fine and absolutely not expected to remain like this forever. 

Let me explain this from the perspective of a self-hoster: 

We do not want/need any of those warnings to be shown to users. So what options do we have to hide it?
With CSS (and without this class being added), we have no handle to simply apply `display: none;`, because we do not hide all alerts - just this specific one.

The alternative is to modify the template. But such modifications add to the debt when updating to newer versions in the future, typically by rebasing. And for this, every little change to `contact.html` that you might do in the future can cause a conflict when doing the rebase.
This is by far more unstable than a possible future change to the css classes on the element, which may happen (and that's fully acceptable), but it will happen much more rarely (if at all) than those rebasing conflicts.

A move to bootstrap 5 probably won't matter at all, since this is just about giving the element a name in order to hide it, it's not used for applying any specific styling. Using css classes to name elements is a common and wide-spread practce anyway, without those class names being used for styling at all.

KarenKonou (approved):

nijel:
Let's merge this for now, but I think we should get rid of this warning soon, see https://github.com/WeblateOrg/weblate/issues/16017.

PR #15906, opened 2025-08-20
Apache License 

clearancej2:


nijel:
Weblate is not under the Apache license.

PR #15907, opened 2025-08-21
fix (formats): removes fallback value for file format param that can be falsy

### LINKED ISSUE #15898: Respect JSON indentation format setting.
### Describe the issue

"JSON indentation" settings == 0.

Image

Since some time I've got this:

Image

### I already tried

Check the result PR with translations and got reformatted .json file.

### Expected behavior

No 4 spaces indentation (JSON) if it is set to zero.

### Screenshots

_No response_

### Exception traceback

```pytb

```

### How do you run Weblate?

weblate.org service

### Weblate versions

_No response_

### Weblate deploy checks

```shell

```

### Additional context

_No response_

gersona:
- fixes #15898


weblate/trans/file_format_params.py
```diff
@@ -240,11 +239,11 @@ class YAMLOutputIndentation(BaseYAMLFormatParam):
     label = gettext_lazy("YAML indentation")
     field_class = forms.IntegerField
     default = 2
-    field_kwargs = {"min_value": 0, "max_value": 10}
+    field_kwargs = {"min_value": 1, "max_value": 10}
```
gersona, line 245: Minimal value for YAML indentation was actually 1 and it was incorrectly migrated to file format parameter
https://github.com/WeblateOrg/weblate/blob/fea203addad13e3608652927d59b1ce1d6ad1f6f/weblate/addons/forms.py#L242-L246

nijel: For YAML it makes sense—the indentation is part of the syntax, so having 0 indentation would most likely generate an invalid YAML or merge all keys into a single mapping.

nijel: To give an example:

```yaml
foo:
  bar: baz
```

would become something different with no indentation:

```yaml
foo:
bar: baz
```

PR #15937, opened 2025-08-25
feat(api): Added support for Component's File format parameters in REST API

### LINKED ISSUE #15877: Expose "File format parameters" through components API
### Describe the problem

Since 5.13, addon `weblate.gettext.customize` has been replaced by _File format parameters_. ([doc](https://docs.weblate.org/en/latest/admin/addons.html#customize-gettext-output))

> File format parameters provide a way to configure settings related to the file format. They are configured at **component level** and allow you to customize how file parsing and serialization are handled.
(https://docs.weblate.org/en/latest/formats.html#file-format-params)

I cannot find how to set `file_format_params` via API https://docs.weblate.org/en/latest/api.html#components

I tried `POST /api/projects/<project>/components/` with :
```
{
    ...
    "file_format": "po",
    "file_format_params": {"po_line_wrap": 65535},
    "file_format_params_po_line_wrap": 65535,
    ...
}

```
But still default value in UI `/settings/<project>/<component>/#files`

Then, I updated `po_line_wrap` value in UI and tried to see the change via `GET /api/components/<project>/<component>/ | grep -i "format"`, but only found `"file_format":"po",`.

Sorry if I missed something.

### Solution brainstorm

- Set `file_format_params` on component creation via API ?
- Set default `file_format_params` in _settings-override.py_ ?
- Add an example in the API documentation ?

- Staying in 5.12, with `weblate.gettext.customize` addon in _settings-override.py_.
- Manually updating each component via UI.

### Screenshots

_No response_

### Additional context

Weblate 5.13.0 via official docker image

gersona:
- fixes #15877

### DIFF
--- a/weblate/trans/models/component.py
+++ b/weblate/trans/models/component.py
@@ -3277,7 +3277,7 @@ class Component(
         ]:
             if self.file_format not in param.file_formats:
                 message = gettext(
-                    "The parameter '%(param)s' is not applicable for the '%(format)s'"
+                    "The parameter '%(param)s' is not applicable for the file format '%(format)s'."
                 ) % {"param": param.name, "format": self.file_format}
                 raise ValidationError({"file_format_params": message})
 


weblate/api/tests.py
```diff
@@ -2391,6 +2391,29 @@ def test_create_component_enforced(self) -> None:
...
+                "file_format": "po",
+                "file_format_params": {"po_line_wrap": -1, "yaml_indent": 8},
+                "push": "https://username:password@github.com/example/push.git",
+                "new_lang": "none",
+            },
+            format="json",
+        )
+        component = Component.objects.get(slug="api-project", project__slug="test")
+        self.assertEqual(component.file_format_params["po_line_wrap"], -1)
+        self.assertNotIn("yaml_indent", component.file_format_params)
```
nijel: Can you please add a test case for using a parameter with an invalid value as well?

nijel: Also, I'm not sure if swallowing unknown parameters is a good idea here. It can cause silent breakages, failing with ValidationError in this case might be more desired.

gersona: You're right, it's better to have predictable behavior.

weblate/trans/file_format_params.py
```diff
@@ -96,6 +96,14 @@ def get_default_params_for_file_format(file_format: str) -> dict[str, str | int
     return {param.name: param.default for param in params}
 
 
+def strip_unused_file_format_params(file_format: str, file_format_params: dict) -> dict:
+    """Clean file format parameters, removing those not applicable to the given file format."""
+    for param in FILE_FORMATS_PARAMS:
+        if file_format not in param.file_formats:
+            file_format_params.pop(param.name, None)
+    return file_format_params
```
gersona, line 110: After a second thought,  this is still necessary because of the dynamic display of format param in the Component form on the frontend and of how `MultiValueField` behaves. It is just moved higher to the form level, and stricter validation is done at the model level

weblate/trans/forms.py
```diff
@@ -1874,15 +1874,11 @@ def __init__(self, request: AuthenticatedHttpRequest, *args, **kwargs) -> None:
...
-
-        # only applicable fields are saved to model
         if "file_format_params" in data:
-            selected_format = data["file_format"]
-            for param_format in FILE_FORMATS_PARAMS:
-                if selected_format not in param_format.file_formats:
-                    data["file_format_params"].pop(param_format.name, None)
+            data["file_format_params"] = strip_unused_file_format_params(
+                data["file_format"], data["file_format_params"]
+            )
```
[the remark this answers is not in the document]
nijel: In case this was this intentional, it should be documented.

weblate/trans/validators.py
```diff
@@ -60,3 +60,21 @@ def validate_language_code(code: str | None, filename: str, required: bool = Fal
         raise ValidationError({"filemask": message})
 
     return Language.objects.auto_get_or_create(code=code, create=False)
+
+
+def validate_file_format_parameters(value: dict | None) -> None:
+    from weblate.trans.file_format_params import FILE_FORMATS_PARAMS
+
+    name_to_file_format_parmas = {param.name: param for param in FILE_FORMATS_PARAMS}
```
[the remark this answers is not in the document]
nijel: Should be fixed.

weblate/api/tests.py
```diff
@@ -2391,6 +2391,68 @@ def test_create_component_enforced(self) -> None:
...
+        }
+        self.do_request(
+            "api:project-components",
+            self.project_kwargs,
+            method="post",
+            code=400,
+            superuser=True,
+            request=payload,
+            format="json",
+        )
```
nijel, line 2472: Test for passing non-existing parameter is missing here.

weblate/trans/validators.py
```diff
@@ -60,3 +60,21 @@ def validate_language_code(code: str | None, filename: str, required: bool = Fal
...
+
+    if value is None:
+        return
+
+    if not isinstance(value, dict):
+        raise ValidationError(
+            gettext("File format parameters must be a dictionary of key-value pairs.")
+        )
+
+    for param_name, param_value in value.items():
```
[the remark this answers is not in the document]
nijel: IMHO a valid suggestion, as this value can be any JSON data (and this is not tested, see another comment).

weblate/trans/models/component.py
```diff
@@ -3268,9 +3271,19 @@ def clean_repo(self) -> None:
             )
             raise ValidationError({"push_branch": msg})
 
+    def clean_file_format_params(self) -> None:
+        for param in [
+            p for p in FILE_FORMATS_PARAMS if p.name in self.file_format_params
+        ]:
+            if self.file_format not in param.file_formats:
+                message = gettext(
+                    "The parameter '%(param)s' is not applicable for the '%(format)s'"
```
```suggestion
                    "The parameter '%(param)s' is not applicable for the file format '%(format)s'."
```

nijel (approved):

nijel:
Scheduled for merge, thanks!

PR #15942, opened 2025-08-25
docs: clarify glossary 'terminology' flag semantics (first-time behav…

### LINKED ISSUE #15683: Unclear Semantics of "terminology" Flag
## Unclear Semantics of "terminology" Flag

While the other two are about defining policies for the translation process, that doesn't seem to be the case for the "terminology" flag:

### Operational first-time behavior

When the flag is set for the first time, empty translation strings are created for all language which don't have an entry yet for that string.
That's how a command works, but not a property change of an entity, even more when considering that the operation is irreversable, which means that when the flag is removed later, the outcome of the operation will not be reverted.

### State Indication?

Besides the above, I'm wandering which kind of state the presence of that flag would indicate?

What's the difference between a unit (glossary) where translations have been imported initially and units already have all translations - but without that flag?

What's the change in business logic when the flag is removed from a unit?

### I already tried

m

### Expected behavior

_No response_

### Screenshots

_No response_

### Exception traceback

```pytb

```

### How do you run Weblate?

_No response_

### Weblate versions

_No response_

### Weblate deploy checks

```shell

```

### Additional context

_No response_

niyati34:
…ior, state, removal, deletion constraints) #<issue-number>

PR body
Closes #15683

Summary

- Clarifies the meaning and effects of the glossary terminology flag.
- Documents first-time population behavior, ongoing state/management, what happens when unmarked, and deletion constraints.
- No functional changes; documentation only.

What this improves

- Addresses “Unclear Semantics of ‘terminology’ Flag”.
- Explains that marking a source glossary term with the terminology flag:
    - Creates entries for that term in all glossary languages that lack it (with empty targets).
    - Continues to create missing entries when new languages are added while the flag is present.
    - Prevents deleting individual non‑source terminology entries (remove the source term or unmark first).
- Clarifies that removing the flag later:
    - Does not delete previously created entries; it only stops further auto‑creation for newly added languages.
    - Re-enables deleting individual per‑language entries (subject to component/file-format rules).
- Differentiates “terminology” terms from regular glossary entries (regular entries only exist where created/imported; terminology ensures presence across all languages).
- Notes that creating terminology entries may require “Add glossary terminology” permission, depending on configuration.

Scope of change

- Documentation only. No user-visible behavior or API changes.

Files changed

docs/user/glossary.rst: Expanded the “Terminology” section with:

- One-time population behavior
- State/management semantics while flagged
- Unmark/removal behavior
- Deletion protection and how to remove
- Differences from regular entries
- Permission note

Checklist

- [x]  Documentation updated
- [x]  No code changes
- [x]  No tests required (docs-only)
- [x]  Linked the issue (replace <issue-number>)


niyati34:
any updates? @orangesunny

orangesunny:
> any updates? @orangesunny

This PR’s content and description, both LLM-generated, would deliver an LLM-generated, bloated, long article that adds little information and value for its reader while reducing the readability and searchability of the content. Because of this, it will not be merged.

PR #16002, opened 2025-09-01
fix(auth): shorten session expiry while in 2fa

nijel:
The 2fa should have a short session expiry as this is expected to happen interactively. Previously 2fa input got the session expiry for authenticated users what was wrong.

### DIFF
--- a/weblate/accounts/forms.py
+++ b/weblate/accounts/forms.py
@@ -785,14 +785,14 @@ class LoginForm(forms.Form):
                     )
                     % lockout_period
                 )
-            self.user_cache = cast(
+            user = self.user_cache = cast(
                 "User | None",
                 authenticate(self.request, username=username, password=password),
             )
-            if self.user_cache is None:
-                for user in try_get_user(username, True):
+            if user is None:
+                for failed_user in try_get_user(username, True):
                     audit = AuditLog.objects.create(
-                        user,
+                        failed_user,
                         self.request,
                         "failed-auth",
                         method="password",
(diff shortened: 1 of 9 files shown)


weblate/accounts/forms.py
```diff
@@ -803,14 +803,14 @@ def clean(self):
                 raise forms.ValidationError(
                     self.error_messages["invalid_login"], code="invalid_login"
                 )
-            if not self.user_cache.is_active or self.user_cache.is_bot:
+            if not user.is_active or self.user_cache.is_bot:
```
amCap1712: minor nit for consistency
```suggestion
            if not user.is_active or user.is_bot:
```

weblate/auth/models.py
```diff
@@ -594,6 +598,10 @@ def has_usable_password(self):
     def is_anonymous(self):
         return self.username == settings.ANONYMOUS_USER_NAME
 
+    def is_verified(self) -> bool:
+        # django_otp integration (via OTPMiddleware)
```
amCap1712: nit: Took me a while to realize that this method is not just used but overridden by OTPMiddleware, would be nice to mention that explicitly.

weblate/accounts/utils.py
```diff
@@ -151,14 +151,24 @@ def cycle_session_keys(request: AuthenticatedHttpRequest, user: User) -> None:
...
 ) -> None:
     """
     Adjust session expiry based on scope.
 
     - Set longer expiry for authenticated users.
     - Set short lived session for SAML authentication flow.
     """
+    is_2fa = False
+    if user is None:
+        user = request.user
```
amCap1712: Small suggestion. Only one place doesn't pass in user explicitly.

https://github.com/WeblateOrg/weblate/blob/a23327aa9471e8ee4d0688555c56921919ab7905/weblate/accounts/middleware.py#L69

Can pass the user explicitly there as already available and then here make it mandatory. Also, would be able to remove the None type as not logged in users will be sent as AnonymousUser. As a result, simplified to:

`is_2fa = user.profile.has_2fa and not user.is_verified()`

PR #16102, opened 2025-09-08
feat: push filter to database in commit pending detection

### LINKED ISSUE #15846: Smarter detecting of pending changes
### Describe the problem

The current logic for detecting components with pending changes is naive, and it goes through all the components with pending changes:

https://github.com/WeblateOrg/weblate/blob/3f72c059857fc375fb57f40b1d486af6032bf371/weblate/trans/tasks.py#L151-L175

### Solution brainstorm

Include pending age in the query for existing pending changes. It can be either provided as the parameter and defaults `Component.commit_pending_age`.

_No response_

### Screenshots

_No response_

### Additional context

https://github.com/WeblateOrg/weblate/issues/15691

amCap1712:
Push the age cutoff filter to the database when detecting which components might be eligible for a commit.
Fixes #15846.

### DIFF
--- a/weblate/trans/functions.py
+++ b/weblate/trans/functions.py
@@ -45,7 +45,7 @@ class MySQLTimestampAdd(Func):
         function: str | None = None,
         template: str | None = None,
         arg_joiner: str | None = None,
-        **extra_context: Any,
+        **extra_context: Any,  # noqa: ANN401
     ) -> _AsSqlType:
         interval_sql, interval_params = self.source_expressions[0].as_sql(
             compiler, connection


weblate/trans/tasks.py
```diff
@@ -153,22 +154,16 @@ def commit_pending(
     else:
         components = Component.objects.filter(translation__pk__in=pks)
 
-    # All components with pending units
+    hours_expr = F("commit_pending_age") if hours is None else Value(hours)
+    age_cutoff = ExpressionWrapper(
+        Now() - hours_expr * timedelta(hours=1), output_field=DateTimeField()
+    )
```
nijel: This doesn't work for MySQL/MariaDB:

```
query = b'SELECT DISTINCT `trans_component`.`id`, `trans_component`.`name`, `trans_component`.`slug`, `trans_component`.`proje...nge`.`timestamp` < (CURRENT_TIMESTAMP(6) - (`trans_component`.`commit_pending_age` * INTERVAL 3600000000 MICROSECOND))'
...
django.db.utils.ProgrammingError: (1064, "You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '))' at line 1")
```

Probably you can't multiply a field by an INTERVAL there.

amCap1712: Tried out a few ways to fix this and creating a custom function that calls `TIMESTAMPADD` in mysql is the only solution that worked so far.

nijel: Hmm, maybe we should again revisit the decision to support MySQL, these workarounds are getting annoying...

amCap1712: If we could collect information about how many clients are using MySQL/MariaDB vs Postgres, that would be super.

weblate/trans/functions.py
```diff
@@ -0,0 +1,50 @@
...
+            compiler, connection
+        )
+        timestamp_sql, timestamp_params = self.source_expressions[1].as_sql(
+            compiler, connection
+        )
+
+        # override default template to avoid addition of unnecessary parentheses
+        # around function arguments
+        sql = f"{self.function}({self.unit}, {interval_sql}, {timestamp_sql})"
+        params = interval_params + timestamp_params
```
amCap1712: Will look into type errors and fix it.

weblate/trans/tasks.py
nijel: We deploy on PostgreSQL only, but last time this was brought up, several community members wanted to keep MySQL support. The previous discussion on this topic happened in https://github.com/WeblateOrg/weblate/issues/2825.

We don't have numbers for all installs because we do not track them for privacy reasons.

weblate/trans/functions.py
```diff
@@ -0,0 +1,50 @@
...
+            raise ValueError(msg)
+        self.unit = unit
+        super().__init__(interval, timestamp)
+
+    def as_sql(
+        self,
+        compiler,
+        connection,
+        **kwargs,
+    ):
```
nijel: Remove extra kwargs:

```suggestion
    def as_sql(self, compiler, connection):
```

weblate/trans/functions.py
nijel: Ignore type error here, IMHO there is something wrong in Django stubs, there is simialr code in weblate/utils/db.py.

```suggestion
        params = interval_params + timestamp_params  # type: ignore[operator]
```

weblate/trans/tasks.py
```diff
@@ -153,22 +156,19 @@ def commit_pending(
     else:
         components = Component.objects.filter(translation__pk__in=pks)
 
-    # All components with pending units
+    hours_expr = F("commit_pending_age") if hours is None else Value(hours)
+    if using_postgresql():
```
nijel, line 168: Type annotate prior to first use:

```suggestion
    age_cutoff: Expression
    if using_postgresql():
```

weblate/trans/tasks.py
```diff
@@ -17,7 +17,8 @@
 from django.contrib.staticfiles.storage import staticfiles_storage
 from django.core.cache import cache
 from django.db import IntegrityError, transaction
-from django.db.models import Count, F
+from django.db.models import Count, DateTimeField, ExpressionWrapper, F, Value
```
nijel: Add import for type annotation:

```suggestion
from django.db.models import Count, DateTimeField, Expression, ExpressionWrapper, F, Value
```

weblate/trans/functions.py
```diff
@@ -0,0 +1,45 @@
...
+        self.unit = unit
+        super().__init__(interval, timestamp)
+
+    def as_sql(self, compiler, connection):
+        interval_sql, interval_params = self.source_expressions[0].as_sql(
+            compiler, connection
+        )
+        timestamp_sql, timestamp_params = self.source_expressions[1].as_sql(
+            compiler, connection
+        )
```
nijel:
```suggestion
    def as_sql(self, compiler, connection, function, template, arg_joiner, **extra_context):
        interval_sql, interval_params = self.source_expressions[0].as_sql(
            compiler, connection, function, template, arg_joiner, **extra_context
        )
        timestamp_sql, timestamp_params = self.source_expressions[1].as_sql(
            compiler, connection, function, template, arg_joiner, **extra_context
        )
```

weblate/trans/functions.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def as_sql(
+        self, compiler, connection, function, template, arg_joiner, **extra_context
+    ):
+        interval_sql, interval_params = self.source_expressions[0].as_sql(
+            compiler, connection, function, template, arg_joiner, **extra_context
+        )
+        timestamp_sql, timestamp_params = self.source_expressions[1].as_sql(
+            compiler, connection, function, template, arg_joiner, **extra_context
+        )
```
nijel:
```suggestion
        interval_sql, interval_params = self.source_expressions[0].as_sql(compiler, connection)
        timestamp_sql, timestamp_params = self.source_expressions[1].as_sql(compiler, connection)
```

weblate/trans/functions.py
```diff
@@ -0,0 +1,47 @@
...
+        # error if the unit argument to TIMESTAMPADD is quoted.
+        if unit not in VALID_UNIT_VALUES:
+            msg = f"Invalid unit: {unit}"
+            raise ValueError(msg)
+        self.unit = unit
+        super().__init__(interval, timestamp)
+
+    def as_sql(
+        self, compiler, connection, function, template, arg_joiner, **extra_context
+    ):
```
nijel:
```suggestion
    def as_sql(
        self,
        compiler: SQLCompiler,
        connection: BaseDatabaseWrapper,
        function: str | None = None,
        template: str | None = None,
        arg_joiner: str | None = None,
        **extra_context: Any,
    ) -> _AsSqlType:
```

weblate/trans/functions.py
```diff
@@ -0,0 +1,47 @@
+# Copyright © Michal Čihař <michal@weblate.org>
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from django.db.models import DateTimeField, Expression, Func
```
nijel, line 9:
```suggestion
from __future__ import annotations
from typing import Any
from django.db.models import DateTimeField, Expression, Func
from django.db.backends.base.base import BaseDatabaseWrapper
from django.db.models.sql.compiler import SQLCompiler, _AsSqlType
```

weblate/trans/tasks.py
amCap1712: Thanks for sharing. I wonder if things have changed since then. We'd have to ask the community somehow if they are using mysql/mariadb with weblate since we don't have numbers but that's a separate discussion.

weblate/trans/functions.py
```diff
@@ -0,0 +1,61 @@
...
+        super().__init__(interval, timestamp)
+
+    def as_sql(
+        self,
+        compiler: SQLCompiler,
+        connection: BaseDatabaseWrapper,
+        function: str | None = None,
+        template: str | None = None,
+        arg_joiner: str | None = None,
+        **extra_context: Any,
```
nijel:
```suggestion
        **extra_context: Any,  # noqa: ANN401
```

nijel:

weblate/trans/tasks.py
nijel: Created https://github.com/WeblateOrg/weblate/issues/16140 for that.

PR #16109, opened 2025-09-08
fix(formats): pass language codes and parameters to try_load

### LINKED ISSUE #16086: API import of es_XX and pt_XX Ruby YAML files with plurals issue (many/other)
### Describe the issue

When importing a Ruby YAML files for `es_XX` (e.g. `es_AR`) or `pt_XX` (e.g. `pt_BR`), I get some weird behavior where the `many` value is mapped to `other`, and the `other` value is dropped.

I initially noticed the error when uploading using the API, but this also occurs when uploading through the UI.

This might be related to https://github.com/WeblateOrg/weblate/issues/4484, but I'm not entirely sure.

This does not seem to be an issue for `es`, even though it has the same plural rules as `es_XX`, so that seems related to the `_XX`.

### I already tried

1. upload a `pt_BR` Ruby YAML file with CLDR plurals (one/many/other) as a translation file in a project
2. notice that `other` is missing and its value ended up in `many`

### Expected behavior

Mapping should work properly

### Screenshots

Image

Image

### Exception traceback

```pytb

```

### How do you run Weblate?

Docker container

### Weblate versions

5.13.1

### Weblate deploy checks

```shell

```

### Additional context

This is a bit of a blocker as we need to import a bunch of files with a lot of plurals. If there's no easy fix, is there a way around the issue at least?

nijel:
This ensures that the uploaded files are parsed using the same settings and when loaded directly in Weblate. This fixes wrong plurals handling for some formats and lack of parameters that influence parsing (such as flat XML ones).

Fixes #16086


nijel:
Ideally this should have a test case (or at least verification by the original reporter).

PR #16122, opened 2025-09-09
feat: list administered projects when removing account

nijel:
This way user would be better aware of the consequences.

### DIFF
--- a/weblate/accounts/tests/test_remove.py
+++ b/weblate/accounts/tests/test_remove.py
@@ -23,7 +23,8 @@ class AccountRemovalTest(ViewTestCase, RegistrationTestMixin):
         # Verify confirmation URL
         response = self.client.get(url, follow=True)
         self.assertContains(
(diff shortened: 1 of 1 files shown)


weblate/templates/accounts/removal.html
```diff
@@ -17,16 +17,33 @@ <h4 class="panel-title">{% translate "Account removal" %}</h4>
...
+                {% endfor %}
+              </ul>
+            </li>
+          {% endif %}
+          {% if has_billing %}
+            <li>{% translate "Any billing you might have created will not be terminated by account removal." %}</li>
+          {% endif %}
+          {% if is_confirmation %}
+            <li>
+              {% blocktranslate %}By pressing following button, your will no longer be able to use this account.{% endblocktranslate %}
```
nijel:
```suggestion
              {% blocktranslate %}By pressing the following button, you will no longer be able to use this account.{% endblocktranslate %}
```

PR #16145, opened 2025-09-10
feat: run auto translate more proactively

### LINKED ISSUE #15944: Automatic translation add-on triger
### Describe the problem

The automatic translation add-on currently triggers on update or daily. But with more complex string filters that are now available, this might not bring desired results in time.

For example, with `source_state:approved AND state:empty` user would expect that automatic translation would trigger immediately once the source string was approved.

### Solution brainstorm

The add-on could additionally listen to the generic `EVENT_CHANGE` and match the unit changes against the filter and trigger translation for matching strings only (in the background).

_No response_

### Screenshots

_No response_

### Additional context

This has become more of an issue since https://github.com/WeblateOrg/weblate/pull/15585.

amCap1712:
Just a first draft to check whether on the right track.

In addition to the changes in the PR I intend to modify the AutoTranslate class to support running it on particular units and not just whole translations.

https://github.com/WeblateOrg/weblate/blob/c804931126ea572cda957b7c2cca062a45ea58ee/weblate/trans/autotranslate.py#L38

Fixes #15944

### DIFF
--- a/weblate/addons/migrations/0012_add_change_event_to_autotranslate.py
+++ b/weblate/addons/migrations/0012_add_change_event_to_autotranslate.py
@@ -15,7 +15,7 @@ def add_autotranslate_change_event(apps, schema_editor) -> None:
         schema_editor,
         ["weblate.autotranslate.autotranslate"],
         [AddonEvent.EVENT_CHANGE],
-        False,
+        [],
     )
 
 


weblate/addons/autotranslate.py
```diff
@@ -13,16 +13,18 @@
...
 if TYPE_CHECKING:
-    from weblate.trans.models import Component
+    from weblate.trans.models import Change, Component
 
 
 class AutoTranslateAddon(BaseAddon):
     events: set[AddonEvent] = {
         AddonEvent.EVENT_COMPONENT_UPDATE,
         AddonEvent.EVENT_DAILY,
+        AddonEvent.EVENT_CHANGE,
```
nijel, line 28: This will need a migration to sync events in the database. There is a (currently unused) helper for this in `weblate/addons/utils.py`.

weblate/addons/autotranslate.py
```diff
@@ -61,3 +63,41 @@ def daily(self, component: Component) -> None:
             return
 
         self.component_update(component)
+
+    def change_event(self, change: Change) -> None:
+        translation_ids = set()
+        if change.action in ACTIONS_CONTENT:
+            if change.unit is not None:
+                translation_ids.add(change.unit.translation.id)
```
nijel: The change in source string (`Unit.is_source`) should trigger for all the translations. 

The similar logic will apply to all the events.

amCap1712:
Also, I think this might not cover all cases. For one example, I don't think there are events for labels but it is possible to create filter queries for specific labels.

nijel:
I think there should be an event for that anyway. https://github.com/WeblateOrg/weblate/issues/16146 will track that :-).

amCap1712:
Noting more events that might need to be created here:

Currently, there are only events for adding a new screenshot and editing the existing screenshot. Further, the change event for screenshots is created before the attaching it to the unit.

https://github.com/WeblateOrg/weblate/blob/b5f978df24129a7f8cc84f11c52d4b59b812c8c7/weblate/screenshots/views.py#L233-L239

I can change the order here likely without issue but attaching units from the screenshot detail page would still be missed, so  I think we need an event for:

1. assigning a screenshot to a unit.
2. removing a screenshot from a unit.

nijel:
Yes, please do that in a separate pull request for easier review.

> attaching units from the screenshot detail page would still be missed

Probably the change generation should be pushed to the model so that the individual views don't have to handle this.

nijel:
@amCap1712 Any blockers on this? I'd like to have this in 5.14 in 10 days.

Adding missing change events can be done later.

amCap1712:
@nijel ready except for the new events.

amCap1712:
I'll add the migration.

weblate/addons/tests.py
```diff
@@ -1378,6 +1378,44 @@ def test_auto(self) -> None:
...
+
+        translation_2 = component_2.translation_set.get(language_code="cs")
+        unit_2 = translation_2.unit_set.get(source="one")
+        Comment.objects.create(unit=unit_2, comment="Foo")
+        change = unit_2.change_set.latest("timestamp")
+
+        addon_change.run([change.pk])
+
+        unit_2 = translation_2.unit_set.get(source="one")
+        self.assertEqual(unit_2.target, "jeden")
```
nijel, line 1417: This assertion fails.

weblate/addons/migrations/0012_add_change_event_to_autotranslate.py
```diff
@@ -0,0 +1,31 @@
...
+from weblate.addons.utils import adjust_addon_events
+
+
+def add_autotranslate_change_event(apps, schema_editor) -> None:
+    adjust_addon_events(
+        apps,
+        schema_editor,
+        ["weblate.autotranslate.autotranslate"],
+        [AddonEvent.EVENT_CHANGE],
+        False,
```
nijel: It would work, but the intention was an empty list:

```suggestion
        [],
```

I will add type annotations to make it clear.

nijel:
Scheduled for merge, thanks!

PR #16166, opened 2025-09-11
Add global permissions for viewing team, user and role info

### LINKED ISSUE #15950: Site-wide Permission for viewing all Users
### Describe the problem

On our dedicated Weblate instance we'd like to collect information about all users via the API for reporting purposes.

The user calling the API needs the `user.edit` permission. It would be more secure and less error-prone if the API key used would only have read-only permissions.

### Solution brainstorm

Introduce a `user.view` global permission, and possibly more where it makes sense, like `group.view`.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

gersona:
- fixes #15950 

Most global permissions are not listed in the `access.rst` file, is that on purpose or something that can be added here ?


nijel:
Not intentional; it is a bug in the rendering. https://github.com/WeblateOrg/weblate/pull/16174 fixes that.

nijel:

PR #14531, opened 2025-04-11
feat: backup categories

amCap1712:
Include categories in project backups.

Fixes #13743.

Depends on https://github.com/WeblateOrg/weblate_schemas/pull/426.

PR #16172, opened 2025-09-11
fix(projectbackup): correctly name JSON with component data

nijel:
Using just slug is not enough because it does not have to be unique across categories.

Fixes WEBLATE-354X


nijel:
@amCap1712 Thanks to changes in #15776, I was able to discover and fix a bug in #14531.

PR #16187, opened 2025-09-12
feat(addon): Improve Addon Activity Log for LanguageConsistency Addon

### LINKED ISSUE #12738: Improve language consistency add-on logs
### Describe the problem

Right now the add-ons log only contain information about the add-on being executed. While this is useful, it would be useful to contain more information.

Let's start with the language consistency add-on. The log should include:

* Information on added languages
* Information on languages which could not be added

The background task should extend/append the log entry.

_No response_

### Screenshots

_No response_

### Additional context

Probably all logging from the task should make it into add-on log:

https://github.com/WeblateOrg/weblate/blob/9db0980cb2291b077fb3b5dd7012ac22b546c803/weblate/addons/tasks.py#L77-L115

gersona:
- fixes #12738


weblate/addons/tasks.py
```diff
@@ -112,6 +115,8 @@ def language_consistency(
         )
     ).exclude(translation_count=languages.count())
 
+    log_result = ""
```
nijel: The repeated string concatenation will perform badly with a log of log entries, use list and merge it with join in the end:

```suggestion
    log_result : list[str] = []
```

weblate/addons/autotranslate.py
```diff
@@ -34,7 +34,7 @@ class AutoTranslateAddon(BaseAddon):
     multiple = True
     icon = "language.svg"
 
-    def component_update(self, component: Component) -> None:
+    def component_update(self, component: Component, **kwargs) -> None:
```
nijel: I don't really like using `**kwargs` here; it makes it easier to break things when the API changes, so use the new arg here directly.

weblate/addons/base.py
```diff
@@ -194,21 +206,38 @@ def post_configure_run_component(self, component: Component) -> None:
...
             # The post_update typically operates on files, so make sure these are updated
             component.commit_pending("add-on", None)
-            self.post_update(component, "", False)
+            self.run_handler(
+                self.post_update,
+                AddonEvent.EVENT_POST_UPDATE,
+                component=component,
+                previous_head="",
+                skip_push=False,
+            )
```
nijel, line 218: Couldn't we use `execute_addon_event` instead of reimplementing the logic here?

weblate/addons/tasks.py
```diff
@@ -159,6 +165,19 @@ def daily_callback(addon: Addon, component: Component) -> None:
...
+def update_addon_activity_log(
+    pk: int, result: str = "", error_occurred: bool = False, pending: bool | None = None
+) -> None:
+    addon_activity_log = AddonActivityLog.objects.select_for_update().get(id=pk)
+    addon_activity_log.details["error"] = error_occurred
+    if result:
+        addon_activity_log.update_result(result)
+    if pending is not None:
+        addon_activity_log.pending = pending
+    addon_activity_log.save()
```
nijel: Update only changed fields:

```suggestion
    addon_activity_log.save(update_fields=["details", "pending"])
```

weblate/addons/tasks.py
```diff
@@ -159,6 +165,19 @@ def daily_callback(addon: Addon, component: Component) -> None:
     )
 
 
+@app.task(trail=False)
+def update_addon_activity_log(
```
nijel, line 174: What is the motivation for having this in a task? It is a single database update, so the task overhead will be most likely bigger than the task itself.

gersona: makes sense, that was a failed over-optimization attempt

nijel:

gersona:
@nijel I don't recognize the `ResetAddon` name,  is this a custom add-on on the hosted instance that just needs to be updated with the new arguments ?
The compatibility section of the changelog also needs to be updated to reflect that

nijel:
Yes , it's in the hosted repo, I'll look into that tomorrow.

nijel:
https://github.com/WeblateOrg/hosted/commit/caf8046a933921c008c2c960b3268056e477664a should address this, please update the changelog to mention this.

PR #16209, opened 2025-09-15
fix: make team admins behave as documented

### LINKED ISSUE #16186: Team member managers with rights to only modify member list
### Describe the problem

Teams can have members that are administrators, which have the right to change everything about the team, including the roles.

It is not possible to delegate membership management in teams to some members while ensuring the team does not get more permissions.

### Solution brainstorm

Add a new _Team member managers_ list to the Teams which allows to only add and remove members. Add additional option "Grant member management" besides the "Grant administration" functionality in the member listing.

When being a member manager, on the Members tab, allow:
- adding
- removing
- granting member management
but do not allow granting administration rights.

Being granted both _administration_ and _member management_ should not be a issue because the rights are overlapping (admin has same and more rights than member manager).

_No response_

### Screenshots

_No response_

### Additional context

When e.g. contracting a third party to do translations over a longer period of time, this would allow to delegate the membership management to the third party and make their personnel changes more self-service.

nijel:
It was possible to change team properties by the team admins, but the desired behavior was only to change team members. This also changes the admin view visibility, it is now added to the project menu.

Fixes #16186

### DIFF
--- a/weblate/templates/project.html
+++ b/weblate/templates/project.html
@@ -91,7 +91,7 @@
             <a href="#bulk-edit" data-toggle="tab">{% translate "Bulk edit" %}</a>
           </li>
         {% endif %}
-        <li role="separator" class="divider"></li>
+        {% if replace_form and bulk_state_form %}<li role="separator" class="divider"></li>{% endif %}
         {% if user_can_see_repository_status %}
           <li>
             <a href="#repository"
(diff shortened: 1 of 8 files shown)


weblate/templates/project.html
nijel, line 1: @KarenKonou This will probably have to be incorporated in #15862.

PR #16213, opened 2025-09-15
feat(lang): extend plural mapper for subset matching

nijel:
This helps getting a viable translation for more plurals - where translation is subset of source. It might be not accurate enough, but still better than ommitting one of the plurals.

### DIFF
--- a/weblate/lang/models.py
+++ b/weblate/lang/models.py
@@ -1194,6 +1194,7 @@ class PluralMapper:
         for i in range(target_plural.number):
             examples = target_plural.examples.get(i, [])
             if len(examples) == 1:
+                # Map plurals 1:1
                 number = int(examples[0])
                 if number in exact_source_map:
                     result.append((exact_source_map[number], None))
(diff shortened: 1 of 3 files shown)


nijel:
@Changaco Any thoughts on this? The motivation is to provide more plural translations for languages with additional plurals compared to the source. It might be mixed/replaced with the approach outlined in https://github.com/WeblateOrg/weblate/pull/15701/files#diff-1139fa26bfde0f7283b09cf831f4c9160d4f30a050e7d2256d8cc978154883a1R1210 to always fall back to the last plural form.

Changaco:
Sadly this patch doesn't actually improve the plural mapper. In the only affected test case, English to Czech, the result would be an inaccurate translation in the second plural form, no better than simply falling back to duplicating the translation of the last plural form of the source language.

I originally felt it was more appropriate to leave empty the plural forms that can't be accurately translated due to the lack of support for plural form hints in machine translation APIs, but if other users would rather have them filled with inaccurate translations than stay empty, I'm okay with that. It could be an option, either opt-in or opt-out, enabling everyone to pick the behaviour they prefer.

nijel:
The question is what we're trying to achieve here. The incomplete plurals behavior varies in different frameworks, but in many you end up using a blank string in the UI. This was my motivation for this change.

PS: The other thing is that you never know *which* of the plural forms will be the result of machine translation if the target language has way more plurals than the source. So this should be anyway subject of review.

nijel:
Let's merge this. The translation of additional plural will indeed be inaccurate, but that is often the case with machine translations, and human review is desired. On the other side, missing plurals break some use cases, and I think it's better to have inaccurate translation rather than a blank string in the UI.

PR #16233, opened 2025-09-17
ui: add option to hide local repositories in admin repo status page

michael-smt:
Similar to #9950 this adds another checkbox to hide all components without a remote VCS (like glossaries) in the admin repositories page.

This helps to diagnose VCS integration issues by hiding all other components.


nijel:

PR #16238, opened 2025-09-18
docs: clarify that there can be multiple team administrators

michael-smt:
Just a very small change to make it clear that there can be multiple team administrators.


nijel:

PR #16303, opened 2025-09-24
fix(formats): close internally open file in convert format

nijel:
There is no need to keep it open once the conversion has been completed.

### DIFF
--- a/weblate/formats/convert.py
+++ b/weblate/formats/convert.py
@@ -152,6 +152,7 @@ class ConvertFormat(TranslationFormat):
                 store = self.convertfile(handle, template_store)
         else:
             store = self.convertfile(storefile, template_store)
+        # Adjust store to have translations
         if self.needs_target_sync(template_store):
             for unit in store.units:
                 if unit.isheader():


weblate/formats/convert.py
```diff
@@ -148,9 +148,10 @@ def load(
     ) -> TranslationStore:
         # Did we get file or filename?
         if not hasattr(storefile, "read"):
-            storefile = open(storefile, "rb")  # noqa: SIM115
-        # Adjust store to have translations
-        store = self.convertfile(storefile, template_store)
+            with open(storefile, "rb") as f:
+                store = self.convertfile(f, template_store)
```
nijel:
```suggestion
            with open(storefile, "rb") as handle:
                store = self.convertfile(handle, template_store)
```

weblate/formats/convert.py
```diff
@@ -148,9 +148,10 @@ def load(
     ) -> TranslationStore:
         # Did we get file or filename?
         if not hasattr(storefile, "read"):
-            storefile = open(storefile, "rb")  # noqa: SIM115
-        # Adjust store to have translations
-        store = self.convertfile(storefile, template_store)
+            with open(storefile, "rb") as handle:
+                store = self.convertfile(handle, template_store)
+        else:
+            store = self.convertfile(storefile, template_store)
```
nijel, line 154:
```suggestion
            store = self.convertfile(storefile, template_store)
        # Adjust store to have translations
```

PR #16322, opened 2025-09-25
Use batch operations for slow source checks

### LINKED ISSUE #14174: Batch triggering of source checks
### Describe the problem

The source checks like MultipleFailingCheck or LongUntranslatedCheck are currently the slowest paths in the `update_checks` task. These could probably be done using a single SQL query instead of doing a query per string.

Similarly to the target checks, there should be a way to batch these. It should be used in both translation update and in scheduled checks update.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

gersona:
- fixes #14174


weblate/checks/source.py
```diff
@@ -112,20 +138,36 @@ def get_description(self, check_obj):
...
+
+        result = []
+        for unit in source_units:
+            if states := source_unit_to_states[unit.pk]:
+                total = len(states)
+                not_translated = states.count(STATE_EMPTY) + states.count(STATE_FUZZY)
+                translated_percent = 100 * (total - not_translated) / total
+                if (
+                    2 * translated_percent
+                    < unit.translation.component.stats.translated_percent
```
gersona: This line would still be performing one request to the cache DB for each unit. 
Should it be converted to something like batch loading of stats , or it is OK to leave as it is because it's only a the cache DB ?

weblate/checks/source.py
```diff
@@ -112,20 +138,36 @@ def get_description(self, check_obj):
...
+    def check_source_unit(self, sources: list[str], unit: Unit) -> bool:
+        return self.handle_batch(unit, unit.translation.component)
+
+    def check_component(self, component: Component) -> Iterable[Unit]:
+        from weblate.trans.models import Unit
+
+        units = Unit.objects.filter(
+            translation__component=component,
+            source_unit__timestamp__lt=timezone.now() - timedelta(days=90),
+        ).select_related("source_unit__translation__component")
```
nijel: I believe prefetch would work better here as it fetches each object just once instead of fetching the component separately for each translation.

```suggestion
        ).prefetch_related("source_unit__translation__component")
```

weblate/checks/source.py
nijel: There is `prefetch_stats` which could be used for that. It would be still beneficial with projects with a lot of components. But with prefetch_related (see previous comment), each component will have just one instance.

weblate/utils/stats.py
```diff
@@ -288,7 +288,7 @@ def aggregate_get(self, name: str) -> StatItem:
         Fast getter for aggregation.
 
         - expects data is present
-        - expects valid basic keys only
+        - expects valid §asic keys only
```
nijel: Seems not intended :-).

weblate/checks/tests/test_source_checks.py
```diff
@@ -52,18 +55,61 @@ class LongUntranslatedCheckTestCase(FixtureTestCase):
     def test_recent(self) -> None:
         unit = self.get_unit(language="en")
         unit.timestamp = timezone.now()
+        unit.save()
+        LongUntranslatedCheck().perform_batch(self.component)
```
nijel: This should not be necessary, the single run_checks behavior should not be changed.

weblate/checks/source.py
```diff
@@ -61,22 +64,45 @@ class MultipleFailingCheck(SourceCheck):
...
+        return Check.objects.filter(unit_id__in=unit_ids).select_related(
+            "unit__translation"
+        )
 
     def check_source_unit(self, sources: list[str], unit: Unit):
-        related = (
-            self.get_related_checks(unit)
-            .values_list("unit__translation", flat=True)
-            .distinct()
+        return self.handle_batch(unit, unit.translation.component)
```
nijel: Should be conditional only:

```suggestion
        if component.batch_checks:
            return self.handle_batch(unit, unit.translation.component)
        ... direct check execution.
```

weblate/checks/source.py
nijel: ...actually there will be just a single component object here, so no need to prefetch stats.

weblate/checks/source.py
```diff
@@ -112,20 +138,36 @@ def get_description(self, check_obj):
...
-            < unit.translation.component.stats.translated_percent
-        )
+    def check_source_unit(self, sources: list[str], unit: Unit) -> bool:
+        return self.handle_batch(unit, unit.translation.component)
+
+    def check_component(self, component: Component) -> Iterable[Unit]:
+        from weblate.trans.models import Unit
+
+        units = Unit.objects.filter(
+            translation__component=component,
```
nijel, line 180: The source check should operate on source strings only:

```suggestion
            translation__component=component,
            translation_language=component.source_language,
```

weblate/checks/source.py
gersona: oh that's right, I can't believe I missed that

weblate/checks/source.py
gersona: I does operate on source strings only, but gathers the states from the children units as well. 
Former code was
```python
states = list(unit.unit_set.values_list("state", flat=True))
total = len(states)
```

weblate/checks/source.py
```diff
@@ -6,19 +6,23 @@
 
 import re
 from collections import defaultdict
+from collections.abc import Iterable
```
nijel:
```suggestion
```

weblate/checks/tests/test_source_checks.py
```diff
@@ -52,18 +55,88 @@ class LongUntranslatedCheckTestCase(FixtureTestCase):
...
+            if not unit.is_source:
+                self.edit_unit(
+                    unit.source,
+                    unit.source,
+                    unit.translation.language.code,
+                    True,
+                    unit.translation,
+                )
+        source_unit = self.get_unit(language="en")
+        child_unit = unit.unit_set.exclude(pk=source_unit.id).first()
```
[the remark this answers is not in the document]
nijel: @gersona Using the loop variable outside of it seems confusing, shouldn't it use `source_unit` here?

weblate/checks/tests/test_source_checks.py
```diff
@@ -52,18 +55,88 @@ class LongUntranslatedCheckTestCase(FixtureTestCase):
...
```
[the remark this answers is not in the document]
nijel: ...and here?

nijel:

PR #16441, opened 2025-10-06
Add Vietnamese exceptions for consecutive duplicated words

IoeCmcomc:
This commit aims to add two Vietnamese words "luôn luôn" (always) and "song song" (parallel) to the ignore list of the duplicated words check.


nijel:
Can you please also add matching tests?

nijel:

PR #16491, opened 2025-10-09
feat(ci): test with free-threaded Python 3.14

nijel:


nijel:
At least psycopg is missing, see https://github.com/psycopg/psycopg/issues/1095.

PR #16519, opened 2025-10-10
fix(memory): Imported memory content marked as active

### LINKED ISSUE #16456: Imported translation memory should not be pending
### Describe the issue

The translation memory matches based on the imported file get a low score because of penalty for pending strings.

### I already tried

1. Go to translation memory.
2. Import TMX file.
3. See that all imported entries are marked as pending.

### Expected behavior

Imported translation memory content should be marked as active, not as pending.

### Screenshots

_No response_

### Exception traceback

```pytb

```

### How do you run Weblate?

Git checkout

### Weblate versions

_No response_

### Weblate deploy checks

```shell

```

### Additional context

https://github.com/orgs/WeblateOrg/discussions/16448

gersona:
- fixes #16456

### DIFF
--- a/weblate/memory/tests.py
+++ b/weblate/memory/tests.py
@@ -298,8 +298,34 @@ class MemoryModelTest(TransactionsTestMixin, FixtureTestCase):
 
     def test_import_po(self) -> None:
         """Test the import of an GNU PO file."""
-        self.import_file_with_languages_test(get_test_file("cs.po"), "en", "cs", 1)
-        memory = Memory.objects.first()
+        with tempfile.NamedTemporaryFile(suffix=".po") as temp_file:
+            temp_file.write(
+                rb"""
+msgid ""
+msgstr ""
+"Project-Id-Version: Weblate Hello World 2012\n"
+"Report-Msgid-Bugs-To: <noreply@example.net>\n"
+"POT-Creation-Date: 2012-03-14 15:54+0100\n"
+"PO-Revision-Date: 2013-08-25 15:23+0200\n"
+"Last-Translator: testuser <>\n"
+"Language-Team: Czech <http://example.com/projects/test/test/cs/>\n"
+"Language: cs\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=3; plural=(n==1) ? 0 : (n>=2 && n<=4) ? 1 : 2;\n"
+"X-Generator: Weblate 1.7-dev\n"
+
+#: main.c:11
+#, c-format
+msgctxt "Greeting"
+msgid "Hello, world!\n"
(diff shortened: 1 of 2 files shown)


gersona:
The different context penalty factor would still apply though. Is it something we also want to avoid for imported memory content ?
https://github.com/WeblateOrg/weblate/blob/f85e19c487e32a19af441fa14a8493f68e83bed8/weblate/memory/machine.py#L50-L51

nijel:
Good point, it's probably reasonable to avoid it if imported context is empty.

gersona:
I also noticed (but need a confirmation) that memory context doesn't seem to be loaded when importing from admin command. Is it a blind spot or something I don't fully understand?
https://github.com/WeblateOrg/weblate/blob/563737ddbf5b3c1c43cf0e80e3012cc077bbe396/weblate/memory/models.py#L261-L272

weblate/memory/management/commands/import_memory.py
```diff
@@ -54,6 +54,8 @@ def handle(self, *args, **options) -> None:
                         langmap,
                         source_language=options["source_language"],
                         target_language=options["target_language"],
+                        from_file=True,
```
nijel: I think `import_file` should add this flag. Right now it seems that only `import_tmx` does that.

weblate/memory/management/commands/import_memory.py
```diff
@@ -54,6 +54,8 @@ def handle(self, *args, **options) -> None:
...
+                        status=Memory.STATUS_ACTIVE,
```
nijel: The same here, this should be done by `import_file`. Leaving this to the caller will cause inconsistencies (you didn't update the file upload view with the args).

gersona:
Support for context in TMX file is still to be implemented though (see #9947)

weblate/trans/tests/data/cs.po
```diff
@@ -15,6 +15,7 @@ msgstr ""
 
 #: main.c:11
 #, c-format, max-length:100
+msgctxt "Greeting"
```
nijel: This breaks other test cases. Better to have a separate test file for this.

weblate/memory/tests.py
```diff
@@ -246,16 +292,20 @@ def test_import_xliff(self) -> None:
...
-        self.import_file_with_languages_test("ids-translated.xliff", "en", "cs", 2)
+        self.import_file_with_languages_test(
+            get_test_file("ids-translated.xliff"), "en", "cs", 2
+        )
 
     def test_import_po(self) -> None:
         """Test the import of an GNU PO file."""
-        self.import_file_with_languages_test("cs.po", "en", "cs", 1)
+        self.import_file_with_languages_test(get_test_file("cs.po"), "en", "cs", 1)
+        memory = Memory.objects.first()
```
nijel: In my experience, using `first()` is not reliable long-term; it's better to get the actual entry by get.

nijel:

PR #16568, opened 2025-10-15
fix: rescan cleans out changes when commit policy is enabled

### LINKED ISSUE #16458: Updating the source language cleans out unapproved pending translations
### Describe the issue

(related community discussion here: https://github.com/orgs/WeblateOrg/discussions/16392 and maybe related bug report here: #16227)
We've set up a [translation portal](https://translation.botc.app/) with English as the source language and a Github repository that has a webhook and push permissions for persisting approved changes to VCS. The project only includes approved translations in commits and we have 2 components set up to manage 2 separate translation projects that use the same Git repository. In case it matters: the [Game Content](https://translation.botc.app/projects/botc-app/game-content/) component contains the main VCS configuration, while the [App Interface](https://translation.botc.app/projects/botc-app/app-interface/) component is linked to the VCS configuration from the Game Content component.

The problem we're currently facing is that any change to the App Interface English source file automatically cleans out a lot of unapproved translations in other languages. Here's a good example for what happened: https://translation.botc.app/translate/botc-app/app-interface/da/?&offset=4#history
The affected string has been originally translated into Danish, but not approved yet. A completely unrelated change to the English source file suddenly cleaned out most of the Danish translation strings. I've manually reverted / fixed the clean-up and submitted the translations again, yet the next unrelated commit deleted those translations once more.

At no point was the Danish translation file on the VCS changed or touched, the commits only contained changes to the English source file. This cleaning of translation strings happened across a multitude of languages and thousands of completely unrelated strings: https://translation.botc.app/checks/translated/botc-app/app-interface/
https://translation.botc.app/changes/?page=158&limit=20&user=webhook%3Agithub

### I already tried

1. Set up a project with a Git VCS repository, Translation quality filter: Only include approved translations
2. Create 2 components, A & B from 2 English language files in the repository
3. Link component B to the VCS settings of component A
4. Create a new language for the project, Danish (this will create an empty da.json file for each component in the repository)
5. add a bunch of Danish translations and leave them at "needs to be reviewed"
6. add a new string to the *English* source file of component B on the VCS

### Expected behavior

The webhook should automatically add the new String to the English source language and allow it to be translated to Danish in the translation interface. Any previously submitted, unrelated translation strings for Danish should be kept in the "needs review" state.

### Screenshots

Git VCS setup
Image

A sample commit that triggered the mass purge
Image

What the history of a purged string in Danish looks like
Image

### Exception traceback

```pytb

```

### How do you run Weblate?

Docker container

### Weblate versions

 * Weblate: 5.13
 * Django: 5.2.5
 * siphashc: 2.5
 * translate-toolkit: 3.15.6
 * lxml: 6.0.0
 * pillow: 11.3.0
 * nh3: 0.3.0
 * python-dateutil: 2.9.0.post0
 * social-auth-core: 4.7.0
 * social-auth-app-django: 5.5.1
 * django-crispy-forms: 2.4
 * oauthlib: 3.3.1
 * django-compressor: 4.5.1
 * djangorestframework: 3.16.1
 * django-filter: 25.1
 * django-appconf: 1.1.0
 * user-agents: 2.2.0
 * filelock: 3.19.1
 * RapidFuzz: 3.13.0
 * openpyxl: 3.1.5
 * celery: 5.5.3
 * django-celery-beat: 2.8.1
 * kombu: 5.5.4
 * translation-finder: 2.23
 * weblate-language-data: 2025.8
 * html2text: 2025.4.15
 * pycairo: 1.28.0
 * PyGObject: 3.52.3
 * diff-match-patch: 20241021
 * requests: 2.32.4
 * django-redis: 6.0.0
 * hiredis: 3.2.1
 * sentry-sdk: 2.35.0
 * Cython: 3.1.3
 * mistletoe: 1.4.0
 * GitPython: 3.1.45
 * borgbackup: 1.4.1
 * pyparsing: 3.2.3
 * ahocorasick_rs: 0.22.2
 * charset-normalizer: 3.4.3
 * cyrtranslit: 1.1.1
 * drf-spectacular: 0.28.0
 * Python: 3.13.3
 * Git: 2.50.1
 * psycopg: 3.2.9
 * psycopg-binary: 3.2.9
 * phply: 1.2.6
 * ruamel.yaml: 0.18.14
 * tesserocr: 2.8.0
 * boto3: 1.40.10
 * aeidon: 1.15
 * iniparse: 0.5
 * mysqlclient: 2.2.7
 * google-cloud-translate: 3.21.1
 * openai: 1.99.9
 * Mercurial: 7.1
 * git-svn: 2.50.1
 * git-review: 2.5.0
 * Redis server: 8.2.0
 * PostgreSQL server: 17.6
 * Database backends: django.db.backends.postgresql
 * PostgreSQL implementation: psycopg3 (binary)
 * Cache backends: default:RedisCache, avatar:FileBasedCache
 * Email setup: django.core.mail.backends.smtp.EmailBackend: smtp.mandrillapp.com
 * OS encoding: filesystem=utf-8, default=utf-8
 * Celery: redis://cache:6379/1, redis://cache:6379/1, regular
 * Platform: Linux 6.8.0-84-generic (x86_64)

### Weblate deploy checks

```shell
System check identified some issues:

INFOS:
?: (weblate.I021) Error collection is not set up, it is highly recommended for production use
	HINT: https://docs.weblate.org/en/weblate-5.13/admin/install.html#collecting-errors
?: (weblate.I028) Backups are not configured, it is highly recommended for production use
	HINT: https://docs.weblate.org/en/weblate-5.13/admin/backup.html
?: (weblate.I031) New Weblate version is available, please upgrade to 5.13.3.
	HINT: https://docs.weblate.org/en/weblate-5.13/admin/upgrade.html

System check identified 3 issues (5 silenced).
```

### Additional context

_No response_

amCap1712:
Fixes #16458.

a) When a unit is translated, check if `disk_state` already exists in `details` of the `Unit`.
b) If not, write the status of the `Unit` before the translation to it and then create the `PendingUnitChange`.
c) If `disk_state` already exists in the `details`, then don't change it and just create a new `PendingUnitChange`.
d) During the committing of pending changes to disk, update the `disk_state` in the `details` column with the current state of the unit in the translation store. (As the the translation store is already open and available during commit, this should be fine performance wise.)
e) During rescan of files from the repository, we have access to the Unit and hence its details. Use it to compare the incoming VCS changes with the unflushed pending changes.
f) If changes are incompatible, delete the unflushed changes and reset the `disk_state` of the `Unit`.

### DIFF
--- a/weblate/trans/tests/test_remote.py
+++ b/weblate/trans/tests/test_remote.py
@@ -223,9 +223,11 @@ class MultiRepoTest(ViewTestCase):
         translation = self.component2.translation_set.get(language_code="cs")
         self.assertEqual(translation.stats.all, 1)
 
+
+class FileScanTest(ViewTestCase):
     def test_file_scan_commit_policy(self) -> None:
         """Test file scan does not discard pending changes blocked by commit policy."""
-        self.component2.delete()
+        request = self.get_request()
         self.project.commit_policy = CommitPolicyChoices.APPROVED_ONLY
         self.project.translation_review = True
         self.project.save()
@@ -299,9 +301,9 @@ class MultiRepoTest(ViewTestCase):
         ttk_unit.set_target(disk_target)
         translation.store.save()
         with transaction.atomic():
-            translation.git_commit(self.request.user, "TEST <test@example.net>")
+            translation.git_commit(request.user, "TEST <test@example.net>")
 
-        self.component.do_file_scan(self.request)
+        self.component.do_file_scan(request)
 
         # check that changes are not lost and disk state is retained for future
         unit_1 = Unit.objects.get(pk=unit_1.pk)


amCap1712:
Changes are incomplete, opened the PR to clarify a doubt but figured it while typing it out.

weblate/trans/models/unit.py
```diff
@@ -607,6 +607,52 @@ def store_old_unit(self, unit) -> None:
             "explanation": unit.explanation,
         }
 
+    def store_disk_state(self) -> None:
+        """
+        Store a snapshot of unit state before pending changes are created.
+
+        Only stores the state if it doesn't already exist (first pending change).
+        """
+        # TODO: ensure this function is called before all relevant PendingChange creations
```
amCap1712: Still working on this.

weblate/trans/models/unit.py
```diff
@@ -989,6 +1052,7 @@ def update_from_unit(  # noqa: C901,PLR0914
                 only_save=True,
                 update_fields=["location", "explanation", "note", "position"],
             )
+            # TODO: delete pending changes and details for the unit in that case ?
```
amCap1712: Should existing pending changes be discarded in this case? IIUC, this means unflushed changes to `explanation` will be lost.

weblate/trans/models/unit.py
```diff
@@ -1007,6 +1071,7 @@ def update_from_unit(  # noqa: C901,PLR0914
             same_content=same_source and same_target,
             run_checks=not same_source or not same_target or not same_state,
         )
+        # TODO: delete pending changes and details for the unit in that case ?
```
amCap1712: Should existing pending changes be discarded in this case always or only if the `pending` flag is true?

weblate/trans/models/unit.py
nijel: Well, this logic is wrong :smile:. It was written at a time when explanation was not editable. With editable explanations, it should no longer be considered metadata. So this should not delete pending changes, but the code path should not be triggered for explanation changes in formats that can edit it. Created https://github.com/WeblateOrg/weblate/issues/16571 to track this issue.

weblate/trans/models/unit.py
nijel: As the content has changed, the pending changes cannot be safely applied anymore, so they should be deleted here. The pending flag just marks that there is additional change on top of what was parsed from the file (string is marked for editing).

amCap1712:
> Proposal for dealing with this:
> 
>     1. Detect that the string was not actually changed in `Unit.update_from_unit` when the only differences are tracked `PendingUnitChange`.
> 
>     2. Discard related `PendingUnitChange` objects when the bilingual `Unit` was actually changed.
> 
>     3. When a monolingual string changes, keep the current pending translation and add a `PendingUnitChange` marking it fuzzy (if not already).

Are there any existing tests for these conditions to your knowledge? If not, I'll write up new ones. Also, any other cases that might need to tested?

nijel:
I don't think there are any tests for this. This could not happen before introduction of pending changes model and we didn't add tests back then.

weblate/trans/models/unit.py
```diff
@@ -709,7 +756,13 @@ def update_variants(self) -> None:
             else:
                 component.needs_variants_update = True
 
-    def get_unit_state(self, unit, flags: str | None, string_changed: bool = False):
+    def get_unit_state(
+        self,
+        unit,
+        flags: str | None,
```
nijel:
```suggestion
        flags: Flags | str | None,
```

(Not really a new issue, but I spotted it now)

weblate/trans/models/unit.py
```diff
@@ -730,14 +783,16 @@ def get_unit_state(self, unit, flags: str | None, string_changed: bool = False):
 
         # We need to keep approved/fuzzy state for formats which do not
         # support saving it
-        if unit.is_fuzzy(self.fuzzy and not string_changed):
+        is_existing_fuzzy_state = self.fuzzy or disk_unit_state == STATE_FUZZY
```
nijel, line 788: I'm not sure about this logic. The intention is to keep the fuzzy/approved flag for formats that do not support saving it. Why mix disk state to that? If the string was not changed, the existing state should be preserved without involving disk state. The same applies to the approved state below.

weblate/trans/models/translation.py
```diff
@@ -778,6 +784,30 @@ def _commit_pending(self, reason: str, user: User | None) -> bool:
...
+        )
+
+        units_to_actually_clear = (
+            units_to_clear_disk_state - units_with_remaining_changes
+        )
+
+        if units_to_actually_clear:
+            units_to_update = list(
+                Unit.objects.filter(
+                    id__in=units_to_actually_clear, details__has_key="disk_state"
```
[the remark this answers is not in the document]
nijel: I have no clue if Copilot is right here, but that just shows we need to test this code.

weblate/trans/models/unit.py
```diff
@@ -1007,6 +1071,9 @@ def update_from_unit(  # noqa: C901,PLR0914
             same_content=same_source and same_target,
             run_checks=not same_source or not same_target or not same_state,
         )
+        self.clear_disk_state()
+        PendingUnitChange.objects.filter(unit=self).delete()
```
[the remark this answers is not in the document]
nijel: This is wrong, the cleanup here is correct.

weblate/trans/models/unit.py
amCap1712: Assuming the commit policy is set to commit approved only, the uncommitted changes to the disk for that unit would be the ones that in translated state that are not yet approved.  So for such a unit that the states would be as follows:
1. `disk_state` would have `STATE_APPROVED` since the last flushed translation would have been approved.
2.  `Unit.state` would be `STATE_TRANSLATED` as there are changes pending approval for the unit in the database.
3. The unit read from incoming VCS changes would be `STATE_TRANSLATED` if the format doesn't support storing the approved/fuzzy state.

Without any changes to the above code, because of 2 and 3 `get_unit_state` returns `STATE_TRANSLATED` whereas the `disk_state` has `STATE_APPROVED`. The `same_state` check fails and the changes in Weblate are discarded.

https://github.com/WeblateOrg/weblate/blob/e88b0f36b66a702ba4f705bacd6b92c361091967/weblate/trans/models/unit.py#L1010

I didn't see `get_unit_state` being used outside `update_from_unit` so I thought this would be a fine place to update the code but I can also move the changes outside here to `same_state`.

(FWIW, this is the doubt that I was initially confused about so would be nice if you could verify my thinking is correct. This change was needed to make things work in my local testing.)

nijel: But this would set the state to approved in this case when there is content change in the file just because the previous disk state is approved.

For the case when the content is not changed, the behavior you describe is indeed an issue, but to avoid changes in other situations, it needs to be handled differently.

The solution might be to the pass `disk_unit_state` only when `same_target` is true, but I still might be missing some cases.

amCap1712: `string_changed`  considers `same_target` and that is `AND`ed to the check in the next line but I'll try to simplfy and double check and test these conditions.

amCap1712:
While working on tests, came across another bug. This needs to use `pending_change.state` instead of `unit.state` to correctly write the state to disk if there are unapproved changes.
https://github.com/WeblateOrg/weblate/blob/30411ce60825d25c075e4c02b339830f4dabb539/weblate/trans/models/translation.py#L1039

Should I open a separate PR to fix this and add a test or just do it in this one?

weblate/trans/models/unit.py
nijel: Right, then it's probably correct, but I still find it weird that *both* `is_existing_fuzzy_state` and `is_existing_approved_state` could be true—one because of `disk_unit_state` and one because of `Unit.state`.

nijel:
https://github.com/WeblateOrg/weblate/pull/16595 is the pull request addressing issue that was described here, right?

weblate/trans/models/unit.py
amCap1712: True :/.  I am wondering if we can remove the comparison with `Unit.state` here.

For `Unit.state = STATE_APPROVED` it is not an issue with current commit policies because those changes will always be flushed before file scan is done. However, for `STATE_FUZZY` there can be issues already. If a unit was translated and then later changed to needs-editing, removing the comparison with `Unit.state` here would result in losing that in Weblate IIUC.

(FWIW, that is already a separate issue that I realized while working on this. If an existing translation is marked as needs-editing, and the commit policy is skip those changes we don't do anything. Maybe we should untranslate the string on the disk?)

amCap1712:
>https://github.com/WeblateOrg/weblate/pull/16595 is the pull request addressing issue that was described here, right?

Yes.

nijel (approved):

weblate/trans/models/unit.py
nijel: I don't think we should change the file content in this case. Let's keep it this way.

nijel:

PR #16612, opened 2025-10-17
fix(unit): do not trigger enforced checks when string was not saved

nijel:
Fixes WEBLATE-373Q


nijel:
Actually adding tests discovered that this patch is wrong—it would skip enforcing the check in this case.

amCap1712:
Ah! Nice catch :D

PR #16644, opened 2025-10-21
chore(deps): bump weblate-schemas to 2025.6

nijel:


nijel:
https://github.com/WeblateOrg/weblate/pull/16645 instead.

PR #16746, opened 2025-10-28
fix(auth): correct SAML certificate field typo

M0Rf30:
Fix typo in Docker SAML configuration where 'x508cert' was used instead of 'x509cert', causing KeyError: 'IDP must contain x509cert or x509certMulti' when authenticating via SAML.

This resolves the SAML authentication regression introduced in Weblate 5.14 where the certificate field name was misspelled.


nijel:
There turned out to be more issues, https://github.com/WeblateOrg/weblate/pull/16780 addresses this with tests.

PR #16765, opened 2025-10-30
feat(api): show admins of group

michael-smt:
This makes the [team administrators](https://docs.weblate.org/en/latest/admin/access.html#team-administrators) of a group available via API.

I would like to have this data available to be able to automatically create a overview table of who is admin in each team.


michael-smt:
Not sure about the access rights needed to view this data...

nijel:

There should be no additional permissions needed, seeing admin of a group I'm member of seems reasonable for regular users.

PR #16781, opened 2025-11-01
fix(accounts): exclude other user events from logging detailed info

nijel:
This avoids potential information leak of admins.

### DIFF
--- a/weblate/accounts/models.py
+++ b/weblate/accounts/models.py
@@ -406,7 +406,8 @@ class AuditLogManager(models.Manager):
         user_agent: str = ""
         # Log only address for own actions (unauthenticated or when the request user matches audit user)
         if request and (
-            not hasattr(request, "user") or not request.user
+            not hasattr(request, "user")
+            or not request.user
             or not request.user.is_authenticated
             or request.user == user
         ):


weblate/accounts/models.py
```diff
@@ -400,10 +400,14 @@ def is_new_login(self, user: User, address, user_agent) -> bool:
...
     def create(  # type: ignore[override]
-        self, user: User, request: HttpRequest | None, activity, **params
+        self, user: User, request: HttpRequest | None, activity: str, **params
     ):
-        address = get_ip_address(request)
-        user_agent = get_user_agent(request)
+        address: str | None = None
+        user_agent: str = ""
+        # Log only address for own actions (unauthenticated or when the request user matches audit user)
+        if request and (not request.user.is_authenticated or request.user == user):
```
nijel:
```suggestion
        if request and (not getattr(request, "user") or not request.user.is_authenticated or request.user == user):
```

weblate/accounts/models.py
```diff
@@ -400,10 +400,18 @@ def is_new_login(self, user: User, address, user_agent) -> bool:
...
-        self, user: User, request: HttpRequest | None, activity, **params
+        self, user: User, request: HttpRequest | None, activity: str, **params
     ):
-        address = get_ip_address(request)
-        user_agent = get_user_agent(request)
+        address: str | None = None
+        user_agent: str = ""
+        # Log only address for own actions (unauthenticated or when the request user matches audit user)
+        if request and (
+            not request.user
```
nijel:
```suggestion
            not hasattr(request, "user") or not request.user
```

PR #16794, opened 2025-11-03
Change Weblate installation command for yunohost in docs

DeMiro5001:
Apps in the yunohost catalog are installed via their names directly. Using the repo url is for installing apps that aren't in the catalog or installing testing branch (for beta testers)


nijel:

PR #16798, opened 2025-11-03
fix(update from source): include `original_state` in comparison state

amCap1712:
Include `original_state` in comparison_state for checking during `update_from_unit` when updating weblate with incoming changes from repository.


weblate/trans/tests/test_remote.py
```diff
@@ -311,7 +329,12 @@ def test_file_scan_commit_policy(self) -> None:
         self.assertEqual(unit_1.state, STATE_TRANSLATED)
```
amCap1712, line 329: Test fails at L328 because the change is overwritten during file scan. The unit does not match because `original_state` read from the incoming unit in the repository is `STATE_TRANSLATED` whereas the in `unit.details["disk_state"]` it is `STATE_APPROVED`.

amCap1712:
closed in favor of #16805

PR #16829, opened 2025-11-06
fix(glossary): do a full preload when adding

nijel:
The terms are then rendered and when not using full reload the additional fields need to be fetched. This hits some bug insude Django with prefetch and only() lookups.

Fixes WEBLATE-37PQ


nijel:
https://code.djangoproject.com/ticket/36713 describes the issue in Django.

PR #16867, opened 2025-11-10
feat: Add ability to disable admin contact form

### LINKED ISSUE #16864: Disable admin contact form
### Describe the problem

I would like a settings/environment variable that disables the admin contact form to avoid exposing a vector for spam and abuse.

### Solution brainstorm

Add a `WEBLATE_ENABLE_ADMIN_CONTACT_FORM` environment variable which defaults to true for backwards compatibility.

_No response_

### Screenshots

_No response_

### Additional context

This was requested in https://github.com/WeblateOrg/weblate/issues/9479 but closed as a duplicate of https://github.com/WeblateOrg/weblate/issues/9353, which I don't think is the case. I don't just want to restrict the form to authenticated users. I want to _completely_ disable the form.

TudorGR:
Allow administrators to completely disable the contact form by setting CONTACT_FORM to 'disabled', preventing spam and abuse vectors.

This extends the existing CONTACT_FORM setting with a new option instead of introducing a separate configuration variable.

Closes #16864

## Changes

- **`weblate/accounts/views.py`**: Modified `contact()` view to check if `CONTACT_FORM == "disabled"` and raise `Http404` when disabled
- **`weblate/accounts/tests/test_views.py`**: Added `test_contact_disabled()` to verify both GET and POST requests return 404
- **`docs/admin/config.rst`**: Updated documentation for `CONTACT_FORM` setting to include the new `"disabled"` option
- **`docs/changes.rst`**: Added changelog entry under "New features" for Weblate 5.15

## Implementation Details

- **Minimal changes**: Only 3 lines added to the main view function
- **Backward compatible**: Default behavior remains `"reply-to"`
- **No new settings**: Uses existing `CONTACT_FORM` setting as suggested by @nijel
- **Docker support**: Works with existing `WEBLATE_CONTACT_FORM` environment variable
- **Comprehensive testing**: Tests for both GET and POST requests

## Usage

Administrators can disable the contact form by adding to `settings.py`:
```python
CONTACT_FORM = "disabled"
```

Or in Docker:
```
WEBLATE_CONTACT_FORM=disabled
```

When disabled, any attempts to access the contact form will result in a 404 error.


nijel (changes requested):
Disabling the form should also hide the links to it.

nijel:

PR #16870, opened 2025-11-10
fix(memory): exclude blank context from matching

nijel:
The historical translation memory entries do not store context and are now suddenly 95% matches instead of 100% ones. This is confusing to users, as nothing has changed and the automatic translation with a 100 quality threshold no longer works.

The downside of this is that we ignore the actual blank context when translating strings with a non-blank context.

@gersona What do you think about this approach? I'm not sure if it will work better long-term, but for the existing entries, it might work better.


gersona:
I think it is good enough to address the issue. 
Long term, we could potentially introduce a parameter to control how we handle memories without context.

Issue #16869, opened 2025-11-10
Disable "must be logged in to translate" message


sloria:
### Describe the problem

My project's translation workflow relies on translators making suggestions only and therefore should not encourage translators to register and log in. So this banner can lead to misunderstanding for my collaborators

<img width="2134" height="238" alt="Image" src="https://github.com/user-attachments/assets/a5fdbc99-2d3f-4b3b-8ac3-95ffbbbb91bc" />

It would be nice if this could be hidden.

### Solution brainstorm

Add a setting/environment variable that controls the visibility of the alert here:

https://github.com/WeblateOrg/weblate/blob/2ca596de52c6ea5b297b4b9514812854a0e4b712/weblate/templates/dashboard/anonymous.html#L7-L10

### Describe alternatives you have considered

Could also change the text of the warning to "You must be signed in to save translations."

nijel:
Adding settings is something most admins will not do, so I'd like to avoid that. Using more general text as you suggest is definitely a practical approach.

sloria:
Updating the copy seems like a good first step 👍 . And understandable re: adding yet another setting.

If adding a setting isn't in the cards, then I'd argue to remove the alert altogether. It feels a bit like "overonboarding". And for workflows like mine where contributors will never be logged in, it's distracting UI that can never be dismissed.

nijel:
The onboarding is intentional here with the expectation that regular contributors will authenticate and thus not see this again. If your workflow doesn't expect authentication, the alert is indeed annoying.

sloria:
I'd argue that regular contributors will already know to register, either through their projects' onboarding docs or just implicitly given that auth-to-contribute is such a common UX pattern.

sloria:
If you want to nudge people towards authenticating, then IMO it should be done at the time it's relevant, e.g. by adding a tooltip to the disabled "Save translation" UI explaining that the user needs to log in.

nijel:
I don't have a strong opinion on this. We do have contextual tooltips on the disabled save button to guide users to authenticate, so it might no longer be necessary. In the end this is something we're carrying for 13 years since f35de82bf1b6aa90a2ee9b8c10bd127d0c5bbb50 🙂.

sloria:
Understandable! Given that there are already tooltips, and it's been 13 more years for users to get accustomed to auth conventions, maybe it's time to remove it? If I send a PR for this, would you consider merging it?

PR #16874, opened 2025-11-11
feat(ui): remove "must be logged in to translate" alert

sloria:
Closes https://github.com/WeblateOrg/weblate/issues/16869


sloria:
The translation files need to be updated to remove the string, but I couldn't figure out the right incantation to get that to happen 🤔

nijel:

PR #16875, opened 2025-11-11
feat(ui): don't style "sign in" and "lock" messages like errors

sloria:
Being unauthenticated is not an error state [^1], nor is a locked translation, so the messages should not be styled like errors.

**Before**

Screenshot 2025-11-10 at 21 06 30@2x

**After**

Screenshot 2025-11-10 at 21 05 46@2x

[^1]: As described in https://github.com/WeblateOrg/weblate/issues/16869, my project's translation workflow _encourages_ translators to be unauthenticated

### DIFF
--- a/weblate/templates/translate.html
+++ b/weblate/templates/translate.html
@@ -186,9 +186,9 @@
             </div>
             <div class="panel-footer">
               {% if locked %}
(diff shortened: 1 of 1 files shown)


nijel:

PR #16876, opened 2025-11-11
feat(ui): de-emphasize "Skip" button

sloria:
Improve the visual hierarchy of the translate buttons by de-emphasizing the "skip" action.

**Before**

before screenshot

**After**
after screenshot

### DIFF
--- a/weblate/templates/translate.html
+++ b/weblate/templates/translate.html
@@ -259,7 +259,7 @@
                     {% icon "suggest.svg" %} {% translate "Suggest" %}
                   </button>
                 {% endif %}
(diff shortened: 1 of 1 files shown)


nijel:
Mixed color makes it look weird. Maybe using `btn-link` instead would work better?

sloria:
Hm, `btn-link` doesn't seem to make a difference. Unfortunately won't have time to look into this today, so feel free to make updates as you see fit.

nijel:
Let's merge this as is for now, we can tweak the CSS later (once #16882 is merged).

PR #16894, opened 2025-11-11
feat: Basic structure for LLM machinery + Initial Ollama machinery tools for local Ollama AI model integration

urtzai:
This PR integrates [Ollama ](https://ollama.com/) to use local AI models (and recently added Ollama cloud models) in translation machinery. 

For this first approach, it is necessary to install Ollama locally because API key settings and integrations are not yet implemented.

All code bases were copied from OpenAI machinery due to its similarity.

No official Ollama Python library was used for simplicity and to avoid dependency issues. 

Here are some screenshots:

### DIFF
--- a/weblate/machinery/forms.py
+++ b/weblate/machinery/forms.py
@@ -353,7 +353,8 @@ class DeepLMachineryForm(KeyURLMachineryForm):
 
 class LLMBasicMachineryForm(BaseMachineryForm):
     base_url = WeblateServiceURLField(
-        label=pgettext_lazy("Automatic suggestion service configuration", "API URL"), required=False,
+        label=pgettext_lazy("Automatic suggestion service configuration", "API URL"),
+        required=False,
     )
     model = forms.CharField(
         label=pgettext_lazy(


urtzai:
Hi nigel, 

Makes all the sense to me. I agree that it's better to avoid third-party libraries if we can keep things simple and reduce the number of dependencies. Maintaining 6-10 libraries could become quite cumbersome. 

If you're comfortable with it, I think closing this PR is a good idea. I'm happy to help with a new approach for LLMs. Using OpenAI integration as a base with Ollama as a starting point would be easiest for me because I don't have an OpenAI account or experience with testing yet.

Let me know what you think

nijel:
It's not needed to implement everything at once, it was merely mentioned where we might want to go in the future. 

Having a base class for LLMs that would be extracted from the current OpenAI class and used in Ollama class to avoid copying the logic would be a good starting point in this PR.

urtzai:
Hi nijel, 

Take a look at these changes. I reused the code from OpenAI implementation and abstracted two main customizable methods to be implemented on each LLM case (the third one is `get_model` which was already implemented): 

```python
def translation_split(self, text: str) -> list[str]:
    """Splits the input text into individual translations based on a separator."""
    return SEPARATOR_RE.split(text)

def fetch_llm_translations(self, prompt: str, content: str) -> str:
    """Fetches translations from an LLM instance using the provided prompt and content."""
    return ""
```

Let me know what you think.

nijel:
Looks good at first sight. Can you please add tests for the Ollama integration?

urtzai:
Tests and doc added. Let me know if you need something else.

nijel:
Thanks, I think it's good to merge given the tests pass. Adding a changelog entry would be great.

weblate/machinery/tests.py
```diff
@@ -2689,6 +2690,163 @@ def mock_response(self) -> None:
...
+
+    def mock_error(self) -> NoReturn:
+        self.skipTest("Not tested")
+
+    def mock_response(self) -> None:
+        respx.post(
+            "http://localhost:11434/api/generate",
+        ).mock(
+            httpx.Response(
+                200,
```
```suggestion
        responses.add(
                responses.POST,
                "http://localhost:11434/api/generate",
                status=200,
```

weblate/machinery/tests.py
```diff
@@ -2689,6 +2690,163 @@ def mock_response(self) -> None:
...
+        "persona": "",
+        "style": "",
+    }
+
+    def mock_response(self) -> None:
+        respx.post(
+            "http://localhost:11434/api/generate",
+        ).mock(
+            httpx.Response(
+                200,
```
```suggestion
        responses.add(
                responses.POST,
                "http://localhost:11434/api/generate",
                status=200,        
```

weblate/machinery/tests.py
```diff
@@ -2689,6 +2690,163 @@ def mock_response(self) -> None:
...
+                        98764,
+                    ],
+                    "total_duration": 3946971317,
+                    "load_duration": 3325185239,
+                    "prompt_eval_count": 73,
+                    "prompt_eval_duration": 107465065,
+                    "eval_count": 11,
+                    "eval_duration": 503286987,
+                },
+            )
```
```suggestion
```

weblate/machinery/tests.py
```diff
@@ -2689,6 +2690,163 @@ def mock_response(self) -> None:
...
+                    "done": True,
+                    "done_reason": "stop",
+                    "total_duration": 5740856828,
+                    "prompt_eval_count": 63,
+                    "eval_count": 481,
+                },
+            )
```
```suggestion
```

weblate/machinery/forms.py
```diff
@@ -457,3 +457,51 @@ class AzureOpenAIMachineryForm(BaseOpenAIMachineryForm):
...
+        label=pgettext_lazy(
+            "Automatic suggestion service configuration",
+            "Translator style",
+        ),
+        widget=forms.Textarea,
+        help_text=gettext_lazy(
+            "Describe the style of translation. For example: “Use informal language.”"
+        ),
+        required=False,
+    )
```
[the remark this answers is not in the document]
nijel: Makes sense to reuse the form definitions as well.

weblate/machinery/ollama.py
```diff
@@ -0,0 +1,49 @@
...
+            payload = {
+                "model": self.get_model(),
+                "system": prompt,
+                "prompt": content,
+                "stream": False,
+            }
+            api_url = urljoin(self.settings["base_url"], self.end_point)
+            response = requests.request(
+                "post", api_url, json=payload, timeout=self.request_timeout
+            )
```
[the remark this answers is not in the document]
nijel: Good point, it should use it for consistency.

weblate/machinery/ollama.py
```diff
@@ -0,0 +1,49 @@
...
+        except Exception as error:
+            raise MachineTranslationError(error) from error
+
+        return response.json().get("response") or None
```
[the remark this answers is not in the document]
nijel: Using `self.request` instead will indded address this.

weblate/machinery/forms.py
```diff
@@ -351,7 +351,16 @@ class DeepLMachineryForm(KeyURLMachineryForm):
     )
 
 
-class BaseOpenAIMachineryForm(KeyMachineryForm):
+class LLMBasicMachineryForm(BaseMachineryForm):
+    base_url = WeblateServiceURLField(
+        label=pgettext_lazy("Automatic suggestion service configuration", "API URL"),
```
nijel, line 356:
```suggestion
        label=pgettext_lazy("Automatic suggestion service configuration", "API URL"), required=False,
```

weblate/machinery/forms.py
```diff
@@ -351,7 +351,16 @@ class DeepLMachineryForm(KeyURLMachineryForm):
...
+    )
+    model = forms.CharField(
+        label=pgettext_lazy(
+            "Automatic suggestion service configuration",
+            "LLM model",
+        ),
```
nijel, line 363:
```suggestion
        ),
        required=False,
```

nijel:

nijel:
https://github.com/WeblateOrg/weblate/pull/17146 rebases OpenAI on top of generic LLM.

urtzai:
Hardest PR ever :rofl: 

Thanks @nijel I learned a lot.

PR #16897, opened 2025-11-12
Add Modern MT support for sr_Latn language code

datenspast:
This issue is about the ModernMT compatibility.

Weblate uses the language code "sr_Latn", ModernMT uses "sr-Latn".
This leads to issues with auto suggestions.

An additional mapping in `weblate/machinery/modernmt.py` should fix the issue.

Thanks for keeping up the great work!


nijel:
This type of conversion happens in `map_language_code` just below, why do you think this is needed?

datenspast:
> This type of conversion happens in `map_language_code` just below, why do you think this is needed?

Yes, I have not seen this. Thanks a lot!

PR #16905, opened 2025-11-12
feat(ui): remove suggest icon from button

sloria:
The icon for "Suggest" makes it look like a destructive action, and is also inconsistent with the "Save" buttons, which don't have an icon. This removes the icon.

Before:

before

After:

after


nijel:
The icon is used in other places as well.

sloria:
Ah, you're right...added back the icon so those other usages aren't affected

nijel:

PR #16960, opened 2025-11-17
Fix inconsistencies in _FILE environment based configuration

ffried:
Currently, [the docs state](https://docs.weblate.org/en/weblate-5.11.4/admin/install/docker.html#passing-secrets) that Docker secrets (and thus also configs) can be used for configuring Weblate. However, this is only partially true. E.g. `WEBLATE_GITHUB_TOKEN` cannot be configured using `WEBLATE_GITHUB_TOKEN_FILE` because the helpers are inconsistent (both in itself and in their usage).
This PR fixes this in two steps:

1. Use the `get_env_str` helper also in the other `get_env_*` helpers, to make sure all env vars are potentially read from a file.
2. Fix remaining `os.environ` usages in `settings_docker.py`. They all use the helpers now.

Note that I had to add a couple of `*_or_none` overloads for places that want an optional value. If needed, I can consolidate those with the original method and adding `None` as potential return type.

### DIFF
--- a/weblate/utils/environment.py
+++ b/weblate/utils/environment.py
@@ -72,6 +72,8 @@ def get_env_map_or_none(name: str) -> dict[str, str] | None:
     """
     parsed_list = get_env_list_or_none(name)
     if parsed_list is not None:
+        if parsed_list == [""]:
+            return {}
         return dict(e.split(":") for e in parsed_list)
     return None
 
--- a/weblate/utils/tests/test_environment.py
+++ b/weblate/utils/tests/test_environment.py
@@ -144,7 +144,7 @@ class EnvTest(SimpleTestCase):
         os.environ["TEST_DATA"] = "foo"
         self.assertEqual(get_env_list("TEST_DATA"), ["foo"])
         os.environ["TEST_DATA"] = ""
-        self.assertEqual(get_env_list("TEST_DATA"), [])
+        self.assertEqual(get_env_list("TEST_DATA"), [""])
         del os.environ["TEST_DATA"]
         self.assertEqual(get_env_list("TEST_DATA"), [])
         self.assertEqual(get_env_list("TEST_DATA", ["x"]), ["x"])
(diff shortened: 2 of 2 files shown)


weblate/settings_docker.py
```diff
@@ -272,7 +273,7 @@
 OTP_WEBAUTHN_ALLOW_PASSWORDLESS_LOGIN = False
 OTP_WEBAUTHN_HELPER_CLASS = "weblate.accounts.utils.WeblateWebAuthnHelper"
 
-if "WEBLATE_NO_EMAIL_AUTH" not in os.environ:
+if not get_env_str("WEBLATE_NO_EMAIL_AUTH"):
```
nijel, line 278: Boolean would be more appropriate here:

```suggestion
if not get_env_bool("WEBLATE_NO_EMAIL_AUTH"):
```

weblate/utils/environment.py
```diff
@@ -39,11 +39,27 @@ def get_env_str(
     return result
 
 
+def get_env_list_or_none(name: str) -> list[str] | None:
+    """Get list from environment."""
+    if string_value := get_env_str(name):
+        return string_value.split(",")
+    return None
```
nijel, line 56: This changes behavior. Before, `WEBLATE_LICENSE_FILTER=""` would translate to an empty list; now it translates to `None`.

weblate/settings_docker.py
ffried: Correct, but it changes the behavior.

nijel: Good point, even the documentation states "Disables e-mail authentication when set to any value".

weblate/utils/environment.py
ffried: I'm not sure about that. `get_env_string` should return `""` in this case (since `os.environ.get` should return that) which would still result in an empty list. Only if `WEBLATE_LICENSE_FILTER` is unset, it would result in `None`.

ffried: Ah, you're right. The empty string fails the if check. I'll adjust it.

ffried: The same problem existed for empty lists / dicts. I've adjusted the handling to be the same as before, using temporary variables.

ffried:
@nijel Done. While doing so, I've noticed that the fallback variable name wasn't respecting files either. That's now also fixed (including tests).

nijel:

PR #16882, opened 2025-11-11
Finish migrating frontend to Bootstrap 5

KarenKonou:
- [x]  Remove Bootstrap 3
- [x]  Migrate any remaining templates
- [x]  Check JS and make fixes if necessary
- [x] Check templates with dark theme

PR #16972, opened 2025-11-18
feat(state): use needs-checking/needs-rewriting states

amCap1712:
Use `STATE_NEEDS_CHECKING` and `STATE_NEEDS_REWRITING` where possible instead of `STATE_FUZZY`.

Fixes https://github.com/WeblateOrg/weblate/issues/2240

### DIFF
--- a/weblate/trans/autotranslate.py
+++ b/weblate/trans/autotranslate.py
@@ -95,7 +95,6 @@ class AutoTranslate(BaseAutoTranslate):
         self.progress_base = 0
         self.target_state = STATE_TRANSLATED
         if self.mode == "fuzzy":
-            # TODO: decide which fuzzy state to use?
             self.target_state = STATE_FUZZY
         elif self.mode == "approved" and translation.enable_review:
             self.target_state = STATE_APPROVED
--- a/weblate/trans/models/translation.py
+++ b/weblate/trans/models/translation.py
@@ -1015,7 +1015,6 @@ class Translation(
                         project=self.component.project,
                         skip_sentry=True,
                     )
-                    # TODO: decide which fuzzy state to use?
                     unit.state = STATE_FUZZY
                     # Use update instead of hitting expensive save()
                     Unit.objects.filter(pk=unit.pk).update(state=STATE_FUZZY)
@@ -1243,7 +1242,6 @@ class Translation(
 
             state = STATE_TRANSLATED
             if add_fuzzy or set_fuzzy:
-                # TODO: decide fuzzy state?
                 state = STATE_FUZZY
             elif add_approve:
                 state = STATE_APPROVED


weblate/addons/flags.py
```diff
@@ -53,7 +57,8 @@ def unit_pre_create(self, unit: Unit, activity_log_id: int | None = None) -> Non
             and unit.state >= STATE_TRANSLATED
             and not unit.readonly
         ):
-            unit.state = STATE_FUZZY
+            # TODO: needs-checking or needs-rewriting as this is source?
+            unit.state = STATE_NEEDS_CHECKING
```
amCap1712, line 64: @nijel

weblate/trans/models/comment.py
```diff
@@ -68,10 +68,11 @@ def add(
         if scope == "report":
             if component.has_template():
                 if unit_scope.translated and not unit_scope.readonly:
+                    # TODO: needs-checking for target and needs-rewriting for source?
```
amCap1712: @nijel

weblate/addons/flags.py
nijel: Needs checking seems appropriate here.

weblate/trans/models/comment.py
nijel: Needs checking for source, the target should get needs rewriting once the source is edited.

amCap1712:
image

1. Wondering what needs states should be shown here. Since source strings can be marked as needs checking and target strings as needs rewriting, should the needs checking/rewriting be hidden when its not possible to apply them? Or should the user always be given the option to choose needs editing and then the backend chooses the right state to apply?
2. What about needs editing? IIUC, it can only come from gettext's fuzzy or possibly a failed pending change but the user cannot manually mark a unit as STATE_FUZZY. When should `needs editing` be shown to the user? Always or only when a unit is already in STATE_FUZZY.
3. If a unit is marked as needs checking/rewriting should the info be written to the disk as fuzzy state?

nijel:
The ultimate goal here is to be able to distinguish what triggered the change from the state and what is the desired action. On the other side, we would rather not expose three similarly named states for selection to the translator. What about using one state selection with another input for detailed state, like outlined in https://github.com/WeblateOrg/weblate/issues/2240#issuecomment-424369487?

In the end, what we're trying to represent is a subset of what `state` and `state-qualifier` from XLIFF can (see https://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html#state). But the two-field is confusing as well because most of the combinations do not make sense. So they removed this in XLIFF 2 (there is subState, which is user-defined and not subject to the specification).

- Flag add-ons -> needs checking
- Report source string -> needs checking
- Source string changed in monolingual file (detected by Weblate) -> needs rewriting
- GNU gettext fuzzy -> needs rewriting
- XLIFF states might need to be investigated as well

That makes me think whether the names we currently have are clear enough to understand...

The states, IMHO, should be written to file when possible. 

PS: Beware that any UI changes here will most likely run into conflicts with #16882.

amCap1712:
>  Flag add-ons -> needs checking
    Report source string -> needs checking
    Source string changed in monolingual file (detected by Weblate) -> needs rewriting
    GNU gettext fuzzy -> needs rewriting
    XLIFF states might need to be investigated as well

Do I understand it correctly there won't be any cases for `needs-editing` then or the cases not mentioned here are `needs-editing`?

>But the two-field is confusing as well because most of the combinations do not make sense. 
We could the hide the subfield when the state is set to a non fuzzy state? Then when a user chooses, `Needs editing` they can be shown another field that lets them specify `Needs checking` or `Needs rewriting`. But I am unsure if this is good UX :/

I took a look at Xliff states but I think we should avoid adding more granular states until users request.

nijel:
"Needs editing" is more a superset of "Needs checking" and "Needs rewriting". I'm not actually convinced we need to expose all of these in the UI.

amCap1712:
Hmm I see. I am a bit confused. Do we only want to expose "needs editing" to the user and track all states internally?

nijel:
I think https://github.com/WeblateOrg/weblate/issues/2240#issuecomment-424369487 might be the best approach—it will show a detailed state, but the user should only select from a subset of the states. I don't really have a strong opinion here besides the need for more detailed states :-).

weblate/trans/autotranslate.py
```diff
@@ -95,6 +95,7 @@ def __init__(self, *, translation: Translation, **kwargs) -> None:
         self.progress_base = 0
         self.target_state = STATE_TRANSLATED
         if self.mode == "fuzzy":
+            # TODO: decide which fuzzy state to use?
```
amCap1712: This is the AutoTranslate addon, in name `needs checking` seems more appropriate here but `needs rewriting` would be more consistent with other addons.

weblate/trans/models/translation.py
```diff
@@ -1242,6 +1243,7 @@ def merge_translations(
 
             state = STATE_TRANSLATED
             if add_fuzzy or set_fuzzy:
+                # TODO: decide fuzzy state?
```
amCap1712: `needs rewriting` but unsure.

weblate/trans/models/translation.py
```diff
@@ -1015,6 +1015,7 @@ def update_units(
                         project=self.component.project,
                         skip_sentry=True,
                     )
+                    # TODO: decide which fuzzy state to use?
```
amCap1712: I don't think there is any good option for this one, as the unit itself is missing from the file.

amCap1712:
@nijel, in doubt about the three comments but otherwise the PR seems ready to me.

I will take a look at the docs changes in the other open PR.

weblate/trans/models/translation.py
nijel: Indeed, this is an error state. I think it can stay as is until we want to introduce a specific state for this.

weblate/trans/models/translation.py
nijel: I think it can stay as is. This is result of manual user action (selected in the upload form).

weblate/trans/autotranslate.py
nijel: I think this can stay as well. This is result of user action who selected automatic translation to be added this way.

nijel:
Thanks, I will think over the weekend whether to push this to 5.15.2 (because the feature is unfinished there) or to 5.16 (becaues these are bigger changes).

PR #17014, opened 2025-11-21
feat(add-ons): validate webhook secret length

nijel:
- validate secret length to make sure it is specificiation compliant
- reimplement signing to avoid runtime dependency as it pulls out too many additional dependencies (including type checking ones)


gersona (approved):
This looks good to me

PR #17037, opened 2025-11-23
feat(unit): replace auto-translated label with dedicated boolean field

### LINKED ISSUE #15762: Use flag instead of Automatically translated label
### Describe the issue

Prerequisites: Glossary Component with untanslatable strings and the "Fill read-only strings with source" add-on installed

- After the add-on has done its job a yellow label is shown in the translation editor.
- When I open the edit dialog, the label-checkbox is NOT checked - but the lablle is shown
- When I check the lable and save there's no change, the  label is shown
  in the dialog, the checkbox is shown checked now
- When I uncheck then and save -> the label is no longer shown

https://github.com/user-attachments/assets/cabee335-8fab-4e09-80f7-c31999e908fc

amCap1712:
Replace the ManyToMany label-based tracking of automatically translated units with a dedicated `automatically_translated` boolean field on the Unit model. This improves query performance when searching for automatically translated units and provides a cleaner data model. Add `is:automatically_translated` search query support and expose the field in REST API. The auto-translated status is shown as a check in the UI.

Fixes #15762.


docs/user/search.rst
```diff
@@ -69,6 +69,8 @@ Fields
       Untranslated strings, same as ``state:<translated``.
    ``pending``
       Pending strings not yet committed to the file (see :ref:`lazy-commit`).
+   ``automatically_translated``
```
nijel: We typically use - in the search queries, so let's use it here as well for consistency:

```suggestion
   ``automatically-translated``
```

Supporting both in the queries and documenting only `-` version might be even better.

weblate/templates/translate.html
```diff
@@ -676,6 +676,16 @@ <h5 class="list-group-item-heading">
                 </div>
               {% endif %}
 
+              {% if unit.automatically_translated %}
+                <div class="list-group-item check check-item">
+                  <h5>
+                    <span class="green">{% icon "check.svg" %}</span>
+                    {% translate "Automatically translated" %}
+                  </h5>
+                  <p class="list-group-item-text check-description">{% translate "This string was translated automatically." %}</p>
```
nijel: Perhaps we could look into the last change (`Unit.get_last_content_change()` should work well) and mention the author (service) that did the automatic translation.

weblate/trans/models/unit.py
```diff
@@ -362,18 +362,9 @@ def annotate_stats(self):
...
-
-        # Delete translation labels when not checked
-        new_labels = {label.name for label in data}
-        through = getattr(instance, self.attname).through.objects
-        for label in TRANSLATION_LABELS:
-            if label not in new_labels:
-                through.filter(unit__source_unit=instance, label__name=label).delete()
+class LabelsField(ManyToManyField):
+    # unused but needed until migrations are sqaushed
+    pass
```
nijel: Can't we rewrite the migrations to use `ManyToManyField` directly?

weblate/trans/models/unit.py
```diff
@@ -1868,11 +1867,10 @@ def translate(
...
 
         if change_action == ActionEvents.AUTO:
-            self.labels.add(component.project.automatically_translated_label)
+            self.automatically_translated = True
         else:
-            self.labels.through.objects.filter(
-                unit=self, label__name="Automatically translated"
-            ).delete()
+            self.automatically_translated = False
+        self.save(update_fields=["automatically_translated"])
```
nijel: Can't we move this above just before the `save_backend` call to avoid additional database hit?

weblate/utils/search.py
nijel: This should be no longer necessary as the labels are now set on source unit only. There might be more of such lookups no longer needed in the codebase.

weblate/trans/models/unit.py
amCap1712: I think in this particular case, yes should be fine to update the existing migration files directly.

weblate/trans/models/unit.py
```diff
@@ -1827,6 +1827,11 @@ def translate(
         if new_state != STATE_READONLY:
             self.original_state = self.state
 
+        if change_action == ActionEvents.AUTO:
+            self.automatically_translated = True
+        else:
+            self.automatically_translated = False
+
         # Save to the database
         saved = self.save_backend(
```
nijel, line 1845: The `save_backend` now needs to detect changes in the `automatically_translated` field via `old_unit`. Without that no-op edit on the automatically translated string does not remove the flag.

weblate/templates/translate.html
```diff
@@ -676,6 +676,18 @@ <h5 class="list-group-item-heading">
...
               {% endif %}
 
+              {% if unit.automatically_translated %}
+                <div class="list-group-item check check-item">
+                  <h5>
+                    <span class="green">{% icon "check.svg" %}</span>
+                    {% translate "Automatically translated" %}
+                  </h5>
+                  <p class="list-group-item-text check-description">
+                    {% blocktranslate with author=unit.get_last_author %}This string was translated automatically by {{ author }}.{% endblocktranslate %}
```
nijel, line 686: Adding a dismiss button (which would really just save the string without changes) here would make it obvious how to get rid of this alert.

amCap1712: IIUC, this would need to render the translate form here but with all inputs hidden, which seems a bit ugly but the alternative would be to add a new `/js/` endpoint to dismiss the automatically translated check or assign a special id to this and update the dismiss check endpoint? I am leaning in favor of a new endpoint, do you have a preference?

nijel: That's right, a new endpoint sounds good.

weblate/trans/migrations/0053_migrate_automatically_translated_label_to_unit_field.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    Label = apps.get_model("trans", "Label")
+
+    auto_labels = Label.objects.filter(name="Automatically translated")
+
+    if not auto_labels.exists():
+        return
+
+    for label in auto_labels:
+        label.unit_set.update(automatically_translated=True)
```
nijel, line 23: I think the label should be removed here as well.

amCap1712: Just remove the label from the unit or delete the labels?

nijel: Delete the label, it was automatically created and assigned, so there is no need to keep it (unless I'm missing something).

nijel:

nijel:
Ah, there is now conflict in migration...

amCap1712:
on it.

Issue #14397, opened 2025-03-30
feat: Allow a comment/note when blocking a user


nikclayton:
### Describe the problem

When blocking a user I want to note why I'm blocking them. At the moment I have to do that outside Weblate.

### Describe the solution you would like

When blocking a user provide a field that would allow a short (1K characters?) comment/note to be included. Show this in the UI when reviewing blocked users, and return it in the relevant API responses.

yussypu:
Hi! 👋 I’m looking to make my first contribution to Weblate, and this looks like a great issue to work on.
Would it be alright if I take this on?

— Yahya

nijel:
Yes, you're welcome!

The block model is here:

https://github.com/WeblateOrg/weblate/blob/def5773810603b64cd153ff2af7cbb9f1a47fbf4/weblate/auth/models.py#L956-L966

variantbyx:
Please assign me with this issue, I would love to conttibute

nijel:
Contributions are welcome; we don't assign issues to outside collaborators (because they often disappear).

PR #17038, opened 2025-11-24
allow users to leave a note when blocking a user

abtinmo:
Hello there, 

this PR references the following issue and Ames to implement a note feature for blocking the user.
[#14397](https://github.com/WeblateOrg/weblate/issues/14397)

I have implemented a test for POST method `block-user`, let me know if its needed to test the html response of `manage-access` and `user/<name:user>/` as well.

one of the requirements was `Show this in the UI when reviewing blocked users, and return it in the relevant API responses.` i could not find any API that interacts with UserBlock model, please let me know if I'm missing something.

### DIFF
--- a/weblate/trans/forms.py
+++ b/weblate/trans/forms.py
@@ -1361,7 +1361,12 @@ class UserBlockForm(forms.Form):
         required=False,
     )
     note = forms.CharField(
-        required=False, widget=forms.Textarea, label=gettext_lazy("Block note")
+        required=False,
+        widget=forms.Textarea,
+        label=gettext_lazy("Block note"),
+        help_text=gettext_lazy(
+            "Internal notes regarding blocking the user that are not visible to the user."
+        ),
     )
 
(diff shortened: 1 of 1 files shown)


nijel:
I don't think we have an API for this, so it looks complete for now. What I'd like to see is clarification in the help texts that this is an internal note not sent to the user. And please add a changelog entry.

weblate/trans/forms.py
```diff
@@ -1360,6 +1360,9 @@ class UserBlockForm(forms.Form):
         ),
         required=False,
     )
+    note = forms.CharField(
+        required=False, widget=forms.Textarea, label=gettext_lazy("Block note")
```
nijel: As this is not a model form to inherit the help text, it should be copied here as well.

abtinmo: copied the same text here as well. ✅

nijel:

PR #17053, opened 2025-11-25
feat(pending change): display detailed count on the repository status page

amCap1712:
Display the detailed count of units with pending change on the repository status page. This includes the total count of units with pending changes, count of units with changes skipped due to commit policy and count of units skipped due to errors.

### DIFF
--- a/weblate/templates/js/git-status.html
+++ b/weblate/templates/js/git-status.html
@@ -31,18 +31,22 @@
             {{ total_count }} pending units
           {% endblocktranslate %}
           {% with policy=pending_units.commit_policy_skipped errors=pending_units.errors_skipped %}
+            {% blocktranslate count counter=policy asvar policy_count_text with policy_count=policy|intcomma trimmed %}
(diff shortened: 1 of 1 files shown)


weblate/templates/js/git-status.html
```diff
@@ -25,7 +25,26 @@ <h4 class="card-title">
...
+                ({{ policy_count }} skipped by the commit policy, {{ errors_count }} skipped because of errors)
+              {% endblocktranslate %}
+            {% elif policy %}
+              {% blocktranslate with policy_count=policy|intcomma trimmed %}
+                ({{ policy_count }} skipped by the commit policy)
+              {% endblocktranslate %}
+            {% elif errors %}
+              {% blocktranslate with errors_count=errors|intcomma trimmed %}
+                ({{ errors_count }} skipped because of errors)
+              {% endblocktranslate %}
```
nijel: All these should use plurals as well, but having two counts in a single string makes it impossible to do it properly.

nijel:

PR #17064, opened 2025-11-26
chore(deps): update translate-toolkit to 3.17.1

nijel:
This changes Laravel keys to not include PHP structure.


nijel:
There seems to be some performance regression in translate-toolkit handling of PO files, investigating.

Issue #16806, opened 2025-11-04
Get comments via API


michael-smt:
### Describe the problem

While it is possible to [post a comment via API](https://docs.weblate.org/en/latest/api.html#post--api-units-(int-id)-comments-), it is not possible to get the comments on a Unit via the API.

### Solution brainstorm

Allow a `GET` on `/api/units/(int: id)/comments/`.

### Additional context

> @nijel with my example described in https://github.com/WeblateOrg/weblate/issues/7601#issuecomment-1316610725 above, this only addresses "meta data" as the state. But when doing such operations as mentioned, it would be good to act on further "meta data" as `suggestions` and `comments` , so that potentially API would allow to act on them as well.
> 
> 
> 
> Would this fit to this issue or shall I write a new one? 

 _Originally posted by @rhofer in [#7601](https://github.com/WeblateOrg/weblate/issues/7601#issuecomment-1316614310)_

nijel:
This should be pretty easy to implement; `/api/units/(int: id)/comments/` would be the endpoint to list/add suggestions. Probably a separate endpoint `/api/comments/(int: id)/` for removal/accepting. `/api/comments/` could be used to list site-wide comments (filtered for user accessible content).

biplopghimire:
Hi, we are a group of students from Carnegie Mellon University. As part of an open source contribution project for a class, our team wanted to work on this issue. If this issue is still available for contribution, we would be happy to take it and begin working.

nijel:
@biplopghimire contributions are welcome!

PR #17108, opened 2025-12-01
api :  Added support for listing comments on a unit

aadityajo:
## Proposed changes
Fixes: https://github.com/WeblateOrg/weblate/issues/16806
## Checklist


aadityajo:
Hey @nijel,
I am excited to take this up! I am currently working on the list support for unit comments.
I had a couple of doubts regarding the design:
- Should we add a new permission for viewing comments, or is the existing comment.add permission sufficient for the GET request?
- Can a user list all comments on a unit, or should they only see their own? (Currently fetching all).

Also, could you let me know what the checklist looks like for merging this? I want to make sure I cover everything.

Thank you!

docs/api.rst
```diff
@@ -2361,6 +2361,9 @@ and XLIFF.
     :>json string user: URL of the commenter's object
     :>json string timestamp: creation timestamp of the comment
 
+    .. versionchanged:: 5.15
+        Returns a list of comments on a given ttranslation unit
```
nijel: Instead the `.. http:get:: /api/units/(int:id)/comments/` should be described with versionadded annotation.

docs/changes.rst
```diff
@@ -7,6 +7,7 @@ Weblate 5.15
 
 * Batch automatic translation can now be manually triggered at project language, category or component level, see :ref:`auto-translation`.
 * Added ability to completely disable the admin contact form by setting :setting:`CONTACT_FORM` to ``"disabled"``.
+* Added :http:get:`/api/units/(int:id)/comments/` to get a list of comment for the given translation unit.
```
nijel: This points to nowhere (see previous comment).

weblate/api/views.py
```diff
@@ -2213,33 +2213,46 @@ def translations(self, request: Request, *args, **kwargs):
         methods=["post"],
         request=CommentSerializer,
     )
-    @action(detail=True, methods=["post"], serializer_class=CommentSerializer)
+    @extend_schema(
+        description="Return a list of comments on the unit.", methods=["get"]
+    )
```
nijel, line 2222: I think the OpenAPI schema will be wrong, it needs to know it returns list of the comments.

weblate/api/views.py
```diff
@@ -2213,33 +2213,46 @@ def translations(self, request: Request, *args, **kwargs):
...
+                not user.has_perm("project.edit", unit.translation.component.project)
+            ):
+                self.permission_denied(request)
+
+            serializer.save()
+            return Response(serializer.data, status=HTTP_201_CREATED)
+
+        # Do we need a new permission or does permission to add comment grant the user to list all comments/user specific comment?
+        if not user.has_perm("comment.add", unit.translation):
             self.permission_denied(request)
```
nijel: I think only access to the unit needs to be checked here and any user that can access the unit, can see the comments (same as in the UI).

docs/api.rst
aadityajo: Understood, thank you - updated the doc to use versionadded annotation.

weblate/api/views.py
aadityajo: Yep - agreed, updated the code to return responses with many set to true.

weblate/api/views.py
aadityajo: The filter_queryset checks for the user's access to the unit. Calling get object will raise 404 if the user does not have access. Simply removed the permission check. Thanks!

aadityajo:
Hey @nijel,
Could you please review the PR once you get a chance?

weblate/api/views.py
```diff
@@ -2213,33 +2213,45 @@ def translations(self, request: Request, *args, **kwargs):
         methods=["post"],
         request=CommentSerializer,
     )
-    @action(detail=True, methods=["post"], serializer_class=CommentSerializer)
+    @extend_schema(
+        description="Return a list of comments on the unit.",
+        methods=["get"],
+        responses={200: CommentSerializer(many=True)},
+    )
```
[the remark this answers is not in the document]
nijel: I think Copilot is wrong here.

weblate/api/views.py
```diff
@@ -2213,33 +2213,45 @@ def translations(self, request: Request, *args, **kwargs):
...
-        serializer.save()
-        return Response(serializer.data, status=HTTP_201_CREATED)
+            serializer.save()
+            return Response(serializer.data, status=HTTP_201_CREATED)
+
+        # If user doesn't have access to the unit, get_object will raise 404.
+        qs = unit.comment_set.all()
+
+        serializer = CommentSerializer(qs, context={"request": request}, many=True)
+        return Response(serializer.data, status=HTTP_200_OK)
```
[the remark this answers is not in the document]
nijel: I think it makes sense to paginate this, even when a situation with many comments is not that frequent.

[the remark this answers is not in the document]
aadityajo: Agreed, best to support pagination (and the OpenAPI spec also mentions this). I will make the change (not a big fan of how Copilot does it).
Thanks!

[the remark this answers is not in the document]
aadityajo: Updated the code @nijel.

nijel:

Issue #16140, opened 2025-09-10
Drop support for MySQL/MariaDB


nijel:
### Describe the problem

When using more complex ORM features, we often run into problems with MySQL compatibility:

* https://github.com/WeblateOrg/weblate/pull/16102
* https://github.com/WeblateOrg/weblate/pull/15869

While it was always possible to address it, it needs an extra effort and introduces extra code, which means extra maintenance in the future.

So we might ask a question about whether MySQL/MariaDB support is still something worth maintaining.

### Solution brainstorm

Drop support for MySQL/MariaDB and focus on PostgreSQL only. This would allow us to drop MySQL specific code and introduce more of [PostgreSQL specific features](https://docs.djangoproject.com/en/5.2/ref/contrib/postgres/).

### Additional context

Years ago, this was discussed in https://github.com/WeblateOrg/weblate/issues/2825.

nijel:
When removing MySQL support, https://github.com/WeblateOrg/weblate/issues/9315 could be addressed as well.

nijel:
As there are no objections so far, let's make the removal happen in the 5.16 release.

PR #17142, opened 2025-12-03
fix(component): discard disk state on reset

nijel:
Without that the reset content from the repository is not loaded.

### DIFF
--- a/weblate/trans/models/component.py
+++ b/weblate/trans/models/component.py
@@ -2080,12 +2080,14 @@ class Component(
 
             if keep_changes:
                 # Mark all strings as pending when keeping changes
-                self.do_file_sync(request, do_commit=False)
+                self.do_file_sync(request, do_commit=False, store_disk_state=False)
             else:
                 # Explicitly remove all pending changes
                 PendingUnitChange.objects.filter(
                     unit__translation__component=self
                 ).delete()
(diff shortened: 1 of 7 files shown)


weblate/trans/models/component.py
```diff
@@ -2080,12 +2080,19 @@ def do_reset(
...
                 PendingUnitChange.objects.filter(
                     unit__translation__component=self
                 ).delete()
+            # Remove disk state as we are going to change that
+            units = Unit.objects.filter(
+                translation__component=self, details__has_key="disk_state"
+            ).select_for_update()
+            for unit in units:
+                del unit.details["disk_state"]
+            Unit.objects.bulk_update(units, fields=["details"])
```
nijel: @amCap1712 Do you see a better way to perform this?

nijel: Ah, I didn't notice that you've implemented a similar thing before Copilot mentioned it, so let's move it to a separate method...

amCap1712: Yes, and there is also a possibility to push the update completely into the database once we only support Postgres.

nijel: Do you want to annotate it somehow for https://github.com/WeblateOrg/weblate/issues/16140?

Issue #17255, opened 2025-12-10
Unmigrated Boostrap 5


nijel:
### Describe the issue

I've done some grepping over templates and CSS and I think these need to be migrated: 

* weblate/legal/templates/legal/confirm.html (shown upon registration with https://docs.weblate.org/en/latest/admin/optionals.html#legal-module)
* weblate/templates/manage/backups.html (/manage/backup/, can be used with local repo, see https://docs.weblate.org/en/latest/admin/backup.html#local-filesystem)
* `pull-right` in CSS rules should either be rewritten or removed
* weblate/templates/mail/style.css will probably need adjustments, as that is standalone CSS for notification e-mails

### I already tried

- [x] I've read and searched [the documentation](https://docs.weblate.org/).
- [x] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

```
git grep 'class.*panel\|pull-right' -- '*.html' '*.css'
```

### Exception traceback

```pytb

```

### Weblate deploy checks

```shell

```

KarenKonou:
Unsure how i missed this since I did very similar greps, will fix

PR #17261, opened 2025-12-10
fix(ui): unmigrated templates fixes

KarenKonou:
Should fix the issues in #17255

### DIFF
--- a/weblate/templates/manage/backups.html
+++ b/weblate/templates/manage/backups.html
@@ -59,131 +59,131 @@
                 </div>
               </div>
             </div>
-            {% endif %}
+          {% endif %}
+          <div class="card">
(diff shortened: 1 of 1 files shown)


weblate/templates/manage/backups.html
nijel, line 19: Lint error: H025 19:6 Tag seems to be an orphan. <div class="card">

KarenKonou: I'm going to double check this but there was a superfluous `</div>` that caused the `.main-content` div to be closed prematurely, could be the linter getting confused. The page appears to render fine

nijel, line 60: I think the missing `</div>` is actually here:

```suggestion
              </div>
            </div>
```

PR #17276, opened 2025-12-11
fix(vcs): simplify git remote update to persist shallow clones

nijel:
Since 9666bbd6b7c89e4d2b178d524d28ad86bb434ddb this should not be used for initial fetch, so the code can be simplified to deal with the update only. Keep fetch depth args because this will fetch new branches and we do not want to fetch full history for these.

See also #17239


nijel:
I messed up history, https://github.com/WeblateOrg/weblate/pull/17278 is replacement.

PR #17294, opened 2025-12-12
feat: detect automatically translated units from XLIFF state-qualifier

### LINKED ISSUE #14039: Leverage  state-qualifier from XLIFF
### Describe the problem

The [state-qualifier](http://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html#state-qualifier) attribute from XLIFF could be used to determine origin of the string translation.

Most of them are probably not useful, but `leveraged-mt` and `mt-suggestion` could add the `Automatically translated` label to the string.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

amCap1712:
Add support for detecting machine-translated units in XLIFF files based on the state-qualifier attribute. Units with "leveraged-mt" or "mt-suggestion" state qualifiers are now marked as automatically translated when imported. To implement this, a `is_automatically_translated` method has been added to the base TranslationUnit class and it is used to set the `automatically_translated` flag in database during unit import.

Fixes #14039

### DIFF
--- a/weblate/trans/tests/test_remote.py
+++ b/weblate/trans/tests/test_remote.py
@@ -634,7 +634,12 @@ class FileScanTest(ViewTestCase):
         self.assertEqual(unit_1.state, STATE_TRANSLATED)
         self.assertEqual(
             unit_1.details["disk_state"],
-            {"target": target_1, "state": STATE_APPROVED, "explanation": explanation_1},
+            {
+                "target": target_1,
+                "state": STATE_APPROVED,
+                "explanation": explanation_1,
+                "automatically_translated": False,
+            },
         )
         self.assertEqual(PendingUnitChange.objects.filter(unit=unit_1).count(), 1)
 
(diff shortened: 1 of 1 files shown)


weblate/trans/models/unit.py
```diff
@@ -981,6 +983,7 @@ def update_from_unit(  # noqa: C901,PLR0914
         unit = unit_attributes["unit"]
         created = unit_attributes["created"]
         pos = unit_attributes["pos"]
+        automatically_translated = unit_attributes["automatically_translated"]
```
amCap1712, line 1000: Wondering if this field should be checked as well in the `same_metadata` check now? I don't think most formats support writing the `automatically_translated` flag so we'll need the same treatment as `is_fuzzy` at least if we do.

(https://github.com/WeblateOrg/weblate/blob/b85f8d2c35aad35fa29bba11cb2ff4b275672498/weblate/trans/models/unit.py#L1059)

amCap1712:
If a unit is marked automatically translated in Weblate, should we write it to the xliff file as well? If yes, which of the two state qualifiers should we choose?

nijel:
Oh, and I don't have preference out of these two stay, it should not flip them if one is already present.

nijel:
@amCap1712 Do you think this is good to include in 5.15.1?

amCap1712:
@nijel Yes, I did some manual tests and added some to automated tests as well and am happy with the changes.

nijel:
Some tests are failing, can you check it?

amCap1712:
on it.

nijel:

PR #17330, opened 2025-12-16
fix(validators): reject certain paths from being used

nijel:
Restrict based on the translation-finder blacklist which covers files we do not want to touch.

### DIFF
--- a/weblate/utils/tests/test_validators.py
+++ b/weblate/utils/tests/test_validators.py
@@ -139,6 +139,16 @@ class FilenameTest(SimpleTestCase):
     def test_empty(self) -> None:
         validate_filename("")
 
+    def test_prohibited(self) -> None:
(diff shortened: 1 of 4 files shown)


weblate/utils/tests/test_validators.py
```diff
@@ -139,6 +139,10 @@ def test_simplification(self) -> None:
     def test_empty(self) -> None:
         validate_filename("")
 
+    def test_prohibited(self) -> None:
+        with self.assertRaises(ValidationError):
+            validate_filename(".git/config")
```
nijel, line 144:
```suggestion
        with self.assertRaises(ValidationError):
            validate_filename(".git/config")
        with self.assertRaises(ValidationError):
            validate_filename("path/.git/config")```

PR #17333, opened 2025-12-16
docs(state): add new states

michael-smt:
Adds documentation and changelog entry for #16947.

In my understanding they are functionally equivalent and a commit policy of "Skip translations marked as needing editing" should also skip "Needs rewriting" and "Needs checking".

### DIFF
--- a/docs/workflows.rst
+++ b/docs/workflows.rst
@@ -63,10 +63,10 @@ Each translated string can be in one of the following states:
 Untranslated
     Translation is empty, it might or not be stored in the file, depending
     on the file format.
-Needs editing
+Needs editing / Needs rewriting / Needs checking
     Translation needs editing, this is usually the result of a source string change, fuzzy matching or translator action.
-    The translation is stored in the file, depending on the file format it might
-    be marked as needing edit (for example as it gets a ``fuzzy`` flag in the gettext file).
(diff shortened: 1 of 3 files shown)


nijel:
Unfortunately this ended up half-finished in the release, let's try to polish it for 5.15.1.

docs/workflows.rst
```diff
@@ -63,10 +63,10 @@ Each translated string can be in one of the following states:
 Untranslated
     Translation is empty, it might or not be stored in the file, depending
     on the file format.
-Needs editing
+Needs editing / Needs rewriting / Needs checking
     Translation needs editing, this is usually the result of a source string change, fuzzy matching or translator action.
-    The translation is stored in the file, depending on the file format it might
-    be marked as needing edit (for example as it gets a ``fuzzy`` flag in the gettext file).
+    The translation is stored in the file if the :ref:`project-commit_policy` allows it.
+    Depending on the file format it might be marked as needing edit (for example as it gets a ``fuzzy`` flag in the gettext file).
```
amCap1712, line 69: Not related to the documentation of new states, but I think all of the state descriptions here need to be updated because of project commit policy. I think we can remove the mention of whether the translation is stored in the file or not here and add a note at the end explaining it once.

nijel: Indeed, it now depends on both file format and commit policy.

Issue #6334, opened 2021-07-25
Quality check: spaces around tags


miroslavrehor:
**Problem:**
The translator sometimes does not make correct spaces around tags. For example:
```
<p>Click on the links<a href="...">here</a>or<a href="...">here</a>.</p>
```

**Solution suggestion:**
Create new quality check. Logic like:
* if tag: a, strong, b, i, ...
* and if following chars are before or after the tag: \p{L}
* then show a failed check

nijel:
The logic probably should be: If there is a space in the source string before/after the tag, it should be in the translation as well.

miroslavrehor:
> The logic probably should be: If there is a space in the source string before/after the tag, it should be in the translation as well.

That is what I thought as well, at first. But then I came across a translated string in our project where this rule does not hold:

```
msgid "Do fotografie <strong>nevkládejte žádné texty</strong> v jakékoli formě"
msgstr "<strong>Nie wstawiaj żadnego tekstu</strong> do zdjęć w jakiejkolwiek formie"
```

But for sure, _some_ check is better than none. :-)

nijel:
Okay, space or string boundary is probably a better definition.

evanjhoward11:
Hello, if this issue is still relevant, can I work on it?

nijel:
Yes, contributions are welcome!

evanjhoward11:
Great! @nijel or @miroslavrehor, can you point me to the relevant file(s) to work on? Also, should any other chars that appear before/after the inline tag result in a failed check? I'm thinking about numbers, colons, slashes, etc.

nijel:
The existing XML checks are in `weblate/checks/markup.py` so this should probably belong there as well.

evanjhoward11:
@nijel, I submitted PR [#17373](https://github.com/WeblateOrg/weblate/pull/17373) for this issue. It looks like the workflow checks are not passing even though I was able to pass the checks from ruff and pre-commit on my machine, so let me know if that is an issue.

Let me know how I can fix or improve my code!

PR #17373, opened 2025-12-18
Checks: Add XMLCharsAroundTagsCheck for issue #6334

evanjhoward11:
issue [#6334](https://github.com/WeblateOrg/weblate/issues/6334)

Adds a new check (XMLCharsAroundTagsCheck) to ensure that the translation provided by the user is consistent with the original source. Specifically, it checks that the character(s) adjacent to each XML tag is a letter or non-letter in both the source and translation. The main goal is to prevent the user from adding or removing spaces around tags during their translation.

Added unit tests to weblate/checks/tests/test_markup_checks.py
Added the new check to weblate/settings_test.py
Added the new check to weblate/checks/models.py


weblate/checks/markup.py
```diff
@@ -282,6 +283,60 @@ def check_highlight(self, source: str, unit: Unit):
...
+                tgt_next_char = target[tgt_end_idx]
+
+                if self.char_check(src_prev_char, tgt_prev_char):
+                    return True
+                if self.char_check(src_next_char, tgt_next_char):
+                    return True
+        return False
+
+    def char_check(self, src_char: str, tgt_char: str) -> bool:
+        any_letter = regex.compile(r"\p{L}")
```
nijel: The original issue was focused on whitespace, but this uses regex to match against letters. This will not fire when whitespace is added instead of punctuation. For example, when `Hello <strong>world</strong>.` it translated to `Hello <strong>world</strong> .` it would still pass. Or am I missing something? What is the reason for using regex here instead of simple `.isspace()`?

evanjhoward11: Yes you are correct, but I intentionally coded it this way because of the edge case shown [here](https://github.com/WeblateOrg/weblate/issues/6334#issuecomment-886496980). Since it's possible for white space and special characters to be "interchangeable" (for lack of a better term) based on the example, I didn't want to prevent a translator from adding a technically accurate translation. So my thought process became to enforce letters, but let everything else pass by. I agree it's not perfect. And it also makes me wonder if there are any examples out there where letters and white space or letters and special characters can be interchanged. Does this help clear things up? I'm open to any ideas or improvements.

nijel: Okay, makes sense, I just wanted to clarify the behavior.

weblate/checks/markup.py
```diff
@@ -282,6 +283,60 @@ def check_highlight(self, source: str, unit: Unit):
...
+            return False
+
+        for i, tag in enumerate(src_tags):
+            src_start_idx = tag.start()
+            tgt_start_idx = tgt_tags[i].start()
+            src_end_idx = tag.end()
+            tgt_end_idx = tgt_tags[i].end()
+
+            # if string starts with tag, only check char following this tag
+            if src_start_idx == 0 or tgt_start_idx == 0:
```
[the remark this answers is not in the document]
evanjhoward11: Using the conditions with `or` is correct. It's very likely that there are cases where a source starts with a tag, but the target doesn't. Leaving this with `or` gives the translator more flexibility.

docs/user/checks.rst
```diff
@@ -1885,13 +1885,35 @@ Checks that XML tags are replicated between both source and translation.
...
 
+.. note::
+
+   This check is disabled by the ``safe-html`` flag as the HTML cleanup done by
+   it can produce HTML markup which is not valid XML.
+
+.. _check-xml-chars-around-tags:
+
+XML surrounding characters
+~~~~~~~~~~
```
nijel:
```suggestion
XML surrounding characters
~~~~~~~~~~~~~~~~~~~~~~~~~~
```

nijel:

PR #17420, opened 2025-12-20
fix(permission): make announcement.delete a real permission

### LINKED ISSUE #15132: Make announcement.delete a real permission
> 2. There is a dummy `announcement.delete` permission. https://github.com/WeblateOrg/weblate/blob/6a5a25c34cd8a1f8a49e29f3d7f3de696b4ce8af/weblate/auth/permissions.py#L530-L531
>    I think that permission should either be merged with the `announcement.add` permission or converted to separate real permission.

It grants the permission based on the announcement scope, but the question is how much is that used. Anyway, component scoped team with the `announcement.delete` permission would work as replacement. It just needs to be documented, as this is not something we can easily handle in the migrations. So probably having a real `announcement.delete` permission is a way to go.

----
Make announcement.delete a real permission and add it to translation coordinator role.

amCap1712:
Make announcement.delete a real permission like announcement.add and .edit.

Fix #15132.

### DIFF
--- a/docs/admin/access.rst
+++ b/docs/admin/access.rst
@@ -805,6 +805,10 @@ List of privileges
 | Announcements                | Post announcements                        | :guilabel:`Administration`            |
 |                              |                                           +---------------------------------------+
 |                              |                                           | :guilabel:`Translation coordinator`   |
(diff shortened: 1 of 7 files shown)


nijel:

PR #17426, opened 2025-12-21
fix(file sync): handle explanation change for TBX

### LINKED ISSUE #16571: Metadata changes detection when parsing TBX
### Describe the issue

Weblate differs in the update path of the unit depending on whether data (content) or metadata was changed:

https://github.com/WeblateOrg/weblate/blob/00887266cee789f71f54dbc74f069d02e7fc5e92/weblate/trans/models/unit.py#L947-L964

Since allowing to edit explanations in some formats, the split is wrong here, because changes to the explanation actually change the content.

### I already tried

1. Create TBX file
2. Edit explanation
3. Change explanation in the file

Depending on the situation, the changes to explanation could be lost.

### Expected behavior

Explanation should be considered data for formats that support editing of it.

### Screenshots

_No response_

### Exception traceback

```pytb

```

### How do you run Weblate?

_No response_

### Weblate versions

_No response_

### Weblate deploy checks

```shell

```

### Additional context

Discovered while reviewing https://github.com/WeblateOrg/weblate/pull/16568

amCap1712:
Certain formats like TBX support writing the explanation to the file. When syncing from these file formats, explanation read from disk should be treated as a data change and not a metadata change.

Fixes #16571.

### DIFF
--- a/weblate/trans/models/unit.py
+++ b/weblate/trans/models/unit.py
@@ -1076,7 +1076,6 @@ class Unit(models.Model, LoggerMixin):
         same_state = state == comparison_state["state"] and flags == Flags(self.flags)
         same_metadata = (
             location == self.location
-            and explanation == comparison_state["explanation"]
             and note == self.note
             and pos == self.position
             and automatically_translated == self.automatically_translated
@@ -1092,6 +1091,13 @@ class Unit(models.Model, LoggerMixin):
             and self.source_unit == old_source_unit
             and old_source_unit is not None
         )
+
+        supports_explanation = component.file_format_cls.supports_explanation
+        same_explanation = explanation == comparison_state["explanation"]
+        if supports_explanation:
+            same_data &= same_explanation
+        else:
+            same_metadata &= same_explanation
 
         # Conditionally check original state changes if it would be used. It is not
         # properly tracked in PendingUnitChange, so this would not work for units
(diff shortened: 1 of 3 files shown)


weblate/trans/models/unit.py
```diff
@@ -1093,6 +1092,13 @@ def update_from_unit(  # noqa: C901,PLR0914
             and old_source_unit is not None
         )
 
+        supports_explanation = component.file_format_cls.supports_explanation
+        same_explanation = explanation == comparison_state["explanation"]
+        if supports_explanation:
+            same_data &= same_explanation
+        else:
+            same_metadata &= same_explanation
```
amCap1712, line 1100: I think we can just omit the check for `same_explanation` if the file format doesn't support explanation. As it would always come from the current explanation in the database.

https://github.com/WeblateOrg/weblate/blob/c443fd688ea676934c0a4928aabf336295ea4e92/weblate/trans/models/unit.py#L929

weblate/trans/models/unit.py
```diff
@@ -1122,16 +1128,18 @@ def update_from_unit(  # noqa: C901,PLR0914
         # Metadata update only, these do not trigger any actions in Weblate and
         # are display only
         if same_data and not same_metadata:
+            update_fields = [
+                "location",
+                "note",
+                "position",
+                "automatically_translated",
```
amCap1712, line 1135: I think we'll need the same treatment for `automatically_translated` with `supports_automatically_translated` on the file formats.

weblate/trans/models/unit.py
nijel: I'm not sure if the flag actually doesn't mean support for writing the explanation. Needs to be checked.

weblate/trans/models/unit.py
nijel: Currently automatically translated, fuzzy and approved states are handled in format with fallback from a database, it would also make sense to make it a feature flag on the format (that would be useful for documentation purposes as well) and handle it on the model side. But that's definitely a separate issue from this one 😉

amCap1712: Yup, makes sense.

nijel: Will you please create an issue to track this? Feel free to self-assign if interested in addressing it..

amCap1712: Done.

Issue #17450, opened 2025-12-23
Add support for DeepL beta languages


shyraptor:
### Describe the problem

DeepL recently added beta languages (e.g., Acehnese, Balinese, etc.) that require passing enable_beta_languages: true in the API request.

https://developers.deepl.com/docs/getting-started/supported-languages

### Solution brainstorm

A simple toggle in the DeepL service configuration (/machinery/<project>/deepl/) that would mark the parameter to be added to any DeepL request would be enough.

PR #17457, opened 2025-12-26
Add support for DeepL beta languages (#17450)

shyraptor:
Implements [#17450](https://github.com/WeblateOrg/weblate/issues/17450) - Add support for DeepL beta languages

Adds a toggle to DeepL service configuration.

When enabled, requests to DeepL are sent with enable_beta_languages parameter set to true.

See https://developers.deepl.com/docs/getting-started/supported-languages#beta-languages


nijel (commented):
How does this play with the glossaries, shouldn't the parameter be included there as well? In that case it might be better to move the parameter build to a separate method so that it is consistently used in all requests.

shyraptor:
> In that case it might be better to move the parameter build to a separate method so that it is consistently used in all requests.

https://developers.deepl.com/docs/getting-started/supported-languages#beta-languages
"Beta languages can be enabled by adding enable_beta_languages... Note that these languages do not support glossaries or formality yet."

EDIT: Is it fine as is then, or should I make any adjustments?

nijel:
I think this is good to go once the documentation is updated with the new paratemeter and lack of glossary support is documented.

shyraptor:
@nijel will probably need one last go-ahead to finish the merge

nijel:
Thanks! Looks good, scheduled for merge.

PR #17477, opened 2025-12-29
chore(pending change): add test for update of non-translation files

amCap1712:
Add test case that pending database changes are not overwritten when remote commits non-translation files.

Relates to https://github.com/orgs/WeblateOrg/discussions/15594.


nijel:
Thanks for taking care of this!

PR #17506, opened 2026-01-05
fix(unit): gracefully handle missing automatically_translated flag

nijel:
It was not stored before in the disk_state field.

Fixes WEBLATE-38C1

### DIFF
--- a/weblate/trans/models/unit.py
+++ b/weblate/trans/models/unit.py
@@ -691,7 +691,12 @@ class Unit(models.Model, LoggerMixin):
 
         """
         if "disk_state" in self.details:
-            return self.details["disk_state"]
+            disk_state = self.details["disk_state"]
+            # Compatibility code to handle disk_state without automatically_translated flag
+            # TODO: remove this in future release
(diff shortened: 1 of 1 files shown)


weblate/trans/models/unit.py
```diff
@@ -691,7 +691,12 @@ def get_comparison_state(self) -> dict[str, Any]:
 
         """
         if "disk_state" in self.details:
-            return self.details["disk_state"]
+            disk_state = self.details["disk_state"]
+            # Compatibility code to handle disk_state without automatically_translated flag
+            # TODO: remove this in future release
+            if "automatically_translated" not in disk_state:
+                disk_state["automatically_translated"] = False
```
amCap1712: Instead of False, maybe set based on `unit.automatically_translated`'s current value?

nijel: Good point, changed.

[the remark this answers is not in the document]
nijel: This should be safe to mutate.

PR #17526, opened 2026-01-06
fix(googlev3): ensure glossary path is set properly

Korpyc:
Advanced Google Translate with glossary did not function correctly because the integration passed only the glossary name (ID) rather than the required full glossary resource path. This change ensures the full path is used, restoring correct glossary-based translations.


nijel:

PR #17576, opened 2026-01-12
Revert "feat(machinery) Add support for DeepL beta languages (#17450)…

nijel:
… (#17457)"

This reverts commit 823ad4f0abfab9e4b9d33356661a3ad3ecded86e.

The new flag is no longer needed and these languages are now available based on the configured model type.


nijel:
@shyraptor This reverts #17457 as it seems no longer needed.

PR #17684, opened 2026-01-19
docs: Automated documentation update

### LINKED ISSUE #16632: Automate generated documentation updates
### Describe the problem

The documentation often gets out of sync with the code, typically:

* Newly added Change events are not added in the docs such as https://github.com/WeblateOrg/weblate/pull/16631/files#diff-1b7c0dd152dcd4310ac87eaee705ea22d3d00e93eed7b62ccea1fc89043ac607L1040
* Changed permissions are missing in the docs

Generating portions of the documentation is already automated (see https://docs.weblate.org/en/latest/contributing/documentation.html#documenting-permissions-checks-add-ons-and-automatic-suggestions), but it still needs somebody to manually apply the changes to the file.

### Solution brainstorm

Automated solution that would update the generated portions of the documentation that could be executed both locally and in CI. This will need some restructuring of the docs to make automated updates possible without removing manually written content.

Some commands generate standalone content that could be saved to a snippet that is included. This is what is currently used with file format parameters.

Documentation for add-ons or machinery is more complex because the `list_adons` and `list_machinery` commands only generate the structure of the document, but the additional documentation is then manually added.

We have manual updates now and we tend to forget them, so that is really not a solution 😄.

### Screenshots

_No response_

### Additional context

_No response_

gersona:
- fixes #16632


.github/workflows/docs.yml
```diff
@@ -20,8 +20,50 @@ permissions:
...
+      CI_DB_PASSWORD: weblate
+      CI_DB_HOST: 127.0.0.1
+      CI_DB_PORT: '60000'
+      CI_SELENIUM: '1'
+      DJANGO_SETTINGS_MODULE: weblate.settings_test
+    steps:
+    - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8   # v6.0.1
+      with:
+        token: ${{ secrets.WEBLATE_CI_TOKEN }}
+        persist-credentials: true
```
nijel, line 39: This will need similar (complex) logic as https://github.com/WeblateOrg/weblate/blob/main/.github/workflows/uv-upgrade.yml to work in all situations:

- WEBLATE_CI_TOKEN is used to update when run on the main branch and peter-evans/create-pull-reques creates pull request then
- pre-commit-ci/lite-action is used for outside collaborators as it can push changes to their forks and CI can be triggered
- stefanzweifel/git-auto-commit-action is used for in-repository pull requests to update the branch directly (this is typically used for dependency updates)

It would be great if we could factor this out into reusable action as this pattern is already used several times...

gersona: Implemented this in #17730

nijel:
Merged, created follow-up issue to cover other portions: https://github.com/WeblateOrg/weblate/issues/17761

PR #17688, opened 2026-01-19
fix(unit): honor source unit based state even when there are no flags

nijel:
I don't see a reason to look for flags when inferring state from the source unit. It magically worked for most of the time because we started to pass empty Flags instead instead of None in some cases.


amCap1712 (approved):
LGTM, Thanks!

Issue #17587, opened 2026-01-13
Refactor frontend theming


KarenKonou:
### Describe the problem

Currently the theming implementation is a mix of several approaches, likely stemming from being implemented by different people over several different library versions. This can cause issues such as #16616, as well as generally unclean and messy code. I propose unifying theming approach based on Bootstrap 5 color modes https://getbootstrap.com/docs/5.3/customize/color-modes/ which should result in cleaner code and greater flexibility in customization and adding the possibility to add more themes/color modes in the future.

### Solution brainstorm

1. Refactor current CSS to use variables and Bootstrap 5 color modes wherever possible.
2. Clean up unused rules in the CSS files.

nijel:
It definitely makes sense to use upstream support for this.

jermanuts:
Looks like it is now permanent instead of just being flashes

<img width="1902" height="643" alt="Image" src="https://github.com/user-attachments/assets/c26a0e08-c330-4154-b348-512ca789ecd1" />

<img width="1902" height="817" alt="Image" src="https://github.com/user-attachments/assets/0a2499be-e352-4033-b8a2-9cacb22626c3" />

Basically Weblate is unusable on Dark theme, please prioritize this issue.

KarenKonou:
@jermanuts For what it's worth, these issues should only occur when theme is set to automatically synchronize with system theme, setting it to dark in manually in the preferences should alleviate this.

Either way we will work on fixing this.

KarenKonou:
Also the current state of the auto theme is definitely a regression

PR #17704, opened 2026-01-20
chore: pin jquery version to 3.7.1 for compatibility reasons

KarenKonou:
This pull request: https://github.com/WeblateOrg/weblate/pull/17679 brought in the newly released 4.0 version of jQuery, which unfortunately doesn't seem to work with the daterangepicker library. This breaks our js files causing issues such as the theme breakage described here: https://github.com/WeblateOrg/weblate/issues/17587#issuecomment-3767858056

Sticking with jQuery 3.7.1 appears to fix this for the time being

### DIFF
--- a/docs/specs/sbom/partial/javascript.json
+++ b/docs/specs/sbom/partial/javascript.json
@@ -2,7 +2,7 @@
   "$schema": "http://cyclonedx.org/schema/bom-1.5.schema.json",
   "bomFormat": "CycloneDX",
   "specVersion": "1.5",
-  "serialNumber": "urn:uuid:33796d06-f0d8-595f-810f-c77533fb98d9",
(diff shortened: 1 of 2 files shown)


KarenKonou:
Slight correction: the problem stemmed from two different versions of jQuery being pulled in. It's possible that we can upgrade to jQuery 4.0 but I'd like to make sure everything works first.

nijel:
Ah, another outdated dependency library kicks in...

nijel:
@KarenKonou It doesn't seem to work as expected; it is still being pulled via `yarn upgrade`: https://github.com/WeblateOrg/weblate/pull/17705/changes/67888d5398f9a63398d8395a33ecf162aac3065e

KarenKonou:
@nijel should be fixed by #17708

PR #17730, opened 2026-01-22
refactor(ci): Reusable Github Actions action for auto-commit changes

gersona:
Suggestion from  https://github.com/WeblateOrg/weblate/pull/17684#discussion_r2703807336

### DIFF
--- a/.github/workflows/uv-upgrade.yml
+++ b/.github/workflows/uv-upgrade.yml
@@ -2,7 +2,6 @@
 #
 # SPDX-License-Identifier: CC0-1.0
 
-# This file is maintained in https://github.com/WeblateOrg/meta/
 name: uv lock update
 
 on:
--- a/.github/workflows/yarn-update.yml
+++ b/.github/workflows/yarn-update.yml
@@ -2,7 +2,6 @@
 #
 # SPDX-License-Identifier: CC0-1.0
 
-# This file is maintained in https://github.com/WeblateOrg/meta/
 name: yarn update
 
 on:


nijel:
Looks good! Some of these scripts were kept at https://github.com/WeblateOrg/meta/ and synced to other repos, but as these never spread, let's move them here. Can you please update headers to remove references to maintenance of these in the meta repo? I've removed them in https://github.com/WeblateOrg/meta/commit/848ae840f36f5b96092602fd4a0ebb8b0fda5cbc

.github/actions/auto-commit/action.yml
```diff
@@ -0,0 +1,60 @@
+# Copyright © Michal Čihař <michal@weblate.org>
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+name: Auto Commit Changes
+description: Automatically commit and push changes based on workflow context
+
```
nijel, line 7: Shouldn't this start with:

```
on:
  workflow_call:
```

See https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows

.github/actions/auto-commit/action.yml
```diff
@@ -0,0 +1,60 @@
...
+    description: Branch name for pull request (if created)
+    required: false
+    default: create-pull-request/auto-update
+  pr-labels:
+    description: Labels to apply to pull request (one per line)
+    required: false
+    default: ''
+  github-token:
+    description: GitHub token for creating pull requests
+    required: true
```
nijel, line 22: It should be in `secrets` instead.

.github/actions/auto-commit/action.yml
gersona: I wasn't really aware of reusable workflows. I actually used a composite action
https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action

AFAIU, main difference is that reusable workflow can include multiple jobs, while composite action only allow multiple steps

Should I then move it to reusable workflow ?

nijel: And I was not aware of composite actions :-). I think using it is okay, so let's stick with this.

gersona: After double checking, composite actions is the way to go here, as reusable workflows start with a fresh checkout, so changes made in the previous steps will be lost

.github/actions/auto-commit/action.yml
gersona: That's one downside of composite actions, they don't have access to secrets so they need to be passed as inputs

.github/workflows/uv-upgrade.yml
nijel:
```suggestion
```

.github/workflows/yarn-update.yml
nijel:
```suggestion
```

nijel:

Issue #17720, opened 2026-01-21
Filtering by label with an AND condition does not work as expected


datenspast:
### Describe the issue

I guess this is a bug and it worked in previous versions, but I`m not 100 % sure.

We have several strings which are tagged with several labels. 
For example one source string is labled with "ABC" and "XYZ".

If I search for "label:ABC", the search works.
If I search for "label:XYZ", the search works.
If I search for "label:ABC OR label:XYZ", the search works as expected.
If I search for **"label:ABC AND label:XYZ"**, **no results are shown**, although there are strings which have both labels.

On the other hand searching for something like "changed:>='7 days ago' AND changed:<='2 days ago'" works as expected.

Thanks for your help!

### I already tried

- [x] I've read and searched [the documentation](https://docs.weblate.org/).
- [x] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

1. Add two labels to one string.
2. Try to search for this with the query "label:ABC AND label:XYZ".

### Expected behavior

I would expect to see all strings which match the condition, e.g. which have both labels.

### Exception traceback

```pytb

```

### How do you run Weblate?

Docker container

### Weblate versions

[Weblate 5.15]

### Weblate deploy checks

```shell

```

nijel:
Confirmed, it wrongly translates to:

```sql
SELECT "trans_unit"."id", "trans_unit"."translation_id", "trans_unit"."id_hash", "trans_unit"."location", "trans_unit"."context", "trans_unit"."note", "trans_unit"."flags", "trans_unit"."source", "trans_unit"."previous_source", "trans_unit"."target", "trans_unit"."state", "trans_unit"."original_state", "trans_unit"."details", "trans_unit"."position", "trans_unit"."num_words", "trans_unit"."priority", "trans_unit"."timestamp", "trans_unit"."last_updated", "trans_unit"."extra_flags", "trans_unit"."explanation", "trans_unit"."variant_id", "trans_unit"."automatically_translated", "trans_unit"."source_unit_id" 
FROM "trans_unit" 
INNER JOIN "trans_unit" T2 ON ("trans_unit"."source_unit_id" = T2."id") 
INNER JOIN "trans_unit_labels" ON (T2."id" = "trans_unit_labels"."unit_id") 
INNER JOIN "trans_label" ON ("trans_unit_labels"."label_id" = "trans_label"."id")
[...]

PR #17767, opened 2026-01-23
fix(search): handle multiple label filtering with AND

amCap1712:
Use Exists with a subquery to ensure each label condition gets its own join so that filtering on multiple labels with AND works.

Fixes https://github.com/WeblateOrg/weblate/issues/17720

### DIFF
--- a/weblate/utils/tests/test_search.py
+++ b/weblate/utils/tests/test_search.py
@@ -438,11 +438,6 @@ class UnitQueryParserTest(SearchTestCase):
             Q(check__name__iexact="ellipsis") & Q(check__dismissed=True),
         )
 
-    def test_labels(self) -> None:
-        self.assert_query(
-            "label:'test label'", Q(source_unit__labels__name__iexact="test label")
-        )
-
     def test_screenshot(self) -> None:
         self.assert_query(
             "screenshot:'test screenshot'",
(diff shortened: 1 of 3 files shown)


nijel:
Did you check if there are other search conditions that would be affected by this?

weblate/utils/tests/test_search.py
```diff
@@ -438,11 +438,6 @@ def test_checks(self) -> None:
             Q(check__name__iexact="ellipsis") & Q(check__dismissed=True),
         )
 
-    def test_labels(self) -> None:
```
amCap1712, line 441: Hard to add a test verifying the query part parsed because of its nested nature but the functional test in the other modified test file tests the functionality fine.

amCap1712:
Ah, good point. I took a look now and I think all one to many relations, checks (name), comments (text and author), suggestions (text and author), screenshots (name), are affected as well. However, to my understanding they didn't work with AND'ing previously either.

I guess we could add support for AND with all of them or wait until someone requests it. Personally, I do see value in being able to search for units with multiple checks.

nijel:
For checks it definitely makes sense to fix. For comments and suggestions it can stay as is. I'm unsure with screenshots, but it is probably more useful to search for string with multiple named screenshots, so let's fix that as well.

weblate/utils/tests/test_search.py
nijel: What exactly fails on this? It just compares the Q objects and tries whether they can be executed.

nijel:
I've decided to merge this to avoid the fix being blocked by additional fixes, reopened https://github.com/WeblateOrg/weblate/issues/17720 instead.

PR #17778, opened 2026-01-25
Account: Allow choosing commit name attribution

devimarj:
## Summary

This PR introduces anonymized commit names to complement the existing private e-mail feature, improving GDPR compliance. 

It implements an explicit state logic for this new field. Currently, with e-mails, users relying on a "private-by-default" setting become unintentionally public if the administrator later toggles the global default. This PR ensures that a user's commit name privacy preference is locked in and preserved, regardless of future changes to the global `PRIVATE_COMMIT_NAME_OPT_IN` setting.

## Changes

* **Models:** Added `commit_name` to `Profile` with fallbacks to prevent Git fatal errors (empty idents).
* **Logic:** Implemented explicit state saving for privacy persistence.
* **Integration:** Updated `weblate.auth.models.User.get_author_name()` for proper Git attribution.
* **Forms:** Added UI in profile settings that dynamically adjusts to global defaults and template availability.
* **Settings:** Introduced `PRIVATE_COMMIT_NAME_TEMPLATE` and `PRIVATE_COMMIT_NAME_OPT_IN`. Added `{user_id}` and `{site_domain}` support for both name and e-mail templates.
* **Tests & Docs:** Updated configuration docs and added comprehensive tests.

## Checklist

- [x] Tested the changes locally.
- [x] Added tests for the new logic.
- [x] Updated documentation (`docs/admin/config.rst`).


docs/admin/install/docker.rst
```diff
@@ -1018,6 +1018,18 @@ Generic settings
 
    Configures :setting:`PRIVATE_COMMIT_EMAIL_OPT_IN`.
 
+.. envvar:: WEBLATE_PRIVATE_COMMIT_NAME_TEMPLATE
+
+   .. versionadded:: 5.1x
```
nijel: Let's focus on 5.16

weblate/accounts/models.py
```diff
@@ -856,6 +862,11 @@ class Profile(models.Model):
         blank=True,
         max_length=EMAIL_LENGTH,
     )
+    commit_name = models.CharField(
+        verbose_name=gettext_lazy("Commit name"),
+        blank=True,
+        max_length=200,
```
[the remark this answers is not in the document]
nijel: I think using `IntegerChoices` and `IntegerField` will work better for this purpose.

weblate/accounts/models.py
```diff
@@ -1135,6 +1146,34 @@ def get_commit_email(self) -> str:
...
+        """Return the generated private commit name from the site template."""
+        if not settings.PRIVATE_COMMIT_NAME_TEMPLATE:
+            return ""
+
+        return settings.PRIVATE_COMMIT_NAME_TEMPLATE.format(
+            user_id=self.user.id,
+            username=self.user.username,
+            site_title=settings.SITE_TITLE,
+            site_domain=settings.SITE_DOMAIN.rsplit(":", 1)[0],
+        )
```
nijel, line 1175: I think renaming `format_private_email`, extending it to support `site_title` and reusing it here would be a better approach.

nijel (commented):
Thanks for the feature. This might be useful for some users even though you can use already any name in the profile (unlike with e-mail which is also used for notifications or password recovery).

Please refer to my review comments and add a changelog entry.

docs/changes.rst
```diff
@@ -11,6 +11,7 @@ Weblate 5.16
 * Added new management command :wladmin:`list_change_events`, which lists all possible change events, :ref:`addon-choice-events`.
 * Added Anthropic machinery integration, see :ref:`mt-anthropic`.
 * Encoding for :ref:`formats` can now be configured using :ref:`file_format_params` (e.g., ``csv_encoding``, ``properties_encoding``).
+* Added support for anonymous commit names via :setting:`PRIVATE_COMMIT_NAME_TEMPLATE`.
```
devimarj, line 14: should we also update  [/accounts/snippets/login-info.html](https://github.com/WeblateOrg/weblate/blob/main/weblate/templates/accounts/snippets/login-info.html)?

weblate/accounts/forms.py
```diff
@@ -223,6 +234,24 @@ def __init__(self, *args, **kwargs) -> None:
...
+            default_label = gettext_lazy("Use account name")
+
+        name_choices = [
+            (Profile.CommitNameChoices.DEFAULT, default_label),
+            (Profile.CommitNameChoices.PUBLIC, visible_name),
+        ]
+
+        if site_name:
+            name_choices.append((Profile.CommitNameChoices.PRIVATE, site_name))
+
```
devimarj, line 252: should we add a comprobation here?

`        if instance.commit_name == Profile.CommitNameChoices.PRIVATE and not site_name:
            self.initial["commit_name"] = Profile.CommitNameChoices.DEFAULT`

this implies that if the user saves the form, their 'Private' setting will be overwritten with 'Default'. The setting wouldn't automatically revert to 'Private' if the template is re-enabled later, but I assume this implicit cleanup is the desired behavior when the feature is disabled.

docs/changes.rst
nijel: Good catch, it indeed makes sense to adjust it.

weblate/accounts/forms.py
nijel: The cleanup would only happen for users saving the settings, so I don't think it's worth of the effort.

AliceVisek (approved):
I'm happy with the login snippet, it is clear and concise.

nijel:

Issue #16690, opened 2025-10-23
Extended validation of VCS settings


nijel:
### Describe the problem

Even though there are a lot of options to configure the VCS integration (see https://docs.weblate.org/en/latest/admin/continuous.html#pushing-changes-from-weblate), some of them are invalid, and there is currently no validation for that.

The documentation doesn't specifically mention such optiosn, because there are ambiguous. There is no way to create pull request to a branch from the same branch (it would be already updated).

* Merge request with a push URL and push and pull branch being the same.
* Merge request with a push URL and no push branch.

### Solution brainstorm

Extend `Component.clean_repo` to handle these cases.

### Additional context

* https://github.com/orgs/WeblateOrg/discussions/16223
* https://github.com/WeblateOrg/weblate/issues/16689

gersona:
Isn't the 1st case already covered here ?
https://github.com/WeblateOrg/weblate/blob/97ae7576ea603d2daa786f56d0d0ffae5af93db6/weblate/trans/models/component.py#L3385-L3393

nijel:
Mostly, as `repo` and `push` actually might be different URLs pointing to the same repo (for example, `https` and `ssh` URLs). I don't think we will ever be able to automatically detect this, so let's simplify the condition to check only for the push URL being set?

PR #17866, opened 2026-02-02
fix: clarify repo validation and help texts

nijel:
Make it easier to understand where the problem comes from.

Followup for #16690


gersona (approved):

micz:
Changing from

`URL of a push repository, pushing is turned off if empty.`

to

`URL of a push repository, pushing is turned off if empty. When using pull/merge requests, keep empty to fork the repository.`

is better, but I suspect someone could not understand what "using a fork" means.

Maybe it should be linked also this page from there: https://docs.weblate.org/en/latest/admin/continuous.html#push-changes
This documentation is what helped me understand why I was no more able to change branch, because I was using Github PR and I populated that field that was useless in this case.

Thank you!

PR #17880, opened 2026-02-03
docs: clarify tmserver maintenance status

nijel:


nijel:
pre-commit.ci autofix

PR #17933, opened 2026-02-08
Fix: make access radio options fully clickable

SWAPNILVERMA108:
This change makes the entire access option row clickable
by wrapping the content in a single label and improving layout.

I was not able to run a full Weblate instance locally due to
complex dependencies, but this change only affects templates
and styling.


weblate/templates/mail/style.css
```diff
@@ -1,3 +1,40 @@
...
+  font-weight: 500;
+}
+
+.radio-description {
+  margin-left: 28px;
+  margin-top: 4px;
+  font-size: 14px;
+  color: #666;
+}
+
```
nijel, line 37: This is definitely the wrong place; this CSS is only used in e-mail notifications.

weblate/templates/bootstrap5/layout/radioselect_access.html
```diff
@@ -19,68 +19,91 @@
...
-          {% if has_billing %}
-            <a href="{% url 'billing' %}">{% translate "Check your billing status" %}</a>
-          {% endif %}
-        </p>
-      </div>
-    {% endif %}
 
   </div>
-
-</div>
```
nijel, line 86: This only changes rendering for the access control selection in project settings, not for a generic radioselect.

weblate/templates/mail/style.css
SWAPNILVERMA108: Thanks for pointing that out.
I’ve removed the unrelated mail CSS change.

nijel:
@KarenKonou Please review this if this is something reasonable to maintain.

KarenKonou:
> @KarenKonou Please review this if this is something reasonable to maintain.

I don't think I have much to add in addition to your original review. I think I'd prefer to reimplement this myself if the PR author doesn't respond.

nijel:
Closing as per @KarenKonou comment.

PR #17936, opened 2026-02-08
fix(bug): Ensure check_cache isolation for suggestions

### LINKED ISSUE #12723: Cache key collision in render checks for suggestions
### Describe the issue

There is a cache key collision between renders generated in checks for the unit and for its suggestions.

### I already tried

1. Create a component with the max-size check
2. Create a unit in the source language and translate it in a target language
3. The render check (whether valid or incorrect) will display the render for the target string
4. Add a different suggestion to the same unit
5. The render check will display the render for the suggestion instead of the target string

### Expected behavior

_No response_

### Screenshots

_No response_

### Exception traceback

_No response_

### How do you run Weblate?

weblate.org service

### Weblate versions

_No response_

### Weblate deploy checks

_No response_

### Additional context

Originally addressed by @vinceh121 at https://github.com/WeblateOrg/weblate/pull/12673 but that proven to be causing other issues, see https://github.com/WeblateOrg/weblate/issues/12721

What we really need:

* Add a test case covering this scenario (rendering check triggering on a string and on a suggestion for that string)
* Verify that all views work
* Verify that the image is different in each case

Without that, we're changing code that is not covered by tests and easy to make things wrong.

devimarj:
### Description
This PR addresses the cache key collision between parent unit checks and their suggestions, as described in #12723.

### Technical Details
- **The Issue:** `Suggestion.get_checks()` was using a shallow copy of the parent `Unit`. Since the `check_cache` dictionary is a mutable object, the `fake_unit` and the original `Unit` shared the same memory reference. This led to cache poisoning, where suggestion validation results (including renders) overwrote the parent unit's data.
- **The Fix:** Explicitly re-initializing `fake_unit.check_cache = {}` ensures a clean, isolated memory space for each suggestion's validation, preventing any leak to the parent unit or other suggestions.

### Changes
- **Model:** Updated `weblate/trans/models/suggestion.py` to reset `check_cache`.
- **Tests:** Added `test_rendering_check_isolation` in `weblate/trans/tests/test_suggestions.py` to verify memory isolation and data integrity.
- **Documentation:** Added a note to `docs/changes.rst`.

Fixes #12723

### DIFF
--- a/weblate/trans/models/suggestion.py
+++ b/weblate/trans/models/suggestion.py
@@ -129,9 +129,8 @@ class Suggestion(models.Model, UserDisplayMixin):
     timestamp = models.DateTimeField(auto_now_add=True)
 
     votes = models.ManyToManyField(
-    settings.AUTH_USER_MODEL, through="Vote", related_name="user_votes"
+        settings.AUTH_USER_MODEL, through="Vote", related_name="user_votes"
     )  # type: ignore[var-annotated]
-
 
     objects = SuggestionManager.from_queryset(SuggestionQuerySet)()
 


weblate/trans/models/suggestion.py
```diff
@@ -144,7 +144,7 @@ def __str__(self) -> str:
 
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
-        self.fixups = []
+        self.fixups: list = []
```
nijel: We can make it more specific:

```suggestion
        self.fixups: list[str] = []
```

weblate/trans/models/suggestion.py
```diff
@@ -128,7 +128,7 @@ class Suggestion(models.Model, UserDisplayMixin):
     userdetails = models.JSONField(default=dict)
     timestamp = models.DateTimeField(auto_now_add=True)
 
-    votes = models.ManyToManyField(
+    votes: models.ManyToManyField = models.ManyToManyField(
```
nijel: Does this really improve anything for type checkers?

devimarj: that's actually trash

nijel:

PR #17937, opened 2026-02-08
feat: Add button to copy suggestion to translation field (fix #17827)

### LINKED ISSUE #17827: Add button to copy suggestion to translation field
### Describe the problem

Very useful if you want to add edits to suggested variant if user lacks privilege to save the translation and can only add suggestions.

### Solution brainstorm

Add button somewhere at the right of suggestion text field to copy suggestion to translation field

_No response_

### Screenshots

_No response_

### Additional context

_No response_

_Originally posted by @Blueberryy in https://github.com/WeblateOrg/weblate/discussions/17785_

devimarj:
### Description
This PR implements a button to copy suggestion text directly into the translation field, requested in #17827

This feature improves the UX for translators who need to modify a suggestion before saving it, or for users without save privileges who want to use a suggestion as a base for a new one.

### Implementation
* **Backend (`Suggestion` model):** Added a `@property target_list` that uses `split_plural` to correctly expose suggestion targets as a list. This ensures compatibility with languages that have multiple plural forms.
* **Frontend (JS):** 
    * Implemented logic in `full.js` to handle the copy action.
    * Used Weblate's internal `.replaceValue` method instead of standard `.val()`. **This preserves the Undo/Redo history** and triggers necessary change events properly.
    * Added logic to handle multiple plural fields (populating `data-text-0`, `data-text-1`, etc.).
    * Included `autosize.update()` to resize textareas after insertion.
* **UI:** Added a copy button to the suggestion snippet template.

Fixes #17827


weblate/templates/snippets/suggestions.html
```diff
@@ -35,6 +35,14 @@
...
                         title="{% translate "Vote against" %}">{% icon "thumb-down.svg" %}</button>
               {% endif %}
+              {# Copy suggestion feature #}
+              <button type="button"
+                      class="btn btn-link green js-copy-suggestion"
+                      title="{% translate "Clone to translation" %}"
+                      aria-label="{% translate "Clone to translation" %}"
+                      {% for text in suggestion.target_list %}data-text-{{ forloop.counter0 }}="{{ text|force_escape }}"{% endfor %}>
+                {% icon "clone.svg" %}
+              </button>
```
nijel: This block should be in `if user_can_suggest`.

devimarj: thanks for the feedback!

nijel:

PR #17712, opened 2026-01-20
Add bulk accept suggestions feature

Hendrik292:
This PR adds a "bulk accept suggestions" feature that allows users to accept all suggestions from a specific user with a single button click.

### Features
- Backend view (`bulk_accept_user_suggestions`) handling suggestion acceptance
- Frontend button appearing next to each suggestion author
- Shows progress and status ("X accepted", percentage, "Done")
- Auto-reloads page after completion

### Implementation
- **New view:** `weblate/trans/views/bulk_suggestions.py` (91 lines)
- **New JavaScript:** `weblate/static/js/bulk-accept-suggestions.js` (74 lines)
- **New CSS:** `weblate/static/styles/bulk-accept.css` (56 lines)
- **Modified:** `weblate/urls.py` (added route)
- **Modified:** `weblate/templates/snippets/suggestions.html` (added button)

### Related
- Anna's Archive bounty: https://software.annas-archive.li/AnnaArchivist/annas-archive/-/issues/312
- Implements official integration (vs. external bookmarklet/nginx injection from first bounty)

### Testing
- Tested with multiple suggestions from different users
- Permissions properly checked (requires `suggestion.accept`)
- Works for all suggestions from a user across the translation
- Error handling for failed accepts
- Button appears only for authenticated users with proper permissions^

Screenshot_73

PR #17938, opened 2026-02-08
feat: Enhance bulk accept with quality checks & Django messages

devimarj:
### Description
This PR proposes some enhancements to the excellent bulk accept feature introduced in #17712 by @Hendrik292.

While Weblate offers robust tools to easily revert changes if something goes wrong, "Bulk Accept" is still a powerful action. I thought it might be beneficial to add an extra validation step to prevent potential issues upfront, rather than relying on reverting them later.

To align the feature further with Weblate's workflow and prepare it for potential UI updates, I implemented the following ideas:

### Key Changes

1.  **Quality Checks Validation:**
    * The view now verifies `suggestion.get_checks()` before accepting.
    * **Behavior:** Any suggestion failing a quality check is **skipped** and remains in the list for manual review, while valid ones are processed.
    * **Goal:** To help maintainers filter out suggestions that might need small adjustments without stopping the whole batch.

2.  **Standardized Feedback (Django Messages):**
    * Transitioned the feedback logic to **`django.contrib.messages`**.
    * **Why:** This keeps the feedback consistent with other Weblate backend actions. It also decouples the logic from the specific button location—if we decide to move this action to the **"Tools" dropdown** in the future to save space, the feedback mechanism will still work perfectly.
    * **States:** Handles success, partial success (warnings), and errors natively.

3.  **Accessibility (a11y) & UX:**
    * Minor adjustments to the frontend status text ("Processed") to improve the screen reader experience during the page reload transition.

### Summary
I hope these additions help make the feature even more robust and consistent with the rest of the framework.

Thanks again to @Hendrik292 for the solid foundation!

### DIFF
--- a/docs/user/translating.rst
+++ b/docs/user/translating.rst
@@ -61,7 +61,8 @@ those matching current translations.
 If you have the :guilabel:`Accept suggestions` permission, you can bulk accept
 all suggestions from a specific user by clicking the account check icon next
 to their suggestion. This is useful for quickly processing suggestions from
-trusted contributors.
+trusted contributors. Suggestions failing quality checks are skipped and
+remain for manual review.
 
 .. _user-comments:
 


nijel:

PR #17980, opened 2026-02-11
Force XWiki File Formats to use a template fixes #14712

### LINKED ISSUE #14712: regression: bad support of XWikiPageProperties in 9.5.2
### Describe the issue

Since the upgrade to Weblate 9.5.2 I noticed a regression when adding a new language in components that uses XWikiPageProperties translation format: this format is supposed to remove some parts of the source XML files and to transform some fields: e.g. the `language` attribute is suppose to contains the actual language code. 
Apparently it's not the case anymore, see this test PR showing the problems: https://github.com/xwiki-contrib/application-activitypub/pull/46

I investigated a bit and I locally tried to improve a bit the test_formats to test this, but I haven't been able to reproduce the failure in those tests: apparently there the format is properly taken into account. 

### I already tried

1. Create a component using the XWikiPageProperties format 
2. Add a new language and perform a translation
3. Check the obtained file

### Expected behavior

The file shouldn't be a mirror of the original XML file, but it should contain quite a few modifications, like the language attribute.

### Screenshots

_No response_

### Exception traceback

```pytb

```

### How do you run Weblate?

PyPI module

### Weblate versions

_No response_

### Weblate deploy checks

```shell

```

### Additional context

_No response_

surli:

### DIFF
--- a/weblate/formats/ttkit.py
+++ b/weblate/formats/ttkit.py
@@ -2136,7 +2136,7 @@ class XWikiPropertiesFormat(PropertiesBaseFormat):
     loader = ("properties", "xwikifile")
     language_format = "linux"
     autoload: tuple[str, ...] = ("*.properties",)
-    empty_file_template = "\n"
+    empty_file_template = None
     can_add_unit: bool = False
     can_delete_unit: bool = False
     set_context_bilingual: bool = True
@@ -2146,7 +2146,6 @@ class XWikiPropertiesFormat(PropertiesBaseFormat):
     # comments are preserved as in the original source file.
     def save_content(self, handle) -> None:
         current_units = self.all_units
-        store_units = self.store.units
 
         # We empty the store units since we want to control what we'll serialize
         self.store.units = []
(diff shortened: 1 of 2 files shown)


weblate/formats/ttkit.py
```diff
@@ -2136,7 +2136,7 @@ class XWikiPropertiesFormat(PropertiesBaseFormat):
     loader = ("properties", "xwikifile")
     language_format = "linux"
     autoload: tuple[str, ...] = ("*.properties",)
-    empty_file_template = "\n"
+    empty_file_template = None
```
nijel, line 2139: Is it intentional for all formats? The properties formats should work just fine.

nijel (commented):
The tests do not pass, probably need to be adjusted to match changed file format behavior.

weblate/formats/ttkit.py
surli: Yes it's intentional: even for the base XWiki Java Properties we want to use a template to get the missing properties and comments from source translation file. It's actually quite cleaner like that.

surli:
@nijel tests are now fixed

nijel:

PR #18074, opened 2026-02-17
feat: use logical location for importing translations

### LINKED ISSUE #18023: Improve initial import of HTML and Markdown translations
### Describe the problem

Presently, the initial import of translations for Markdown and HTML is likely to go wrong if the documents were generated by something besides translate-toolkit as the source and translation locations do not match. The reason is that it includes line numbers, which is likely to be different in (manually) translated HTML or Markdown files.

### Solution brainstorm

Introduce some kind of logical identification of the strings to match when parsing the initial translation. This could be based on element position (for example, “third paragraph”).

_No response_

### Screenshots

_No response_

### Additional context

https://github.com/WeblateOrg/weblate/blob/4e5cda450ffe866c29c0a8182f241befc54952b4/weblate/formats/convert.py#L253-L260

nijel:
This shold work more reliable if file structure is not changed, only formatting.

Fixes #18023

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -11,6 +11,7 @@ Weblate 5.16.1
 .. rubric:: Improvements
 
 * Improved documentation for translation states to clarify the difference between :guilabel:`Needs editing`, :guilabel:`Needs rewriting`, and :guilabel:`Needs checking` states.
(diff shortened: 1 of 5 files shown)


nijel:
Needs https://github.com/translate/translate/pull/6098 to let the intermediate pofile handle docpath.

PR #18106, opened 2026-02-19
feat(docs): Automatically generate format features documentation

### LINKED ISSUE #9077: Move file format features from table to format pages

gersona:
- fixes #9077

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -21,6 +21,7 @@ Weblate 5.17
 * Improved performance of project and category search result pages with very large match sets.
 * :envvar:`WEBLATE_COMMIT_PENDING_HOURS` is now available in Docker container.
 * Improved documentation with auto-generated snippets for :ref:`addons`, :ref:`fmt_capabs`, :ref:`checks`, and :ref:`machine-translation` machines.
+* Added :setting:`PROJECT_WEB_RESTRICT_PRIVATE` to reject project website and repository browser URLs targeting non-global IP ranges, and exposed it in Docker as :envvar:`WEBLATE_PROJECT_WEB_RESTRICT_PRIVATE`.
 * Improved performance of :ref:`mt-weblate` lookups.
 * Screenshot and font upload forms now honor :setting:`ALLOWED_ASSET_SIZE` which now defaults to 10 MB.
 
@@ -47,6 +48,7 @@ Weblate 5.17
 * Improved :ref:`backup` status reporting while keeping maintenance after failed backup attempts.
 * POT update add-ons now fall back to the component URL for the ``Report-Msgid-Bugs-To`` header when the component setting is empty.
 * Improved repository lock error handling when deleting units.
+* :ref:`check-max-size` previews now keep the configured text box visible and render overflowing text in red.
 
 .. rubric:: Compatibility
 
@@ -62,6 +64,7 @@ Weblate 5.17
 Please follow :ref:`generic-upgrade-instructions` in order to perform update.
 
 * There are several changes in :file:`settings_example.py`, most notably :setting:`ADMINS` syntax has changed in Django and ``SOCIAL_AUTH_PIPELINE`` and ``INSTALLED_APPS`` need adjustments; please adjust your settings accordingly.
+* Project website and repository browser URLs pointing to private or other non-global IP ranges are now rejected by default. If your setup intentionally links to internal addresses, set :setting:`PROJECT_WEB_RESTRICT_PRIVATE` to ``False`` or :envvar:`WEBLATE_PROJECT_WEB_RESTRICT_PRIVATE` to ``0``.
 
 .. rubric:: Contributors
 
--- a/docs/admin/config.rst
+++ b/docs/admin/config.rst
@@ -1681,6 +1681,7 @@ Default configuration:
 
    * :ref:`project-web`
    * :setting:`PROJECT_WEB_RESTRICT_NUMERIC`
+   * :setting:`PROJECT_WEB_RESTRICT_PRIVATE`
    * :setting:`PROJECT_WEB_RESTRICT_RE`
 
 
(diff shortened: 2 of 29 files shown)


gersona:
@nijel could you briefly have a look at the implementation suggestion before I move on to implement this for the other formats ?

gersona:
That's a fair point, many flags can definitely be read directly from the format class.
My concern is with format-specific features that don't map to any existing class attribute (and they can get pretty long if we keep them in the format class), like
 https://github.com/gersona/weblate/blob/c3cdbee736cdea21861c917a24d3372a0cb1ed43/weblate/formats/docs.py#L67-L72
The intent was to split concern and keep this kind of doc-only metadata out of the format class and avoid adding attributes to it purely for documentation purposes.

nijel:
To make this easier to deal with, let's start with the features we already have flags for. Some are obvious, like `supports_explanation`, and some are a bit hidden, like `SUPPORTS_FUZZY` and will need conversion.

gersona:
I actually already got something quite ready for review, with a different approach than what I initially proposed..
I can push it later today.

nijel:
I think this is a good direction. Can we get rid of existing SUPPORTS_FUZZY and use the new flags instead?

weblate/formats/external.py
```diff
@@ -28,6 +28,7 @@
 class XlsxFormat(CSVFormat):
     name = gettext_lazy("Excel Open XML")
     format_id = "xlsx"
+    monolingual = True
```
gersona: The monolingual attribute has been added to match what was previously defined in the documentation. 
https://github.com/WeblateOrg/weblate/blob/89e3d86d1cbc60d5a8b0e04080c267e4c70c1cee/docs/formats.rst?plain=1#L245-L246
I assumed that the documentation is correct and the Python classes had to be updated accordingly.
Same for `JSONFormat, GWTFormat` and a couple of others.

Is this assumption correct ? because those edits also impact the class behavior, which now causes some tests to fail https://github.com/WeblateOrg/weblate/actions/runs/23444649999/job/68205005080?pr=18106

gersona:
one of the comments in #9077 mentions a doc page with an overview of features and the file formats that support them. Should it also be in the scope of this PR ?

weblate/formats/external.py
nijel: JSON and XLS can be both mono- and bilingual. GWT is monolingual.

nijel:
If all info from the “Translation types capabilities” is now covered directly in the formats doc, it should be removed.

docs/snippets/format-features/gwt-features.rst
```diff
@@ -0,0 +1,26 @@
...
+   See :ref:`format-context`
+:Supports location: `No`
+   See :ref:`format-location`
+:Supports flags: `No`
+   See :ref:`format-flags`
+:API identifier: `gwt`
+:Supports read-only strings: `No`
+   See :ref:`read-only-strings`
+
+.. AUTOGENERATED END: format-features gwt
```
nijel, line 28: I've taken this as an example to comment on the stying:

* Use double backticks for string literals such as extension or API indentifier
* Use the full name for monolingual/bilingual instead of just the short name. This is already confusing enough for new users, and shortening it makes it even worse.
* The See ... is rendered next to the yes/no, which looks weird.
* Overall, the features listing seems to consume too much space: https://weblate--18106.org.readthedocs.build/en/18106/formats/gwt.html#supported-features

gersona: Using a list table might be a better solution then,

gersona: It would look like this, does it look good ?
Capture d’écran 2026-03-25 à 10 57 06

nijel: Looks great. I'd also add a `.` before the extension to make it more obvious (`.properties` instead of just `properties`).

nijel: And perhaps calling that "File extensions" so that the scope is clear.

weblate/formats/management/commands/list_format_features.py
```diff
@@ -0,0 +1,213 @@
...
+
+        def new_list_table_row(
+            lines, feature_name: str, value: str | bool, doc_link: str | None = ""
+        ) -> None:
+            if isinstance(value, bool):
+                value = "Yes" if value else "No"
+            if not value:
+                return
+
+            doc_link = f":ref:`↗ <{doc_link}>`" if doc_link else ""
```
nijel: How about an info icon, which is the same as we use in the UI?

```suggestion
            doc_link = f":ref:`🛈 <{doc_link}>`" if doc_link else ""
```

weblate/formats/ttkit.py
```diff
@@ -1982,6 +2022,7 @@ class YAMLFormat(DictStoreFormat):
     unit_class = MonolingualSimpleUnit
     autoload: tuple[str, ...] = ("*.pyml",)
     empty_file_template = "{}\n"
+    monolingual = True
```
nijel: This can also be bilingual (see the testsuite failure):

```suggestion
```

But the subclasses should be monolingual.

docs/changes.rst
```diff
@@ -15,6 +16,7 @@ Weblate 5.17
 * Improved error messages in some of the :ref:`api` endpoints.
 * Improved performance of project and category search result pages with very large match sets.
 * :envvar:`WEBLATE_COMMIT_PENDING_HOURS` is now available in Docker container.
+* Improved documentation with auto-generated snippets for :ref:`addons`, :ref:`fmt_capabs`, :ref:`checks`, and :ref:`machine-translation` machines.
```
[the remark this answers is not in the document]
gersona: the other snippet generators are included in PR #18152, but the changelog was not updated then

nijel:
There is some issue with generating extensions, as many files have `.php` in it (for example, https://weblate--18106.org.readthedocs.build/en/18106/formats/gotext.html).

Also, the extensions should be as individual literals, not as whole comma joined:

```
``.ini, .json`` is wrong
``.ini``, ``.json`` is correct
```

In general, string literals should be only used for actual literals (extensions and API identifier), rest should be plain text.

weblate/formats/management/commands/list_format_features.py
```diff
@@ -0,0 +1,216 @@
...
+        value: str | bool | list[str],
+        doc_link: str | None = "",
+    ) -> None:
+        if isinstance(value, bool):
+            value = "Yes" if value else "No"
+
+        if isinstance(value, list):
+            value = ", ".join([f"``{item}``" for item in value])
+        else:
+            value = f"``{value}``"
```
nijel: Avoid using code blocks for non-literal values:

```suggestion
```

PS: I hope I did read the code correctly that both API identifiers and extensions are passed as list, anything else as boolean or string.

gersona: My bad, I skipped that

nijel:

PR #18108, opened 2026-02-19
docs: update Python environment documentation

nijel:
Use "Python environment" instead of a bit cryptic "virtualenv". This also matches how uv documents it.

### DIFF
--- a/docs/admin/install.rst
+++ b/docs/admin/install.rst
@@ -1538,9 +1538,9 @@ Sample configuration for NGINX and uWSGI
 
 To run production webserver, use the WSGI wrapper installed with Weblate (when
 using a Python environment it is installed as
-:file:`~/weblate-env/lib/python3.9/site-packages/weblate/wsgi.py`). Don't
-forget to set the Python search path to your Python environment as well (for example
-using ``virtualenv = /home/user/weblate-env`` in uWSGI).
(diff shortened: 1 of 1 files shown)


docs/admin/management.rst
```diff
@@ -19,15 +19,15 @@ Invoking management commands
 
 As mentioned before, invocation depends on how you installed Weblate.
 
-If using virtualenv for Weblate, you can either specify the full path to
-:command:`weblate`, or activate the virtualenv prior to invoking it:
+If using Python environment for Weblate, you can either specify the full path to
```
nijel:
```suggestion
If using a Python environment for Weblate, you can either specify the full path to
```

docs/admin/upgrade.rst
```diff
@@ -62,11 +62,11 @@ releases should work, but is not as well tested as single version upgrades!
...
         cd weblate-src
         git pull
-        # Update Weblate inside your virtualenv
+        # Update Weblate inside your Python environment
         . ~/weblate-env/bin/uv pip install -e '.[all]'
-        # Install dependencies directly when not using virtualenv
+        # Install dependencies directly when not using Python environment
         uv pip install --upgrade -e .
-        # Install optional dependencies directly when not using virtualenv
+        # Install optional dependencies directly when not using Python environment
```
nijel:
```suggestion
        # Install optional dependencies directly when not using a Python environment
```

PR #18114, opened 2026-02-19
fix(vcs): correctly detect stale Git locks

nijel:
The regex had a typo and that was not caught because of missing tests.


nijel:
@sentry review

PR #18117, opened 2026-02-19
docs: add rsvg dependency to suse install command

michael-smt:
Without it you get a `ValueError: Namespace Rsvg not available`.


nijel:

nijel:
That reminded me that I'd like to get rid of this dependency: https://github.com/WeblateOrg/weblate/issues/18118

PR #18133, opened 2026-02-20
fix(import): use source explanation correctly during component creation

### LINKED ISSUE #17643: Explanation is not parsed from TBX
### Describe the issue

TBX parsing doesn't seem to work as described in https://github.com/WeblateOrg/weblate/issues/12795

### I already tried

* Add glossary component from https://github.com/openSUSE/suse-terminology/blob/
* Check for strings having an explanation, for example, https://github.com/openSUSE/suse-terminology/3001e87143f447d85d58e73977721572c4065f73/TBX/Weblate_single_lang/Export_for_Weblate_DE-DE_(TBX).xml#L18815

### Expected behavior

The explanation should be parsed.

### Screenshots

_No response_

### Exception traceback

```pytb

```

### How do you run Weblate?

weblate.org service

### Weblate versions

_No response_

### Weblate deploy checks

```shell

```

### Additional context

* https://github.com/WeblateOrg/weblate/issues/12795
* https://github.com/WeblateOrg/weblate/pull/14998
* https://github.com/translate/translate/pull/5601

amCap1712:
During component creation, `Component.bulk_create_sources` is used to create source units. This method wrongly used the `explanation` attribute read from TTKUnit instead of `source_explanation` for setting the explanation on the source db unit.

`Unit.update_from_unit` which is used for updating source units at other places correctly passed `source_explanation` so sync-ing existing components worked fine and probably why the issue was not caught the time TBX extended metadata was implemented.

Fixes #17643.


nijel:
Good catch! Scheduling merge.

PR #18136, opened 2026-02-20
feat(search): AND multiple checks and screenshots in one search

### LINKED ISSUE #17720: Filtering by label with an AND condition does not work as expected

amCap1712:
Similar to the changes introduced for multi label AND search in https://github.com/WeblateOrg/weblate/pull/17767.

Fix #17720

### DIFF
--- a/weblate/utils/search.py
+++ b/weblate/utils/search.py
@@ -721,6 +721,7 @@ class UnitTermExpr(BaseTermExpr):
         with AND using exists ensures each check condition gets its own subquery.
         """
         from weblate.checks.models import Check
+
         lookup = "name__iexact" if self.operator == ":=" else "name__icontains"
         return Q(
             Exists(
@@ -756,7 +757,7 @@ class UnitTermExpr(BaseTermExpr):
         with AND using exists ensures each screenshot condition gets its own subquery.
         """
         from weblate.screenshots.models import Screenshot
-        
+
         lookup = "name__iexact" if self.operator == ":=" else "name__icontains"
         screenshot_query = Screenshot.objects.filter(**{lookup: text})
         return Q(
--- a/weblate/utils/search.py
+++ b/weblate/utils/search.py
@@ -33,10 +33,8 @@ from pyparsing import (
     one_of,
 )
 
-from weblate.checks.models import Check
 from weblate.checks.parser import RawQuotedString
 from weblate.lang.models import Language
-from weblate.screenshots.models import Screenshot
 from weblate.trans.models import Category, Component, Label, Project, Translation
 from weblate.trans.util import PLURAL_SEPARATOR
 from weblate.utils.db import re_escape, using_postgresql
@@ -722,6 +720,7 @@ class UnitTermExpr(BaseTermExpr):
         This is needed because filtering on a reverse ForeignKey relation
         with AND using exists ensures each check condition gets its own subquery.
         """
+        from weblate.checks.models import Check
         lookup = "name__iexact" if self.operator == ":=" else "name__icontains"
         return Q(
             Exists(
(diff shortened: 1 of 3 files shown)


weblate/utils/search.py
```diff
@@ -717,6 +715,52 @@ def label_field(self, text: str, context: dict) -> Q:
...
+        )
+
+    def screenshot_field(self, text: str, context: dict) -> Q:
+        """
+        Handle screenshot filtering.
+
+        This is needed because filtering on ManyToMany relations
+        with AND using exists ensures each screenshot condition gets its own subquery.
+        """
+        lookup = "name__iexact" if self.operator == ":=" else "name__icontains"
```
nijel, line 761:
```suggestion
        from weblate.screenshots.models import Screenshot
        
        lookup = "name__iexact" if self.operator == ":=" else "name__icontains"
```

weblate/utils/search.py
```diff
@@ -717,6 +715,52 @@ def label_field(self, text: str, context: dict) -> Q:
...
         return Q(Exists(label_query.filter(unit__id=OuterRef("source_unit_id"))))
 
+    def check_field(self, text: str, context: dict) -> Q:
+        """
+        Handle check filtering.
+
+        This is needed because filtering on a reverse ForeignKey relation
+        with AND using exists ensures each check condition gets its own subquery.
+        """
+        lookup = "name__iexact" if self.operator == ":=" else "name__icontains"
```
nijel, line 725:
```suggestion
        from weblate.checks.models import Check
        lookup = "name__iexact" if self.operator == ":=" else "name__icontains"
```

weblate/utils/search.py
```diff
@@ -717,6 +715,52 @@ def label_field(self, text: str, context: dict) -> Q:
...
+        )
+
+    def dismissed_check_field(self, text: str, context: dict) -> Q:
+        """
+        Handle dismissed check filtering.
+
+        This is needed because filtering on a reverse ForeignKey relation
+        with AND using exists ensures each check condition gets its own subquery.
+        """
+        lookup = "name__iexact" if self.operator == ":=" else "name__icontains"
```
nijel, line 743:
```suggestion
        from weblate.checks.models import Check

        lookup = "name__iexact" if self.operator == ":=" else "name__icontains"
```

weblate/utils/search.py
```diff
@@ -33,8 +33,10 @@
     one_of,
 )
 
+from weblate.checks.models import Check
 from weblate.checks.parser import RawQuotedString
 from weblate.lang.models import Language
+from weblate.screenshots.models import Screenshot
```
nijel: The imports need to be inside the code to avoid circular deps.

```suggestion
from weblate.checks.parser import RawQuotedString
from weblate.lang.models import Language
```

PR #18138, opened 2026-02-20
docs(detailed state): improve documentation on needs checking/rewriting

### LINKED ISSUE #17865: Add state:source-changed filter (or equivalent) for Automatic translation add-on
### Problem

I use the **Automatic translation add-on** with machine translation to keep my translations up to date. My ideal workflow is:

- **Empty strings** → auto-translate via MT, mark as needs-editing for human review
- **Source string changed** → discard old translation, re-translate via MT, mark as needs-editing for human review

There is no way to distinguish "needs editing because the source changed" from "needs editing for any other reason" in the add-on's filter field.

### Solution brainstorm

Add a filter 'state:source-changed'.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

amCap1712:
Document the role of needs rewriting and needs checking state. And also, include it in the search docs.

Fix #17865

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -9,6 +9,8 @@ Weblate 5.16.1
 
 .. rubric:: Improvements
 
+* Improved documentation for translation states to clarify the difference between :guilabel:`Needs editing`, :guilabel:`Needs rewriting`, and :guilabel:`Needs checking` states.
+
 .. rubric:: Bug fixes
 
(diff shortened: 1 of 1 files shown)


amCap1712:
Should we also include an example on the search page in Weblate itself? Haven't added one yet because the addon should also be installed for the states to be assigned needs checking/rewriting in the first place.

nijel:
I think this is good to go.

PR #18200, opened 2026-02-24
chore(db): remove mysql/mariadb support

### LINKED ISSUE #16140: Drop support for MySQL/MariaDB

amCap1712:
fixes #16140

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -11,6 +11,8 @@ Weblate 5.17
 
 .. rubric:: Compatibility
 
+* Dropped support for MySQL and MariaDB as the database engine.
+
 .. rubric:: Upgrading
 
 Please follow :ref:`generic-upgrade-instructions` in order to perform update.
--- a/docs/changes/v4.rst
+++ b/docs/changes/v4.rst
@@ -938,7 +938,7 @@ Notable configuration or dependencies changes:
 
 * There is a change in :setting:`django:INSTALLED_APPS`, ``weblate.configuration`` has to be added there.
 * Django 3.1 is now required.
-* In case you are using MySQL or MariaDB, the minimal required versions have increased, see :ref:`mysql`.
+* In case you are using MySQL or MariaDB, the minimal required versions have increased.
 
 .. versionchanged:: 4.4.1
 


nijel:
`weblate/trans/migrations/0001_squashed_weblate_5.py` migration still imports MySQL constants.

amCap1712:
There are the following places where we can simplify code or move some of the computation to the database. Should I do that now or in a later PR?

1. pending unit error query
https://github.com/WeblateOrg/weblate/blob/45401006f492f1cb5b20db9fea8c5eb3d1266156/weblate/trans/models/pending.py#L178-L185
2. case sensitive comparison
https://github.com/WeblateOrg/weblate/blob/9862b6320bf48af281b3b054934f5d5955944465/weblate/trans/tasks.py#L247-L253
https://github.com/WeblateOrg/weblate/blob/1693021ebbd720817bcd06562986ee356d417b9c/weblate/trans/forms.py#L946
https://github.com/WeblateOrg/weblate/blob/8a492512c4fc4d4fd741f19cb2092533978bee98/weblate/trans/util.py#L129
https://github.com/WeblateOrg/weblate/blob/81ba5ef270a752113882789a32c313bd9ca75ed9/weblate/trans/models/suggestion.py#L60
3. varint index
https://github.com/WeblateOrg/weblate/blob/365d9ed591e9f50926307a081b3750aa4c518cea/weblate/trans/models/variant.py#L16-L21
4. exception handling
https://github.com/WeblateOrg/weblate/blob/45401006f492f1cb5b20db9fea8c5eb3d1266156/weblate/utils/search.py#L480-L482

nijel:
We can do that separately, each in a pull request. We can also simplify the CI setup as well by using GitHub actions services rather than spawning them manually using docker compose.

amCap1712:
I looked into it and that should be possible but it has the downside that there is no way to modify the postgres command in the service container currently so fsync can't be disabled and other perf improvements for tests cannot be applied. This has been a long standing issue but it seems just two weeks ago the GH actions team started working on it: https://github.com/orgs/community/discussions/160506#discussioncomment-15789600. So I think we should do that change later when the feature is available.

nijel:
Honestly, I'm not sure how much these actually affect performance in the CI environment, so removing these is definitely something to test.

Anyway, 5.16.1 is out, so we can merge this. Will you please look into the merge conflicts?

amCap1712:
Resolved the merge conflict, can do the CI without fsync testing in a separate PR.

nijel:

PR #18229, opened 2026-02-26
search: add mentions to query example

michael-smt:
This allows to quickly find all unresolved comments where one is mentioned.


nijel:

Issue #18234, opened 2026-02-26
Simplifications after mysql support removal


amCap1712:
With mysql support removal in #18200, we can simplify the following:

Varint index
https://github.com/WeblateOrg/weblate/blob/365d9ed591e9f50926307a081b3750aa4c518cea/weblate/trans/models/variant.py#L16-L21

Exception handling
https://github.com/WeblateOrg/weblate/blob/45401006f492f1cb5b20db9fea8c5eb3d1266156/weblate/utils/search.py#L480-L482

We can also simplify the CI setup as well by using GitHub actions services rather than spawning them manually using docker compose.

PR #18243, opened 2026-02-26
ci(services): use service containers instead of custom docker compose files

amCap1712:
#18234

### DIFF
--- a/.github/workflows/migrations.yml
+++ b/.github/workflows/migrations.yml
@@ -22,6 +22,8 @@ jobs:
     runs-on: ubuntu-24.04
     name: postgresql
     env:
+      # This is needed for older versions to migrate, can be removed once migration from 5.17+ only is supported
+      CI_DATABASE: postgresql
       CI_DB_PASSWORD: weblate
       CI_DB_HOST: 127.0.0.1
       CI_DB_PORT: '5432'


.github/workflows/test.yml
```diff
@@ -59,14 +59,36 @@ jobs:
...
+          POSTGRES_PASSWORD: weblate
+        ports:
+        - 5432:5432
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+      cache:
+        image: ${{ matrix.datastore == 'redis' && 'redis:8.6.1-alpine@sha256:2afba59292f25f5d1af200496db41bea2c6c816b059f57ae74703a50a03a27d0' || 'valkey/valkey:9.0.3-alpine@sha256:ad09f25978db13ed3ed7c635a0e43118568deca056ed8f7a5020f40632e2923e' }} # zizmor: ignore[unpinned-images]
```
nijel: I wonder if we still need this. I added it when switching to valkey as default to see if it has any impact, but their promise is to keep API consistent, so maybe sticking with valkey only would be good enough (and simplify the test matrix). In the end we're not doing any low-level calls here.

amCap1712:
I would like to remove the CI_DATABASE variable. Are you fine with that?

https://github.com/WeblateOrg/weblate/blob/86e15de60298424c3fc39b083f85f2312d4f72a8/weblate/settings_test.py#L19-L27

nijel:
I don't see an issue with this.

amCap1712:
Yeah, CI runtimes seem similar to before.

.github/workflows/migrations.yml
```diff
@@ -20,28 +20,35 @@ permissions:
...
     runs-on: ubuntu-24.04
-    strategy:
-      matrix:
-        database: [postgresql]
-      fail-fast: false
-    name: ${{ matrix.database }}
+    name: postgresql
     env:
-      CI_DATABASE: ${{ matrix.database }}
+      CI_DATABASE: postgresql
```
nijel:
```suggestion
```

amCap1712: Migrations need this variable because they checkout code from older versions where the variable is mandated.

nijel: You're right, sorry for messing this up.

.github/workflows/migrations.yml
```diff
@@ -20,28 +20,34 @@ permissions:
 jobs:
   migrations:
     runs-on: ubuntu-24.04
-    strategy:
-      matrix:
-        database: [postgresql]
-      fail-fast: false
-    name: ${{ matrix.database }}
+    name: postgresql
     env:
```
nijel, line 24:
```suggestion
    env:
      # This is needed for older versions to migrate, can be removed once migration from 5.17+ only is supported
      CI_DATABASE: postgresql
```

nijel:

PR #18247, opened 2026-02-26
chore(ci): simplify pip-install ci script

amCap1712:
Binaries for lxml/xmlsec are now available on pypi.

Merge after https://github.com/WeblateOrg/weblate/pull/18243.

### DIFF
--- a/ci/pip-install
+++ b/ci/pip-install
@@ -9,7 +9,7 @@ set -e -x
 if [ "${1:-latest}" = minimal ]; then
     # Adjust deps to force minimal version
     sed -i '/^ *"/ s/>=/==/' pyproject.toml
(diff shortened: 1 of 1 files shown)


nijel:
The binaries are there, but the challenge is to get a compatible install. Both wheels need to be built against the same libxml2. See the test failure with the minimal deps (lxml & xmlsec libxml2 library version mismatch). We might adjust lowest supported versions to avoid this for now, but the issue will pop up again when new lxml is released and new xmlsec is not yet built against the same libxml2.

nijel:
See also https://github.com/xmlsec/python-xmlsec/issues/356

amCap1712:
makes sense.

PR #18260, opened 2026-02-28
chore(pending): remove MySQL/MariaDB workarounds from pending change query

### LINKED ISSUE #18232: Simplify pending unit error query
With the removal of mysql support in #18200, the following query can be simplified:

https://github.com/WeblateOrg/weblate/blob/45401006f492f1cb5b20db9fea8c5eb3d1266156/weblate/trans/models/pending.py#L178-L185

amCap1712:
Now that MySQL/MariaDB support has been dropped (#18200), simplify the retry eligibility query by removing the `CAST(..., TextField())` around `KT()` which are not needed for postgres.

Fix #18232.

### DIFF
--- a/weblate/trans/models/pending.py
+++ b/weblate/trans/models/pending.py
@@ -18,7 +18,6 @@ from django.db.models import (
     OuterRef,
     Q,
     Subquery,
-    TextField,
     When,
 )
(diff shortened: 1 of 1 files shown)


nijel:

Issue #9315, opened 2023-05-26
"value too long for type character varying" when adding a variant


burner1024:
### Describe the issue

An exception when adding a variant to a specific string.

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar issues in this repository.

### Steps to reproduce the behavior

1. Try to add a variant. A long one, I guess.

### Expected behavior

No exception, variant added.

### Exception traceback

```pytb
Internal Server Error: /new-unit/infinity-engine/ascension/pl/

DataError at /new-unit/infinity-engine/ascension/pl/
value too long for type character varying(576)


Request Method: POST
Request URL: https://hive.bgforge.net/new-unit/infinity-engine/ascension/pl/
Django Version: 4.0.4
Python Executable: /usr/local/bin/python
Python Version: 3.10.4
Python Path: ['/', '/usr/local/lib/python3.10/site-packages/git/ext/gitdb', '/', '/usr/local/bin', '/usr/local/lib/python310.zip', '/usr/local/lib/python3.10', '/usr/local/lib/python3.10/lib-dynload', '/usr/local/lib/python3.10/site-packages', '/app/data/python', '/usr/local/lib/python3.10/site-packages/gitdb/ext/smmap']
Server time: Thu, 25 May 2023 18:12:42 +0000
Installed Applications:
['customize',
 'weblate.addons',
 'weblate.auth',
 'weblate.checks',
 'weblate.formats',
 'weblate.glossary',
 'weblate.machinery',
 'weblate.trans',
 'weblate.lang',
 'weblate_language_data',
 'weblate.memory',
 'weblate.screenshots',
 'weblate.fonts',
 'weblate.accounts',
 'weblate.configuration',
 'weblate.utils',
 'weblate.vcs',
 'weblate.wladmin',
 'weblate.metrics',
 'weblate',
 'weblate.gitexport',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django.contrib.admin.apps.SimpleAdminConfig',
 'django.contrib.admindocs',
 'django.contrib.sitemaps',
 'django.contrib.humanize',
 'social_django',
 'crispy_forms',
 'compressor',
 'rest_framework',
 'rest_framework.authtoken',
 'django_filters']
Installed Middleware:
['weblate.middleware.RedirectMiddleware',
[...]

Issue #7661, opened 2022-05-18
Slow variant update


burner1024:
### Describe the issue

So I'm marking variants through API (`PATCH /api/units/${id}/`), a few thousands of them.
It started with ~20-30 seconds per API call when there were a few variants overall. Now there's over 600, and each call takes about 1.5 minutes.
(I also short-circuited `invalidate_cache` in `component.py`, because each patch would trigger it, slowing the process further).

By my estimation, at this rate, the process is going to take over 2 weeks.

Most of the time seems to be spent in postgresql, with queries like this
```sql
UPDATE "trans_unit" SET "variant_id" = 39 WHERE "trans_unit"."id" IN (SELECT U0."id" FROM "trans_unit" U0 INNER JOIN "trans_translation" U1 ON (U0."translation_id" = U1."id") WHERE (U1."component_id" = 16 AND U0."variant_id" IS NULL AND (U0."source" = 'I said... ' OR U0."id_hash" IN (9191887938771624418)))) 
```
It looks like each API call triggers reprocessing for every `variant_id`, which results in quickly degrading performance.

### I already tried

- [X] I've read and searched [the documentation](https://docs.weblate.org/).
- [X] I've searched for similar issues in this repository.

### Steps to reproduce the behavior

1. Mark variants.
2. Keep marking variants.

### Expected behavior

Faster variant marking in the first place, and not degrading performance with number of variants.

### How do you run Weblate?

Docker container

### Weblate versions

* Weblate: 4.12.2
 * Django: 4.0.4
 * siphashc: 2.1
 * translate-toolkit: 3.6.2
 * lxml: 4.6.5
 * Pillow: 9.1.0
 * bleach: 5.0.0
 * python-dateutil: 2.8.2
 * social-auth-core: 4.2.0
 * social-auth-app-django: 5.0.0
 * django-crispy-forms: 1.14.0
 * oauthlib: 3.2.0
 * django-compressor: 4.0
 * djangorestframework: 3.13.1
 * django-filter: 21.1
 * django-appconf: 1.0.5
 * user-agents: 2.2.0
 * filelock: 3.6.0
 * jellyfish: 0.9.0
 * openpyxl: 3.0.9
 * celery: 5.2.6
 * kombu: 5.2.4
 * translation-finder: 2.12
 * weblate-language-data: 2022.2
 * html2text: 2020.1.16
 * pycairo: 1.21.0
 * pygobject: 3.42.1
 * diff-match-patch: 20200713
[...]

PR #18262, opened 2026-02-28
misc: change index to TextField and remove OperationalError

### LINKED ISSUE #18234: Simplifications after mysql support removal

amCap1712:
fix #18234

### DIFF
--- a/weblate/trans/models/unit.py
+++ b/weblate/trans/models/unit.py
@@ -810,8 +810,7 @@ class Unit(models.Model, LoggerMixin):
         if new_variant:
             variant = Variant.objects.filter(
                 key__md5=MD5(Value(new_variant))
-            ).get_or_create(key=new_variant, component=component, variant_regex=""
-            )[0]
+            ).get_or_create(key=new_variant, component=component, variant_regex="")[0]
             variant.defining_units.add(self)
 
         # Update variant links


weblate/trans/models/variant.py
```diff
@@ -13,9 +13,7 @@ class Variant(models.Model):
         "trans.Component", on_delete=models.deletion.CASCADE, db_index=False
     )
     variant_regex = RegexField(max_length=VARIANT_REGEX_LENGTH, blank=True)
-    # This really should be a TextField, but it does not work with unique
-    # index and MySQL
-    key = models.CharField(max_length=VARIANT_KEY_LENGTH)
```
amCap1712, line 18: I wonder if we need to add manual validation for this at application level now because even with Postgres there will be a max limit on total size of an entry in b-tree index.

nijel:
There is also https://github.com/WeblateOrg/weblate/issues/9315 that this is fixing.

We could also use MD5 in the unique index to avoid limits fully (and that whould probably also perform better on larger texts). That could also be utilized for faster lookup (we do something similar with units already). I doubt that MD5 collisions are something that is likely to happen in this scope.

nijel:
See also https://github.com/WeblateOrg/weblate/issues/7661 which might be addressable with the index change as well.

amCap1712:
Looking into the slow variant updates, 

https://github.com/WeblateOrg/weblate/blob/35c9675a67229132a2be382e27c6ecee42f0885b/weblate/trans/models/component.py#L3722-L3724

variant is assigned here but not used. Not sure if its a bug or if it should be removed.

Edit: removed.

nijel:

Issue #18309, opened 2026-03-02
Component discovery do not copy all properties from base component


bogusdominica:
### Describe the issue

Component discovery on create do not copy "Contribute to project translation memory" and "Do not show glossary matches" value from parent component.
If base component have option "Contribute to project translation memory" disabled it still create new components with this status enabled.
If base component have option "Do not show glossary matches" enabled it still create new components with status as disabled.

### I already tried

- [x] I've read and searched [the documentation](https://docs.weblate.org/).
- [x] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

Create component, disable option "Contribute to project translation memory" and enable option "Do not show glossary matches".
Install component discovery addon.
Newly created child components will have this options in other states.

### Expected behavior

Options statuses copied from parent correctly.

### Exception traceback

```pytb

```

### How do you run Weblate?

Docker container

### Weblate versions

5.16.1

### Weblate deploy checks

```shell

```

PR #18310, opened 2026-03-02
Fixed bug #18309

bogusdominica:
Copy attributes hide_glossary_matches, contribute_project_tm of parent component in component discovery.


nijel:

PR #18404, opened 2026-03-05
fix(unit): restore translation state when source string is no longer readonly

### LINKED ISSUE #18181: Translations stay read-only when source string leaves needs editing state when reviews and intermediate file active
### Describe the issue

When the source string is set to "_Needs editing_" all translations are also set to read-only with the note "_The source string needs review_". But since Weblate 5.16, when the source string state changes from "_Needs editing_" to another state, the translations remain read-only.

This only appears to happen when both reviews are enabled and a intermediate file is configured.

### I already tried

1. Use a component with intermediate file in a project where reviews are enabled
2. Set a source string to "_Needs editing_"
3. Check that other languages are in "_Source in review_"
4. Set the source string to "_Waiting for review_"
5. Check another language, it is ready-only.

### Expected behavior

Translations are no longer read-only.

### Screenshots

_No response_

### Exception traceback

```pytb

```

### How do you run Weblate?

Docker container

### Weblate versions

5.16.0.0

### Weblate deploy checks

```shell

```

### Additional context

_No response_

amCap1712:
`get_unit_state()` included the source-based readonly check when computing `original_state` during file syncs, corrupting it to STATE_READONLY. When `update_state()` later tried to restore the state, it saw state == original_state == STATE_READONLY and did nothing. Do not consider weblate originating readonly states when computing `original_state` in file sync.

Fix #18181.

### DIFF
--- a/weblate/trans/models/unit.py
+++ b/weblate/trans/models/unit.py
@@ -833,23 +833,26 @@ class Unit(models.Model, LoggerMixin):
         flags: Flags | str | None,
         string_changed: bool = False,
         disk_unit_state: StringState | None = None,
+        include_weblate_readonly: bool = True,
     ) -> StringState:
         """Calculate translated and fuzzy status."""
-        # Read-only from the file format
+        # Read-only from the file format is always considered for the unit state
         if unit.is_readonly():
             return STATE_READONLY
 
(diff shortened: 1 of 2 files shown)


nijel:

PR #18415, opened 2026-03-06
fix(unit): preserve detailed state on files rescan

### LINKED ISSUE #18371: Detailed needs editing states do not survive parsing
### Describe the issue

When string is marked as "Needs rewrite" parsing from the file discards the state to regular "Needs editing".

### I already tried

1. Create component with PO file (or anything else that stores needs editing state)
2. Change string to "Needs rewrite" state
3. Trigger repository parsing (for example by other string change)
4. See the string state is changed to "Needs editing"

### Expected behavior

The "Needs rewrite" state should be preserved as long as the string is marked `fuzzy` in the PO file.

### Screenshots

_No response_

### Exception traceback

```pytb

```

### How do you run Weblate?

Docker container

### Weblate versions

Weblate 5.16.1 

### Weblate deploy checks

```shell

```

### Additional context

Can be seen on https://localizationlab.weblate.cloud/translate/shira-public/ready-made-quiz/ready-made/fr/?checksum=29c79aefdf012274

amCap1712:
Instead of returning STATE_FUZZY in `get_unit_state`, return the more specific substate if unit has one assigned to it in Weblate.

Fixes #18371

### DIFF
--- a/weblate/trans/models/unit.py
+++ b/weblate/trans/models/unit.py
@@ -858,6 +858,11 @@ class Unit(models.Model, LoggerMixin):
         # support saving it
         is_existing_fuzzy_state = self.fuzzy or disk_unit_state in FUZZY_STATES
         if unit.is_fuzzy(is_existing_fuzzy_state and not string_changed):
+            # Preserve specific fuzzy sub-state when possible
+            if self.state in FUZZY_STATES:
+                return self.state
+            if disk_unit_state in FUZZY_STATES:
+                return disk_unit_state
             return STATE_FUZZY
 
         if not unit.is_translated():
--- a/weblate/trans/tests/test_models.py
+++ b/weblate/trans/tests/test_models.py
@@ -47,6 +47,7 @@ from weblate.utils.state import (
     STATE_APPROVED,
     STATE_FUZZY,
     STATE_NEEDS_CHECKING,
+    STATE_NEEDS_REWRITING,
     STATE_READONLY,
     STATE_TRANSLATED,
 )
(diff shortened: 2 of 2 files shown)


amCap1712:
While working on this, I realized a separate issue.

Consider a project with a commit policy that prevents fuzzy pending changes to be committed and a unit in that project that never been translated before or changes for which have not been written to file yet (because waiting for review).  The `unit.details["disk_state"]["state"]` will have `STATE_EMPTY`.

If the first or the latest change to such a unit is in a fuzzy state, `get_unit_state` will return `STATE_FUZZY` (from fallback) for formats that don't support writing fuzzy state to the disk but `STATE_EMPTY` (because has_translation is False) for formats that do.

https://github.com/WeblateOrg/weblate/blob/370adc360759d3453598eb0b3b4074539093456a/weblate/trans/models/unit.py#L857-L858
https://github.com/WeblateOrg/weblate/blob/370adc360759d3453598eb0b3b4074539093456a/weblate/formats/ttkit.py#L158-L166

In the case of `STATE_FUZZY` being returned, `update_from_unit` wipes all the pending changes which is incorrect behavior.

There are two possible solutions I think, either `has_translation` in `TTKUnit` is always checked and not just for formats that support fuzzy state in file. Or in `Unit.get_unit_state` we check for STATE_EMPTY before checking for other states.

nijel:
Okay, let's address this as well, but do not hurry for today's release. I'm a bit lost in your description (even after reading several times); can you start with a test case showing it?

In general, the behavior should be: 

* If the string and its state in the file weren't changed, do not change the state in Weblate. 
* If the string or its state was changed in the file, discard Weblate pending changes and use the state from the file.

amCap1712:
I think we can merge this PR for now if you are satisfied, I'll open an another one with those test cases.

nijel:
Yes, this one looks good.

PR #18421, opened 2026-03-06
fix(unit): preserve fuzzy state for formats without fuzzy support

amCap1712:
Consider the following: There is an **untranslated** unit in a file format that doesn't support writing fuzzy state to the file. The project has been configured with approved only commit policy. The unit is translated to a fuzzy state. As it is the first uncommitted change to the unit, `unit.details["disk_state"]["state"]"` will be updated and set to `STATE_EMPTY`. `unit.state` is `STATE_FUZZY`. There is a `PendingUnitChange` in the database to commit the translation to disk.

When a file sync is triggered, it attempts a commit. The pending change above is not committed because it is not approved and nothing changes so far. During file sync, `Unit.update_from_unit` is called on this unit which in turn calls `Unit.get_unit_state` to get current unit state. `unit.state` is `STATE_FUZZY` so `unit.is_fuzzy(fallback=True)` is called.

https://github.com/WeblateOrg/weblate/blob/370adc360759d3453598eb0b3b4074539093456a/weblate/trans/models/unit.py#L857-L858

JSON format is not `SUPPORTS_FUZZY` so the fallback is returned directly.

https://github.com/WeblateOrg/weblate/blob/370adc360759d3453598eb0b3b4074539093456a/weblate/formats/ttkit.py#L158-L166

As a result, `Unit.get_unit_state` returns `STATE_FUZZY`. `comparison_state["state"]` is STATE_EMPTY. `same_state` evaluates to `False` and the pending change is deleted from the database.

https://github.com/WeblateOrg/weblate/blob/370adc360759d3453598eb0b3b4074539093456a/weblate/trans/models/unit.py#L1102

### DIFF
--- a/weblate/formats/ttkit.py
+++ b/weblate/formats/ttkit.py
@@ -159,10 +159,12 @@ class TTKitUnit[U: TranslateToolkitUnit, F: "BaseTTKitFormat"](TranslationUnit[U
         """Check whether unit needs editing."""
         if not self.has_unit():
             return fallback
+        if not self.has_translation():
+            return False
         # Most of the formats do not support this, but they
         # happily return False
         if isinstance(self.unit, SUPPORTS_FUZZY):
-            return self.has_translation() and self.unit.isfuzzy()
+            return self.unit.isfuzzy()
         return fallback
 
     def has_content(self) -> bool:


weblate/trans/tests/test_remote.py
```diff
@@ -809,6 +814,41 @@ def test_tbx_explanation_sync(self) -> None:
...
+
+        unit = translation.unit_set.get(source="Hello, world!\n")
+        self.assertEqual(unit.state, STATE_FUZZY)
+
+        # pending changes for unit are not discarded
+        count = PendingUnitChange.objects.filter(unit=unit).count()
+        self.assertEqual(count, 1)
+        self.assertEqual(unit.details["disk_state"]["state"], STATE_EMPTY)
+
+    def test_fuzzy_state_commit_policy_json(self) -> None:
```
amCap1712, line 844: The json format test should fail without a fix demonstrating the bug.

amCap1712:
My suggested fix is to take `self.has_translation()` out of the `SUPPORTS_FUZZY` if so that its always called in `TTKUnit.is_fuzzy`.

nijel:
The proposed solution makes sense, something like:

```py
     if not self.has_unit() or not self.has_translation(): 
         return fallback
```

amCap1712:
Actually, that won't work. If the unit is empty, existing behavior (here and in the custom implement in XliffUnit too) is to return False instead of fallback.

nijel:

PR #18450, opened 2026-03-11
fix(ui): change multi widget scaling behavior

### LINKED ISSUE #17694: Multi-select widgets are too big
### Describe the issue

The multi-select widgets seem to be sized based on the screen size and number of items. This is especially annoying with large screens and selects that render on few lines.

### I already tried

1. Go to /manage/teams/
2. Click on any team
3. See the selection taking the whole screen while there are just few lines of items to select

### Expected behavior

- The widget should not take more space than needed for currently shown items.
- It might scroll if there is too many items to show.
- We should revisit whether there isn't some better implementation of this functionality. The currently used `multi.js` hasn't received updates since 2021.

### Screenshots

Image

### Exception traceback

```pytb

```

### How do you run Weblate?

Git checkout

### Weblate versions

_No response_

### Weblate deploy checks

```shell

```

### Additional context

This probably exists since https://github.com/WeblateOrg/weblate/pull/12426, but I've only noticed it now with a larger screen 😄

KarenKonou:
fixes #17694

### DIFF
--- a/weblate/static/styles/new-language.css
+++ b/weblate/static/styles/new-language.css
@@ -6,7 +6,8 @@
 
 .multi-wrapper .non-selected-wrapper,
 .multi-wrapper .selected-wrapper {
-  height: 35vh;
+  height: auto;
+  max-height: 35vh;
   min-height: 200px;
 }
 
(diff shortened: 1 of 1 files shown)


nijel:

PR #18458, opened 2026-03-12
fix: correct color styling for button links

### LINKED ISSUE #16919: Link button icon colors
### Describe the issue

The Skip button on the translate page has mixed colors because the inline SVG is not styled the same as the link text.

### I already tried

1. https://hosted.weblate.org/translate/weblate/languages/cs/?checksum=20f8357285a09729
2. See the skip button

### Expected behavior

The color should be the same as text.

### Screenshots

Image

### Exception traceback

```pytb

```

### How do you run Weblate?

Git checkout

### Weblate versions

_No response_

### Weblate deploy checks

```shell

```

### Additional context

This could be fixed by adding a `green` class to the element, but I think it should not be needed for the default rendering where the SVG path inside `btn-link` should match the text.

The `green` might then be removed from all `btn-link` elements, because it was apparently used to address this before.

KarenKonou:
fixes #16919

### DIFF
--- a/weblate/static/style-dark.css
+++ b/weblate/static/style-dark.css
@@ -1013,6 +1013,12 @@ fieldset[disabled] .btn-link {
   text-decoration-color: currentcolor;
   background-color: transparent;
 }
+a.btn-link:hover {
+  color: #6cf8d8;
+  path {
+    fill: #6cf8d8;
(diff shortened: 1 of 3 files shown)


nijel:

PR #18460, opened 2026-03-12
feat(ui): clarify styling of explanation box

### LINKED ISSUE #16534: Clarify style of "explanation" box
### Describe the problem

Since I added an explanation to a string, several of our translators have translated the explanation instead of translating the actual string. The reason for this is likely that the explanation looks exactly like the source language box. If you don't realize that there are suddenly 2 boxes and you still look at the topmost one, you can easily miss it.

Image

### Solution brainstorm

I think adding an info icon and changing the background color could help to make it more clear

Image

Maybe yellow like a warning message? Then people definitely won't translate it. Also, they will maybe more likely read+follow the explanation

Image

### Screenshots

_No response_

### Additional context

_No response_

KarenKonou:
fixes #16534


nijel:

PR #18467, opened 2026-03-13
fix: use effective default reviewers for bitbucket cloud

subnix:
Bitbucket Cloud has a hierarchical system of settings: workspace -> project -> repository. Some settings, like default reviewers, may be configured at a project level and inherited by a repository. Thus, we should use [effective default reviewers](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-pullrequests/#api-repositories-workspace-repo-slug-effective-default-reviewers-get) to obtain reviewers for a pull request.

### DIFF
--- a/weblate/vcs/git.py
+++ b/weblate/vcs/git.py
@@ -2417,7 +2417,12 @@ class BitbucketCloudRepository(GitMergeRequestBase):
         except RepositoryError:
             return []
 
-        return [reviewer["user"]["uuid"] for reviewer in reviewers]
+        result = []
+        for reviewer in reviewers:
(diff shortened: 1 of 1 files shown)


nijel:
Can you please adjust tests to the changed behavior? Won't it break existing configurations?

subnix:
> Can you please adjust tests to the changed behavior?

Done

> Won't it break existing configurations?

This change should be backward compatible since the required privileges remain the same, and default reviewers are only used for PR creation. I've tested this on a local instance and it works fine. Additionally, I've added some defensive code to safeguard against potential response parsing issues.

nijel:

PR #18524, opened 2026-03-18
fix(utils): consistently use regex for user provided expressions

nijel:
This provides consistent experience and allows us to limit time used to evaluate the expressions.


weblate/trans/models/component.py
```suggestion
expression = "".join(result)
return re.compile(f"^{expression}$")
```

PR #18542, opened 2026-03-19
Support data URI auth images and fix auth name display

felixfon:
- Added support for data URIs in the image parameter.
- Replaced spacers with CSS margins for better layout control and to prevent link underlines from touching the image.
- Implemented vertical alignment for images and text.

### DIFF
--- a/weblate/accounts/templatetags/authnames.py
+++ b/weblate/accounts/templatetags/authnames.py
@@ -12,7 +12,6 @@ from django import template
 from django.conf import settings
 from django.contrib.staticfiles.storage import staticfiles_storage
 from django.utils.html import format_html
-from django.utils.safestring import mark_safe
 from django.utils.translation import gettext_lazy
 
 from weblate.accounts.utils import get_key_name
@@ -59,15 +58,11 @@ SECOND_FACTORS: dict[DeviceType, StrOrPromise] = {
     "recovery": gettext_lazy("Use recovery codes"),
 }
 
-IMAGE_SOCIAL_TEMPLATE = """
-<img class="auth-image" src="{image}" />
-"""
+IMAGE_SOCIAL_TEMPLATE = (
+    """<img class="auth-image" src="{image}" alt="" aria-hidden="true" />"""
+)
 
-SOCIAL_TEMPLATE = """
-{icon}
-{separator}
-{name}
-"""
+SOCIAL_TEMPLATE = """{icon}<span class="auth-name">{name}</span>"""
 
 
 def get_auth_params(auth: str) -> dict[str, StrOrPromise]:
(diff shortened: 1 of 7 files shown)


weblate/accounts/templatetags/authnames.py
```diff
@@ -59,15 +58,11 @@
     "recovery": gettext_lazy("Use recovery codes"),
 }
 
-IMAGE_SOCIAL_TEMPLATE = """
-<img class="auth-image" src="{image}" />
-"""
+IMAGE_SOCIAL_TEMPLATE = """<img class="auth-image" src="{image}" />"""
```
nijel: The name should be next to the image, so clearly hide this from the screen readers:

```suggestion
IMAGE_SOCIAL_TEMPLATE = """<img class="auth-image" src="{image}" alt="" aria-hidden="true"/>"""
```

weblate/accounts/templatetags/authnames.py
```diff
@@ -92,22 +87,20 @@ def get_auth_params(auth: str) -> dict[str, StrOrPromise]:
     return params
 
 
-auth_name_default_separator = mark_safe("<br />")
-
-
 @register.simple_tag
-def auth_name(auth: str, separator: str = auth_name_default_separator, only: str = ""):
+def auth_name(auth: str, only: str = ""):
```
nijel, line 91: Removing the default separator breaks the add new identity tab in the user profile which relied on this.

felixfon: The template isn't passing that argument anymore, so it should be safe! I've checked [the code here](https://github.com/WeblateOrg/weblate/blob/main/weblate/templates/accounts/profile.html#L249) and it seems fine.
It works perfectly on my device, too. Did I miss a spot somewhere else?
image

nijel: Before this PR, the names were below the icons because it used the default separator `<br>`.

felixfon: > Before this PR, the names were below the icons because it used the default separator `<br>`.

Thanks for the heads-up. I've fixed it.
image

weblate/accounts/templatetags/authnames.py
```diff
@@ -96,30 +89,26 @@ def get_auth_params(auth: str) -> dict[str, StrOrPromise]:
...
 
 @register.simple_tag
-def auth_name(auth: str, separator: str = auth_name_default_separator, only: str = ""):
+def auth_name(auth: str, only: str = ""):
     """Create HTML markup for social authentication method."""
     params = get_auth_params(auth)
 
-    if not params["image"].startswith("http"):
+    if not params["image"].startswith(("http", "data:")):
         params["image"] = staticfiles_storage.url(f"auth/{params['image']}")
```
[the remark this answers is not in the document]
nijel: While this is valid feedback, this setting is in the hands of admin, so we can assume the value to be safe.

nijel:

PR #18556, opened 2026-03-20
docker: support setting default commit pending hours

michael-smt:


nijel:

Issue #10101, opened 2023-10-05
Support overriding password reset URL


nijel:
We have a docker installation, and the login is with LDAP.

The "Forgot your password?" link goes to: `https://https://weblate.my.org/accounts/reset/`
But the location of the of the LDAP registration and edition is located elsewhere like `https:/ldap.my.org/`

Looking at this [line](https://github.com/WeblateOrg/weblate/blob/ac50f44f72e75e37ec51f27ea8a0667808b32238/weblate/templates/accounts/login.html#L58) **reset_url** is the one I need to change to go to the real LDAP registration/reset page.

This documentation does not show how to go to the LDAP registration/reset page
https://docs.weblate.org/en/latest/admin/auth.html#ldap-authentication

How can I solve the problem?
Regards

_Originally posted by @cvvergara in https://github.com/WeblateOrg/weblate/discussions/10042_

KunalBhat55:
Hey the link doesn't seem to work? `https:/ldap.my.org/`

nijel:
The goal is to be able to customize the password reset link, not to hard-code a specific link, which was provided as an example.

kk271kg:
Hello,
Can I contribute on this issue? I would like to try to solve this.

nijel:
Yes, all contributions are welcome.

Shashankpantiitbhilai:
I am interested please assign this to me

nijel:
@Shashankpantiitbhilai you're welcome to work on the issue.

PR #18564, opened 2026-03-21
Configure login password-reset link via PASSWORD_RESET_URL

baltenaxis:
### Motivation

Fix issue #10101 by allowing LDAP/external-auth deployments to send users to the correct password reset page.

### What this PR changes

- Adds a configurable `PASSWORD_RESET_URL` for the login page reset link.
- Updates login logic/template so **“Forgot your password?”** uses the configured URL, or falls back to internal `password_reset` when not set.
- Adds tests for configured vs. non-configured reset URL behavior.
- Updates docs for settings, Docker env var (`WEBLATE_PASSWORD_RESET_URL`), and LDAP example.

### Testing

- Added unit tests in `weblate.accounts.tests.test_views` for reset-link behavior.


weblate/accounts/views.py
```diff
@@ -898,8 +898,12 @@ def show_login_form(self) -> bool:
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         auth_backends = get_auth_keys()
+        reset_url = getattr(settings, "PASSWORD_RESET_URL", None) or getattr(
+            settings, "EXTERNAL_PASSWORD_RESET_URL", None
+        )
```
nijel: Please include PASSWORD_RESET_URL in WeblateAccountsConf, so that it can be accessed directly from the settings.

weblate/accounts/views.py
```diff
@@ -898,8 +898,12 @@ def show_login_form(self) -> bool:
...
         context["login_backends"] = [x for x in sorted(auth_backends) if x != "email"]
-        context["can_reset"] = self.has_email_auth
+        context["can_reset"] = self.has_email_auth or bool(reset_url)
+        context["reset_url"] = reset_url
```
nijel: This might be better to pass using the existing context processor in weblate/trans/context_processors.py.

PR #18624, opened 2026-03-24
fix(accounts): use wall-clock time for 2fa password recheck

nijel:
### Motivation
- A 2FA re-verification gate used `time.monotonic()` to timestamp session verification, which is relative to system boot and can differ across hosts or after reboot, allowing the password-change 2FA check to be bypassed in multi-host or restart scenarios.

### Description
- Replace monotonic-based comparisons with wall-clock UNIX timestamps derived from `timezone.now().timestamp()` in `weblate/accounts/views.py` for both the password-change freshness check and when storing the second-factor completion timestamp.
- Remove the `monotonic` import and ensure the session stores an `int(timezone.now().timestamp())` value instead of `int(monotonic())` to make cross-request/session validation stable across workers and reboots (`weblate/accounts/views.py`).
- Add regression tests in `weblate/accounts/tests/test_twofactor.py` (`test_password_requires_recent_second_factor` and `test_password_allows_recent_second_factor`) that patch `weblate.accounts.views.timezone.now` and use `SECOND_FACTOR_VERIFY_SECONDS` and `SESSION_SECOND_FACTOR_TIMESTAMP` to validate stale and fresh timestamp behavior.

### Testing
- `ruff check weblate/accounts/views.py weblate/accounts/tests/test_twofactor.py` completed successfully.
- `python -m compileall weblate/accounts/views.py weblate/accounts/tests/test_twofactor.py` completed successfully.
- Attempted `pytest weblate/accounts/tests/test_twofactor.py -k 'password_requires_recent_second_factor or password_allows_recent_second_factor'`, but pytest could not be executed to completion in this environment due to a missing `gi` module referenced by the test config (PytestConfigWarning), so the tests could not be run here; the added tests are deterministic and should pass in a normal development CI environment.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69c2a2de50488329b5f8734d0ac34658)

### DIFF
--- a/weblate/accounts/tests/test_twofactor.py
+++ b/weblate/accounts/tests/test_twofactor.py
@@ -244,9 +244,9 @@ class TwoFactorTestCase(FixtureTestCase):
             mocked_now.return_value = now()
 
             session = self.client.session
-            session[SESSION_SECOND_FACTOR_TIMESTAMP] = int(
-                mocked_now.return_value.timestamp()
-            ) - SECOND_FACTOR_VERIFY_SECONDS
+            session[SESSION_SECOND_FACTOR_TIMESTAMP] = (
+                int(mocked_now.return_value.timestamp()) - SECOND_FACTOR_VERIFY_SECONDS
+            )
             session.save()
 
             response = self.client.get(reverse("password"))


nijel:
Obsoleted by https://github.com/WeblateOrg/weblate/pull/18625

PR #18633, opened 2026-03-25
fix: 'Save and stay' button color improvement

sloria:
The previous yellow coloring created an odd hierarchy.

## Before

"Save and stay" is disabled in looks clickable and has too much visual weight. The yellow also communicates a potentially dangerous action.

light-before

dark-before

## After

Both disabled buttons look disabled and have equal visual weight, leading unauth'd users correctly to the "Suggest" button.

light-after

dark-after

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -38,7 +38,7 @@ jobs:
         python-version: '3.14'
     - uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7.6.0
       with:
-        version: 0.11.0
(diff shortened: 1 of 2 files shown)


sloria:
Looks like all CI builds are currently failing due to a failure in building pyroma ultimately caused by a [bug](https://github.com/pypa/build/pull/980) in `build` that was released today. The [fix](https://github.com/pypa/build/pull/1003) was just merged, so builds should hopefully work again once that's released

nijel:

gaborbernat:
https://github.com/pypa/build/releases/tag/1.4.2 is out now fixing this, thanks!

PR #18638, opened 2026-03-25
fix(formats): align implicit encoding fallback defaults

nijel:
### Motivation

- The encoding used to select Translate Toolkit loader variants is taken from `file_format_params`, and when no encoding param is provided `get_encoding_param` returns `None`, so `TTKitFormat.get_class()` falls back to the first entry in a format's `loader` dict.  
- The `loader` dict ordering for `StringsFormat`, `PropertiesFormat`, and `GWTFormat` did not match the documented defaults (`utf-16`, `iso-8859-1`, `utf-8` respectively), causing components created without explicit `file_format_params` to pick the wrong parser.

### Description

- Reordered the `loader` mapping keys in `StringsFormat`, `PropertiesFormat`, and `GWTFormat` so the first mapping matches each format's documented default encoding.  
- Added a small regression test `test_encoding_loader_defaults` in `weblate/formats/tests/test_formats.py` that asserts the first key of each affected `loader` equals the documented default.  
- Added a short changelog entry in `docs/changes.rst` noting the restored implicit encoding fallback behavior.

### Testing

- Ran the focused unit tests with `uv run pytest weblate/formats/tests/test_formats.py -k "get_class or encoding_loader_defaults"`, and the selected tests passed (`2 passed`).  
- Ran the linter with `uv run ruff check --fix weblate/formats/tests/test_formats.py weblate/formats/ttkit.py` to apply trivial fixes, which completed successfully and addressed the reported issues.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69c3a89354b48329beb6dee7d1ba56e7)


nijel:
@gersona This went out of Codex scanning. Please check if that actually makes sense.

weblate/formats/ttkit.py
```diff
@@ -533,6 +533,8 @@ def get_class(cls, encoding: str | None = None) -> type[S]:
             if encoding in cls.loader:
                 module_name, class_name = cls.loader[encoding]
             else:
+                # The first loader variant is the implicit default used when no
+                # explicit encoding is configured for the component.
```
nijel, line 537: What is really not clear to me is how `default` plays with this.

gersona: The intuition is probably good, it would cover very specific edge-cases. (e.g An older component loaded from backups where no file format parameter is defined yet, or the user manually changes the `Component.file_format_params` attribute).

In practice, `TTKitFormat.get_class()` is called without encoding parameter when fetching data unrelated to encoding (e.g extension, mimetype).

A more robust solution would be to load the default param values for a given format before trying to get the encoding param, so the fallback value is always the same regardless of the `loader` orders.

Something like
```python
def get_encoding_param(file_format: str, file_format_params: FileFormatParams | None) -> str | None:
    """Get encoding parameter from file format parameters."""
    defaults = get_default_params_for_file_format(file_format)
    file_format_params = defaults | file_format_params
    for param_name, value in file_format_params.items():
        try:
            if get_param_for_name(param_name).is_encoding():
                return cast("str", value)
        except ValueError:
            continue
    return None
```

gersona: This would also ensure `None` is returned only when there's no encoding variant for the format

nijel:
https://github.com/WeblateOrg/weblate/pull/18716 uses suggested approach instead.

PR #18647, opened 2026-03-25
fix(machinery): treat empty Anthropic base_url as unset and fall back to default

nijel:
### Motivation

- The Anthropic machinery form allows the `base_url` to be left empty, but runtime URL construction used `urljoin` on the stored empty string producing a relative URL and causing requests to raise `MissingSchema`, breaking the integration. 
- The change restores the documented behaviour of "leave empty to use the default URL" by normalizing empty values before building the endpoint URL.

### Description

- Normalize `base_url` in `weblate/machinery/anthropic.py` to use `self.settings.get("base_url") or "https://api.anthropic.com"` before calling `urljoin` so empty strings are treated as unset. 
- Add a regression test `test_empty_base_url_uses_default` in `weblate/machinery/tests.py` that configures `base_url` as an empty string and verifies requests target `https://api.anthropic.com/v1/messages`. 
- Add a changelog entry in `docs/changes.rst` under the unreleased Weblate 5.17 section describing the bug fix.

### Testing

- Ran the new test selectively with `uv run pytest weblate/machinery/tests.py -k empty_base_url_uses_default`, which could not complete in this environment because the PostgreSQL test database is unavailable and pytest errored during Django DB setup. 
- No further automated test failures were observed in this environment, but full test verification requires a configured test database and CI environment to run the complete test suite.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69c3ad3af2dc8329899350902e03ac14)

### DIFF
--- a/weblate/machinery/tests.py
+++ b/weblate/machinery/tests.py
@@ -2856,7 +2856,12 @@ class AnthropicTranslationTest(BaseMachineTranslationTest):
         )
 
         machine = self.MACHINE_CLS({**self.CONFIGURATION, "base_url": ""})
-        machine.translate_batch(self.ENGLISH, self.SUPPORTED, [self.SOURCE_BLANK])
+        self.assert_translate(
+            self.SUPPORTED,
+            self.SOURCE_BLANK,
+            self.EXPECTED_LEN,
+            machine=machine,
+        )
 
     @responses.activate
     def test_error_non_json(self) -> None:


docs/changes.rst
```diff
@@ -28,6 +28,7 @@ Weblate 5.17
 * Improved API access control for pending tasks.
 * Faster category and project removals.
 * Project backup restore no longer trusts repository-local VCS configuration and hooks from the uploaded archive.
+* :doc:`/admin/machine` Anthropic backend now falls back to the default API URL when base URL is empty.
```
nijel:
```suggestion
* :doc:`/admin/machine` now falls back to the default API URL when base URL is empty.
```

weblate/machinery/tests.py
```diff
@@ -2829,6 +2829,35 @@ def mock_response(self) -> None:
...
+                "stop_sequence": None,
+                "usage": {
+                    "input_tokens": 25,
+                    "output_tokens": 5,
+                },
+            },
+        )
+
+        machine = self.MACHINE_CLS({**self.CONFIGURATION, "base_url": ""})
+        machine.translate_batch(self.ENGLISH, self.SUPPORTED, [self.SOURCE_BLANK])
```
```suggestion
        self.assert_translate(
            self.SUPPORTED,
            self.SOURCE_BLANK,
            self.EXPECTED_LEN,
            machine=machine,
        )
```

PR #18684, opened 2026-03-27
fix(machinery): limit allowed URLs

nijel:
Allow only public URLs by default to reduce risk of SSRF.

### DIFF
--- a/docs/admin/config.rst
+++ b/docs/admin/config.rst
@@ -81,6 +81,30 @@ This is currently used in the following places:
 
    * :setting:`ALLOWED_ASSET_SIZE`
 
+.. setting:: ALLOWED_MACHINERY_DOMAINS
+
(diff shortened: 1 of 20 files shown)


nijel:
This grew too much by trying to handle all possible situations conveniently (like keeping useful error messages for trusted endpoints).

amCap1712:
I agree :sweat_smile: !

amCap1712:
The fix for dns-rebinding is quite complex but I guess necessary. I was wondering if adding dnsmasq (to reject private IP resolution from public DNS servers) to the weblate docker image. It would not help users setting up without the image though, unless they follow specific instructions to add it to their setups.

nijel:
Yes, some issues might be addressed outside Weblate, but in practice that would leave them unresolved for some part of installations.

Issue #17713, opened 2026-01-20
Improve client testing


nijel:
### Describe the problem

Currently issues like https://github.com/WeblateOrg/weblate/issues/17711 go unnoticed through our CI because there is really no testing of client code.

### Solution brainstorm

There must be a way to test that client code works well. I'd like to have some direct testing of JavaScript code that updating deps didn't break anything, but our code might not be ready for that.

### Describe alternatives you have considered

We might look into extending Selenium tests. It might be even possible to fail on JavaScript errors there.

PR #18685, opened 2026-03-27
test: Add check that JS bundle is active

KarenKonou:
Adds a simple check that the JS bundle loads successfully, should be able to catch some potential breakages from deps updates
For https://github.com/WeblateOrg/weblate/issues/17713


nijel:

PR #18732, opened 2026-03-31
Restore format encoding defaults and fix TTKit encoding/loader handling

nijel:
### Motivation

* Restore the documented default encoding fallback for certain file formats (e.g. Apple `strings` and Java `properties`) so format encoding comes from parameter defaults when parameters are not explicitly set.

### Description

* Change `get_encoding_param` signature to `get_encoding_param(file_format, file_format_params)` and make it merge format defaults with provided params and return the format-specific encoding parameter using the parameter class `is_encoding()`/`get_value()` logic.
* Update `weblate.formats.ttkit` to call `get_encoding_param` with the format id and to use `self.get_encoding()` when creating store instances and new files, and add a defensive fallback to `TTKitFormat.get_class()` for missing/unknown encoding variants.
* Update `Component.configure_repo` to pass the component's `file_format` into `get_encoding_param` when creating default file contents.
* Add `test_encoding_loader_defaults` to `weblate.formats.tests.test_formats` and update docs (`docs/changes.rst`) to note the restored default encoding behavior.

### Testing

* Ran the formats unit tests including the new `test_encoding_loader_defaults` in `weblate.formats.tests.test_formats`, and they passed.
* Ran the modified code paths for `TTKit` store creation and `Component.configure_repo` in unit tests, and no regressions were observed.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69c3a89354b48329beb6dee7d1ba56e7)

PR #18734, opened 2026-03-31
fix(formats): properly handle default params

nijel:
* When no encoding is specified, the defaults should come from the file format parameters.
* This removes xwiki page properties parametrization as it was not consistent with other properties and the format actually used hardcoded utf-8 before.

Foolowup for #18732

Needs https://github.com/WeblateOrg/weblate/pull/18735 for tests to pass.


nijel:
@gersona I think this is actually ready for review. Sorry for the previous mess with this.

Issue #16906, opened 2025-11-12
Increase contrast of default colors


sloria:
### Describe the problem

The current default colors make text difficult to read[^1] due to the low contrast. WCAG [recommends](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html) a minimum contrast ratio of 4.5:1, with 7:1 being the [standard](https://www.w3.org/WAI/WCAG21/Understanding/contrast-enhanced.html) for AAA.

Currently, the buttons are 2:1 (checked with [pika](https://github.com/superhighfives/pika))

<img height="626" alt="Image" src="https://github.com/user-attachments/assets/d5a0edef-db50-484c-8eed-988af3b8017a" />

The body text is 3.2:1

<img width="1442" height="678" alt="Image" src="https://github.com/user-attachments/assets/09946f32-8dce-439b-b2bb-40bdbb4e2e0a" />

[^1]: Even for me as a person with no vision impairments

### Solution brainstorm

Update the default colors to meet at least AA contrast ratio.

### Describe alternatives you have considered

I'm currently overriding the default colors in the Appearance settings. It's an imperfect solution though. "Hover color" gets applied for both dark backgrounds (e.g. in the navbar) and light backgrounds (e.g. in the main content), so it ends up incorrect in places.

Here's what I'm using for now:

![Image](https://github.com/user-attachments/assets/be74dc57-2fe3-4a26-abe7-4ab1c6300f0c)

### Screenshots

Current:

![Image](https://github.com/user-attachments/assets/84057a5b-e3b2-4e84-bc2d-c437401a1379)

Possible update:

![Image](https://github.com/user-attachments/assets/a8e5c3ff-1c6d-4daa-9382-b98b0440ba08)

PR #18751, opened 2026-03-31
feat: improve color contrast to meet WCAG AA guidelines

sloria:
Addresses https://github.com/WeblateOrg/weblate/issues/16906. This increases the contrast ratio for most places to be greater than 4.5:1 . The change is primarily in light mode; dark mode already met the threshold for the most part.

This is mostly a find-and-replace job to update the CSS variables and the hardcoded colors in all stylesheets and config.

## Before

before-checks
before-strings-status
before-translate
before-translation-status

## After

after-checks
after-strings-status
after-translate
after-translation-status

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -166,14 +166,14 @@ html_theme_options = {
     "dark_css_variables": {
         "font-stack": '"Source Sans 3", sans-serif',
         "font-stack--monospace": '"Source Code Pro", monospace',
-        "color-brand-primary": "#1fa385",
-        "color-brand-content": "#1fa385",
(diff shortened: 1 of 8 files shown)


nijel:
Does this affect the dark mode as well?

sloria:
in dark mode, the Glossary label text colors were changes since some of them were too low-contrast.

Before:

before-labels-dark

After:

after-labels-dark

nijel:
I think it's okay to change colors slightly for accessibility reasons. We should consider accessibility from beginning when doing new design.

nijel:

PR #18759, opened 2026-04-01
feat(checks): offload batch checks processing to background task

nijel:
- make the project update task trigger checks update separately improving observed performance
- the checks update no longer needs to hold the component lock now

Issue #13623

### DIFF
--- a/weblate/trans/models/component.py
+++ b/weblate/trans/models/component.py
@@ -951,7 +951,7 @@ class Component(
         self.stats = ComponentStats(self)
         self.needs_cleanup = False
         self.alerts_trigger: dict[str, list[dict]] = {}
-        self.updated_sources: dict[int, Unit] = {}
+        self.updated_sources: set[int] = set()
         self.old_component_settings = self.get_old_component_settings()
         self._sources: dict[int, Unit] = {}
         self._sources_prefetched = False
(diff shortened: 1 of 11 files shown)


weblate/trans/models/component.py
```diff
@@ -3122,18 +3117,43 @@ def start_batched_checks(self) -> None:
...
+            finalize_component_checks(
+                self.id,
+                source_unit_ids,
+                batched_checks,
+                batch_mode=batch_mode,
+                component=self,
+            )
+        else:
+
+            def schedule() -> None:
```
amCap1712: Unsure about why this wrapping and subsequent if check is needed.

nijel: Good catch. It's not needed here now. I had more complex code here before, but after removing it, it is not useful anymore.

weblate/trans/tests/test_tasks.py
```diff
@@ -101,6 +102,29 @@ def test_cleanup_old_comments_enabled(self) -> None:
 
 
 class TasksTest(ViewTestCase):
+    @override_settings(CELERY_TASK_ALWAYS_EAGER=False)
```
amCap1712: I think this test can also be simplified in that case? Directly mock and assert finalize_component_checks.delay_on_commit was called.

nijel: I think it can be fully removed now.

PR #18813, opened 2026-04-07
docs: mention PostgreSQL 18 volume change

nijel:
Include it in PostgreSQL upgrading instructions as that is where people will likely run into this.

See #18498

### DIFF
--- a/docs/admin/install/docker.rst
+++ b/docs/admin/install/docker.rst
@@ -275,7 +275,7 @@ Upgrading PostgreSQL container
 
    Leaving the old mount target unchanged without setting ``PGDATA`` can cause
    PostgreSQL to write its data outside the persisted volume.
-   
+
    See `PGDATA documentation <https://hub.docker.com/_/postgres#pgdata>`_ for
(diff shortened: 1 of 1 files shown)


docs/admin/install/docker.rst
```diff
@@ -268,6 +262,22 @@ not needed in most case. See :ref:`docker-postgres-upgrade` for upgrading the Po
...
+   common older setup mounted the database volume at
+   ``/var/lib/postgresql/data``, while PostgreSQL 18 now uses
+   ``/var/lib/postgresql`` by default.
+
+   If you are upgrading from an older version, either update the mount target
+   in your Docker configuration to the new path, or keep the old mount target
+   and set ``PGDATA`` accordingly.
+
+   Leaving the old mount target unchanged without setting ``PGDATA`` can cause
+   PostgreSQL to write its data outside the persisted volume.
```
nijel, line 277:
```suggestion
   PostgreSQL to write its data outside the persisted volume.
   
```

PR #18847, opened 2026-04-08
fix(vcs): improved symlink validation

nijel:

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -31,6 +31,7 @@ Weblate 5.17
 
 .. rubric:: Bug fixes
 
+* Tightened repository boundary checks for symlink targets.
 * Matching exporters now honor component file format parameters.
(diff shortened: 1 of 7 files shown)


M9nx:
Hi @nijel ,

I reviewed and tested the patch locally from PR 18847 (commit 66bab48).

Validation performed:
- Re-ran the original prefix-collision PoC scenario.
- Tested symlink/junction prefix-collision regression cases.
- Tested additional path-containment edge behavior on Windows path normalization.

Result:
- The patched logic blocks the previously reproducible bypass.
- In my regression comparison, the old startswith checks allowed the prefix-collision case, while the new path-segment-aware containment check rejected it.
- I did not find a bypass in the tested variants.

The added test coverage in the PR for discovery, VCS symlink resolution, and path utility checks aligns with this behavior.

Thanks for the fix and fast response.

PR #18973, opened 2026-04-13
fix(component): consistenly handle settings for linked components

nijel:
- Always use repository settings from the linked component
- Expose linked component settings when editing
- Do not allow to diverge when editing
- Make the behavior clearly described in the documentation

### DIFF
--- a/weblate/api/views.py
+++ b/weblate/api/views.py
@@ -1383,7 +1383,11 @@ class ProjectViewSet(
                     },
                 )
 
-        queryset = obj.component_set.filter_access(self.request.user).order_by("id")
+        queryset = (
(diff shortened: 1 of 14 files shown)


weblate/api/views.py
```diff
@@ -1383,7 +1383,9 @@ def components(self, request: Request, **kwargs):
                     },
                 )
 
-        queryset = obj.component_set.filter_access(self.request.user).order_by("id")
+        queryset = (
+            obj.component_set.filter_access(self.request.user).prefetch().order_by("id")
```
nijel:
```suggestion
            obj.component_set.filter_access(self.request.user).order_by("id")
```

PR #18982, opened 2026-04-13
feat(file_format_params): Gettext PO header configuration with file format parameters

### LINKED ISSUE #16971: Migrate gettext specific settings to file format parameters
### Describe the problem

The project has the *Set "Language-Team" header* setting that affects gettext only.

There is also component setting *Source string bug reporting address*, but that is also used for e-mail notifications of source string problems, so that one should probably stay.

### Solution brainstorm

As we now have file format-specific parameters, it should be moved there.

Maybe we could have some more fine-grained control over gettext headers as users might want to customize this in more detail, for example exclude “Last-Translator”, “Language-Team” and “Report-Msgid-Bugs-To”. 

### Screenshots

_No response_

### Additional context

_No response_

gersona:
- Fixes #16971
- **Add file format params for .po file headers**
- **migrate set_language_team attribute to File format params**
- **documentation update**
- **update backup schema**


gersona:
Should the gettext file format parameters also apply to the headers set in the PO exporters? For example, `weblate/formats/exporters.py` (lines 71–78) always sets `x_generator`, `language_team`, etc., without consulting `file_format_params`.

nijel:
Yes, exporters honor params since https://github.com/WeblateOrg/weblate/pull/18830

gersona:
This will probably wait for the next release of [Weblate schemas](https://github.com/WeblateOrg/weblate_schemas/) for all tests to pass

nijel:
@gersona Can you please update this on top of the current `main`? https://github.com/WeblateOrg/weblate/pull/19327 pulled in updated schema.

nijel:

Issue #5155, opened 2021-01-06
Administration and presets of notifications


funktionierbar:
**Is your feature request related to a problem? If so, please describe.**

New user by default do not watch any projects, they have to activate this feature manually before they receive email notifications e.g. for new strings or components. 
This leads to a lot of users that register once, forget about the project and never come back. Having the monthly digest for notification of new components auto-enabled for every users would help me engange my idle users, I guess. 
As an admin, I would like to be able to enable the notification for monthly digest for new component for all my users, or for specific users, just as I can add them to a group.
Also, I would like to change the presets for notifications for newly registering users. I would like that when a user registers, the notification for "monthly digest for new component" should be auto-on by default.  

**Describe the solution you'd like**
When I am logged in as Admin/Superuser, there should be a tab in Administration -> Users or in the django admin interface, that allows me to edit the notification presets for new users, and/or to administrate/override single user/group notification settings.

**Describe alternatives you've considered**

I tried to find the presets in the django/accounts module and in the database, I thought they must exist somewhere, at least hard-coded but I failed. If you don't plan to implement this feature, I would be thankful for pointers where I can find these presets, or where the notification settings are stored.

nijel:
Related to https://github.com/WeblateOrg/weblate/issues/4866, we indeed need to find reasonable way to engage new contributors. On the other side we don't want to annoy them by sending unsolicited notifications.

funktionierbar:
Its true, one has to be careful there. But I would think, at least for my user base, receiving one digest per month would be ok, and they would still be able to turn notification off if they dont want it.

nijel:
Yes, both approaches make sense.

orangesunny:
Enforcement by server/project admin needs to be implemented, too.

davidvavra:
Hi everyone. Great to meet @orangesunny on PyCon CZ in person. I didn't know Weblate is a Czech project!

This issue is troubling us, our translators were used to email notifications from previous translation management software we were using. Now the don't receive any, they don't turn it on by themselves and became inactive. Would you be open to collaborate on a pull request from our side?

nijel:
Sure, collaboration is always welcome!

There are few things to distinguish here:

* User languages are used to determine for which translations he receives notification.
* Adding a project to watched makes the user receive notifications.
* Site-wide notifications can also be configured per user.

Weblate tries to configure these based on user actions

- upon initial login the user language is automatically added (if it can be detected)
- upon translating the language is added if no other is configured
[...]

PR #19005, opened 2026-04-14
feat(accounts): add default notification configuration to settings

michael-smt:
Allows configuring the default notification configuration for newly created users in settings.

Implements some of #5155


michael-smt:
The configuration syntax is probably not ideal, and it might be a bit too much documentation :wink:

nijel:
I'm not really sure about this. This used to be an implementation detail, and we should probably think about this more before exposing it as configuration.

michael-smt:
I'm also not so sure about this :slightly_smiling_face: But I thought better something than nothing... 

Maybe we could keep it undocumented, that would allow coming up with a better way to expose it as configuration later...

nijel:
I wasn't able to come to a better conclusion here, so let's make this happen for the next milestone.

The only thing I don't like here is manually written documentation for possible values, which is super easy to become outdated.

There are definitely better and more complex approaches to this, but we know we're going to change the notifications subsystem heavily in the not-so-distant future (at least to support web notifications), so let's not spend much time on something that will have to change anyway.

nijel:
@michael-smt Can you rebase this on current main? If you have time to add documentation generation (see `update-docs` in `docs/Makefile`), that would be great.

michael-smt:
Yes, will do. Thanks for the pointer with the docs generation.

weblate/accounts/management/commands/list_notification_config.py
```diff
@@ -0,0 +1,54 @@
...
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+    from weblate.accounts.notifications import (
+        Notification,
+    )
+
+
+def sorted_handlers(handlers: Iterable[Notification]) -> Iterable[Notification]:
```
nijel:
```suggestion
def sorted_handlers(handlers: Iterable[type[Notification]]) -> Iterable[type[Notification]]:
```

weblate/accounts/management/commands/list_notification_config.py
```diff
@@ -0,0 +1,54 @@
...
+            scope_content.extend(
+                [f"``{scope.value}``", f"   :guilabel:`{scope.label}`"]
+            )
+        self.add_section("notification-scopes", scope_content)
+
+        frequency_content = []
+        for scope in NotificationFrequency:
+            frequency_content.extend(
+                [f"``{scope.value}``", f"   :guilabel:`{scope.label}`"]
+            )
```
nijel:
```suggestion
        for frequency in NotificationFrequency:
            frequency_content.extend(
                [f"``{frequency.value}``", f"   :guilabel:`{frequency.label}`"]
            )
```

weblate/accounts/models.py
```diff
@@ -160,6 +159,61 @@ class WeblateAccountsConf(AppConf):
 
     MAXIMAL_PASSWORD_LENGTH = defaults.DEFAULT_MAXIMAL_PASSWORD_LENGTH
 
+    DEFAULT_NOTIFICATIONS: ClassVar[
```
nijel: The default value should go to the defaults module and it can be then reused as default value in the Docker settings.

michael-smt: I tried, but it means we can't use Django's IntegerChoices anymore to define the defaults, only raw values. Because when they are used in the Docker settings, the Django apps aren't loaded yet. Maybe there is another way to solve this?

weblate/utils/environment.py
```diff
@@ -316,3 +329,28 @@ def get_email_config() -> tuple[int, bool, bool]:
...
+            raise ImproperlyConfigured(msg)
+
+        scope, frequency, _handler = config
+        try:
+            scope = int(scope)
+            frequency = int(frequency)
+        except ValueError as error:
+            msg = f"{name}: not an integer: {error}"
+            raise ImproperlyConfigured(msg) from error
+
```
michael-smt: I would have liked to extend the validation here with something like this:

```python
        if frequency not in NotificationFrequency.values:
            msg = f"{name}: invalid notification frequency {frequency}"
            raise ImproperlyConfigured(msg)
        if scope not in NotificationScope.values:
            msg = f"{name}: invalid notification scope {scope}"
            raise ImproperlyConfigured(msg)
        if handler not in [n.__name__ for n in NOTIFICATIONS]:
            msg = f"{name}: unknown handler {handler}"
            raise ImproperlyConfigured(msg)
```

But can't use these classes here because `django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.`

Maybe there is a better place for the validation?

nijel: I think AppConfig.ready() could be the place. That way it would cover the Python code settings, not just the environment in Docker. The validation in this file should be focused on the environment values parsing.

weblate/accounts/models.py
michael-smt: Figured it out, moving the IntegerChoices into `data.py` did the trick :slightly_smiling_face:

weblate/utils/environment.py
michael-smt: Thanks a lot for the hint and your patience. I tried to implement that now.

Validation of the notification handlers via `NOTIFICATIONS` (the list that gets built from the decorators) is not possible due to `AppRegistryNotReady`... I wanted to do something like

```python
                if handler not in [n.__name__ for n in NOTIFICATIONS]:
                    errors.append(
                        weblate_check(
                            f"{name} contains a unknown notification handler {handler}",
                        )
                    )
```

michael-smt: @nijel is it acceptable to not validate the notification handlers? If not, do you have a idea how it could be done?

michael-smt:
Please reopen, I still want to implement this.

In my [last comment](https://github.com/WeblateOrg/weblate/pull/19005#discussion_r3488148401) I was not clear that I needed feedback, so it got stale.

weblate/utils/environment.py
nijel: The validation and import should IMHO work inside `ready()`. The import has to be in the method, not at the module level.

Issue #16042, opened 2025-09-04
Review javascript dependencies


nijel:
### Describe the problem

Many of our JavaScript dependencies haven't received updates for years:

| Datasource | Name | Last Updated |
|------------|------|-------------|
| npm | `@popperjs/core` | `2023-05-26` |
| npm | `autosize` | `2023-02-13` |
| npm | `css-loader` | `2024-05-22` |
| npm | `daterangepicker` | `2020-05-24` |
| npm | `moment` | `2023-12-27` |
| npm | `mousetrap` | `2020-01-23` |
| npm | `mousetrap-global-bind` | `2016-10-25` |
| npm | `multi.js` | `2021-02-22` |
| npm | `slugify` | `2023-03-26` |
| npm | `source-code-pro` | `2021-04-12` |
| npm | `source-sans` | `2021-09-07` |
| npm | `tributejs` | `2020-03-25` |

### Solution brainstorm

We should review those and look for maintained alternatives.

### Additional context

https://github.com/WeblateOrg/weblate/issues/8486 has the full list of possibly abandoned dependencies.

KarenKonou:
One more thing we could consider is moving away from jQuery. It doesn't enable much compared to modern vanilla JS and it'd allow us to remove another fairly big JS dependency.

nijel:
Yes, that's definitely desirable. Actually, portions of code are already written without jQuery.

KarenKonou:
These are the dependencies that are unmaintained that we should aim to replace and some ideas on their replacements:

- DateRangeTime: In progress replacement with a custom implementation in #19046
- Autosize: Could possibly also be reimplemented or the functionality replicated with CSS
- Mousetrap and mousetrap-global-bind: Could be replaced with [hotkeys-js](https://www.npmjs.com/package/hotkeys-js)
- Multi.js: Maybe we could use something like [tom-select](https://tom-select.js.org)
- Tribute-js: We could potentially implement this functionality with `autocomplete.js` that we already have

Additionally, `popperjs` is bundled with bootstrap, so we could skip bundling it separately ourselves.

After these deps issues are resolved, we can focus on replacing jQuery usage, eventually deprecating it completely.

nijel:
Sounds good. Let's try to do this for the 5.18 release.

nijel:
I think all actionable deps are now handled.

PR #19022, opened 2026-04-15
deps: replace daterangepicker with flatpickr

KarenKonou:
Replaces the daterangepicker dependency with flatpickr
The main benefit of this is that flatpickr doesn't depend on jquery or moment, allowing us to remove those dependencies in the future
For #16042


nijel:
I see it has the last release and commit from 2022 (https://github.com/flatpickr/flatpickr). Is this something we want to depend on?

KarenKonou:
I agree it is not ideal, however I couldn't really find any other viable option. Maybe we could simply build something around an HTML date input without any extra dependencies.

nijel:
I'm afraid that it's quite likely that we run into one of the unfixed issues like https://github.com/flatpickr/flatpickr/issues/3094.

KarenKonou:
Fair point. In that case I'm leaning towards a minimal HTML/JS implementation suited for our needs, without using an external library.

nijel:
Sounds good.

PR #19028, opened 2026-04-15
fix(add-ons): avoid repeated commit pending on linked components

nijel:


nijel:
Probably too much micro-optimization here.

PR #19046, opened 2026-04-16
feat: new date range picker implementation

KarenKonou:
Removed the old jQuery-based date range picker and replaced it with a new vanilla JavaScript implementation.
For #16042

image


nijel:
Have you considered some approach to test the client-side code?

KarenKonou:
Good point, I've been meaning to adapt a more test-driven approach to frontend code, I'll add some tests.

nijel:
Looks great; let's ship it in 5.17.1.

PR #19128, opened 2026-04-20
feat(client): replace mousetrap with hotkeys.js

KarenKonou:
for https://github.com/WeblateOrg/weblate/issues/16042


KarenKonou:
There is a difference in behavior between `mousetrap` and `hotkeys.js`, while mousetrap fired on `keypress` events, hotkeys.js works on `KeyboardEvent.code` (physical keys). This should make the hotkeys work on all keyboard layouts, but on the other hand on non-qwerty layouts the hotkeys might not match the physical key labels (for example the `ctrl+y` hotkey will always trigger on the key to the right of `T`). We could add a message to the keyboard shortcuts modal clarifying this.

We could also maybe work around it for the conflicting hotkeys on major keyboard layouts but things such as DVORAK layout experience might not be great.

(Maybe worth noting that the `keypress` events that mousetrap used are deprecated, and `event.key` would be completely broken on non-latin layouts)

PR #19184, opened 2026-04-23
Fix: translation repository browser

### LINKED ISSUE #12308: Allow setting a separate repository browser url for translation files
### Describe the problem

We use a separate repository for translation files, different from the source string repository.

For example in <https://hosted.weblate.org/translate/godot-engine/godot-docs/zh_Hans/?q=>:

- The source string comes from <https://github.com/godotengine/godot-docs>
- The translation file comes from <https://github.com/godotengine/godot-docs-l10n>

The "Souce string location" and "Translation file" links on that page seems all based on the "Repository browser" setting. So one of the two is always wrong.

Add a separate repository browser setting for "Translation file".

_No response_

### Screenshots

_No response_

### Additional context

_No response_

BrenABarn:
Fixes #12308

What this PR does

Adds a new optional "Repository browser for translations" field to the component settings. When set, this URL is used to generate links to translation files instead of the existing "Repository browser" field. If left empty, it falls back to the existing field, maintaining backward compatibility.

Changes

- New `repoweb_translations` model field on `Component`
- Database migration
- Updated `get_repoweb_link` to accept `is_translation` parameter
- Updated `translation_file_link` in edit view to use the new field
- Documentation added for the new field
- Tests added for the new behavior and fallback logic


docs/admin/projects.rst
```diff
@@ -457,6 +457,18 @@ might want to strip leading directory by ``parentdir`` filter (see
...
+Repository browser for translations
++++++++++++++++++++++++++++++++++++++
+
+URL of repository browser used to display translation files. When empty, the
+:ref:`component-repoweb` URL will be used as a fallback. This is useful when
+the source files and translation files are hosted in different repositories.
+You can use :ref:`markup`.
+
+For example on GitHub, use something like:
+``https://github.com/WeblateOrg/translations/blob/{{branch}}/{{filename}}#L{{line}}``
```
nijel, line 468: I think it will break rendering without newline:

```suggestion
``https://github.com/WeblateOrg/translations/blob/{{branch}}/{{filename}}#L{{line}}``

```

Also, the `seealso` should be present in both of thse.

weblate/trans/views/edit.py
```diff
@@ -782,15 +781,16 @@ def translate(request: AuthenticatedHttpRequest, path):
...
+            "translation_file_link": lambda: (
+                unit.translation.component.get_repoweb_link(
+                    unit.translation.filename,
+                    # '1' as a placeholder, because `get_repoweb_link` can't currently
+                    # generate links without line specified. Although it's ok to use
+                    # '' or '0' on GitHub or GitLab, let's play it safe for now.
+                    "1",
+                    is_translation=True,
+                    user=user,
+                )
```
nijel: This change removes support for direct https links.

nijel (commented):
Thanks for the pull request. Besides the bits I've commented inline, this needs to be exposed in the API and deserves a changelog entry.

I've also triggered copilot review now.

weblate/trans/migrations/0074_component_repoweb_translations.py
nijel, line 1: This migration is not based on the current main:

```
Conflicting migrations detected; multiple leaf nodes in the migration graph: (0069_component_repoweb_translations, 0073_alter_change_action in trans).
```

Please rename it and adjust dependencies.

PR #19228, opened 2026-04-24
fix(ui): display file format params settings correctly

### LINKED ISSUE #17646: File format parameters layout
### Describe the issue

The file format parameters section is formatted differently than rest of the component settings.

### I already tried

Open any component with file format parameters settings. The good choice is gettext PO format which has both dropdowns and checkboxes.

### Expected behavior

The file format parameters should be formatted same as other inputs:

- checkboxes on the left and label next to it
- dropdowns on multiple lines
- it might be even desirable to move this to a separate tab (but this won't work for creating new component where no tabs are used)

### Screenshots

Image

### Exception traceback

```pytb

```

### How do you run Weblate?

weblate.org service

### Weblate versions

_No response_

### Weblate deploy checks

```shell

```

### Additional context

https://github.com/WeblateOrg/weblate/pull/17417 fixed only part of the issue

KarenKonou:
fixes #17646

### DIFF
--- a/weblate/static/styles/main.css
+++ b/weblate/static/styles/main.css
@@ -2732,3 +2732,9 @@ table.table-simple tbody#glossary-terms .alert-info {
 .translate-explanation {
   background-color: var(--bs-secondary-bg) !important;
 }
+
+/* styling fix for file format parameters caused by `display: none` elements as first children in the same container */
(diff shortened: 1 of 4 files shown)


nijel:

PR #19254, opened 2026-04-27
fix(ui): use 50/50 width for side-by-side zen

### LINKED ISSUE #16268: Side-by-side zen 90%/10% width when it should be 50%/50%
### Describe the issue

We're exploring how suitable Weblate is for translating a long HTML text document. Some voiced a preference for side-by-side view. They're used to working with original text on one half of the screen and a translation document on the other. I found in Weblate settings an option to switch the Zen view to side-by-side, which should approximately match the preference of these translators. However, our strings are really long (paragraphs of text), and in this mode the original strings in Weblate take like 90% of the screen width while only the remaining 10% go to the translation, which I think is an edge case bug.

### I already tried

1. Create a project with a component that has long strings, such as paragraphs of text.
2. Turn on the side-by-side view for the zen mode in user settings.
3. Open the component in zen mode.
4. Observe 90% of the screen width taken by the original text, 10% by the translation.

### Expected behavior

The screen should be evenly split: 50% go to the original text, and 50% go to the translation.

### Screenshots

Image

### Exception traceback

```pytb

```

### How do you run Weblate?

Docker container

### Weblate versions

 * Weblate: 5.13.3
 * Only the Weblate version is relevant for this issue I think, because it's a front-end issue

### Weblate deploy checks

```shell

```

### Additional context

Inspecting the table layout element and adding the following style helps:

```css
.zen-horizontal .translatetext, .zen-horizontal .translator {
  width: 50%;
}
```

Note: setting `max-width` on `.translatetext` instead didn't do what I expected.

The translator's browser is `Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:142.0) Gecko/20100101 Firefox/142.0`, I haven't tested how this change works in other people's browsers, other components, for shorter strings etc.

Possibly related to closed issue #2739 - I think fixing my issue would do the opposite of what its author wanted.

KarenKonou:
fixes #16268

### DIFF
--- a/weblate/static/styles/main.css
+++ b/weblate/static/styles/main.css
@@ -798,6 +798,11 @@ kbd {
   width: 100%;
 }
 
+.zen-horizontal .translatetext,
+.zen-horizontal .translator {
+  width: 50%;
+}
+
 .zen-unit:nth-child(even) {
   background-color: #f9f9f9;
 }


nijel:

nykula:
Thank you for the fix!

PR #19305, opened 2026-04-30
Feat replace multi.js

### LINKED ISSUE #19302: Replace multi.js
### Describe the problem

`multi.js` is one of the last remaining jQuery based dependencies. Additionally, it has been unmaintained since 2020. Replacing it with a suitable alternative is therefore important both for security and maintainability reasons, and is also a required step in gradually moving away from jQuery.

However, I wasn't able to find a replacement library that would support the current multi-pane UI. Therefore, we will either have to reimplement the current UX ourselves, or the UX will have to be changed.

### Solution brainstorm

One replacement options could be [tom-select with the checkbox options plugin](https://tom-select.js.org/plugins/checkbox-options/), enabling a searchable tag style UX. This could also be used for other things, such as license select or flags.

The alternative solution could be a lightweight in-house implementation of the current UX, similarly as we did for the date picker. Since we only need to implement the features that we need for Weblate, it might be relatively code-light.

### Screenshots

_No response_

### Additional context

Sub issue of https://github.com/issues/assigned?issue=WeblateOrg%7Cweblate%7C16042
Feedback and ideas welcome!

KarenKonou:
closes #19302


nijel:

PR #19451, opened 2026-05-08
fix(altcha): update altcha widget to match  backend version

### LINKED ISSUE #19448: Registering new account on hosted Weblate - Altcha is not working
### Describe the issue

Impossible to register a new account because "Altcha" keeps saying "Verification failed. Try again later."
I tried with multiple web browsers.

### I already tried

1. Go to "https://hosted.weblate.org/"
2. Disconnect from your account if you are connected
3. Click on Register a new account or go directly to "https://hosted.weblate.org/accounts/register/"
4. Fill in the form  : email address, nickname, full name, resolve the number challenge
5. Try to click on the captcha "Altcha", you got a "Verification failed. Try again later." error message so it's impossible to register a new account.

### Expected behavior

The Altcha checkbox should be clickable as I'm a real human.

### Screenshots

Image

### Exception traceback

```pytb

```

### How do you run Weblate?

Other

### Weblate version

hosted, so : weblate-5.17.1-371-g4775ea108d (I suppose)

### Weblate deploy checks

```shell

```

### Additional context

I'm trying to register the account from France. I tried Brave (with and without Brave Shield enabled), and with Google Chrome.

KarenKonou:
fixes #19448
This also changes the way the widget is localized to use altcha's built in localization instead of passing localized strings as the latter is no longer supported. Custom localized strings can still be set via the JS api but IMO such change would be out of the scope of this fix


KarenKonou:
One more thing we might want to tweak is the difficulty of the challenge - The defaults currently in use (recommended for argon2) take quite a while to solve and make my computer fans spin up :)

PR #19453, opened 2026-05-08
fix: reduce default altcha memory cost

KarenKonou:
The altcha challenge would take a very long time to solve sometimes, worsening the user experience. Reducing the memory cost makes it much faster, hopefully without weakening the deterrence factor too much.


nijel:

PR #19510, opened 2026-05-12
feat: remove autosize

KarenKonou:
Removes autosize, replacing it with a small JS handler.
This could be done with [CSS field-sizing: content property](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/field-sizing) instead, but unfortunately firefox doesn't support it yet (should be landing next month).
Hopefully this covers all of the autosize usages.
Part of https://github.com/WeblateOrg/weblate/issues/16042


nijel:

PR #19530, opened 2026-05-13
feat(client deps): remove tributejs

KarenKonou:
Removes tributejs for username autocompletion and uses autocomplete.js instead.
This is the last remaining legacy frontend dependency
Part of https://github.com/WeblateOrg/weblate/issues/16042


nijel:

PR #19547, opened 2026-05-13
feat: Localize generic user agent strings

### LINKED ISSUE #12916: Localizable user agent strings
### Describe the problem

And will platform names like ``PC``, ``Generic Smartphone`` be translatable?
image

Localize these [generic strings](https://github.com/selwin/python-user-agents/blob/master/user_agents/parsers.py) from the user-agents library.

This needs to be done at display time to reflect current user locales, the string stored in the database should stay as it is now.

- Split string by ` / `
- Try to localize parts
- Merge it back

_No response_

### Screenshots

_No response_

### Additional context

_Originally posted by @Blueberryy in https://github.com/WeblateOrg/weblate/issues/12909_

kristiankunc:
Fixes #12916

- This PR localizes the generic user agent strings when rendering in the audit log

We only localize the first part as the OS and the browser name + version probably don't need translating.

_I am unsure about the tests, I basically just mocked the localization method and then asserted the call - maybe there are better ways to do this_

### DIFF
--- a/weblate/accounts/models.py
+++ b/weblate/accounts/models.py
@@ -34,6 +34,7 @@ from django.utils.translation import (
     gettext_lazy,
     pgettext_lazy,
 )
+from django_stubs_ext import StrOrPromise
 from django_otp.plugins.otp_static.models import StaticDevice
 from django_otp.plugins.otp_totp.models import TOTPDevice
 from django_otp_webauthn.models import WebAuthnCredential
@@ -460,7 +461,7 @@ NOTIFY_ACTIVITY = {
 
 
 # Taken from https://github.com/selwin/python-user-agents/blob/master/user_agents/parsers.py
-USER_AGENT_DEVICE_TYPES: dict[str, str] = {
+USER_AGENT_DEVICE_TYPES: dict[str, StrOrPromise] = {
     # Translators: User agent device type
     "PC": pgettext_lazy("User agent device type", "PC"),
     # Translators: User agent device type
--- a/weblate/accounts/models.py
+++ b/weblate/accounts/models.py
@@ -608,6 +608,21 @@ class AuditLog(models.Model):
             return EXTRA_MESSAGES[self.activity].format(**self.params)
         return None
 
+    def get_user_agent_display(self) -> str:
+        """Return a localized user agent string."""
+        ua_string = self.user_agent
+        if not ua_string:
+            return ""
+
+        parts = ua_string.split(" / ")
+        localized: list[str] = []
+        for part in parts:
+            part = part.strip()
+            if not part:
+                continue
+            localized.append(gettext(part))
+        return " / ".join(localized)
+
     def should_notify(self) -> bool:
         return (
             self.user is not None
--- a/weblate/accounts/models.py
+++ b/weblate/accounts/models.py
@@ -468,7 +468,9 @@ USER_AGENT_DEVICE_TYPES: dict[str, str] = {
     # Translators: User agent device type
     "Generic Smartphone": pgettext_lazy("User agent platform", "Generic Smartphone"),
     # Translators: User agent device type
-    "Generic Feature Phone": pgettext_lazy("User agent platform", "Generic Feature Phone"),
+    "Generic Feature Phone": pgettext_lazy(
+        "User agent platform", "Generic Feature Phone"
+    ),
     # Translators: User agent device type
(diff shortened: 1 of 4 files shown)


nijel:
What is missing is exposing the strings to xgettext so that they can be actually translated.

kristiankunc:
> What is missing is exposing the strings to xgettext so that they can be actually translated.

Done, forgot about that

weblate/accounts/models.py
```diff
@@ -454,6 +459,21 @@ def get_unsubscribe_url(self) -> str:
...
+    # Translators: User agent device type
+    "PC": gettext_noop("PC"),
+    # Translators: User agent device type
+    "Other": gettext_noop("Other"),
+    # Translators: User agent device type
+    "Generic Smartphone": gettext_noop("Generic Smartphone"),
+    # Translators: User agent device type
+    "Generic Feature Phone": gettext_noop("Generic Feature Phone"),
+    # Translators: User agent device type
+    "iOS-Device": gettext_noop("iOS-Device"),
```
nijel: It would be simpler to use `gettext_lazy` here...

weblate/accounts/models.py
```diff
@@ -608,6 +628,22 @@ def get_extra_message(self) -> str | None:
...
+        if not ua_string:
+            return ""
+
+        parts = ua_string.split(" / ")
+        localized = list(parts)
+        if localized:
+            device_type = localized[0].strip()
+            localized[0] = gettext(
+                USER_AGENT_DEVICE_TYPES.get(device_type, device_type)
+            )
```
nijel: ... and use the lazy value directly:

```suggestion
            localized[0] = USER_AGENT_DEVICE_TYPES.get(device_type, device_type)
```

weblate/accounts/models.py
nijel: Based on the copilot feedback, it should actually be `pgettext_lazy`.

weblate/accounts/models.py
```diff
@@ -454,6 +459,21 @@ def get_unsubscribe_url(self) -> str:
 }
 
 
+# Taken from https://github.com/selwin/python-user-agents/blob/master/user_agents/parsers.py
+USER_AGENT_DEVICE_TYPES: dict[str, str] = {
+    # Translators: User agent device type
+    "PC": pgettext_lazy("PC"),
```
nijel: pgettext needs additional parameter for context, something like:

```suggestion
    "PC": pgettext_lazy("User agent platform", "PC"),
```

weblate/accounts/tests/test_models.py
```diff
@@ -36,6 +38,24 @@ def test_superuser_audit_classification(self) -> None:
...
 
+    def test_user_agent_display_empty(self) -> None:
+        audit = AuditLog(user_agent="")
+        self.assertEqual(audit.get_user_agent_display(), "")
+
+    def test_user_agent_display(self) -> None:
+        audit = AuditLog(user_agent="PC / Linux / Chrome 120.0.0")
+        self.assertEqual(audit.get_user_agent_display(), "PC / Linux / Chrome 120.0.0")
+
+    def test_user_agent_display_calls_gettext_for_first_part_only(self) -> None:
```
nijel:
```suggestion
    def test_user_agent_display_localizes_first_part_via_mapping(self) -> None:
```

weblate/accounts/models.py
```diff
@@ -28,7 +28,12 @@
 from django.utils.functional import cached_property
 from django.utils.html import format_html
 from django.utils.timezone import now
-from django.utils.translation import get_language, gettext, gettext_lazy
+from django.utils.translation import (
+    get_language,
+    gettext,
+    gettext_lazy,
+    pgettext_lazy,
+)
```
nijel, line 36:
```suggestion
)
from django_stubs_ext import StrOrPromise
```

weblate/accounts/models.py
```diff
@@ -454,6 +459,23 @@ def get_unsubscribe_url(self) -> str:
 }
 
 
+# Taken from https://github.com/selwin/python-user-agents/blob/master/user_agents/parsers.py
+USER_AGENT_DEVICE_TYPES: dict[str, str] = {
```
nijel:
```suggestion
USER_AGENT_DEVICE_TYPES: dict[str, StrOrPromise] = {
```

kristiankunc:
Sorry I've gotten somewhat busy - thanks for finishing it!

PR #19562, opened 2026-05-14
fix: omit peer deps from sbom

nijel:
These do not have to be satisfied, so omit them.


nijel:
Okay, this proved to exclude more than what I hoped for, so this is probably not a viable approach.

PR #19596, opened 2026-05-15
feat: use tom-select for licence selection

### LINKED ISSUE #15154: Improve license selection
### Describe the problem

Selecting the license in the component configuration can be tedious because many licenses start with the same string or users search based on some keyword, not on the start (`GNU Affero` versus `Affero`).

### Solution brainstorm

Do not use browser native select widget, but instead use something which supports searching.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

KarenKonou:
closes https://github.com/WeblateOrg/weblate/issues/15154

### DIFF
--- a/weblate/static/loader-bootstrap.js
+++ b/weblate/static/loader-bootstrap.js
@@ -966,6 +966,20 @@ $(function () {
     });
   });
 
+  /* Searchable single selects */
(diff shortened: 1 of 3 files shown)


nijel:
Looks great, scheduled for merge.

PR #19634, opened 2026-05-18
chore(client deps): bundle altcha worker in a separate file

### LINKED ISSUE #16207: Do not require CSP changes for Altcha
### Describe the problem

Right now we add `blob:` to `worker-src` CSP which might be considered insecure in some situations:

https://github.com/WeblateOrg/weblate/blob/040921c1d16d9a7e99819bac64cdfe8319680c0d/weblate/middleware.py#L72-L73

This is because Altcha.

### Solution brainstorm

There seems to be a way to avoid that and it is documented at https://altcha.org/docs/v2/widget-integration/#content-security-policy-csp.

_No response_

### Screenshots

_No response_

### Additional context

https://github.com/orgs/WeblateOrg/discussions/16191

KarenKonou:
Bundles altcha's argon2id worker in a separate file, allowing stricter CSP settings (the wasm-eval rule is still required).
Also skips running the worker through terser, which appears to have been causing occasional WASM memory allocation failures.
Closes https://github.com/WeblateOrg/weblate/issues/16207


nijel:

PR #19656, opened 2026-05-19
docs(admin): fix altcha config version number

michael-smt:


nijel:
Thanks, we apparently forgot to update this after the versioning change.

PR #19677, opened 2026-05-21
fix(forms): html escape help text

meesfrensel:
The help text for the 'needs editing' checkbox contains quotes that aren't escaped properly, making the text cutoff (see screenshot) and the generated HTML invalid.

This is my first contribution to weblate, and I'm not 100% sure this is the right place to escape.

I'm happy to add a test, but not sure where that would be.

| Before | After |
|---|---|
| image | image |

Generated HTML after browser parsing (before fix):

```html
<span
  id="id_70ede8292633ed94_fuzzy_helptext" class="form-text"
  title="Strings are usually marked as " needs="" editing"="" after="" the="" source="" string="" is="" updated,="" or="" when="" marked="" as="" such="" manually."=""
>
  <svg ...><path>...</path> </svg>
</span>
```

### DIFF
--- a/weblate/trans/forms.py
+++ b/weblate/trans/forms.py
@@ -33,7 +33,7 @@ from django.forms.utils import from_current_timezone
 from django.template.loader import render_to_string
 from django.urls import reverse
 from django.utils import timezone
-from django.utils.html import format_html, format_html_join
+from django.utils.html import escape, format_html, format_html_join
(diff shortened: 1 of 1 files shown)


nijel:
Thanks for pointing out the issue, but I believe the proper fix is on the template level: https://github.com/WeblateOrg/weblate/pull/19684

PR #19692, opened 2026-05-22
fix(search): limit clause expansion in change query parser

nijel:
### Motivation
- The change-query parser could materialize an exponential number of `ParsedQueryClause` objects when combining OR groups with AND, allowing a short crafted query to exhaust worker memory and CPU. 
- A lightweight parser-side cap is needed to fail early and avoid the Cartesian-product expansion while preserving existing query semantics for reasonable inputs.

### Description
- Add a constant `MAX_PARSED_QUERY_CLAUSES = 4096` and enforce it in `ParsedQuery.combine_and` and `ParsedQuery.combine_or` to guard clause-list growth. 
- When a combination would exceed the cap, raise `SearchQueryError` with a user-facing message (`gettext("Search query is too complex.")`) to abort parsing early. 
- Add a regression test `test_changed_query_complexity_limit` that builds repeated `(changed_by:... OR changed_by:...)` groups joined by `AND` and asserts `parse_query(...)` raises `SearchQueryError`. 
- Changes touch `weblate/utils/search.py` and `weblate/utils/tests/test_search.py` only.

### Testing
- Ran `uv run pytest weblate/utils/tests/test_search.py -k "changed_by or complexity_limit"` and the selected tests passed (`2 passed, 76 deselected`).
- The new test verifies the complexity limit triggers before exponential clause expansion and succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0ffc19a6548329be29d3745a01a416)

### DIFF
--- a/weblate/utils/search.py
+++ b/weblate/utils/search.py
@@ -81,7 +81,7 @@ OPERATOR_MAP = {
     ":>=": "gte",
 }
 
-MAX_PARSED_QUERY_CLAUSES = 4096
+MAX_PARSED_QUERY_CLAUSES = 64
 
 
 @dataclass(slots=True)
--- a/weblate/utils/search.py
+++ b/weblate/utils/search.py
@@ -81,6 +81,8 @@ OPERATOR_MAP = {
     ":>=": "gte",
 }
 
+MAX_PARSED_QUERY_CLAUSES = 4096
+
 
 @dataclass(slots=True)
 class ParsedQueryClause:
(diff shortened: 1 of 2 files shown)


weblate/utils/search.py
```diff
@@ -81,6 +81,8 @@ def __init__(self, tokens) -> None:
     ":>=": "gte",
 }
 
+MAX_PARSED_QUERY_CLAUSES = 4096
```
nijel:
```suggestion
MAX_PARSED_QUERY_CLAUSES = 64
```

PR #19730, opened 2026-05-26
fix(tasks): remove stale scheduled tasks

nijel:
Fixes WEBLATE-39XT


weblate/addons/migrations/0020_remove_obsolete_cleanup_tasks.py
```diff
@@ -0,0 +1,48 @@
...
+    ]
+
+    operations = [
+        migrations.RunPython(
+            remove_obsolete_cleanup_tasks,
+            migrations.RunPython.noop,
+            elidable=True,
+        ),
+    ]
+    ]
```
nijel:
```suggestion
```

PR #19731, opened 2026-05-26
fix(template): avoid duplicate element IDs

nijel:


KarenKonou (approved):
I'm noticing the `shortcuts-table5-container` rules and classes that are leftover from the Bootstrap 5 migrations and can be removed but maybe that should be another PR.

nijel:
Updated the class name as well, thanks for pointing that out.

PR #19774, opened 2026-05-28
Add WEBLATE_SAML_SECURITY_CONFIG Docker environment variable

felixfon:
## Problem

When using SAML authentication with identity providers that enforce multi-factor authentication (e.g. Microsoft Entra ID), login fails with:

> AADSTS75011: Authentication method 'MultiFactor, PasswordlessPhoneSignIn' by which the user authenticated with the service doesn't match requested authentication method 'Password, ProtectedTransport'.

This happens because python3-saml sends a default requestedAuthnContext of urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport, which conflicts with the actual authentication method used.
https://github.com/SAML-Toolkits/python3-saml/blob/52d2ac8da3f35262755f6e1c32ba7c62a6011fe1/src/onelogin/saml2/authn_request.py#L100-L114

## Solution

Expose SOCIAL_AUTH_SAML_SECURITY_CONFIG as the Docker environment variable WEBLATE_SAML_SECURITY_CONFIG, accepting a JSON object. This allows users to customize SAML security settings, for example disabling requestedAuthnContext:

```yaml
WEBLATE_SAML_SECURITY_CONFIG: '{"requestedAuthnContext": false}'
```

### DIFF
--- a/docs/admin/auth.rst
+++ b/docs/admin/auth.rst
@@ -569,6 +569,18 @@ When configuring Weblate SP in your IdP, it is recommended to choose persistent
 
 .. hint::
 
+   Some identity providers (such as Microsoft Entra ID with multi-factor
+   authentication) require disabling the default ``requestedAuthnContext``
+   in the SAML security configuration:
+
+   .. code-block:: python
+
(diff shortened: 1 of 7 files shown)


nijel:

PR #19814, opened 2026-05-29
feat(checks): Add safety checks for MDX format

### LINKED ISSUE #19555: Checks for MDX format
### Describe the problem

The MDX format can include pontentionally dargerous code so there should be a way in Weblate to restrict the impact scope.

### Solution brainstorm

Add MDX specific check that will be turned on by the MDX file format and based on location extension that will ensure that the translated inline code doesn't differ from the source.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

gersona:
- fixes #19555
This only handles the JSX expressions like `{propos.name}`, assuming the other inline codes are already handled by the Markdown check and/or Auto safe HTML checks


gersona:
@nijel I'm not sure if `default_disabled` should be set to `True` as for `MarkdownBaseCheck`
I also assume this will be included in the 2026.7 release.

weblate/formats/base.py
```diff
@@ -242,8 +242,10 @@ def get_extra_flags(self) -> Generator[str | etree._Element | Flags]:
             _, extension = os.path.splitext(location.split(":")[0].strip())
             if extension == ".rst":
                 yield "rst-text"
-            elif extension in {".md", ".markdown", ".mdx"}:
+            elif extension in {".md", ".markdown"}:
                 yield "md-text"
+            elif extension == ".mdx":
+                yield "md-text,safe-mdx"
```
nijel: Does this actually work? Shouldn't it be two yields?

weblate/checks/mdx.py
```diff
@@ -0,0 +1,36 @@
...
+    from weblate.trans.models import Unit
+
+
+class SafeMDXCheck(TargetCheck):
+    """Check for unsafe MDX content."""
+
+    check_id = "safe-mdx"
+    name = gettext_lazy("Safe MDX")
+    description = gettext_lazy("The MDX content is safe.")
+    version_added = "2026.7"
```
nijel, line 172: It should be `default_disabled` as we don't want this to trigger this on all strings, only when turned on by the flag (which the file format does).

weblate/checks/mdx.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def get_jsx_expression_matches(self, text: str):
+        # matches expressions like {props.name.toUpperCase()}
+        jsx_expression_pattern = r"\{[^{}]+\}"
```
nijel: I'm afraid this is too simple; you can have curly braces inside the expression:

```
<button onClick={() => {
  console.log("Clicked!");
  setCount(count + 1);
}}>
  Click Me
</button>

<div style={{ color: 'blue', fontSize: '14px' }}>Text</div>

<div>
  {users.map(({ id, name }) => (
    <p key={id}>{name}</p>
  ))}
</div>
```

PS: I've just quickly Googled for examples, these might be wrong, so first properly evaluate the JSX syntax.

weblate/formats/base.py
gersona: oh my bad, silly mistake

weblate/checks/mdx.py
gersona: You're right, I'll make changes to handle that.
From what I understand, it's a valid example JSX, but the MDX  parser would skip the content inside markups.
But something like the following would be correctly parsed
```mdx
Test {
    [1, 2, 3].map(
        ({ id }) => (<p key={id}>{id}</p>)
        )
}
```

weblate/checks/mdx.py
```diff
@@ -0,0 +1,199 @@
...
+                    i = close + 1
+                    continue
+            if char == "{":
+                # JSX expression
+                close = _scan_expression(text, i)
+                if close is not None:
+                    yield text[i : close + 1]
+                    i = close + 1
+                    continue
+            i += 1
```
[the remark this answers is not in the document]
gersona: AFAIK, code blocks are not extracted from the original file, similar to the import/export statements, so they are skipped at the converter level.

docs/snippets/format-features/mdx-features.rst
```diff
@@ -25,6 +25,6 @@
    * - Supports read-only strings :ref:`ⓘ <read-only-strings>`
      - ``No``
    * - Check flags added by this format :ref:`ⓘ <custom-checks>`
-     - ``auto-safe-html``, ``strict-same``, ``md-text``
+     - ``auto-safe-html``, ``strict-same``, ``md-text``, :ref:`check-safe-mdx`
```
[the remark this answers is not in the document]
gersona: this can probably be ignored

nijel:
@codex review this pull request

weblate/checks/mdx.py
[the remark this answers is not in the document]
gersona: correction, my assuption was incorrect since `md_extract_code_blocks` can actually change that behavior

nijel:

PR #19854, opened 2026-06-01
feat(hooks): warn user if fallback matching was used

nijel:
We want to phase out it, so give users early warning about this.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -1,6 +1,5 @@
 Weblate 2026.7
 --------------
-----------------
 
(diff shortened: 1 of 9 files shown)


nijel:
@amCap1712 This should pave the way to stop supporting the fallback matching in webhooks. I don't think we can drop it right now, but this introduces an alert that makes the fallback matching to users, and they can fix that. https://github.com/WeblateOrg/weblate/issues/19855 will then cover actual code removal.

PR #19870, opened 2026-06-03
feat #17666: Display icons for component priority

### LINKED ISSUE #17666: Display icons for component priority
### Describe the problem

You can select a priority from Very Low to Very High for each component in the component settings.

This is nice to have if you have projects with a huge number of components to direct translators to the more important strings. However, this information of component priority is relatively invisible otherwise.

### Solution brainstorm

I suggest to display a priority icon next to each component list, one for each component. Each icon should also come with a tooltip for the priority.

_No response_

### Screenshots

One idea on how to render it:

Image

The icon is to the left, one for each of the 5 possible priorities from Very High (top) to Very Low (bottom).
All priorities except medium have stylized arrows. Medium is a simple dash.
 Very High is red, High is orange, Medium is orange or yellow, Low is gray, Very Low is dark gray.

This is only one example to get things going, feel free to discuss details.

### Additional context

_No response_

Samuel-k276:
## Changes

- Adds 5 new svgs (1 for each priority)
- Add trumping logic to select the icon based on the component priority
- Display the priority along with the remaining icons related to the component
- Add a test to validate this logic checking each boundary

## Video of all this working

https://github.com/user-attachments/assets/6192b65b-6fa2-4083-bcfe-8f118f23bb8f

Closes #17666

### DIFF
--- a/REUSE.toml
+++ b/REUSE.toml
@@ -144,6 +144,12 @@ SPDX-FileCopyrightText = "Pictogrammers"
 SPDX-License-Identifier = "Apache-2.0"
 
 [[annotations]]
+path = "weblate/static/priorities/**.svg"
+precedence = "aggregate"
+SPDX-FileCopyrightText = "Samuel Gomes <samuel.esteves.gomes@tecnico.ulisboa.pt>"
+SPDX-License-Identifier = "GPL-3.0-or-later"
+
+[[annotations]]
 path = "weblate/static/vendor/prism/**"
 precedence = "aggregate"
 SPDX-FileCopyrightText = "Lea Verou"
--- a/weblate/trans/templatetags/translations.py
+++ b/weblate/trans/templatetags/translations.py
@@ -47,6 +47,7 @@ from weblate.utils.diff import Differ
 from weblate.utils.docs import get_doc_url
 from weblate.utils.hash import hash_to_checksum
 from weblate.utils.html import format_html_join_comma, list_to_tuples
+from weblate.utils.icons import load_icon
 from weblate.utils.markdown import render_markdown
 from weblate.utils.messages import get_message_kind as get_message_kind_impl
 from weblate.utils.random import get_random_identifier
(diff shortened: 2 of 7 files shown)


weblate/trans/templatetags/translations.py
```diff
@@ -1244,6 +1222,35 @@ def indicate_alerts(
     # GhostProjectLanguageStats and GhostCategoryLanguageStats as these would
     # be confusing (showing alert or admin icon on ghost containers).
 
+    priority_html = ""
+    if component and (priority := getattr(component, "priority", None)) is not None:
```
nijel: Why use getattr here?

REUSE.toml
```diff
@@ -143,6 +129,12 @@ precedence = "aggregate"
 SPDX-FileCopyrightText = "Pictogrammers"
 SPDX-License-Identifier = "Apache-2.0"
 
+[[annotations]]
+path = "weblate/static/priorities/**.svg"
+precedence = "aggregate"
+SPDX-FileCopyrightText = "Samuel Gomes <samuel.esteves.gomes@tecnico.ulisboa.pt>"
```
nijel, line 149: Were these icons actually created by you?

weblate/trans/templatetags/translations.py
```diff
@@ -1244,6 +1222,35 @@ def indicate_alerts(
...
+        levels = (
+            (60, "double_arrow_up", "text-danger", gettext_lazy("Priority: Very high")),
+            (80, "single_arrow_up", "text-warning", gettext_lazy("Priority: High")),
+            (100, "dash", "text-info", gettext_lazy("Priority: Medium")),
```
nijel: It might be more reasonable to not show any icon for the default priority. Most components have no priority set, and the dash icon is not really obvious to interpret.

REUSE.toml
Samuel-k276: Yeah created them in figma, is another annotation more adequate?

weblate/trans/templatetags/translations.py
Samuel-k276: Oh yeah, I was initially unsure if priority could be None so I used `getattr`, but after checking its always an int. Fixed!

REUSE.toml
nijel: No, I just wanted to confirm the copyright is actually correct.

nijel:

nijel:
@Samuel-k276 Would you mind looking into https://github.com/WeblateOrg/weblate/issues/20308?

Samuel-k276:
> @Samuel-k276 Would you mind looking into #20308?

Yeah I will have a look

Issue #19891, opened 2026-06-05
License of a workspace doesn't work properly


orangesunny:
### Describe the issue

There are two issues with licenses set for workspaces:

- [x] it seems that all workspaces have the license set to proprietary by default
- [x] scrolling through the list of licenses ends with letter A, other need to ne searched for

### I already tried

- [x] I've read and searched [the documentation](https://docs.weblate.org/).
- [x] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

1. Go to setting of a workspace, like here for Sanbox project https://hosted.weblate.org/workspaces/289c3e8d-5ead-4ec6-bf43-c980f01ada95/#settings
2. see that license is set to proprietary by defalut
3. see that license picker is not fully scrollable

### Expected behavior

- [ ] the license of a workspace is set based on project/components licenses if all have the same
- [ ] license picker is fully scrollable, not just for licenses starting with A

### Exception traceback

```pytb

```

### How do you run Weblate?

weblate.org service

### Weblate version

2026.6

### Weblate deploy checks

```shell

```

### Additional context

Firefox 151.0.3 as Flatpak on Fedora 44.20260602

nijel:
These are two independent issues. @KarenKonou please check the frontend side. The default license is indeed proprietary and there was no attempt to make it match the existing components. I think there will be the same issue on projects with mixed licenses.

orangesunny:
I understand, @nijel.

The license matching is not a big problem; it was just a surprise for me. I hope it won't break anything. Changing this is optional.

Regarding the license picker, that quizes users, so that is the primary issue to be solved here.

PR #19893, opened 2026-06-05
fix: remove options displayed limit in license selector

KarenKonou:
Fixes missing licenses problem mention in #19891


nijel:

PR #19944, opened 2026-06-10
feat(machinery): improve LLMs translation prompting

### LINKED ISSUE #19927: LLM translations improvements
### Describe the problem

Currently we're bumping into a few issues with using LLM translations:

1. There's a fixed prompt with only 2 places to specify extra info. We want to provide different instructions based on the target language. If we need to put it in there for all languages, it's huge and makes it harder for the LLM to process.
2. The glossary passed to the LLM is just sending one translation per string without any extra context/explanation. If you have e.g. "account" twice in the glossary (once for a bank account, and once for a user account), it's hit or miss whether the correct translation will be sent. There's also no way to make glossary strings case-sensitive afaik.
3. The Czech prefilled previous response is sometimes causing Czech translations instead of the target language. (Edge-case, but important)

### Solution brainstorm

1. Giving the ability to add specific instructions per language could help.
2. Providing all glossary translations with additional context/explanation might help here. And allowing case-sensitivity in the glossary.
3. We're using Anthropic that supports an output format instead in later versions. Don't know about other APIs.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

nijel:
- pass additional info on glossaries
- use real examples from the project rather than hard-coded ones
- fallback to Weblate shipped examples that will be localized
- it is now possible to provide language-specific prompt customization

Fixes #19927


weblate/machinery/llm.py
```diff
@@ -331,16 +380,104 @@ def _build_message(
...
+        instructions: str = "",
     ) -> str:
         result = {
             "source_language": source_language,
             "target_language": target_language,
             "glossary": glossary,
             "strings": texts,
         }
+        if instructions:
+            result["instructions"] = instructions
```
dylankiss: Very nice that we can specify instructions per language. 💯 
However, this will send the instructions for each string in the batch. If the batch contains the same target language for each string, this consumes a lot of tokens that can be avoided, right?

Could it be an idea to split the batches per target language and provide this once, or would that be too difficult/not wanted?

weblate/machinery/llm.py
```diff
@@ -724,14 +850,225 @@ def _get_message(
...
+            json.dumps([example["target"] for example in examples]),
+        )
+
+    def _get_previous_messages(
+        self,
+        source_language: str,
+        target_language: str,
+        sources: list[tuple[str, Unit | None]],
+    ) -> tuple[str, str]:
+        project_examples = self._get_project_previous_examples(source_language, sources)
```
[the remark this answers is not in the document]
nijel: Not really worth it because the difference by examples would be small.

weblate/machinery/llm.py
nijel: It sends it in the structured data, not per string. Including it in the prompt might be an option as well.

weblate/machinery/llm.py
```diff
@@ -724,14 +850,225 @@ def _get_message(
...
+        try:
+            candidates = unit_set.filter(
+                state__gte=STATE_TRANSLATED,
+                state__lt=STATE_READONLY,
+            ).exclude(target="")
+            if current_unit_ids:
+                candidates = candidates.exclude(pk__in=current_unit_ids)
+            candidates = candidates.order_by("-last_updated")[
+                : LLM_PREVIOUS_EXAMPLE_LIMIT * 4
+            ]
```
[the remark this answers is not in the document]
nijel: I think we rejected the idea of the last_updated index in the past, as it is updated too frequently and used not that frequently so that the overhead it brings outweighs the update cost. And I don't think this would change it.

weblate/machinery/llm.py
dylankiss: Ah, right, I misread. Then I guess it should be fine.
Is there a difference between sending it in the prompt or in the message?
Both are sent an equal number of times, right? Don't know if there's a difference in token counting.

nijel: I don't think there is a good reason. I was here because I was changing the payload, but it is probably better to do that in the system prompt to be consistent with other settings.

PR #19946, opened 2026-06-10
docs(config): improve allowed asset size configuration section

michael-smt:
It was only noted in the 5.17 changes so far.


nijel:

PR #19958, opened 2026-06-10
add optional user messages on translation changes (#17758)

### LINKED ISSUE #17758: User-provided Messages on Changes
### Describe the problem

The change history is great for seeing **what** changed, but does not help in understanding **why** the change was done.

### Solution brainstorm

Give the user the ability to provide a message when doing changes. For example describing why a certain bulk editing is being done.

This would then be displayed in all the change(s) that resulted from the action.

For single translations comments can be used to leave a trace in the change history. But this does not scale for mass operations.

### Screenshots

_No response_

### Additional context

Something like what commit messages are for git.

dinis-sales:
Allows users to provide an optional justification message when making 
modifications, improving context and accountability in the change history.

Key implementations:

- Adds an optional text area directly to the single translation view.
- Integrates a message field into the Bulk Edit and Search & Replace forms.
- Automatically propagates the batch action message across all individual change records.
- Displays the saved justification notes inside the string's History tab, CSV exports, and REST API.

Closes #17758 

https://github.com/user-attachments/assets/6b3cf3c8-9a62-4f9a-b9f8-12404fd82a90


nijel:
Thanks for your contribution. I think this will need some polishing:

* We already often get feedback that our UI is too complex. Adding a new field to the translate editor will increase cognitive load there. It will also confuse the users to think that the reasoning for the change is needed. I think the message input should be hidden by default.
* I'm not sure whether this requires a dedicated field on the Change model. Storing it inside details should work as well.

I will now also trigger Copilot review, it usually gives a good hints, but still it can make mistakes as well.

weblate/locale/django.pot
nijel, line 1: Avoid updating POT files; these are automatically managed.

weblate/templates/mail/changed_translation.html
```diff
@@ -16,6 +16,12 @@ <h2>{% translate "Translation change" %}</h2>
     {% format_unit_target unit value=target diff=old %}
   {% endif %}
 
+  {% if change.message %}
+    <h2>{% translate "Note:" %}</h2>
```
nijel, line 20: I think we need a better name here to clearly express this is the note made while translating and to avoid confusion with existing content such as comments, source info or explanation.

nijel:
Please avoid changes to the PO and POT files. We will update these once the code is merged to avoid conflicts while still developing the pull request.

nijel:
@codex review

weblate/api/views.py
```diff
@@ -3718,10 +3723,18 @@ class ChangeFilter(filters.FilterSet):
...
     action = filters.MultipleChoiceFilter(choices=ActionEvents.choices)
     user = filters.CharFilter(field_name="user__username")
+    message = filters.CharFilter(
+        field_name="message",
+        lookup_expr="icontains",
+        help_text=gettext_lazy(
+            "Return only changes whose user-provided message contains this text "
+            "(case-insensitive)."
+        ),
+    )
```
nijel, line 3733: I don't think this works - there is no message fields on the model.

weblate/trans/models/change.py
```diff
@@ -776,6 +787,21 @@ def log_event(self) -> None:
...
+        return self.details.get("message", "")
+
+    @message.setter
+    def message(self, value: str) -> None:
+        if self.details is None:
+            self.details = {}
+        if value:
+            self.details["message"] = value
+        else:
+            self.details.pop("message", None)
```
nijel, line 803: Why masking this as a property?

PR #19970, opened 2026-06-11
fix(api): restrict editing for scoped teams

nijel:
The teams with project or workspace scope should not be fully editable and only limited actions should be available here.


amCap1712:
LGTM, Thanks!

PR #20031, opened 2026-06-13
fix: add author language squash mode

### LINKED ISSUE #14440: Squash commits only if author + language match
### Describe the problem

Currently, it is possible to set up squashing by author, language, or all into one.
All of these have issues:

1. Author: says "Translated using Weblate to <language>" but commit may contain many other languages
2. Language: the least worst, but mixes potentially harmful contributions with valid contributions in a single commit, loses authorship
3. Into one commit: Just everything specific is lost in the end.

Instead of squashing based only on those properties specifically, squash based on the author and the language. That way, contributions to multiple files are squashed into one, quick changes in other commits are squashed, but authorship isn't lost, and it becomes easier to find issues.

I can't think of much else, I like this approach, microG seems to do this manually which I found very nice

### Screenshots

_No response_

### Additional context

_No response_

suzunn:
## Problem

The Git squash add-on can squash by author or by language, but neither mode keeps both dimensions intact. Squashing only by author can still combine translations for multiple languages into one commit, while squashing only by language loses the original author grouping.

Fixes #14440.

## Solution

I added a `Per author and language` squash mode. It reuses the existing author squash flow, but groups pending commits by both author email and the language inferred from the translation files changed by each commit. Commits that do not map cleanly to a single language keep their own commit hash in the grouping key, so mixed or non-translation commits are not accidentally merged together.

I also exposed the new mode in the add-on form, updated the help text, and added regression coverage for the expected commit count.

## Testing

- `uvx ruff check weblate/addons/git.py weblate/addons/forms.py weblate/addons/tests.py`
- `uvx ruff format --check weblate/addons/git.py weblate/addons/forms.py weblate/addons/tests.py`
- `py -3 -m py_compile weblate/addons/git.py weblate/addons/forms.py`
- `git diff --check`

I also tried to run `uv run pytest weblate/addons/tests.py::GitSquashAddonTest::test_author_language -q`, but dependency installation failed before tests started because `pyicu==2.16.2` needs `pkg-config` or `ICU_VERSION` in this Windows environment.


suzunn:
Still interested in landing this — rebased onto latest `main` to keep it current. Happy to address any review feedback.

nijel:
I think the Copilot feedback is still valid.

suzunn:
Thanks @nijel — you're right, addressed the Copilot feedback in 5a5732f:

- `get_commit_language()` now takes a precomputed `filename -> language` mapping built once in `squash_author()`, so attribution is O(changed files) per commit instead of scanning every language's full filename list.
- A commit is grouped by language only when *every* touched file maps to the same single language. Commits that also modify non-translation files, or that span multiple languages, are now kept in their own group — matching the documented "mixed or non-translation commits are not merged" behavior.

Added `test_get_commit_language` covering the single-language, multi-language, mixed (translation + non-translation), non-translation-only, and empty cases. Note: I couldn't run the full Django suite locally (the DB/Celery services aren't set up on my machine), so please let CI confirm — happy to adjust if anything's off.

nijel:
@codex review

docs/snippets/addons-autogenerated.rst
nijel:
```suggestion
* Per author
* Per author and language
```

nijel:
Thanks, looks good now. Scheduled for merge if tests pass.

PR #20124, opened 2026-06-17
fix(vcs): avoid duplicating git command on retry

nijel:
### Motivation

- A retry path in the VCS subprocess wrapper could prepend the VCS executable a second time, producing commands like `git git reset --hard` and triggering runtime errors observed in production.

### Description

- Change `_popen` in `weblate/vcs/base.py` to build a separate `cmd` list (`cmd = args if fullcmd else [cls._cmd, *args]`) and pass `cmd` to `subprocess.run` so the original `args` remains unchanged for retry logic.
- Add a regression test `test_popen_retry_does_not_duplicate_command` in `weblate/vcs/tests/test_vcs.py` that mocks `subprocess.run` and `should_retry_popen` to assert both attempts use `["git", "reset", "--hard"]` (no duplicated executable).
- Preserve existing behavior for breadcrumbs, environment handling, and retry decisions; only the subprocess argument construction was adjusted.

### Testing

- Ran the new regression test with `uv run pytest weblate/vcs/tests/test_vcs.py::RepositoryTest::test_popen_retry_does_not_duplicate_command` and it passed.
- Ran the test class with `uv run pytest weblate/vcs/tests/test_vcs.py::RepositoryTest` and all tests passed (`27 passed`).
- Performed a quick repository check (`git diff --check`) with no issues reported.
- Formatting/linting via the repository pre-commit hooks (`uv run prek run ruff-format && uv run prek run ruff`) could not complete in this environment because cloning pre-commit hooks from GitHub failed with a network/403 error, so those checks were not run here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a327dc56b3c8329845f128673eabc9f)

### DIFF
--- a/weblate/vcs/tests/test_vcs.py
+++ b/weblate/vcs/tests/test_vcs.py
@@ -9,6 +9,7 @@ import json
 import os.path
 import re
 import shutil
+import subprocess
 import tempfile
 from contextlib import ExitStack
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


weblate/vcs/tests/test_vcs.py
```diff
@@ -264,6 +266,36 @@ def test_repository_error_negative_retcode_keeps_original_message(self) -> None:
...
+
+        with (
+            patch.object(GitRepository, "should_retry_popen", return_value=True),
+            patch(
+                "weblate.vcs.base.subprocess.run",
+                side_effect=[failed_process, successful_process],
+            ) as run,
+        ):
+            # ruff: ignore[private-member-access, hardcoded-temp-file]
+            GitRepository._popen(["reset", "--hard"], cwd="/tmp")
```
nijel:
```suggestion
            tempfile.TemporaryDirectory() as cwd,
            patch.object(GitRepository, "should_retry_popen", return_value=True),
            patch(
                "weblate.vcs.base.subprocess.run",
                side_effect=[failed_process, successful_process],
            ) as run,
        ):
            # ruff: ignore[private-member-access]
            GitRepository._popen(["reset", "--hard"], cwd=cwd)
```

PR #20138, opened 2026-06-18
feat(add-ons): expose category in Fedora messaging and clarify docs

nijel:
- expose category in Fedora messaging topic
- document in more detail topic and change events


nijel:
@Jibec This adds the missing bits for the Fedora messaging docs (and fixes the topic for categorized components, which is a gap I noticed while reviewing the docs).

Jibec:
thank you!

PR #20154, opened 2026-06-19
fix(checks): preserve MDX expression context in SafeMDX check

nijel:
### Motivation
- The original `SafeMDX` check compared only the multiset of brace-delimited JSX expressions and ignored their JSX/MDX syntactic context, producing false negatives when expressions were moved between text and attribute contexts.
- This could allow translations to change runtime behavior (e.g., move an expression into an `href` attribute) without being detected by the safety check.

### Description
- Replace the multiset comparison with ordered expression signatures that include expression text, the surrounding context (`text`/`tag`/`attribute`), the attribute name when applicable, and a best-effort element stack.
- Add lightweight JSX context helpers: a tag/attribute regex, `get_jsx_expression_signatures`, `get_jsx_expression_context`, `get_jsx_element_stack`, and `_iter_jsx_expressions` while preserving the original expression scanner.
- Adjust `check_single` to compare ordered signatures (`found != expected`) instead of `sorted(found) != sorted(expected)`, so context or order changes fail the check.
- Add regression tests to `weblate/checks/tests/test_mdx.py` that cover moving an expression from element text into an attribute and swapping expressions between attribute/text contexts, and add an assertion that `get_jsx_expression_signatures` exposes the expected contexts.

### Testing
- Ran `uv run pytest weblate/checks/tests/test_mdx.py -q`, and the test file passed with all tests succeeding (16 passed, 4 skipped).
- Ran `uv run python -m compileall -q weblate/checks/mdx.py weblate/checks/tests/test_mdx.py`, which completed successfully.
- Attempted `uv run prek run ruff-format --files weblate/checks/mdx.py weblate/checks/tests/test_mdx.py`, but the pre-commit hook run was blocked by an external repo clone error (network/tunnel 403) and could not complete.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a34d74e8ec88329a51c82dc4dfc1abc)

### DIFF
--- a/weblate/checks/mdx.py
+++ b/weblate/checks/mdx.py
@@ -16,7 +16,7 @@ if TYPE_CHECKING:
 
     from weblate.trans.models import Unit
 
-_JSX_TAG_NAME_RE = re.compile(r"[A-Za-z][\w.:-]*")
+_JSX_TAG_NAME_RE = re.compile(r"[$_A-Za-z][\w.$:-]*")
 _ATTR_NAME_RE = re.compile(r"([A-Za-z_:][\w:.-]*)\s*=\s*$")
 
 # Various lexical contexts tracked while scanning a JSX expression.
--- a/weblate/checks/mdx.py
+++ b/weblate/checks/mdx.py
@@ -4,6 +4,7 @@
 
 from __future__ import annotations
 
+import re
 from typing import TYPE_CHECKING
 
 from django.utils.translation import gettext_lazy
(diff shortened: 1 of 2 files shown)


nijel:
@gersona This came out of Codex Security; please review carefully (I'm don't know enough MDX to evaluate this).

gersona:
This looks like a valid PR to me, both comments are also relevant so it's worth applying fixes to them as well

PR #20161, opened 2026-06-19
chore(tests): rewrite screenshot capture in Selenium

nijel:
Rely on Chrome devtools Page.captureScreenshot to get full page capture rather than resizing the window and waiting fixed delay. That delay was sometimes not enough for content to fully render.


nijel:
Okay, while this looked as a good idea, it turned out to not work that well and would need more hacks (like injecting CSS for grayed out background), so let's revert back to resizing, just fix the timing issues there.

PR #20164, opened 2026-06-19
feat: finish removing jquery

### LINKED ISSUE #19767: Remove jQuery
### Describe the problem

The majority of jQuery usages in weblate can be trivially rewritten with native JS. Removing jQuery will reduce client bundle size, remove another package that needs to be continuously audited and checked when upgrading, and lead to an overall more modern approach to writing frontend code. Most new frontend code is already written without jQuery.

### Solution brainstorm

This can very easily be done in stages, rewriting code blocks and functions one by one, removing the jQuery package only when all usages are migrated. Helper functions can be written for some common patterns (AJAX for example).

_No response_

### Screenshots

_No response_

### Additional context

_No response_

KarenKonou:
closes #19767


KarenKonou:
The Argos diff seems to be from stale screenshots still, the issue it shows should be fixed in the latest commits

nijel:
I think there are real differences

* `labels.png` has different widht of the labels, this might be non-related artifact
* `automatic-translation.png` shows both component-based and engine-based choices, but it should toggle them based on the *Source of automated translations* radio


PS: I removed the links, they point elsewhere than I hoped.

nijel:
The delay really didn't help, perhaps the script is actually broken there?

KarenKonou:
It's strange because my local screenshot looks fine and it displays fine in the browser as well, I'll see if i can figure out what's going on
automatic-translation

nijel:

PR #20196, opened 2026-06-22
feat(add-ons): configure full URL for Fedora Messaging

nijel:
Providing host and SSL flag is not enough in some situations, so allow to configure the full URL.


Jibec:
wow, you do merge fast!

the full URL to use is: amqps://weblate:@rabbitmq.fedoraproject.org/%2Fpubsub
Fedora infra people says it is to specify the "/pubsub" virthost https://forge.fedoraproject.org/infra/tickets/issues/8291#issuecomment-825673

I am not used of using double slashes in URL and wonder if will this URL work? or with `//`?
sorry for the noob question ;)

Issue #20134, opened 2026-06-18
TBX file download does not include metadata


michael-smt:
### Describe the issue

Metadata of the units are not present when all files are downloaded in a glossary component with TermBase eXchange format.

### I already tried

- [x] I've read and searched [the documentation](https://docs.weblate.org/).
- [x] I've searched for similar filed issues in this repository.

### Steps to reproduce the behavior

1. Create a glossary component with TermBase file format
2. Add a term
3. Mark that term as terminology, it gets the `terminology` flag
4. Mark that term as untranslateable, it gets the `read-only` flag
5. Add a custom flag like `max-length:50`
6. Go to "File" -> "Download translation files as ZIP file"

The TBX files do not have any metadata in them.

### Expected behavior

The TBX files contain the metadata so a new component could be created from them with minimal metadata loss.

### Exception traceback

```pytb

```

### How do you run Weblate?

Docker container

### Weblate version

2026.6.1.0

### Weblate deploy checks

```shell

```

jasperdingg:
I'd love to give this a try, any preference on how the flags should be represented in the TBX output?

nijel:
@jasperdingg Most of these (if not all) are properly stored on native TBX storage, so the exporter should follow that.

PR #20199, opened 2026-06-22
docs: add more detailed guidance on using glossaries

### LINKED ISSUE #15610: Trouble understanding the Glossary features
### Describe the issue

The Weblate documentation is pretty good in many ways but when it comes to Glossary features, information is scarce and important points missing, which are surely clear to developers and long-time users, but hard to figure out for new adopters like me. So, I think it might help to share my experience from the point of view of a new user and the difficulties I had (and partially still have) in understanding the intended way of use.

I want to start by saying that I have read the documentation a dozen of times and read through all related issues and discussions on the topic.
One of the primary issues that I'm failing to understand is 

### The difference between Glossary and Terminology items

Docs say 

> Flagging certain glossary terms as terminology [..] adds entries for them to all languages in the glossary. 

Which "entries" does this add? Seemingly empty entries. Okay - but that's more of a one-time action - what's the flag for, then?
You can add translations for normal glossary items as well. 

**When you have added translations to all languages for a glossary item - what's the different then, between this item and an item flagged as "Terminology"?**
**What's the difference between Glossary and Terminology items?**

Trying to understand the intgended logic behind this has driven me nuts... I couldn't and still can't.

### Importing Glossary/Terminology

This sentence in the docs (for terminology)

> Use this for important terms that should be well thought out, and retain a consistent meaning across all languages.

has given me some confidence in matching up with our requirements and so we have prepared three collections for importing:

1. "Glossary"
  Terms which might not be commonly known with English explanations for the translators (English is our source language). 
  Since every translator needs to know English well enough, I see no reason for why any of this should ever be translated to another language (neither terms nor explanations)
2. "Terminology"
  This list includes important terms which should always be translated in the exact same way (button, label, heading and other important UI element captions, feature names, etc.). We don't have translations for them yet, so English-only
3. "Non-Translatables"
  Terms that should never be translated (brand names, technology names, our add-ons which don't have translated names, etc)., including English desciptions

Now, I wanted to import these as three separate glossary components. TBX is the most capable format, so I created TBX files. For (3), I added `<termNote type="translationNote">no</termNote>` to mark them all as non-translatable. Example:

```xml
      <termEntry id="term_3">
        <descrip type="definition">Website that provides subtitle files for movies and TV shows</descrip>
        <langSet xml:lang="en">
          <tig>
            <term>opensubtitles.com</term>
            <termNote type="translationNote">no</termNote>
            <note>NEVER TRANSLATE - Website domain name | Category: Online services</note>
          </tig>
        </langSet>
      </termEntry>
```

I named the file `untranslatable.tbx` - import failed, seemingly due to no language code. 
Then I tried `untranslatable.en.tbx` - then it refused to import due to duplicate languages: en,en
Err - what? Where's the duplication?
I finally got it to import by naming it `untranslatable.en_US.tbx`

**So - in order to import untranslatable terms, I need to import it with translations included?**

**TBX is a multilingual (not bi-) format. But it needs a language code in the file name for importing, even when that language isn't included in the file?**

### Language Files Auto-Creation

After importing the non-translatable strings, it unexpectedly started to create language files - why?
I thought I might have installed the "Add missing language files" add-on accidentally - but that wasn't the case.
I removed the files from VCS, but shortly after it created the files again. **Obviously it doesn't make sense for untranslatable terms to have language files for translation**. 

On top of that, it even showed this in the statistics as "unfinished" - but there is nothing to finish!

I ended up using a language filter.  I still had to allow `en_US`  due to the file name, so I ended up with "50% translated" in the statistics. Better but still not how it should be.

While reading through issues and discussions here, somebody mentioned that language files would be auto-created for glossary components. **I can't remember having read that anywhere in the documentation**.

### Glossary and Untranslateable doesn't work without (empty languages)

After I had imported my glossary and untranslatables - both filtered for en and en-US only - I realized that none of them was actually working. **It seems they do need empty language files to work - even though it's only about the source language. Why that?**

### How is all this supposed to work?

At this point, I decided to create this issue. Still, too many thing are unclear and undocumented.

How do I need to set up the three components from above?

1. English-Only Glossary
  - Does it need all language files?
  - How to exclude those languages from statistics?
  - How to avoid users trying to translate this?
2. "Terminology"
  - Setting the "terminology" flag, doesn't create empty translations - how can I do the same like the item in the tools menu for all items?
3. "Non-Translatables"
  - How to import the terms?
  - Does it need all language files?
  - Do the items need to be set to "terminology" first and then to "read-only", or is read-only sufficient?

I'm confused, thankful for any guidance how these things are supposed to work.

Thanks
sw

### I already tried

.

### Expected behavior

_No response_

### Screenshots

_No response_

### Exception traceback

```pytb

```

### How do you run Weblate?

Other

### Weblate versions

_No response_

### Weblate deploy checks

```shell

```

### Additional context

_No response_

nijel:
- choosing the right content
- using flags to change the behavior
- describe import in more detail
- highlight common usage patterns

Fixes #15610

### DIFF
--- a/docs/user/glossary.rst
+++ b/docs/user/glossary.rst
@@ -3,19 +3,40 @@
 Glossary
 ========
 
-Each project can include one or more glossaries as a shorthand for storing
-terminology. Glossaries help to maintain consistency of the translation.
+Each project can include one or more glossaries for storing terms and their
+preferred translations. Glossaries help keep translations consistent by showing
+matching terms in the translation editor.
 
-A glossary for each language can be managed on its own, but they are
-stored together as a single component which helps project admins
-and multilingual translators to maintain some cross-language consistency as well.
-Terms from the glossary containing words from the currently translated string are
-displayed in the sidebar of the translation editor.
+Glossaries are regular translation components with special behavior. The source
+language stores the terms Weblate searches for in source strings. Each glossary
+translation stores the preferred term, forbidden term, or untranslatable entry
+for one target language.
+
+Terms from the glossary containing words from the currently translated source
+string are displayed in the sidebar of the translation editor.
 
 .. hint::
 
    The glossary terms are not used in quality checks unless you enable that,
    see :ref:`check-check-glossary` for more information.
+
+How glossary matching works
+---------------------------
+
+When you translate a regular component, Weblate searches the source string for
+terms from the source language of the project glossaries. When a term matches,
+Weblate looks up the corresponding glossary entry for the language currently
+being translated and shows it in the editor.
+
+This means glossary entries need to exist in the target glossary language to be
+shown while translating that target language. A source-language-only glossary
+can still store definitions and context, but it will not provide target-language
+matches for languages that do not have glossary entries.
+
+If you want a term to be present in every glossary language, mark the source
+entry as :ref:`glossary-terminology`. If the term should stay unchanged in every
+language, use :ref:`glossary-untranslatable`; combine it with terminology only
+when you want Weblate to create and maintain entries for all glossary languages.
 
 Managing glossaries
 -------------------
@@ -29,10 +50,10 @@ Managing glossaries
 Use any component as a glossary by turning on :ref:`component-is_glossary`.
 You can create multiple glossaries for one project.
 
-An empty glossary for a given project is automatically created with the project.
-Glossaries are shared among all components of the same project, and optionally
-with other projects using :ref:`component-links` from the respective glossary
-component.
+An empty glossary for a given project is automatically created with the
+project. Glossaries are shared among all components of the same project, and
+optionally with other projects using :ref:`component-links` from the respective
+glossary component.
 
 The glossary component looks like any other component in Weblate with added
 colored label:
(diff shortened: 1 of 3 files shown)


michael-smt:
Thanks, this helps a lot. It confirms what I tried to figure out myself last week :sweat_smile:

docs/user/glossary.rst
```diff
@@ -136,6 +214,70 @@ Similar to regular translation components, you can upload existing glossaries
 to Weblate. Formats like :doc:`/formats/csv` or :doc:`/formats/tbx` are
 supported and can be uploaded, see :ref:`upload`.
 
+TBX files can include explanations, read-only metadata, and glossary flags; see
```
michael-smt, line 217: AFAIK arbitrary flags can't be imported in the TBX format, at least that's my understanding from the code and from exporting a TBX component with flags (#20134).

If I'm wrong and it's possible, it would be great if we could document the XML tags needed in the format page.

nijel: The generic XML flags logic should work using the `weblate-flags` attribute, but this is probably only documented on XLIFF now: https://docs.weblate.org/en/latest/formats/xliff.html#specifying-translation-flags

PR #20205, opened 2026-06-22
docs(formats): extend tbx example with types, description and explanations

michael-smt:
It gives a hint about the `weblate-flags` attribute for terminology, see https://github.com/WeblateOrg/weblate/pull/20199#discussion_r3452018701

I did not manage to get variants flags to work yet via importing of a TBX file, so leaving that off for now.

Also drops the duplicate (and conflicting, in the case of read-only) information on the special tags.


nijel:

PR #20245, opened 2026-06-24
fix(checks): avoid repeated SafeMDX context scans

nijel:
### Motivation
- The SafeMDX check performed per-expression prefix rescans and materialized full signature lists, which makes it O(n^2) on attacker-controlled MDX input and enables a CPU exhaustion DoS.
- MDX format enables the `safe-mdx` check by default, so this path is reachable for uploaded/edited MDX translations and must be made linear-time.

### Description
- Compare expected and found JSX expression signatures lazily in `check_single` using `zip_longest` so the check can short-circuit on the first mismatch instead of building full lists. (`weblate/checks/mdx.py`)
- Implement a single-pass `get_jsx_expression_signatures` that maintains the current JSX element `stack` and `open_tag` state while scanning so expressions reuse previously computed context instead of rescanning prefixes. (`weblate/checks/mdx.py`)
- Extend `get_jsx_expression_context` to accept an optional precomputed `stack` and `tag_context` so the scanner can provide context without extra work, while keeping direct callers supported. (`weblate/checks/mdx.py`)
- Add a regression test `test_check_single_stops_on_first_mismatch` to ensure `check_single()` stops after the first mismatch and does not incur repeated context calls, and mark the complex scanner with `# noqa: C901` to satisfy linting. (`weblate/checks/tests/test_mdx.py`)

### Testing
- Ran `uv run pytest weblate/checks/tests/test_mdx.py -q` with final result `20 passed, 4 skipped, 2 subtests passed`.
- Ran formatting and lint checks with `uv run ruff format weblate/checks/mdx.py weblate/checks/tests/test_mdx.py` and `uv run ruff check weblate/checks/mdx.py weblate/checks/tests/test_mdx.py`, both of which passed for the modified files.
- Attempted the repository pre-commit run `uv run prek run ruff-format --files weblate/checks/mdx.py weblate/checks/tests/test_mdx.py`, but it was blocked by an external hook clone failure (`CONNECT tunnel failed, response 403`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3bc0ca8a288329b4417c5d2367484e)

### DIFF
--- a/weblate/checks/mdx.py
+++ b/weblate/checks/mdx.py
@@ -273,6 +273,22 @@ def _find_unclosed_jsx_tag(text: str, start: int) -> tuple[int, str] | None:
     return None
 
 
+def _find_attribute_name(text: str, equals: int) -> str | None:
+    """Find the JSX attribute name before an equals sign."""
+    i = equals - 1
+    while i >= 0 and text[i].isspace():
+        i -= 1
+    end = i + 1
+    while i >= 0 and re.match(r"[\w:.-]", text[i]):
+        i -= 1
+    if end == i + 1:
+        return None
+    name = text[i + 1 : end]
+    if re.match(r"[A-Za-z_:]", name[0]):
+        return name
+    return None
+
+
 class SafeMDXCheck(TargetCheck):
     """Check for unsafe MDX content."""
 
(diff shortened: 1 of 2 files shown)


nijel:
@gersona Another Codex Security produced path for this.

PR #20275, opened 2026-06-25
docs(formats): remove empty translation in untranslateable term from tbx example

michael-smt:
The untranslatable flag does not get applied on import when a translation langSet is present.


nijel:

PR #20282, opened 2026-06-25
feat(machinery): add structured LLM placeholder parts

nijel:


nijel:
Let's take this as it is and revisit the highlighting logic later - https://github.com/WeblateOrg/weblate/issues/20317

PR #20307, opened 2026-06-26
fix(github): bound pending installation storage

nijel:
### Motivation
- Prevent unbounded persistent storage of full GitHub App installation webhooks that can be triggered by public app installations and lead to database growth/DoS.  

### Description
- Add a one-day retention window (`PENDING_GITHUB_INSTALLATION_RETENTION`) and a per-host pending-row cap (`PENDING_GITHUB_INSTALLATION_HOST_LIMIT`) to limit stored unauthorised installation payloads.  
- Replace storing the full signed webhook JSON with a minimized payload via `_pending_github_installation_payload` that keeps only the installation id, app id, account login/type, action, and cached repository entries.  
- Prune stale pending rows in `_prune_pending_github_installation_events` and skip creating new pending rows when the per-host limit is reached.  
- Add regression tests to verify payload minimization and host-limit behavior in `weblate/vcs/tests/test_github_hooks.py` and keep existing setup-apply coverage in `weblate/vcs/tests/test_github_views.py`.  

### Testing
- Ran the focused pytest scenarios: `uv run pytest weblate/vcs/tests/test_github_hooks.py::TestGitHubAppHooks::test_installation_created_without_row_is_pending weblate/vcs/tests/test_github_hooks.py::TestGitHubAppHooks::test_installation_created_pending_host_limit weblate/vcs/tests/test_github_views.py::GitHubInstallationViewTest::test_setup_applies_pending_installation_webhook`, and all tests passed.  
- Ran the repository formatting/lint hooks with `uv run prek run --files weblate/trans/views/hooks.py weblate/vcs/tests/test_github_hooks.py` and the checks passed after formatting fixes.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3e781f064083298f8199eb8b047b99)

### DIFF
--- a/weblate/trans/views/hooks.py
+++ b/weblate/trans/views/hooks.py
@@ -5,12 +5,14 @@ from __future__ import annotations
 
 import re
 from collections.abc import Callable, Mapping
+from datetime import timedelta
 from functools import partial
 from typing import TYPE_CHECKING, ClassVar, NotRequired, TypedDict, cast
 from urllib.parse import quote, urlparse
 
 from django.conf import settings
 from django.db.models import Q
+from django.utils import timezone
 from drf_spectacular.utils import OpenApiParameter, extend_schema
 from rest_framework import parsers, serializers
 from rest_framework.exceptions import (
@@ -73,6 +75,9 @@ BITBUCKET_HG_REPOS = (
     "hg::ssh://hg@{server}/{full_name}",
     "hg::https://{server}/{full_name}",
 )
+
+PENDING_GITHUB_INSTALLATION_RETENTION = timedelta(days=1)
+PENDING_GITHUB_INSTALLATION_HOST_LIMIT = 1000
 
 GITHUB_REPOS = (
     "git://github.com/%(owner)s/%(slug)s.git",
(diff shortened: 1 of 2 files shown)


nijel:
@amCap1712 This Codex Security generated. We definitely need some approach to limit this.

weblate/trans/views/hooks.py
```diff
@@ -74,6 +76,9 @@
     "hg::https://{server}/{full_name}",
 )
 
+PENDING_GITHUB_INSTALLATION_RETENTION = timedelta(days=1)
+PENDING_GITHUB_INSTALLATION_HOST_LIMIT = 1000
```
amCap1712, line 80: Not sure how useful this limit is because most people will be using Github only, at least on Hosted Weblate.

weblate/trans/views/hooks.py
```diff
@@ -507,15 +512,64 @@ def _github_repository_entries(repositories, hostname: str) -> list[dict]:
...
+    return {
+        "action": data.get("action"),
+        "installation": {
+            "id": payload.get("id"),
+            "app_id": payload.get("app_id"),
+            "account": {
+                "login": account.get("login"),
+                "type": account.get("type"),
+            }
+            if isinstance(account, Mapping)
```
amCap1712, line 528: The isinstance checks can be removed. Codex has a habit of being overly defensive.

weblate/trans/views/hooks.py
```diff
@@ -507,15 +512,64 @@ def _github_repository_entries(repositories, hostname: str) -> list[dict]:
...
+            provider=InstallationProvider.GITHUB, hostname=hostname
+        ).count()
+        >= PENDING_GITHUB_INSTALLATION_HOST_LIMIT
+    ):
+        LOGGER.warning(
+            "Skipped pending GitHub account %s/%s webhook; host limit reached",
+            hostname,
+            installation_id,
+        )
+        return
```
amCap1712, line 566: The problem here is that the failure is silent from a user perspective. We need to surface this issue somehow to the user.

Maybe an IP rate limit on the relevant endpoint is a better option?

weblate/trans/views/hooks.py
```diff
@@ -507,15 +512,64 @@ def _github_repository_entries(repositories, hostname: str) -> list[dict]:
...
+            "id": payload.get("id"),
+            "app_id": payload.get("app_id"),
+            "account": {
+                "login": account.get("login"),
+                "type": account.get("type"),
+            }
+            if isinstance(account, Mapping)
+            else {},
+        },
+        "repositories": _github_repository_entries(data.get("repositories"), hostname),
```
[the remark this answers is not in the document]
amCap1712: I would not suggest capping but not storing at all then and making a separate API call to Github when registration completes.

amCap1712:
@nijel Yes, this issue crossed my mind while implementing as well. I think we should reduce the pending time storage to say 15 mins (with a cronjob that clears it) + rate limit on that endpoint triggering the connection.

weblate/trans/views/hooks.py
nijel: The "attack" surface is described as user creating pending installations. I think it makes sense to limit the amount here, but I have no clue how much.

nijel:
If 15 minutes is sensible (I have no clue how big delays to expect), then it's clearly viable approach.

nijel:
@amCap1712 Please let me know if I should iterate with Codex here or if you prefer to deal with this yourself.

amCap1712:
Sure, I'll take care of this.

amCap1712:
Opened #20354

PR #20333, opened 2026-06-27
feat(widgets): add compact language progress matrix

nijel:
This allows to better display progress when many languages are involved. Now used as default widget in the pull requests.


nijel:
Example:

Translation status

Issue #2699, opened 2019-04-19
Per translator review status


comradekingu:
The single most important thing to ensure quality is knowing what strings are seen, and which ones aren't, more importantly, by whom. I think this would improve quality quite a bit, because it makes it easier to focus without putting a lot of time into keeping track of things, and it allows people that only want to review to be part of the process.

comradekingu:
A binary self-review falls short of a confidence level to issue a suggestion/translation/review by, at ones own ability, where "needs review" is now someone's opinion, or extra even more unknown parties, of a bad source string, badly translated string, possible own or mistakes of others, or research needed. Per-reviewer review serves to avoid mistakes of ones own making, to the extent of ones own knowledge of them.

A second pass is better because the first pass is, in so doing, the third and the fourth is too, albeit for diminishing returns, an improvement. What it does beyond ensuring everyone is incapable of finding flaws, is increase the risk that someone disagrees, and I, for one, really want to know who disagrees, and for what reason.

This presents a venue of opportunity in challenging a lot of people at once. One of them will likely take the time to check up on the change, and as certainty increases, likely revert the change, and maybe even take the time to explain why. Or they will at best all learn at once.

Everyone trusting themselves to translate, or review, should know that unless adding value, not only will someone will see it, it will be fixed. In the best interest of adding value that isn't great, there is no real way to say a string, or project is about 85% right, but 85% right sure is a lot better than not translated. Maybe in the future the string complexion could be rated too, but I digress.

Per review status is where to start, it is the only way to improve quality short of miracles, time nobody has, or perfect auto-translation. Adding to that confidence, even knowing who's reviews are overturned, per account doing it, and/or on the whole, adds some level of confidence.

Messing up the one "needs work", or the one "reviewed"-status, is a disaster, especially so if it isn't malicious, because right now someone needs to see and remember why it was applied right away. I don't think there is an e-mail for checking or unchecking "needs work".

What only first party review does is put strings in the category of "no second party review". That is a good place to start reviewing for anyone else, after doing the "no-review" strings.

Reviewing ones own strings, and having a system to do it, to remember what strings have been reviewed, is also better for the same reason knowing who to trust becomes possible. The premise of making mistakes and trying to avoid them is the same all the way down, so it should be no-one, and at best one person, and (only) by extension, many such persons, and automated systems.

Primarily using roles to forego this is a broken system if quality is the issue. Sometimes projects require review for everything, when the (only) practical alternative is superior.

[...]

PR #20370, opened 2026-06-30
Feature/per translator review

Harsha-032:
This PR implements per-translator review tracking, fulfilling the request to explicitly list which users have approved a translation, rather than relying on a binary anonymous "Approved" toggle. 
A new `UnitReview` model has been introduced to track individual approvals. When a user approves a string, a badge reading **Reviewed by: [username]** will be displayed above the source string. If the translation is modified, all existing reviews for that unit are automatically cleared so it can be re-reviewed. 
This also adds support for the `reviewed_by:me` and `reviewed_by:[username]` search queries.
### Changes Made
- Added `UnitReview` model and corresponding migration to track user reviews per translation unit.
- Updated `Unit.translate()` to save a `UnitReview` record when `STATE_APPROVED` is selected, and to clear reviews when the translation text changes.
- Updated the `translate.html` template to render the **Reviewed by: [username]** badge.
- Added `reviewed_by:` search keyword support and tests.
- Updated `docs/changes.rst` with the new feature.
### UI Changes
**Before:**
Previously, the review process was completely anonymous. If translation reviews were turned on, clicking "Approved" would simply set the state to Approved, but there was no visual indicator of *who* actually reviewed the translation on the UI, nor was there a way to track multiple reviewers.
**After:**
image
### Testing Performed
- Validated that `STATE_APPROVED` properly records the review.
- Confirmed multiple users approving the same string displays multiple reviewers.
- Confirmed editing the translation automatically clears all existing `UnitReview` records.
- Tested search queries `reviewed_by:me` to successfully filter approved strings.
- Ran Ruff format and check successfully.


Harsha-032:
@nijel can you review this for the issue #2699

.mypy_cache/3.14/cache.0.db
nijel, line 1: These files should not be here.

weblate/trans/models/review.py
```diff
@@ -0,0 +1,30 @@
...
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+        related_name="unit_reviews",
+    )
+    timestamp = models.DateTimeField(auto_now_add=True)
+
+    class Meta:
+        app_label = "trans"
+        unique_together = [("unit", "user")]
```
nijel, line 25: The ruff is failing here, but you claim that you've run it successfully.

weblate/trans/models/unit.py
```diff
@@ -935,6 +935,14 @@ def has_comment(self) -> bool:
     def has_suggestion(self) -> bool:
         return bool(self.suggestions)
 
+    def is_reviewed_by(self, user: User) -> bool:
+        if not user.is_authenticated:
+            return False
+        return self.reviews.filter(user=user).exists()
```
nijel, line 941: This doesn't seem to be used anywhere?

weblate/trans/models/unit.py
```diff
@@ -2326,6 +2334,18 @@ def translate(
...
+            # Clear reviews if the target translation actually changed
+            if self.target != self.old_unit["target"]:
+                self.reviews.all().delete()
+
+            # Manage per-translator review status
+            if user is not None and user.is_authenticated:
+                if self.state == STATE_APPROVED:
+                    self.reviews.get_or_create(user=user)
+                else:
+                    self.reviews.filter(user=user).delete()
```
nijel, line 2347: This will store at most one user in the reviews model. If there is no change, reviews are not updated (so if the string is already approved, no additional approvals will be added).

weblate/trans/tests/test_search.py
```diff
@@ -262,6 +262,13 @@ def test_review(self) -> None:
         # Review, partial date
         self.do_search({"q": "changed:>=2010-01-"}, None)
 
+        # Test reviewed_by
+        self.unit = self.get_unit()
+        self.unit.reviews.create(user=self.user)
+        self.do_search({"q": "reviewed_by:me"}, "Hello, world")
```
nijel, line 268: The tests fail here; have you run the test suite?

nijel (commented):
I think this does something different from what https://github.com/WeblateOrg/weblate/issues/2699 asked. My understanding there is that the motivation there is for each translator to have it's own review state.

While this pull request implements secondary tracking of who did the last state change to approved (effectively duplicating what can be figured out from the `Change` model already).

Harsha-032:
@nijel  You're completely right—my current implementation just mirrors the global state change rather than tracking independent per-user states.

To properly implement the per-translator review state as requested in #2699, I'd like to align on the architecture before rewriting the logic. My proposed approach would be:

Expand the new UnitReview model to store not just the user, but also the state (e.g., Approved, Needs Editing) that the specific user selected.

Decouple the UI's "Review state" radio buttons from the global Unit.state, so that submitting the form updates the current user's UnitReview record instead.

My main architectural question before proceeding: How should this interact with the global Unit.state? If User A marks their personal state as "Approved", but User B marks theirs as "Needs editing", how should we calculate the global Unit.state? Should we introduce a threshold setting (e.g., requires 2 approvals to globally approve), or should the global state just be removed entirely in favor of an aggregated query?

Let me know your thoughts on the best way to integrate this into the core engine!

Harsha-032:
Hey @nijel any suggestions?

luzpaz:
soft bump. This is a needed feature.

PR #20409, opened 2026-07-01
fix: make workflow customization UX more clear

### LINKED ISSUE #20399: Accessibility: Checkbox is almost invisible in https://hosted.weblate.org/settings/project/-/language/
### Affected page or workflow

In the settings/workflow per-language page, for example in https://hosted.weblate.org/settings/tor/-/de/, the checkbox to enable a specific language workflow is too light, almost invisible. This makes it confusing, it seems like the page is broken.

### Steps to reproduce

1. Open project/settings/workflow/language in a project you administer
2. Only the title:  Customize translation workflow for this language in this project  is clearly visible, without anything else. There is a very light colored checkbox, and the title is also a toggle to enable the feature, but it does not look like one, it looks only as a title.

### Expected behavior

- The checkbox should be darker so it can be seen easily
- Some message saying 'enable configuration to change the options' or similar should appear, instead of an empty page 

### Actual behavior

I wanted to upload a screenshot, but github does not allow me.

### Input method

_No response_

### Environment

_No response_

### Impact

Core workflow has a workaround

### Workaround

_No response_

### Additional context

_No response_

KarenKonou:
Uses bootstrap switch element instead of a checkbox to turn on workflow customization, making it visually clearer and also adds a message clarifying its usage.
image
image
fixes #20399

### DIFF
--- a/weblate/static/loader-bootstrap.js
+++ b/weblate/static/loader-bootstrap.js
@@ -2247,18 +2247,18 @@ onReady(() => {
 
   /* Workflow customization form */
   document.querySelectorAll("#id_workflow-enable").forEach((enableInput) => {
-    enableInput.addEventListener("click", () => {
-      if (enableInput.checked) {
-        document.getElementById("workflow-enable-target").style.visibility =
-          "visible";
-        document.getElementById("workflow-enable-target").style.opacity = 1;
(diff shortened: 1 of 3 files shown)


nijel:

PR #20421, opened 2026-07-02
refactor(vcs): clarify code hosting connection naming

nijel:
Use provider-neutral wording for shared connection pages while keeping GitHub-specific actions explicit.

* Use “Code hosting connection” instead of “VCS integration” / “VCS installation”. I believe that this is easier to understand for users.
* Avoid using GitHub-specific wording in areas that will be shared with other services in the future.


michael-smt:
Agree, nowadays "Code hosting" is easier to understand than "VCS".

orangesunny (approved):
missspelling fixed, new naming is good. Ready to go!

PR #20443, opened 2026-07-03
fix(memory): speed up fuzzy TM lookups

nijel:
Use a GiST KNN prefix index for the primary candidate slice and keep a full-source trigram fallback for saturated long strings so suffix-relevant matches are preserved.

Share the fuzzy selection policy between the machinery and API lookup paths.

### DIFF
--- a/weblate/memory/migrations/0006_memory_source_gist_prefix.py
+++ b/weblate/memory/migrations/0006_memory_source_gist_prefix.py
@@ -1,0 +1,29 @@
+# Copyright © Michal Čihař <michal@weblate.org>
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import annotations
+
+from django.contrib.postgres import indexes as postgres_indexes
+from django.contrib.postgres.operations import BtreeGistExtension
+from django.db import migrations, models
+from django.db.models.functions import Left
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("memory", "0005_memory_scopes"),
+    ]
+
+    operations = [
+        BtreeGistExtension(),
+        migrations.AddIndex(
+            model_name="memory",
+            index=postgres_indexes.GistIndex(
+                models.F("source_language"),
+                models.F("target_language"),
+                postgres_indexes.OpClass(Left("source", 2048), name="gist_trgm_ops"),
+                name="memory_source_gist_prefix",
+            ),
+        ),
+    ]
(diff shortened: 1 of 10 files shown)


weblate/memory/models.py
```diff
@@ -328,62 +332,164 @@ def minimum_similarity(self, text: str, threshold: int) -> float:
...
+
+    def get_scored_fuzzy_candidates(
+        self,
+        text: str,
+        scorer: Callable[[Memory], int],
+        *,
+        threshold: int = MACHINERY_DEFAULT_THRESHOLD,
+        limit: int = MEMORY_LOOKUP_LIMIT,
+    ) -> list[tuple[int, Memory]]:
+        candidates: list[Memory] = list(self.get_fuzzy_candidates(text, limit=limit))
```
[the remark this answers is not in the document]
nijel: I think this is an acceptable trade-off for performance.

weblate/trans/forms.py
```diff
@@ -1793,7 +1794,7 @@ class CostEstimateReportsForm(forms.Form):
     tm_threshold = forms.IntegerField(
         label=gettext_lazy("Translation memory threshold"),
         initial=MACHINERY_DEFAULT_THRESHOLD,
-        min_value=1,
+        min_value=MIN_COST_ESTIMATE_TM_THRESHOLD,
```
gersona, line 1797: It might be too edgy, but this could be documented as an incompatibility change to cover cases where a user programmatically posts a request to the `/costs` view with a `tm_threshold` value below 75

weblate/memory/migrations/0006_memory_source_gist_prefix.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("memory", "0005_memory_scopes"),
+    ]
+
+    operations = [
+        BtreeGistExtension(),
+        migrations.AddIndex(
```
gersona, line 20: More of a question than a review: since we now support only PostgreSQL, would this be a good place to use AddIndexConcurrently? We're specifically targeting better performance on large TMs, though I'm not sure whether it would add meaningful value or just extra complexity.

weblate/trans/forms.py
nijel: The lower value really doesn't make much sense, so I thought it's not worth mentioning (and it's not an API in the end).

weblate/memory/migrations/0006_memory_source_gist_prefix.py
nijel: On most deployments (Docker) the database migration is executed before the application runs, so concurrent index creation would only add its overhead (it does two passes on the data). Moreover, `CREATE INDEX CONCURRENTLY` has another possible failure situation of creating an invalid index (but maybe Django's `AddIndexConcurrently` is handling that).

gersona: Oh OK. That makes sense

PR #20499, opened 2026-07-07
test(utils): fail on unexpected error reports

nijel:
Make report_error calls fail under test settings so hidden handled-error paths surface during CI. Keep explicit reporting backend tests opted out.


nijel:
This turned out not to be practical needing to many overrides.

PR #20541, opened 2026-07-08
Updated POT file (xgettext) plugin: option to use keywords exclusively

### LINKED ISSUE #20528: Update POT file (xgettext) plugin: option to use keywords exclusively
### Describe the problem

I work on an old Java project where unfortunately we use some of the xgettext keywords for non translation related string creation. I solve this in my script by adding a "--keyword" without parameter. According to xgettext manual this deletes all default keywords. I add another "--keyword tr" for my translation method(tr). 

### Solution brainstorm

My idea would be e.g. a checkbox to state that only the keywords defined in the additional keyword field  should be used and not the default ones.

_No response_

### Screenshots

_No response_

### Additional context

from xgettext manual:
`--keyword[=keywordspec]´
    Additional keyword to be looked for (**without keywordspec means not to use default keywords**). If keywordspec is a C identifer id, xgettext looks for strings in the first argument of each call to the function or macro id. If keywordspec is of the form `id:argnum´, xgettext looks for strings in the argnumth argument of the call. If keywordspec is of the form `id:argnum1,argnum2´, xgettext looks for strings in the argnum1st argument and in the argnum2nd argument of the call, and treats them as singular/plural variants for a message with plural handling.
    The default keyword specifications, which are always looked for if not explicitly disabled, are gettext, dgettext:2, dcgettext:2, ngettext:1,2, dngettext:2,3, dcngettext:2,3, and gettext_noop.
    This option has an effect with most languages, namely C, C++, ObjectiveC, Shell, Python, Lisp, EmacsLisp, librep, Java, C#, awk, Tcl, Perl, PHP, GCC-source, Glade.

Kakarot30:
Closes #20528

Problem
The Update POT file (xgettext) addon supports an Additional keyword field that appends --keyword=<value> to the xgettext command. However, there is no way to disable xgettext's built-in default keywords (gettext, dgettext:2, ngettext:1,2, etc.).

This is a real problem for projects (e.g. legacy Java codebases) that reuse default xgettext keyword names for non-translation purposes. Users currently have to work around this outside of Weblate by manually passing --keyword without a value in a custom wrapper script.

Solution
Add a Use keywords exclusively checkbox to the xgettext addon settings form. When enabled alongside an Additional keyword, Weblate passes a bare --keyword to xgettext immediately before the named --keyword=<value>. Per the [xgettext manual](https://www.gnu.org/software/gettext/manual/html_node/xgettext-Invocation.html), --keyword without an argument disables all default keywords, so the resulting command only recognises the user-defined keyword.

### DIFF
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -33,7 +33,7 @@ jobs:
       PYTHONUNBUFFERED: 1
     services:
       database:
-        image: postgres:18.4-alpine@sha256:ecafd34249b5c248f3cb6ebe339584ee6448b36fad3f0a827ae5e8efbae6afda
+        image: postgres:18.4-alpine@sha256:9a8afca54e7861fd90fab5fdf4c42477a6b1cb7d293595148e674e0a3181de15
         env:
           POSTGRES_USER: postgres
           POSTGRES_PASSWORD: weblate
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -34,7 +34,7 @@ jobs:
       DJANGO_SETTINGS_MODULE: weblate.settings_test
     services:
       database:
-        image: postgres:18.4-alpine@sha256:ecafd34249b5c248f3cb6ebe339584ee6448b36fad3f0a827ae5e8efbae6afda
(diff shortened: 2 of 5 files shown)


nijel:
@codex review this

PR #20556, opened 2026-07-09
fix(addons): Use keywords exclusively — fix Meson preset bypass, add edge-case tests, and update docs

### LINKED ISSUE #20528: Update POT file (xgettext) plugin: option to use keywords exclusively

Kakarot30:
## Description
This PR implements the **Use keywords exclusively** checkbox for the `weblate.gettext.xgettext` and `weblate.gettext.meson` add-ons. When enabled, Weblate passes a bare `--keyword` to xgettext immediately before the user-defined `--keyword=<name>`. Per the xgettext manual, a bare `--keyword` disables all built-in default keywords so that only the user's custom-defined keywords are recognized.
### Key Problems Solved
1. **Meson Preset Keywords Bypass**:
   `MesonAddon.get_extra_xgettext_args()` previously prepended preset GLib keywords (e.g. `--keyword=_`, `--keyword=N_`) before the parametrized xgettext arguments. Because a bare `--keyword` only disables built-in defaults (and not custom arguments already specified earlier on the command line), Meson users selecting "Use keywords exclusively" still extracted GLib call patterns.
   * **Fix**: When `keyword_exclusive` is enabled, all `--keyword=...` entries from the selected preset are stripped out entirely, ensuring only the custom keyword is processed.
2. **Add-on Configuration Snippets**:
   The `keyword_exclusive` field was missing from the autogenerated parameter tables in `docs/snippets/addons-autogenerated.rst` for both add-ons. 
   * **Fix**: Regenerated/inserted the missing `keyword_exclusive` rows into the respective configuration tables.
3. **Pre-commit Cleanups**:
   * Removed backslash-escaped quotes (`\"`) in `tests.py` that caused syntax errors on Ruff checks.
   * Replaced non-ASCII ellipsis characters (`…` / U+2026) in python docstrings and comments with ASCII `...` to prevent `doccmd` UnicodeDecodeError crashes.
   * Suppressed Ruff `unused-class-method-argument` warnings in `tests.py` for test mock signature compliance using `# noqa: ARG002`.
---

Closes #20528


nijel:
I think that two commits should be removed from this:

* https://github.com/WeblateOrg/weblate/pull/20556/changes/9f3efaa41585fbc1e17440f31c2ac726cc6b7847 just changes override types, should not be needed.
* https://github.com/WeblateOrg/weblate/pull/20556/changes/326337aef44db152ac2e0a1bbd27043d6b4d1cd3 is changing existing documentation and there should be no issue in using Unicode there.

weblate/addons/tests.py
```diff
@@ -11,7 +11,7 @@
 import os
 import re
 import shutil
-import subprocess  # ruff: ignore[suspicious-subprocess-import]
+import subprocess  # noqa: S404
```
nijel: Please avoid changing existing suppressions. The human-readable identifiers are way easier to understand than the old error codes:

```suggestion
import subprocess  # ruff: ignore[suspicious-subprocess-import]
```

nijel:
@codex review this

nijel:
Do you know you can force push to a branch and update the pull request by that? There is no need to close the old pull request and open a new one just to change the content.

nijel:
@codex review this

Kakarot30:
Yes , Sorry for making so many prs i was just focusing on how to pass that pre commit check because everything else was looking good

nijel:
The pre-commit.ci is broken. They fetched a broken module and apparently keep it cached forever, so you can ignore this one.

weblate/addons/tests.py
```diff
@@ -12,7 +12,7 @@
...
```
nijel:
```suggestion
import subprocess  # ruff: ignore[suspicious-subprocess-import]
```

nijel:
@codex review

nijel:

PR #20717, opened 2026-07-16
Collect Errors From Automatic Suggestion Backends

### LINKED ISSUE #9638: Collect errors from automatic suggestion backends
### Describe the problem

Currently, automatic suggestion backends silent themselves for errors which are most likely caused by rate limiting:

https://github.com/WeblateOrg/weblate/blob/8b514cb50db9a4839c0e48350d4d5c5c3fe13250/weblate/machinery/base.py#L246-L259

This is nice for user viewed suggestions as it doesn't show an error message on each page.

But this makes such errors invisible to admins.

Collect such errors and display them in the management interface.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

shwetasinghh:
Fixes https://github.com/WeblateOrg/weblate/issues/9638

**Implementation summary**

**_weblate/machinery/models.py_**
Added MachineryError model with:
  - engine — CharField storing the service name (e.g. "DeepL"), indexed for filtering
  - project — nullable FK to trans.Project (CASCADE delete), covering project-scoped installations
  - timestamp — DateTimeField defaulting to now, indexed for efficient pruning queries
  - error — TextField storing the cause + exception type + message

**_weblate/machinery/base.py_**
Extended report_error with an optional exc: Exception | None = None parameter. When an exception is provided, `_save_machinery_error` is called, which creates a MachineryError row. The project is pulled from `self.settings.get("_project")` (already set for project-scoped services). DB write failures are caught and logged rather
  than propagating.
  Both existing callers were updated to pass exc=exc:
  - `supported_languages` property (language fetch failures)
  - `_download_pending_translations` (translation fetch failures, including rate limits)

**_weblate/machinery/tasks.py (new)_**
`cleanup_machinery_errors` task deletes records older than 30 days. Registered as a periodic task via `@app.on_after_finalize.connect` running daily at 01:30.

**_weblate/machinery/migrations/0001_initial.py (new)_**
Initial migration creating the MachineryError table. Depends on trans.0095_report (the current latest trans migration).

**_weblate/machinery/views.py_**
`ListMachineryView.get_context_data` now fetches the 50 most recent MachineryError rows (filtered by project when in the project view) and passes them as machinery_errors.

**_weblate/templates/machinery/list.html_**
Added an errors card at the top (only shown when there are errors) with a warning header and a table showing timestamp, service, project (global view only), and error message. Displayed for both /manage/machinery/ and per-project /machinery/{project}/ views.


weblate/machinery/tasks.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@app.task(trail=False)
+def cleanup_machinery_errors() -> None:
+    """Cleanup old machinery errors."""
+    # ruff: ignore[import-outside-top-level]
+    from weblate.machinery.models import MachineryError
+
+    MachineryError.objects.filter(
+        timestamp__lt=timezone.now() - timedelta(days=30)
```
nijel: I'm not sure if it is needed here, but typically the threshold for such cleanups is configurable.

weblate/machinery/base.py
```diff
@@ -364,12 +364,31 @@ def map_language_code(self, code: str) -> str:
...
+        if exc is not None:
+            self._save_machinery_error(exc, cause)
+
+    def _save_machinery_error(self, exc: Exception, cause: str) -> None:
+        # ruff: ignore[import-outside-top-level]
+        from weblate.machinery.models import MachineryError
+
+        try:
+            MachineryError.objects.create(
+                engine=self.name,
```
nijel: Using an identifier is better in this scope, as that does not change over time, while the visible name can change (see `weblate/machinery/microsoft.py` which got several renames over its lifetime, but the identifier is stable).

```suggestion
                engine=self.mtid,
```

weblate/machinery/models.py
```diff
@@ -17,6 +20,29 @@
...
+    """Stores errors from automatic suggestion backends for admin visibility."""
+
+    engine = models.CharField(max_length=100, db_index=True)
+    project = models.ForeignKey(
+        "trans.Project",
+        on_delete=models.deletion.CASCADE,
+        null=True,
+        blank=True,
+    )
+    timestamp = models.DateTimeField(default=timezone.now, db_index=True)
```
nijel: Wouldn't compound index on engine, project and timestamp work better than separate indexes? That is the typical query shape here.

nijel (commented):
Thanks for the pull request. Generally looks good, but I have some remarks. I will also trigger Codex review which is typically more nitpicky than I am.

nijel:
@codex review this pull request

weblate/machinery/base.py
```diff
@@ -364,12 +364,31 @@ def map_language_code(self, code: str) -> str:
...
+                project=self.settings.get("_project"),
+                error=f"{cause}: {type(exc).__name__}: {exc}",
```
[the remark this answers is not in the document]
nijel: I think this feedback is wrong. We want to scope the error to where the service is installed, not to where it is used.

weblate/machinery/tasks.py
shwetasinghh: Yeah, I think making it configurable is better 👍🏽

weblate/machinery/models.py
shwetasinghh: Good point, done. Replaced the separate db_index=True on engine and timestamp with a single compound index on (engine, project, timestamp).

shwetasinghh:
@codex review this pull request

shwetasinghh:
@codex review this pull request

shwetasinghh:
@codex review this pull request

shwetasinghh:
@nijel please review the PR

weblate/templates/machinery/list.html
```diff
@@ -21,6 +21,36 @@
...
+        <h4 class="card-title">{% translate "Recent automatic suggestion errors" %}</h4>
+      </div>
+      <table class="table table-striped">
+        <thead>
+          <tr>
+            <th>{% translate "Timestamp" %}</th>
+            <th>{% translate "Service" %}</th>
+            {% if not project %}
+              <th>{% translate "Project" %}</th>
+            {% endif %}
```
nijel: The list is either global or a single project.  So the project column has no value it either case.

weblate/machinery/views.py
```diff
@@ -172,6 +172,10 @@ def get_context_data(self, **kwargs):
         result = super().get_context_data(**kwargs)
         result["configured_services"] = self.configured_services
         result["available_services"] = self.available_services
+        errors_qs = MachineryError.objects.select_related("project")
+        if self.project:
+            errors_qs = errors_qs.filter(project=self.project)
```
nijel: I don't think we need to show project at all. The site-wide list should show errors from the site-wide installs only.

```suggestion
        errors_qs = MachineryError.objects.filter(project=self.project)
```

nijel (commented):
I wonder if the top of the list is the right place to show this. Actually it might be more appropriate to show it on the detail/settings view for each backend. This is similar approach we use for add-ons since https://github.com/WeblateOrg/weblate/pull/20765.

nijel:
There are some test failures:

```
FAILED weblate/machinery/tests.py::MachineryErrorTest::test_rate_limit_error_creates_db_record_and_sets_cache - TypeError: HTTPStatusError.__init__() missing 1 required keyword-only argument: 'request'
FAILED weblate/machinery/tests.py::MachineryErrorTest::test_report_error_redacts_url_query_parameters - TypeError: HTTPStatusError.__init__() missing 1 required keyword-only argument: 'request'
```

@codex review

shwetasinghh:
@codex review

shwetasinghh:
@codex review

shwetasinghh:
@codex review

shwetasinghh:
@codex review

shwetasinghh:
@nijel please review

nijel:

weblate/machinery/tasks.py
```diff
@@ -0,0 +1,34 @@
...
+    from weblate.machinery.models import MachineryError
+
+    MachineryError.objects.filter(
+        timestamp__lt=timezone.now()
+        - timedelta(days=settings.MACHINERY_ERROR_KEEP_DAYS)
+    ).delete()
+
+
+@app.on_after_finalize.connect
+def setup_periodic_tasks(sender, **kwargs) -> None:
```
nijel:
```suggestion
def setup_periodic_tasks(sender: Celery, **kwargs: object) -> None:
```

weblate/machinery/tasks.py
```diff
@@ -0,0 +1,34 @@
...
+
+from __future__ import annotations
+
+from datetime import timedelta
+
+from celery.schedules import crontab
+from django.conf import settings
+from django.utils import timezone
+
+from weblate.utils.celery import app
```
nijel, line 14:
```suggestion
from weblate.utils.celery import app
from celery import Celery

```

weblate/machinery/tests.py
```diff
@@ -10045,3 +10048,260 @@ def test_batch_translate(self) -> None:
...
+    """Tests for machinery error collection via the MachineryError model."""
+
+    def get_machine(self, settings=None):
+        machine = DummyTranslation(settings or {})
+        machine.delete_cache()
+        return machine
+
+    def test_report_error_without_exc_creates_no_db_record(self) -> None:
+        """report_error without exc does not create a MachineryError row."""
+        self.get_machine().report_error("Something went wrong")
```
```suggestion
        self.get_machine().report_error("Something went wrong", Exception())
```

nijel:
Okay, the tests do pass. Can you please look into that?

PR #20723, opened 2026-07-17
feat(legal): support custom document styles

nijel:
Keep deployment-specific styling out of Weblate core while giving operators a documented way to provide legal documents and CSS.


nijel:
Companion to https://github.com/WeblateOrg/wllegal/pull/1964

PR #20740, opened 2026-07-17
feat(diagnostics): refine community guidance alerts

nijel:
Replace coarse checks with actionable coverage and repository direction details while avoiding false positives for existing projects. Refresh coverage at logical operation boundaries so imports stay accurate without making bulk assignments expensive.

### DIFF
--- a/weblate/trans/models/translation.py
+++ b/weblate/trans/models/translation.py
@@ -2738,7 +2738,9 @@ class Translation(
             self.handle_store_change(request, user, previous_revision)
             if source_unit_deleted:
                 transaction.on_commit(
-                    lambda: update_alerts(component, {"MissingScreenshots", "UnusedScreenshot"})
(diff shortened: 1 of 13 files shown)


weblate/trans/models/translation.py
```diff
@@ -2730,6 +2736,10 @@ def delete_unit(self, request: AuthenticatedHttpRequest | None, unit: Unit) -> N
                     alert.save(update_fields=["details"])
 
             self.handle_store_change(request, user, previous_revision)
+            if source_unit_deleted:
+                transaction.on_commit(
+                    lambda: update_alerts(component, {"MissingScreenshots"})
```
nijel:
```suggestion
                    lambda: update_alerts(component, {"MissingScreenshots", "UnusedScreenshot"})
```

PR #20861, opened 2026-07-27
feat(addon): Base file only component Discovery 

### LINKED ISSUE #2084: Discovery addon: component creation based on the monolingual template or new base

gersona:
- fixes #2084

This Pr introduces some visual changes in the Discovery Addon form


nijel:
@codex review

nijel:
I'm not sure about the overall approach of this—determining the file mask based on the base name template brings some challenges (see review feedback from Copilot) and removes flexibility (users can no longer control what matches the component name such as allowing or not allowing `/` in it).

The alternative approach would be to allow omitting the language part of the regexp and configure a template for the translation file instead. This way the user will have more control, and the behavior would be more explicit.

gersona:
That makes sense, I missed that part that excludes other groups than `component` .
I'll rework the architecture as per your suggesition

PR #20888, opened 2026-07-27
fix(accounts): rotate API token when disabling password

amCap1712:
Regenerate a user's personal API token by default when password authentication is disabled. This invalidates the previous DRF token so it cannot continue authenticating after an administrator-triggered password reset. Expose a default-enabled administrator opt-out matching the existing password-change flow.


nijel:

PR #20932, opened 2026-07-30
fix(addons): adds missing context field in webhook payload

nlepage:
## Problem

The [webhook addon](https://docs.weblate.org/en/latest/admin/addons.html#webhook) is not adding the context field in the payload although [the docs](https://docs.weblate.org/en/latest/contributing/schemas.html#schema-messaging:~:text=context,-Translation) specifies this field should be there.

## Proposal

Add the context field in the webhook's payload.


nijel:
@codex review

nijel:

PR #20999, opened 2026-08-02
fix(machinery): preserve Anthropic base URL paths

### LINKED ISSUE #20992: `anthropic` machinery service discards the path component of a custom `base_url`
### Describe the issue

The `anthropic` machinery service builds its request URL like this (`weblate/machinery/anthropic.py`):

```python
end_point = "/v1/messages"
api_url = urljoin(self.settings.get("base_url") or "https://api.anthropic.com", self.end_point)
```

Because `end_point` starts with a leading slash, `urljoin` treats it as absolute-from-host and discards any path component `base_url` had — this is standard `urllib.parse.urljoin` / RFC 3986 behavior, not Weblate-specific:

```python
>>> from urllib.parse import urljoin
>>> urljoin("https://openrouter.ai/api",  "/v1/messages")
'https://openrouter.ai/v1/messages'
>>> urljoin("https://openrouter.ai/api/", "/v1/messages")
'https://openrouter.ai/v1/messages'
```

So a `base_url` of `https://openrouter.ai/api` (OpenRouter's documented base for its Anthropic-compatible endpoint, which actually lives at `/api/v1/messages`) silently becomes `https://openrouter.ai/v1/messages` — the wrong URL, which returns an HTML page rather than JSON. Weblate then reports a misleading error that names JSON parsing rather than the real cause:

```
Could not fetch translation: Expecting value: line 1 column 1 (char 0)
```

The `openai` machinery service in the same codebase already handles this correctly via a `join_api_url` helper that normalizes the base and joins a **relative** path:

```python
@staticmethod
def join_api_url(base_url: str, path: str) -> str:
    return urljoin(f"{base_url.rstrip('/')}/", path)
```

### I already tried

(Closest prior art: #18647, which fixed the case of an *empty* `base_url`, but didn't address a path-prefixed one.)

1. Configure the `anthropic` machinery service with a `base_url` that includes a path prefix, e.g. `https://openrouter.ai/api` (OpenRouter's documented base URL for its Anthropic-compatible endpoint).
2. Trigger a translation via that service.
3. Observe the request is actually sent to `https://openrouter.ai/v1/messages` (path prefix dropped), which does not serve the expected API, producing a JSON-parsing error that gives no hint the URL itself was wrong.

Verified independently outside of Weblate with a two-line Python REPL check confirming `urljoin`'s behavior on both URL shapes above.

### Expected behavior

`end_point` should not start with a leading slash, and the base/path should be joined the same way `openai.py` already does — normalizing the base with a trailing slash and joining a relative path — so any path component of a custom `base_url` (used for gateways like OpenRouter, LiteLLM, or Bedrock-compatible proxies) is preserved.

### Screenshots

N/A

### Exception traceback

```
Could not fetch translation: Expecting value: line 1 column 1 (char 0)
```

(Not a Python traceback — this is the user-facing error message Weblate surfaces, which is itself part of the problem: it points at JSON parsing rather than the actual wrong-URL cause.)

### How do you run Weblate?

Docker container

### Weblate version

### Weblate deploy checks

N/A — this is a pure URL-construction bug reproducible with the standard library alone.

### Additional context

Workaround: use the `openai` machinery service instead, pointed at an OpenRouter model id (e.g. `anthropic/claude-haiku-4.5`) — that path already joins correctly and works against OpenRouter today.

i4innovationnet:
## Summary

- preserve path prefixes in custom Anthropic base URLs
- share the existing `join_api_url` helper between the OpenAI and Anthropic services
- accept base URLs which already spell out the API version

Fixes #20992.

## Root cause

`end_point` started with `/`, so `urljoin` treated it as an absolute path and discarded any path component of a custom `base_url`. A gateway configured as `https://openrouter.ai/api` was called at `https://openrouter.ai/v1/messages`, which does not serve the API, and the failure surfaced as a JSON parsing error that gives no hint the URL was wrong.

## Approach

`end_point` becomes relative and is joined onto a normalized base, which is what `openai.py` already did through its `join_api_url` static method. That helper moves from `BaseOpenAITranslation` to `BaseLLMTranslation` so both services share one implementation. The OpenAI classes already inherit from `BaseLLMTranslation`, so no call sites change.

Correcting the join changes the result for a `base_url` which itself ends in the endpoint's version segment. `https://api.anthropic.com/v1` resolved correctly before, by accident of the bug, and a plain fix would start sending requests to `/v1/v1/messages`. Since `end_point` carries the version, a duplicate trailing segment is stripped from the base. Only the exact segment the endpoint supplies is stripped, so an unrelated path such as `https://gateway.example/v2` is left intact rather than silently rewritten.

| `base_url` | before | after |
| --- | --- | --- |
| unset | `https://api.anthropic.com/v1/messages` | unchanged |
| `https://api.anthropic.com` | `https://api.anthropic.com/v1/messages` | unchanged |
| `https://api.anthropic.com/v1` | `https://api.anthropic.com/v1/messages` | unchanged |
| `https://openrouter.ai/api` | `https://openrouter.ai/v1/messages` | `https://openrouter.ai/api/v1/messages` |
| `https://gateway.example/x/y/` | `https://gateway.example/v1/messages` | `https://gateway.example/x/y/v1/messages` |

`ollama.py` builds its URL the same way and has the same defect. It is left out of this change to keep the patch focused, and can follow separately.

## Validation

- `uv run pytest weblate/machinery/tests.py -q` — 861 passed, 50 skipped, 26 subtests passed
- LLM engine subset (Anthropic, OpenAI, Azure OpenAI, Mistral, Ollama) — 558 passed, 14 skipped
- both new tests fail against unpatched `main`, confirming they exercise the defect:

  ```
  FAILED AnthropicTranslationTest::test_base_url_path_is_preserved
  FAILED AnthropicTranslationTest::test_base_url_with_version
  FAILED AnthropicCustomModelTranslationTest::test_base_url_path_is_preserved
  FAILED AnthropicCustomModelTranslationTest::test_base_url_with_version
  ```

- mypy reports no new errors relative to `main`

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -40,6 +40,7 @@ Weblate 2026.8
 .. rubric:: Bug fixes
 
 * :ref:`Disabling password authentication <site-wide-user-management>` from site-wide user management now regenerates the user's personal API key by default.
+* :ref:`mt-anthropic` machine translation now preserves the path component of a custom base URL, so API gateways which serve the messages endpoint below a path prefix are reachable.
 * Category, project, and comment statistics now stay consistent after component topology and comment changes, and category metrics are collected correctly.
 * Mercurial repository filenames beginning with a dash are now handled safely.
 * URLs containing backslashes are now rejected as invalid.
--- a/weblate/machinery/anthropic.py
+++ b/weblate/machinery/anthropic.py
@@ -5,7 +5,6 @@
 from __future__ import annotations
 
 from typing import ClassVar
-from urllib.parse import urljoin
 
 from .base import MachineryRateLimitError
 from .forms import AnthropicMachineryForm
@@ -22,7 +21,7 @@ class AnthropicTranslation(BaseLLMTranslation):
 
     name = "Anthropic"
     trusted_error_hosts: ClassVar[set[str]] = {"api.anthropic.com"}
-    end_point = "/v1/messages"
+    end_point = "v1/messages"
     settings_form = AnthropicMachineryForm
     version_added = "5.16"
 
@@ -93,10 +92,15 @@ class AnthropicTranslation(BaseLLMTranslation):
         }
 
     def get_chat_url(self) -> str:
-        return urljoin(
-            self.settings.get("base_url") or "https://api.anthropic.com",
-            self.end_point,
-        )
+        base_url = (
+            self.settings.get("base_url") or "https://api.anthropic.com"
+        ).rstrip("/")
+        # The endpoint carries the API version, so strip it from the base URL
+        # to accept configurations which spell it out there as well.
+        version = self.end_point.partition("/")[0]
+        if base_url.endswith(f"/{version}"):
+            base_url = base_url[: -len(version) - 1]
+        return self.join_api_url(base_url, self.end_point)
 
     @staticmethod
     def parse_chat_response(response_data) -> str | None:
--- a/weblate/machinery/llm.py
+++ b/weblate/machinery/llm.py
@@ -13,6 +13,7 @@ from collections import Counter, defaultdict
 from itertools import chain
 from operator import itemgetter
 from typing import TYPE_CHECKING, Literal, NotRequired, TypedDict, TypeGuard
+from urllib.parse import urljoin
 
 from asgiref.sync import sync_to_async
 from django.utils.html import strip_tags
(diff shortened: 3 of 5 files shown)


nijel:
The ollama seems to suffer similar issue, so perhaps the implementation should be shared across llm backend that allow configuring base_url?

nijel:
The Ollama part is now addressed by https://github.com/WeblateOrg/weblate/pull/21643, scheduled for merge.

PR #21000, opened 2026-08-02
fix(machinery): default optional LLM prompt settings

### LINKED ISSUE #20993: LLM machinery services installed via the API crash with `KeyError: 'persona'` when optional fields are omitted
### Describe the issue

`persona` and `style` are optional fields on `BaseOpenAIMachineryForm` / `LLMBasicMachineryForm` (shared by the `openai`, `azure-openai`, `mistral`, `ollama`, and `anthropic` machinery services). Installing one of these services **through the API** without setting them passes form validation cleanly and returns success:

```
POST /api/projects/{project}/machinery_settings/
{"service": "mistral", "configuration": {"key": "...", "model": "mistral-small-latest"}}
→ 201 {"message": "Service installed: Mistral"}
```

But any subsequent `autotranslate` call using that engine silently does nothing:

```
POST /api/translations/{p}/{c}/{lang}/autotranslate/
{"mode":"translate","q":"state:<translated","auto_source":"mt","engines":["mistral"]}
→ 200 {"details": "Automatic translation completed, no strings were updated."}
```

The real failure only appears in the server log:

```
machinery[Mistral]: Could not fetch translations: KeyError: 'persona'
<project>/<component>/<lang>: failed automatic translation: KeyError: 'persona'
```

Root cause: prompt construction reads `self.settings[name]` as a direct, unconditional dict index (`weblate/machinery/llm.py`, `format_prompt_part`). The **UI** save path populates `persona`/`style` with an empty-string default even when left blank, because it persists `form.cleaned_data`. The **API** save path instead persists the raw submitted configuration dict as-is — so a config installed via the API without these keys never gets the default backfilled, and the first translate call that reads `settings["persona"]` raises `KeyError`. Reproducing via the Django admin UI will **not** trigger this, since the UI path always backfills the defaults; it requires installing the service through the API.

### I already tried

1. Install an LLM-backed machinery service via the API (`POST /api/projects/{p}/machinery_settings/`) with only the required fields — omit `persona` and `style`.
2. Observe: `201`, "Service installed" message.
3. Run `autotranslate` against a translation using that engine.
4. Observe: `200 {"details": "Automatic translation completed, no strings were updated."}` — even though matching source strings exist.
5. Check the server log: `KeyError: 'persona'` at the point the engine attempts to build its translation prompt.

**Workaround confirmed:** explicitly pass `"persona": ""` and `"style": ""` at install time. With those keys present, the same autotranslate call succeeds normally.

### Expected behavior

`format_prompt_part` should use `self.settings.get(name, "")` instead of a direct index, so a config saved via any path (UI or API) behaves consistently regardless of whether the optional field was ever explicitly set.

### Screenshots

N/A

### Exception traceback

```
machinery[Mistral]: Could not fetch translations: KeyError: 'persona'
<project>/<component>/<lang>: failed automatic translation: KeyError: 'persona'
```

(Server-log level, not an HTTP-level traceback — the API call itself returns `200`, which is the core of the problem: the failure is invisible to the caller. I've filed a separate, more general report about machinery failures being reported as API success.)

### How do you run Weblate?

Docker container

### Weblate version

### Weblate deploy checks

### Additional context

Reproduced directly against `mistral`. `openai`, `azure-openai`, `ollama`, and `anthropic` share the same form class and the same unconditional `self.settings[name]` read in `format_prompt_part`, so they are very likely affected the same way, though only `mistral` was runtime-tested end-to-end.

i4innovationnet:
## Summary

- read the optional `persona` and `style` prompt settings through a default
- cover the absent-field configuration for the Mistral and Ollama services

Fixes #20993.

## Root cause

`format_prompt_part` indexed both settings directly, so any configuration lacking them raised `KeyError` on the first translation attempt.

The web interface always stores both, because it persists `form.cleaned_data`. The REST API instead persists the submitted configuration as sent (`weblate/machinery/models.py`), so a service installed through the API without these optional fields validates, saves, and reports success, then fails on every use. The API call itself still returns `200`, so the failure is visible only in the server log.

## Approach

Both settings are read with a default, which is what the other optional prompt setting in `llm.py` already did. After the change there are no remaining direct-index reads of `self.settings` in that module.

This is scoped to the LLM engines. The underlying asymmetry — the API persisting the raw submitted dictionary rather than the cleaned form data — also leaves other services able to store a configuration missing required keys, and that is a broader decision than this fix. Changing it centrally breaks tests which deliberately pin raw persistence (`ProjectAPITest::test_install_machinery`, `CommandTest::test_install_valid_form`), so it is deliberately not attempted here.

## Test approach

Rather than assert a single translation call, the regression coverage subclasses the existing Mistral and Ollama test cases with a configuration that omits the optional fields, so the entire inherited suite runs against that shape.

Ollama is included because its existing fixture sets a non-empty `persona`, making it the one affected family where the absent-field path was otherwise untested.

## Validation

- `uv run pytest weblate/machinery/tests.py -q` — 964 passed, 53 skipped, 26 subtests passed
- `uv run pytest weblate/api/tests.py -q` — 574 passed, 14 subtests passed
- reverting `format_prompt_part` to the direct index fails 70 of the new subclass tests with the reported `KeyError: 'persona'`


nijel:
@codex review

nijel:

PR #21001, opened 2026-08-02
fix(api): document autotranslate request schema

### LINKED ISSUE #20991: OpenAPI schema for `autotranslate` documents the wrong request body
### Describe the issue

`POST /api/translations/{project}/{component}/{language}/autotranslate/` is documented in the published OpenAPI schema (`/api/schema/`, generated via drf-spectacular) as accepting the `Translation` serializer's fields — things like `filename`, `revision`, `is_template`, `language_code`. None of those fields are actually read by the endpoint.

The view instead validates the request body with a Django form, not a DRF serializer:

```python
# weblate/api/views.py
autoform = AutoForm(translation.component, request.user, request.data)
```

The real accepted fields are `AutoForm`'s: `q` (required), `mode`, `auto_source`, `component`, `engines`, `threshold`. Since `AutoForm` is a plain Django form rather than a serializer, drf-spectacular can't introspect it and appears to fall back to describing the URL's resource type instead — producing a schema that actively misleads anyone generating a client from it.

### I already tried

1. Fetch the schema: `GET /api/schema/?format=json`.
2. Find the `autotranslate` operation for the translations resource.
3. Observe its request body resolves to `{"$ref": "#/components/schemas/Translation"}`.
4. Compare against `weblate/api/views.py`, where the same action validates `request.data` against `AutoForm` (`weblate/trans/forms.py`), whose real fields are `q`, `mode`, `auto_source`, `component`, `engines`, `threshold`.
5. None of `Translation`'s ~33 documented properties overlap with `AutoForm`'s actual fields.

### Expected behavior

The schema should describe `AutoForm`'s actual fields, e.g. via `@extend_schema(request=...)` on the view with an explicit serializer or inline schema describing `q`, `mode`, `auto_source`, `component`, `engines`, and `threshold`.

### Screenshots

N/A — raw schema excerpt from a fresh `GET /api/schema/?format=json` instead:

```json
"/api/translations/{component__project__slug}/{component__slug}/{language__code}/autotranslate/": {
  "post": {
    "operationId": "api_translations_autotranslate_create",
    "description": "Trigger automatic translation.",
    "requestBody": {
      "content": {
        "application/json": {
          "schema": { "$ref": "#/components/schemas/Translation" }
        }
      },
      "required": true
    }
  }
}
```

### Exception traceback

N/A — this is a documentation/schema-generation defect, not a runtime error.

### How do you run Weblate?

Docker container

### Weblate version

### Weblate deploy checks

N/A — schema inspection only, no deployment-specific behavior involved.

### Additional context

Also present on the publicly hosted instance's schema at the time of writing (`hosted.weblate.org/api/schema/`), so this isn't specific to a self-hosted deployment quirk.

i4innovationnet:
## Summary

- document the request fields the autotranslate action actually accepts
- regenerate `docs/specs/openapi.yaml`
- align the prose documentation in `docs/api.rst` with the same shape

Fixes #20991.

## Root cause

The action validates its request body with `AutoForm`, a Django form rather than a serializer, which the schema generator cannot introspect. With no explicit request annotation it fell back to the resource serializer, so the published schema described the `Translation` object — `filename`, `revision`, `is_template`, `language_code` and around thirty other properties, none of which the endpoint reads, and none of the fields it does read. Any client generated from that schema was wrong.

## Approach

The action is annotated with an inline request serializer describing `AutoForm`'s fields, following the existing convention in `views.py` (`USER_GROUP_REQUEST_SERIALIZER`, `NEW_UNIT_REQUEST_SERIALIZER`).

The required set is `mode`, `q`, `auto_source` and `threshold`. Only `component` and `engines` carry `required=False` on the form, and Django ignores `initial` on a bound form, so the remaining four are required at runtime — confirmed by posting bodies with each field omitted in turn.

Two runtime constraints are not expressible in a static schema and are documented in `help_text` instead:

- `mode: approved` additionally requires reviews to be enabled and the permission to approve. Publishing the maximal enum matches the existing treatment of `NewUnitStateEnum`, which advertises the approved state unconditionally.
- `component` is a `ChoiceField` over component IDs on instances with few candidate components, and degrades to a free-form `project/component` path above that threshold. It is documented as a string, which makes no claim that every string validates.

## Validation

- `make -C docs update-openapi` run and the regenerated schema committed, so the `Verify OpenAPI spec is up to date` step passes. Without the regenerated file that step exits 1; the same command produces a byte-identical file on unpatched `main`, which confirms the local invocation matches CI.
- `spectacular --fail-on-warn --validate` exits 0 with empty stderr — no new schema warnings, no enum name collisions
- `uv run pytest weblate/api/tests.py::OpenAPITest -q` — 23 passed, 10 subtests passed
- the new test fails against unpatched `main` with `{'$ref': '.../Translation'} != {'$ref': '.../AutoTranslateRequest'}`

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -2459,12 +2459,12 @@ Translations
     :type component: string
     :param language: Translation language code
     :type language: string
-    :<json string mode: Automatic translation mode
+    :<json string mode: Automatic translation mode - ``suggest``, ``translate``, ``fuzzy`` or ``approved``
     :<json string q: Automatic translation search string, see :ref:`search-strings`.
     :<json string auto_source: Automatic translation source - ``mt`` or ``others``
-    :<json string component: Turn on contribution to shared translation memory for the project to get access to additional components.
+    :<json string component: Source component, given as its numeric ID, or as a ``project/component`` path on instances offering many components. Leave blank to use all components sharing translation memory. Turn on contribution to shared translation memory for the project to get access to additional components.
     :<json array engines: Machine translation engines
-    :<json string threshold: Score threshold
+    :<json int threshold: Score threshold (1-100)
 
 .. http:get:: /api/translations/(string:project)/(string:component)/(string:language)/file/
 
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -43,6 +43,7 @@ Weblate 2026.8
 * Category, project, and comment statistics now stay consistent after component topology and comment changes, and category metrics are collected correctly.
 * Mercurial repository filenames beginning with a dash are now handled safely.
 * URLs containing backslashes are now rejected as invalid.
+* The OpenAPI schema now documents the request body actually accepted by the automatic translation endpoint instead of the translation resource. See :http:post:`/api/translations/(string:project)/(string:component)/(string:language)/autotranslate/`.
 * Protected outbound HTTP and Git connections now bind to validated public addresses; protected Git operations fail closed when clients cannot enforce that binding.
 * Permanent same-host Git HTTP redirects are now validated and stored as canonical component repository URLs.
 * Self-service REST API e-mail changes are now restricted to verified addresses.
--- a/docs/specs/openapi.yaml
+++ b/docs/specs/openapi.yaml
@@ -25523,7 +25523,7 @@ paths:
         content:
           application/json:
             schema:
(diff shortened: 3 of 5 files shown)


nijel:
Superseeded by https://github.com/WeblateOrg/weblate/pull/21418

PR #21003, opened 2026-08-02
checks: make Kashida check less aggressive

### LINKED ISSUE #8228: Kashida/Tatweel check still too aggressive
### Describe the issue

This follows up on #6877 in which exceptions were added for certain Arabic prepositions. While it was pointed out that there are limited number of these in Arabic, most languages which use Arabic script characters are not Arabic.

Uses cases of tatweel/kashida which should be permitted:

* Arabic-based scripts have a number of combining characters and diacritics for which the tatweel is used as a "holder" in examples and illustrations, as in ــ٘ـ to highlight the form of the ghunna marker (used in Pakistani languages) without attaching it to surrounding characters. This may come up in translating documentation or applications which have language-specific considerations. Tatweel/kashida + any combining mark should be permitted, as well as tatweel/kashida + combining mark with tatweel/kashida on either side (easier to read with some buffer space around it).

* Any sequence ending in a single tatweel/kashida followed by a spacing character, punctuation, or a combining mark should be permitted. An example I came across is translating an app which has a "Mo Tu We Th Fr Sa Su" header. Weekday abbreviations are not typically used in Punjabi, but there is no space in this context to use full words. So a workaround is مـ or اتـ for example to abbreviate the weekdays in a way that is more legible than just putting the isolated form of each letter. Any number of characters should be permitted before the kashida/tatweel for this, since in many alphabets a combination of multiple characters is required to represent a single sound (for example دھ or ای).

Really what the original kashida/tatweel check was likely trying to prevent is strings like this:

صـــفــــحـــــے

This is fair enough, but the check should be limited to [actual letter] + tatweel/kashida(s) + [actual letter] so that errors are not thrown for other contexts which have more valid use cases.

### I already tried

_No response_

### Expected behavior

_No response_

### Screenshots

_No response_

### Exception traceback

_No response_

### How do you run Weblate?

weblate.org service

### Weblate versions

_No response_

### Weblate deploy checks

_No response_

### Additional context

_No response_

VKspyder2003:
Fixes #8228

## Summary

This change makes `KashidaCheck` less aggressive by warning only when Kashida (Tatweel) is used for decorative stretching within Arabic-script words.

Valid uses of Kashida described in the issue are now allowed, including:

- Abbreviations ending with Kashida (for example, `مـ` and `اتـ`)
- Kashida used as a holder for combining marks (for example, `ــ٘ـ`)
- Standalone preposition placeholders such as `بـ:`

Decorative stretching within words continues to be reported.

## Changes

- Updated the Kashida matching logic in `weblate/checks/chars.py` to use Unicode-aware matching for Arabic-script text.
- Added regression tests in `weblate/checks/tests/test_chars_checks.py` covering the valid and invalid cases described in the issue.

## Testing

- Added regression tests covering the reported scenarios.

### DIFF
--- a/weblate/checks/chars.py
+++ b/weblate/checks/chars.py
@@ -606,12 +606,13 @@ class KashidaCheck(TargetCheck):
     description = gettext_lazy("The decorative kashida letters should not be used.")
 
     kashida_regex = (
-        # Allow kashida after certain letters
-        "(?<![\u0628\u0643\u0644])"
+        # Allow kashida after certain letters, only if sandwiched between actual letters
+        r"(?<=\p{sc=Arab})(?<![\u0640\ufcf2\ufcf3\ufcf4\ufe71\ufe77\ufe79\ufe7b\ufe7d\ufe7f])"
         # List of kashida letters to check
-        "[\u0640\ufcf2\ufcf3\ufcf4\ufe71\ufe77\ufe79\ufe7b\ufe7d\ufe7f]"
+        r"[\u0640\ufcf2\ufcf3\ufcf4\ufe71\ufe77\ufe79\ufe7b\ufe7d\ufe7f]+"
+        r"(?=\p{sc=Arab})(?![\u0640\ufcf2\ufcf3\ufcf4\ufe71\ufe77\ufe79\ufe7b\ufe7d\ufe7f])"
     )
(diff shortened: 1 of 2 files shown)


nijel:
@codex review

nijel:

PR #21018, opened 2026-08-03
fix(addons): redact configuration in change history

amCap1712:
Store versioned, fail-closed snapshots for add-on history instead of copying raw configuration. Add-on forms explicitly allowlist public fields, while all other keys retain a null value and are identified in redacted_fields.

Compare updates with the persisted configuration so changed_fields identifies secret rotation without retaining either credential value. Browser history now names changed settings and marks redacted ones, while the changes API accepts only the validated public schema.

Hide malformed and legacy add-on details at serialization time and scrub existing add-on history in a data migration. Document the public-configuration contract and threat-model condition, and cover webhook, Slack-compatible, Fedora Messaging, rotation, uninstall, API, display, and migration behavior with regression tests.


docs/changes.rst
```diff
@@ -71,6 +71,8 @@ Weblate 2026.8
 * Self-service REST API e-mail changes are now restricted to verified addresses (:ghsa:`x84p-6892-473c`).
 * REST API authorization now consistently protects internal accounts, restricted components, add-on configuration, component sharing, repository links, and review states.
 * :ref:`Project backup imports <projectbackup>` now validate restored data before creating project state, skip repository-linked components the importer cannot access, and no longer load archive-supplied Mercurial configuration or shared-repository indirection (:ghsa:`f66g-8pcg-jm8r`, :ghsa:`327h-qqgm-qv55`).
+* Add-on change history now retains only explicitly public configuration values and marks changed credential fields as redacted.
+* :ref:`Project backup imports <projectbackup>` now validate restored data before creating project state, skip repository-linked components the importer cannot access, and no longer load archive-supplied Mercurial configuration or shared-repository indirection.
```
nijel: Please move this to 2026.8.1.

weblate/trans/migrations/0098_scrub_addon_change_details.py
```diff
@@ -0,0 +1,27 @@
...
+from __future__ import annotations
+
+from django.db import migrations
+
+
+def scrub_addon_change_details(apps, schema_editor) -> None:
+    Change = apps.get_model("trans", "Change")
+    Change.objects.using(schema_editor.connection.alias).filter(
+        action__in=(60, 61, 62)
+    ).update(details={})
```
nijel: Can we make it a bit more fine-grained, or would that make the migration too complex?

nijel (commented):
Looks good. I will also trigger codex review.

nijel:
@codex review

nijel:
`docs/_ext/autodoc_signature_filter.py` will need extensions to avoid documentation build errors: https://github.com/WeblateOrg/weblate/actions/runs/31096950217/job/92601138140?pr=21018#step:7:61

@codex review

weblate/addons/base.py
```diff
@@ -79,6 +89,56 @@ class CompatDict(TypedDict, total=False):
...
+        return False
+    configuration = details.get("configuration")
+    changed_fields = details.get("changed_fields")
+    redacted_fields = details.get("redacted_fields")
+    if (
+        type(configuration) is not dict
+        or type(changed_fields) is not list
+        or type(redacted_fields) is not list
+        or not all(type(key) is str for key in changed_fields)
+        or not all(type(key) is str for key in redacted_fields)
```
nijel:
```suggestion
    if not isinstance(details, dict):
        return False
    if details.get("schema") != ADDON_CHANGE_DETAILS_SCHEMA:
        return False
    configuration = details.get("configuration")
    changed_fields = details.get("changed_fields")
    redacted_fields = details.get("redacted_fields")
    if (
        not isinstance(configuration, dict)
        or not isinstance(changed_fields, list)
        or not isinstance(redacted_fields, list)
        or not all(isinstance(key, str) for key in changed_fields)
        or not all(isinstance(key, str) for key in redacted_fields)
```

weblate/addons/tests.py
```diff
@@ -7288,6 +7333,106 @@ def test_existing_amqp_url_timing_params_are_migrated_to_fields(self) -> None:
...
+            action=ActionEvents.ADDON_CHANGE,
+            target="example.unknown.addon",
+            details={"path": "private-path"},
+        )
+        schema_editor = SimpleNamespace(connection=SimpleNamespace(alias="default"))
+
+        migration.scrub_addon_change_details(apps, schema_editor)
+
+        addon_change.refresh_from_db()
+        self.assertEqual(addon_change.details, {"path": "private-path"})
```
nijel:
```suggestion
        change_object = Change.objects.create(
            action=ActionEvents.ADDON_CHANGE,
            target="example.unknown.addon",
            details={"path": "private-path"},
        )
        schema_editor = SimpleNamespace(connection=SimpleNamespace(alias="default"))

        migration.scrub_addon_change_details(apps, schema_editor)

        change_object.refresh_from_db()
        self.assertEqual(change_object.details, {"path": "private-path"})
```

PR #21020, opened 2026-08-03
Return 202 Accepted for asynchronous project deletion API

### LINKED ISSUE #20998: `DELETE /api/projects/{slug}/` gives no way to know when removal actually completed
### Describe the problem

`DELETE /api/projects/{slug}/` returns `204 No Content` immediately, but the actual project removal runs asynchronously in a background task. A client has no way to distinguish "fully deleted" from "deletion merely queued" from the response alone — the only way to find out is to poll `GET /projects/{slug}/` until it 404s, or dig through the change/audit log after the fact.

This is workable but easy to get wrong: automation that treats `204` as "safe to proceed" (e.g. before recreating a resource, or reporting the operation as complete to a user) can act on stale assumptions about project state while the real deletion is still in flight.

### Solution brainstorm

Return `202 Accepted` instead of a bare `204`, with a task/status URL in the response body — Weblate already has exactly this machinery for other async operations via `/api/tasks/{id}/`. A client can then poll the task URL to know precisely when removal has actually completed, rather than inferring it indirectly.

Polling `GET /projects/{slug}/` until it returns `404` works today and doesn't require an API change, but it's an indirect signal (absence of a resource, not confirmation a specific task completed) and isn't discoverable from the `DELETE` response itself — a caller has to already know to do this.

### Screenshots

N/A

### Additional context

Noticed while testing project lifecycle automation against the REST API — not a data-loss report, just an API-ergonomics gap around an existing async operation that already has a natural fix pattern available in the codebase (`/api/tasks/{id}/`).

VKspyder2003:
Fixes #20998

## Summary

This PR updates the project deletion REST API endpoint (`DELETE /api/projects/{slug}/`) to return **202 Accepted** instead of **204 No Content**, matching the behavior of other asynchronous API endpoints such as project backups and report generation.

The response now includes the task status URL so clients can monitor the deletion task.

### API (`weblate/api/views.py`)

- Return **202 Accepted** after scheduling project deletion.
- Store Celery task metadata using `store_task_metadata()`.
- Include:
  - localized `detail` message
  - `task_url` pointing to the task status endpoint.
- Document the asynchronous response using `@extend_schema`.

### Tests (`weblate/api/tests.py`)

Updated the project deletion API test to:

- expect **HTTP 202 Accepted**
- verify the response contains `detail` and `task_url`
- verify the project is still removed after the asynchronous task completes

### Documentation

Added a changelog entry describing the API response change from **204 No Content** to **202 Accepted**.


nijel:
Please adjust the OpenAPI schema via `@extend_schema` so that it matches new response and update the generate schema.

VKspyder2003:
Hi @nijel, updated as requested. OpenAPI schema now documents the 202 Accepted project deletion response via @extend_schema, and the generated docs/specs/openapi.yaml has been updated accordingly.

nijel:
You probably edited the OpenAPI YAML manually; CI still complains with diff (https://github.com/WeblateOrg/weblate/actions/runs/31310283785/job/93236954296?pr=21020). Either run `make -C docs update-openapi` or allow writes on the pull request so that the CI can update the YAML.

VKspyder2003:
Alright! I have enabled the edits on my PR.

nijel:
@codex review

weblate/api/views.py
```diff
@@ -2288,8 +2300,15 @@ def destroy(self, request: Request, *args, **kwargs):
         if not request.user.has_perm("project.edit", instance):
             self.permission_denied(request, "Can not delete project")
         instance.acting_user = request.user
-        project_removal.delay(instance.pk, request.user.pk)
-        return Response(status=HTTP_204_NO_CONTENT)
+        task = project_removal.delay(instance.pk, request.user.pk)
+        store_task_metadata(task.id, user_id=request.user.id)
+        return Response(
+            {
+                "detail": gettext("Project deletion scheduled."),
```
nijel:
```suggestion
                "detail": "Project deletion scheduled.",
```

PR #21040, opened 2026-08-05
fix(api): return 404 when unit is deleted mid-request

### LINKED ISSUE #20997: `PATCH /api/units/{id}/` can return an unhandled 500 instead of 404 if the unit is deleted mid-request
### Describe the issue

`PATCH /api/units/{id}/` normally 404s cleanly if the id no longer exists at the start of the request — `UnitViewSet` doesn't override `get_object()`, and DRF's default resolves a missing row to a proper `Http404`. That part already works correctly.

However, there is a narrower window where the same operation raises an **unhandled 500** instead: after DRF resolves the unit via `get_object()`, `Unit.translate()` re-fetches the same row a second time via an uncaught `Unit.objects.select_for_update().get(pk=self.pk)`. If the row is deleted **between** those two lookups — e.g. by a concurrent component reload that recreates unit rows with new ids — that second lookup raises a raw `Unit.DoesNotExist`, which propagates as an unhandled `500` with a full traceback instead of a clean `404` or `409`.

This is a narrow time-of-check-to-time-of-use gap, not a claim that unit ids are broadly unstable — reconciliation actually reuses existing rows when a unit's content hash is unchanged, and only recreates rows when the hash itself changes.

### I already tried

(One near-miss found: #10192 has the same `Unit matching query does not exist` message, but for a different endpoint — screenshot association returning 400 — not this code path.)

1. Create a unit and note its `id`.
2. Concurrently:
   - Fire a sustained stream of `PATCH /api/units/{id}/` requests against it, and
   - Trigger something that deletes/recreates that row during the same window — e.g. a file re-upload that changes the unit's content hash, or a component reload.
3. Sustaining concurrent requests across the ~100–400ms window between the delete-triggering action and the reconciliation actually landing produces a mix of results: mostly clean `200`s (request completed before the row moved) and `404`s (request started after the row was already gone), but reliably includes a nonzero number of raw `500`s with `Unit.DoesNotExist` in the traceback, landing specifically inside `Unit.translate()`'s second lookup.

### Expected behavior

The second lookup inside `Unit.translate()` should catch `Unit.DoesNotExist` and translate it into a proper API error response (`404` or a `409 Conflict`, since the resource existed at the start of the request but was concurrently removed) rather than letting the exception propagate as an unhandled `500`.

### Screenshots

N/A

### Exception traceback

```
File "/app/venv/lib/python3.14/site-packages/rest_framework/views.py", line 512, in dispatch
  response = handler(request, *args, **kwargs)
File "/app/venv/lib/python3.14/site-packages/drf_spectacular/drainage.py", line 207, in wrapped_method
  return method(self, request, *args, **kwargs)
File "/app/venv/lib/python3.14/site-packages/rest_framework/mixins.py", line 82, in partial_update
  return self.update(request, *args, **kwargs)
File "/app/venv/lib/python3.14/site-packages/drf_spectacular/drainage.py", line 207, in wrapped_method
  return method(self, request, *args, **kwargs)
File "/app/venv/lib/python3.14/site-packages/rest_framework/mixins.py", line 68, in update
  self.perform_update(serializer)
File "/opt/python/cpython-3.14.6-linux-x86_64-gnu/lib/python3.14/contextlib.py", line 85, in inner
  return func(*args, **kwds)
File "/app/venv/lib/python3.14/site-packages/weblate/api/views.py", line 3459, in perform_update
  unit.translate(user, new_target, new_state)
File "/opt/python/cpython-3.14.6-linux-x86_64-gnu/lib/python3.14/contextlib.py", line 85, in inner
  return func(*args, **kwds)
File "/app/venv/lib/python3.14/site-packages/weblate/trans/models/unit.py", line 2275, in translate
  old_unit = Unit.objects.select_for_update().get(pk=self.pk)
File "/app/venv/lib/python3.14/site-packages/django/db/models/query.py", line 639, in get
  raise self.model.DoesNotExist(
      "%s matching query does not exist." % self.model._meta.object_name
  )
weblate.trans.models.unit.Unit.DoesNotExist: Unit matching query does not exist.
```

This is one of 191 identical tracebacks captured in a single stress run (191 client-side `500` responses matched exactly 191 server-side `Unit.DoesNotExist` tracebacks in the log for that window — no ambiguity between them). The setup: a unit left to stabilize for 20s (well past a separate component-creation-race issue I filed as #20995), then a sustained stream of concurrent `PATCH /api/units/{id}/` requests fired roughly every 15ms for ~700ms per iteration, while concurrently forcing the row to be deleted and recreated via a file re-upload that temporarily removes the unit's key from the source file.

### How do you run Weblate?

Docker container

### Weblate version

### Weblate deploy checks

### Additional context

This needs a dedicated concurrency test to reproduce reliably (it's inherently timing-dependent) — happy to share the stress-test script used to trigger it if useful for a regression test.

alafilearnstocode:
`PATCH /api/units/{id}/` returned an unhandled 500 when the unit row was deleted between DRF's `get_object()` and the `select_for_update()` re-fetch in `Unit.translate()`. The raw `Unit.DoesNotExist` is now turned into a 404, consistent with the response when the unit is already gone at the start of the request.

Fixes #20997

## Test

`test_translate_unit_deleted_mid_request` drives the race deterministically by deleting the row from `invalidate_checks_cache()`, which runs at the top of `Unit.translate()` just before the re-fetch — no threads or sleeps.

Without the fix it reproduces the reported traceback exactly:

```
weblate/api/views.py:3848: in perform_update
    unit.translate(user, new_target, new_state)
weblate/trans/models/unit.py:2385: in translate
    old_unit = Unit.objects.select_for_update().get(pk=self.pk)
E   weblate.trans.models.unit.Unit.DoesNotExist: Unit matching query does not exist.
ERROR django.request: Server Error (500): /api/units/5/
```

With the fix, `weblate/api/tests.py::UnitAPITest` passes (24 tests), and `ruff check` / `ruff format --check` are clean on both changed files.

## Open question

`Unit.translate()` re-fetches the row under `select_for_update()` after the caller has already resolved it, so the same TOCTOU window exists for every caller, not just the API:

- `weblate/api/views.py:3849` (this PR)
- `weblate/trans/views/edit.py:862` and `:976`
- `weblate/trans/views/search.py:140`
- `weblate/trans/views/js.py:144`

I fixed it at the API boundary to keep the diff narrow and leave the model's exception semantics unchanged. The alternative is catching `Unit.DoesNotExist` inside `translate()` itself, which would cover the web editor too but means a model method raising `Http404`.

Also happy to move it if you prefer the model layer fix.

### DIFF
--- a/weblate/api/tests.py
+++ b/weblate/api/tests.py
@@ -11900,6 +11900,28 @@ class UnitAPITest(APIBaseTest):
         # The auto fixer adds the trailing newline
         self.assertEqual(unit.target, "Test translation\n")
 
+    def test_translate_unit_deleted_mid_request(self) -> None:
+        """Unit removed between get_object() and the locking re-fetch."""
+        unit = Unit.objects.get(
+            translation__language_code="cs", source="Hello, world!\n"
+        )
+
+        def delete_unit(self) -> None:
+            Unit.objects.filter(pk=unit.pk).delete()
+
+        # invalidate_checks_cache() runs at the top of Unit.translate(),
+        # before the select_for_update() re-fetch
+        with patch.object(
+            Unit, "invalidate_checks_cache", autospec=True, side_effect=delete_unit
+        ):
+            self.do_request(
+                "api:unit-detail",
+                kwargs={"pk": unit.pk},
+                method="patch",
+                code=404,
+                request={"state": "20", "target": "Test translation"},
+            )
+
     def test_translate_unit_whitespace(self) -> None:
         unit = Unit.objects.get(
             translation__language_code="cs", source="Hello, world!\n"
--- a/weblate/api/views.py
+++ b/weblate/api/views.py
@@ -3845,7 +3845,13 @@ class UnitViewSet(viewsets.ReadOnlyModelViewSet, UpdateModelMixin, DestroyModelM
 
         # Handle translate
         if do_translate:
-            unit.translate(user, new_target, new_state)
+            try:
+                unit.translate(user, new_target, new_state)
+            except Unit.DoesNotExist as error:
+                # The unit can be removed by a concurrent component update between
+                # the initial lookup and the locking re-fetch in Unit.translate()
+                msg = "Unit was removed while processing the request"
+                raise Http404(msg) from error
 
     def destroy(self, request: Request, *args, **kwargs):
         """Delete a translation unit."""


nijel:
@codex review

nijel:

alafilearnstocode:
Thanks, glad to help!

PR #21089, opened 2026-08-07
feat(ui): expand appearance customization options

### LINKED ISSUE #16917: More color customization options
### Describe the problem

As far as I can tell it is currently not possible to modify the primary button color, while it is possible to set the hover color of primary buttons.

Additionally, the color of the progress bar for the translation progress of projects / components / languages can also not be changed.

### Solution brainstorm

Having 2 more customization options in the Appearance settings would be nice (for the primary button color and the progress bar color).

For completeness it might also make sense to be able to set the background color of the progress bar. However, this is optional IMHO since it already uses a neutral color.

I know that it is possible to customize Weblate directly [with python modules](https://docs.weblate.org/en/latest/admin/customize.html) and it should probably be possible to overwrite the CSS file where the colors are defined (probably in [variables.css](https://github.com/WeblateOrg/weblate/blob/main/weblate/static/styles/variables.css)).

However, this is will make updating harder since we would have to check every version if there are style updates.

### Screenshots

Image

Image

### Additional context

_No response_

KarenKonou:
closes https://github.com/WeblateOrg/weblate/issues/16917


nijel:

Issue #21152, opened 2026-08-12
Allow defining multiple keywords for gettext pot extractor


Xenira:
### Describe the problem

The keyword option in gettext can be defined multiple times to define more keywords.
The addon currently only allows defining one additional keyword.

### Solution brainstorm

Allow defining a list of keywords and add the `--keyword` option for each of these keywords.

### Additional context

https://github.com/WeblateOrg/weblate/blob/03f08b9ad749f0954ac61e4cdee6f753ebc24f4a/weblate/addons/gettext.py#L1123

I can try implementing this if desired.

nijel:
Definitely makes sense to add. We just didn't want to implement all that xgettext supports from the beginning to limit scope and to implement only features people actually use.

lavneethora:
Hi! I'd like to work on this as my first contribution. Planning to add support for a list of keywords in the gettext addon per the issue description. Let me know if there's anything specific I should keep in mind before I start.

nijel:
There already is https://github.com/WeblateOrg/weblate/pull/21246 covering this.

PR #21214, opened 2026-08-17
feat: support multiple gettext keywords in xgettext addon

mmustafasenoglu:
## Description

Allow defining multiple extra keywords for the gettext POT extractor. Previously the xgettext addon accepted a single keyword; this change lets users pass several newline-separated keywords, each emitted as `--keyword=<kw>`.

Closes #21152

## Changes

- `weblate/addons/forms.py`: `keyword` field switched to a textarea, one keyword per line; cleaned data stored as a list (backwards compatible with stored string values).
- `weblate/addons/gettext.py`: `get_keyword()` → `get_keywords()` returning a list; each keyword emitted as its own `--keyword=` argument. Bare `--keyword` is still emitted exactly once when exclusive mode is on.
- `weblate/addons/tests.py`: updated existing asserts and added tests for form multi-keyword parsing, list round-tripping, and multi-keyword command generation (normal and exclusive).

## Testing

- `pytest weblate/addons/tests.py -k xgettext` → 48 passed
- `pytest weblate/addons/tests.py -q` → 475 passed, 3 skipped

PR #21218, opened 2026-08-17
feat(xgettext): allow multiple keywords in pot file update

Xenira:
Refs: #21152

See issue. Kept the config option name for backwards compatibility.

First contribution here. Let me know if I missed something.

### DIFF
--- a/weblate/addons/gettext.py
+++ b/weblate/addons/gettext.py
@@ -1103,8 +1103,8 @@ class XgettextAddon(ExtractPotBaseAddon):
             return [str(check) for check in checks]
         return []
 
-    def get_keyword(self) -> str:
-        return str(self.instance.configuration.get("keyword", ""))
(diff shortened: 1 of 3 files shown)


weblate/addons/gettext.py
```diff
@@ -1103,8 +1103,8 @@ def get_checks(self) -> list[str]:
             return [str(check) for check in checks]
         return []
 
-    def get_keyword(self) -> str:
-        return str(self.instance.configuration.get("keyword", ""))
+    def get_keywords(self) -> list[str]:
+        return self.instance.configuration.get("keyword", [])
```
nijel, line 1107: Existing configurations will have a string here, so it needs to be converted.

nijel (commented):
Thanks for your pull request. https://github.com/WeblateOrg/weblate/pull/21214 is a better approach, as it correctly handles existing configuration.

PR #21229, opened 2026-08-18
feat: improve team language assignment

### LINKED ISSUE #18776: Assigning languages to team: Improve UX
### Describe the problem

Goal: I wanted to create a team that has access to only specific roles, specific languages and specific components.

* I started selecting the role
* I started searching for my language and was confused to see so many of them
* after finding the "normal" German, I chose it
* I searched for and selected the components for this language
* I saved
* I went to the team overview and was surprised that no language was selected for the newly created team
* I edited it and added German again
* I saved it
* went back to the team overview and wondered why German was still not selected
* I edited it again
* Added German
* saved
* checked immediately if German was still there, but no it was gone
* I tried to add it again
* saved again
* it was still gone
* I realized that the selection has a dropdown with only two options above that overrides my language selection

Problems with this implementation:

* There is inconsistency between the Language selection and the components. The components infer "all components" by leaving the selection empty, while the languages do not. This is confusing and since the component selection worked, I didn't consider to check if I missed an option but immediately suspected another issue.
* There are a lot of language variants, but in most cases I want the most common one. Instead, I need to scan over more than 15 available options to pick the correct one.
* A dropdown with only two options does not really save space, but requires an unnecessary click to check my options, and a second click to choose it. If a menu only has two options, checkboxes are often a simpler choice, and with few options, radio buttons also make a lot of sense.

### Solution brainstorm

There are multiple options:

1. Only show or enable the language selection when the above selection is "as chosen", as otherwise the state does not get applied.
2. Automatically restrict languages when any selection is made, similar to the Components selection.
3. Convert the dropdown into a boolean option.

My personal favourite would be the following combination, but it might be best to choose whatever best matches Weblate's other menus for a basic level of consistency. Personally, I cannot see much consistency in the options, but I suppose there is a general way how menus are designed.

* Add a checkbox that says "All languages" above the language chooser
* Add the language chooser below
* if the checkbox is enabled, disable the language chooser (but do not hide it, so users scanning for it find the checkbox)
* (optional, if possible): still save the languages, so that enabling the checkbox and later disabling it again does not delete all existing state

see above.

### Screenshots

The situation that I saved is depicted in this screenshot:

Image

### Additional context

I'm managing several Weblate projects on Codeberg's Weblate instance (v5.16.2). Most of them have different needs and I often have to modify the settings.

KarenKonou:
- Adds an "all languages" checkbox, disabling the language picker if it is not set
- Preserves language selection even if all languages is set, improving team language management

closes #18776


nijel:

PR #21245, opened 2026-08-19
ci(pylint): use prebuilt development image

nijel:
Avoid fetching APT dependencies for every pylint run. The pinned development image already contains the required native build packages, reducing exposure to transient repository failures.


nijel:
While this would certainly help in case the APT mirror is unreliable, it would need maintaining a CI compatible image that would mimic a lot from what GitHub provisioned images do.

PR #21250, opened 2026-08-20
fix: detect parameterized and multiline BBCode tags

### LINKED ISSUE #21176: Improve support for BBCode tags (e.g. parameters, self-closing tags)
### Describe the problem

In Godot Engine's class reference, we make great use of [BBCode](https://www.bbcode.org/reference.php) to format text. A lot of go beyond the standard `[b]`, `[i]`, `[code]` tags and require additional parameters.
For example, the `[url]` tag can optionally accept a link, formatted as `[url=https://github.com/WeblateOrg]`. 
Our custom `[codeblock]` tag can also accept a language, formatted as `[codeblock lang=csharp]`.

Unfortunately, the "highlighting" and "error reporting" systems cannot detect these tags.
This can make it quite awkward to change several strings. BBCode-related changes are harder to spot, tags may be mistyped or missing in translations, shortcuts such as `CTRL + number key` to quickly insert tags do not work.

### Solution brainstorm

I believe what mostly needs to change is this regex in this repository:

https://github.com/WeblateOrg/weblate/blob/2b4eab375cd29cb0818922a62b5a9c5ca9d0c527/weblate/checks/markup.py#L76-L78

With this, I figured out that allowing characters after the tag name would do the trick. For some reason, and against any implementation of BBCode I've ever seen myself, `@` followed by any text was already allowed. Allowing white space and `=` matches the [general specifications](https://www.bbcode.org/reference.php) better, allowing `[url]`, `[color]`, `[img]`, etc. tags with their parameters to be detected.

```regex
(?P<start>\[(?P<tag>[^]]+)([@\s=][^]]*)?\])(.*?)(?P<end>\[\/(?P=tag)\])
```

Image

This still doesn't solve all problems, but it'd be a nice start.
That is to say, **tags that do not need a closing tag** would still not be accounted for. While that's not necessarily part of the standard, it would still be very nice (as far as Godot Engine is concerned).

Furthermore, without additional adjustments in the proposed regex, tags that begin in one line and end in another line would also not be accounted for:
```
[custom]My
String[/custom]
```

The ability to define custom BBCode tags in a repository would've also solved this, but it'd be quite the undertaking, especially since Weblate is [well-aware that BBCode support is quite rudimentary](https://docs.weblate.org/en/weblate-5.3.1/user/checks.html#bbcode-markup).

### Screenshots

_No response_

### Additional context

_No response_

mitre88:
BBCode matching now accepts tag parameters (`[url=…]`, `[codeblock lang=…]`) and multiline bodies.

The previous regex only allowed an optional `@…` suffix and used `re.MULTILINE`, which does not make `.` match newlines. This follows the regex from #21176 and adds `re.DOTALL` as noted by @akien-mga.

Self-closing tags are intentionally out of scope.

Fixes #21176


nijel:
@codex review

mitre88:
Addressed the Codex notes: pairing is now a linear open/close stack instead of a DOTALL lazy-dot regex. Nested tags inside parameterized (and multiline) containers are extracted, and unmatched openers no longer rescan the rest of the string.

mitre88:
Addressed the Codex notes: pairing is now a linear open/close stack instead of a DOTALL lazy-dot regex. Nested tags inside parameterized (and multiline) containers are extracted, and unmatched openers no longer rescan the rest of the string.

mitre88:
Follow-up on Copilot: parameterized closers like `[/url=other]` are ignored (only `[/tag]` closes), and the `BBCODE_MATCH` alias is gone because a token regex is not a pair matcher. The reverse stack walk stays; BBCode nesting depth is tiny.

PR #21310, opened 2026-08-25
Use prompt caching in the Anthropic machinery

### LINKED ISSUE #21309: Anthropic machinery: use prompt caching (cache_control) to reduce translation costs
### Describe the problem

The Anthropic machinery sends the system prompt as a plain string, without any cache_control block. Anthropic's prompt caching is opt-in, so nothing gets cached at all right now.

The system prompt built in BaseLLMTranslation (persona + style + language instructions + the JSON schema boilerplate) is identical for every request to a given target language. With a longer style guide it easily reaches 2-3k tokens. When automatic translation runs over a few thousand strings, that same prefix is billed at full input price on every single request. In our setup (~3.4k strings per language, ~8KB style prompt) this repeated prefix is the majority of the MT spend.

### Solution brainstorm

Send the system prompt as a content block with a cache breakpoint in weblate/machinery/anthropic.py:

```python
"system": [
    {"type": "text", "text": prompt, "cache_control": {"type": "ephemeral"}},
],
```

Cached reads are billed at roughly 10% of the normal input price with a 5 minute sliding TTL, which fits automatic translation nicely since the requests come back to back. The variable parts (glossary, few-shot examples, source strings) are already in the messages after the system prompt, so one breakpoint on the system block is enough.

There is a minimum cacheable prefix (about 1024 tokens) below which the API silently ignores cache_control, so the change should be safe to apply unconditionally.

Docs: https://docs.claude.com/en/docs/build-with-claude/prompt-caching

Running a small reverse proxy (configured through the machinery base_url setting) that injects cache_control into outgoing requests. It works, but it adds a moving part that Weblate could just handle out of the box with a one line change.

### Screenshots

_No response_

### Additional context

_No response_

jeenadeepak:
Fixes #21309

The Anthropic machinery sends the system prompt as a plain string, so Anthropic's prompt caching (which is opt-in per request) never kicks in. The system prompt is identical for every request to a given target language, and with a bigger style guide it is easily a few thousand tokens billed at full input price on every request during automatic translation.

This change sends the system prompt as a content block with cache_control set to ephemeral. Cached reads are billed at roughly 10% of the input price with a 5 minute sliding TTL, which matches back-to-back automatic translation requests well. Below the minimum cacheable prefix (about 1024 tokens) the API silently ignores cache_control, so the change is safe unconditionally.

Includes a test asserting the request payload shape and a changelog entry.


nijel:
@codex review

PR #21323, opened 2026-08-27
fix(vcs): uninstall GitHub App on account removal

amCap1712:
Uninstall the GitHub App through its authenticated installation API when the final Weblate workspace connection is removed, including when that connection goes away together with its workspace. Decide whether the connection is the last one under a row lock so that two workspaces removing it at once cannot both leave the App installed.

Keep shared installations active, and remove the connection even when GitHub no longer knows the installation or no longer accepts the App credentials, as there is nothing left to uninstall and keeping it would make the connection impossible to remove. Preserve it only when GitHub could not be reached at all, so that the removal can be retried.

Warn about the components that lose access to their repositories before confirming the removal.


nijel:

PR #21332, opened 2026-08-28
fix: editing regex flags in flag editor

### LINKED ISSUE #21312: Unable to add/remove tag with regex containing comma
### Describe the issue

When attempting to add a new regex flag with a comma from the web interface after #19716 was merged, the flag gets split in half as the comma is also used as a separator character. This also creates an issue with the bulk edit functionality; for existing regex tags with commas, it becomes impossible to remove a large amount of them without manually going through every individual translation.

### I already tried

1. Go to a page with a box for editing labels (e.g. the "Edit additional string info" modal when translating a string, the "Bulk edit" page)
2. Attempt to add a regex tag containing a comma (e.g. `regex:^[-_\p{L}\p{N}\p{sc=Deva}\p{sc=Thai}]{1,32}$`, which is [listed in the documentation](https://docs.weblate.org/en/latest/user/checks.html#check-regex))
3. See that it splits the tag on the comma (with the previous example, the regex tag becomes 2 different tags: `regex:^[-_\p{L}\p{N}\p{sc=Deva}\p{sc=Thai}]{1` and `32}$`)

### Expected behavior

I expected to be able to use a comma inside the regex without its usage as a separator character conflicting, like it was possible to do before #19716 was merged.

### Screenshots

Image

### Exception traceback

```pytb

```

### How do you run Weblate?

Other

### Weblate version

2026.8.1

### Weblate deploy checks

```shell

```

### Additional context

Instance used: https://translate.codeberg.org

KarenKonou:
fixes #21312

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -38,6 +38,7 @@ Weblate 2026.9
 * GitLab merge request forks now disable Git LFS to avoid missing-object push failures. See :ref:`git-lfs`.
 * :ref:`Project backup restores <projectbackup>` now allowlists repository metadata for Git, git-svn, and Mercurial to prevent archives from supplying executable configuration or repository indirection.
 * Notification e-mails now wrap long strings instead of overflowing, keeping the :guilabel:`View` button reachable without horizontal scrolling.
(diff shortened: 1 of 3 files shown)


nijel:

PR #21336, opened 2026-08-30
feat(file-format-params): JSON customize case insensitive sort

### LINKED ISSUE #15591: Customize JSON output / Sorts case sensitive
### Describe the problem

When I use the `Json Output Addon` with `Sort JSON keys` enabled, the keys aren’t sorted the way I expect.

They are sorted like this:
```
AA -> AZ -> Ab
```

But I would expect:
```
AA -> Ab -> AZ
```

Is this the intended behavior?

### Solution brainstorm

Would be great to add an option to customize the logic that is used to sort.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

gersona:
- fixes #15591
- **add JSONDecodder for custom sorting**
- **handle legacy backups with boolean json_sort_keys**
- **documentation update**


nijel:
@codex review

weblate/trans/file_format_params.py
```diff
@@ -205,19 +206,51 @@ class JSONOutputCustomizationBaseParam(BaseFileFormatParam):
...
-            file_format_params
-        )
+        dump_args = cast("JsonFile", store).dump_args
+        sort_mode = self.get_value(file_format_params)
+        if sort_mode == "case_sensitive":
+            dump_args["sort_keys"] = True
+        elif sort_mode == "case_insensitive":
+            # turn off JSONFile sorting which uses Python's default key ordering (cae sensitive)
+            dump_args["sort_keys"] = False
+            dump_args["cls"] = CaseInsensitiveSortingEncoder  # type: ignore[typeddict-item]
```
nijel, line 253: This would deserve fixing the annotation in the translate-toolkit. Please submit pull request there. But this is not a blocker here.

PR #21453, opened 2026-09-04
fix: persist the translate form in more cases

### LINKED ISSUE #19444: Clicking 'View' in the 'Things to check' section causes the loss of any previously made changes
### Describe the issue

Clicking '_View_', and only that in the '_Things to check_' section that is located on the right-sided panel is enough to cause the loss of any previously made changes, even though this action is only meant to display the targeted object.

Based on my observation, there are three **types of change** possible in the typical case, which is when translation approval is disabled:
- modifying translation
- activating _'Needs editing'_ check-box
- modifying translation explanation, which is displayed on glossary pages

### I already tried

1. Make a change
2. Click '_View_'
3. Observe the effect on the previously made change

### Expected behavior

No loss of changes when clicking '_View_'

### Screenshots

_No response_

### Exception traceback

```pytb

```

### How do you run Weblate?

weblate.org service

### Weblate version

5.17.1

### Weblate deploy checks

```shell

```

### Additional context

_No response_

KarenKonou:
also fixes a few more occurances of tab triggers outside of nav blocks
fixes https://github.com/WeblateOrg/weblate/issues/19444

### DIFF
--- a/weblate/static/editor/full.js
+++ b/weblate/static/editor/full.js
@@ -232,25 +232,50 @@
 
     /* Form persistence. Restores translation form upon comment submission */
     const restoreKey = "translation_autosave";
+    const stateSelector =
+      'input[name="fuzzy"], input[name="review"], [name="explanation"]';
     const restoreValue = localStorage.getItem(restoreKey);
     if (restoreValue !== null) {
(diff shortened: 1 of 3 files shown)


nijel:

PR #21509, opened 2026-09-06
feat(api)!: add user and IP rate limit overrides

### LINKED ISSUE #21351: Whitelist an IP for API rate limit
### Describe the problem

We have a server making API requests onto weblate to automate our translation process. However, we often reach the API limit rates. We could increase them but it’s only one server that need an higher rate.
On the `/api.html#api-rate` page, it is mentioned

> Rate limiting can be adjusted in the settings.py; see [Throttling in Django REST framework documentation](https://www.django-rest-framework.org/api-guide/throttling/) for more details how to configure it.

This however does not really describe how to change the throttling mechanism in a more advance way.

### Solution brainstorm

This is the code, we have came up to whitelist an IP:

```py
SAFE_LIST_IPS = [
    "1.2.3.42",
]

from weblate.api.throttling import UserRateThrottle

class CustomUserRateThrottle(UserRateThrottle):
    def allow_request(self, request: Request, view):
        if request.META['REMOTE_ADDR'] in SAFE_LIST_IPS:
            return True
        return super().allow_request(request, view)

REST_FRAMEWORK["DEFAULT_THROTTLE_CLASSES"] = ('weblate.settings_docker.CustomUserRateThrottle', 'weblate.api.throttling.AnonRateThrottle')
```

Let me know which one you prefer:

- adding the above snippet in the doc
- make a PR to add a constant `THROTTLE_WHITELIST_IPS` or something in `weblate.api.throttling.UserRateThrottle` to customise it easily
- make a PR in django rest framework
- do nothing, this is too niche

nijel:
Allow automation accounts and anonymous CI clients to use tailored API limits or exemptions through Weblate settings and Docker environment variables. Document configuration, validation, and migration for 2026.10.

BREAKING CHANGE: Weblate throttles now use API_RATELIMIT_ANON and API_RATELIMIT_USER instead of DEFAULT_THROTTLE_RATES. Remove anon_throttle and user_throttle arguments from get_drf_settings.

Fixes #21351

### DIFF
--- a/docs/admin/config.rst
+++ b/docs/admin/config.rst
@@ -2181,6 +2181,69 @@ The default setting is:
     ]
 
 
+.. setting:: API_RATELIMIT_ANON
+
+API_RATELIMIT_ANON
+------------------
+
+.. versionadded:: 2026.10
+
+Default anonymous :ref:`API rate limit <api-rate>`. Defaults to ``"100/day"``.
+Rates use a request count and a period of seconds, minutes, hours, or days, for
+example ``"100/hour"``. A zero count rejects every request covered by this
+throttle; ``None`` disables this throttle. Anonymous requests without an IP
+override are also subject to :setting:`API_RATELIMIT_USER`.
+
+.. setting:: API_RATELIMIT_USER
+
+API_RATELIMIT_USER
+------------------
+
+.. versionadded:: 2026.10
+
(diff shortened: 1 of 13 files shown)


mart-e:
Thanks for the quick addition :+1:

PR #21542, opened 2026-09-08
refactor(widgets): move font scale constant to where its used

michael-smt:
Makes it easier to see the actual constants used in font rendering.


nijel:

PR #21563, opened 2026-09-09
feat: improve machinery tab

### LINKED ISSUE #16345: Show context in translation memory matches
### Describe the issue

Hi, I noticed that on <https://hosted.weblate.org>, the similarity for _identical_ strings seems to be capped at 95% instead of showing 100%.

Maybe this points to a general issue with the similarity calculation.

### I already tried

1. Go to any project hosted on <https://hosted.weblate.org>
2. On a translatable string that occurs multiple times, click on _Automatic suggestions_.
3. See similarity of just 95% even if the same string as multiple occurrences.

### Expected behavior

The similarity between identical strings should be 100%.

### Screenshots

This is a randomly selected screenshot, it's easy to find other examples:

Image

### Exception traceback

```pytb
n/a
```

### How do you run Weblate?

weblate.org service

### Weblate versions

Live version of <https://hosted.weblate.org> at the time of opening this issue (Sep 28, 2025).

### Weblate deploy checks

```shell

```

### Additional context

_No response_

KarenKonou:
Improves the machinery tab by rewriting it to a suggestion-like view for a cleaner look and to show additional info such as context
closes https://github.com/WeblateOrg/weblate/issues/16345


nijel:
It always shows me the keyboard shortcut indication:

obrazek

When there are multiple matches, the percentage is shown on top while it only applies to translation memory match:

obrazek

So I think the label should be `Service name` or `Service name (99%)` and it should be in bold.

Furthermore, can you please add a Selenium test to cover this so that we can have this covered by Argos? (Which is really noisy here because of the search bar alignment changes.)

nijel:
@codex review

PR #21573, opened 2026-09-10
fix(ux): improve History revert/details button spacing

### LINKED ISSUE #8259: Revert buttons are unclear
### Describe the issue

View button and Revert button are too close together. According to the CSS code below, the gap is 4 pixels. So, I sometimes press the wrong button.
`.btn { padding: 2px 2px; }`

The buttons have around 27 pixels.

Also, I'm unaware reverted it because there is no notice. Perhaps others also have done so. To build a cooperative relationship, this mishandling should be prevented.

### I already tried

- [x] I've searched for similar issues in this repository.

1. Login.
2. This occurs when I am looking at History for all strings or the History tab for each string.
3. In the screenshot below, I try to press the magnifying glass. But I made a mistake and pressed the arrow button. (Or I just press the arrow button.)

image

### Expected behavior

There is no magnifying glass button when displaying each string. It is self-linked.

**Solution 1** : Confirm as an important action. This would be essential.
* Solution 1a : Dialog "Would you like to revert?". It also indicates which string to return to.
* Or, Solution 1b : Simply copy the text into the translation string field. No automatic saving.

**Solution 2**: Notify : But, this doesn't prevent a wrong press and auto saving.
* Notify that "the string has been reverted". Especially in the case of automatic saving. Otherwise, the user may think pressed "View detail" button. So, does not recognize the occurrence of a problem.

**Solution 3** : Button design : But, this doesn't prevent a wrong press and auto saving.
* Widen the gap between View button and Revert button. The gap should be at least 10 pixels. Or/and, If design changes are possible, This button should be at least 36-42 pixels. 
* Or/and, Currently two buttons are placed within one border. Use one border for each button.
* Or/and, Currently the picture changes color when focused. Add to this, the background color will also change when focused.

### Exception traceback

_No response_

### How do you run Weblate?

weblate.org service

### Weblate versions

_No response_

### Weblate deploy checks

_No response_

### Additional context

_No response_

diptomistry:
## Summary
- Scope History `.btn-float` actions to `#changes-table`: 10px gap, per-button border/padding, hover/focus background.
- Changelog entry under Weblate 2026.10 Improvements.
Fixes #8259
## Test plan
- [x] Open project History / changes browse: Revert and View details are spaced and separately bordered
- [x] Open per-string History tab: same button treatment
- [x] Keyboard focus shows background change on the action buttons
- [x] Suggestions/comments `.btn-float` outside History unchanged
## Screenshots
before:

before-btn-float-crop

after: 
after-btn-float-crop


nijel:
https://github.com/WeblateOrg/weblate/issues/8259 also brought another question: the clarity of these icons, so while this might improve the situation, I'm not sure if it fully addresses the concerns brought there.

diptomistry:
Agreed — this addresses the spacing/misclick part of the issue (solution 3, per your original guidance), but icon clarity is a separate, larger question. Happy to leave that as a follow-up issue if it's worth tracking separately, rather than expanding this PR's scope.

weblate/static/styles/main.css
```diff
@@ -868,6 +868,31 @@ form + .login-label {
...
+#changes-table .btn-float .btn:hover,
+#changes-table .btn-float .btn:focus,
+#changes-table .btn-float .btn:focus-visible {
+  background-color: var(--wl-color-row-hover);
+  border-color: var(--wl-color-border-strong);
+  text-decoration: none;
+}
+
+#changes-table .btn-float .btn:focus path,
+#changes-table .btn-float .btn:focus-visible path {
```
KarenKonou: This has higher specificity than `.red path` used by the block user action so it unintentionally changes its color when keyboard focusing it (hover is unaffected)
```suggestion
#changes-table .btn-float .btn:not(.red):focus-visible path {
```

KarenKonou (changes requested):
Overall I think this is fine in terms of making the buttons more visible, if we want to increase their size we can easily do that at any point.

diptomistry:
Good catch, applied — thanks!

KarenKonou (approved):

nijel:

PR #21576, opened 2026-09-10
docs(checks): max-length description update

michael-smt:
I think saying "amount" makes it clearer that this check counts chars instead of measuring something.

Also, the check is already listed in the auto-generated `check-flags-enables` section in the docs snippet, so drop it from the manually maintained listing.


docs/snippets/check-flags-autogenerated.rst
michael-smt, line 129: FYI the auto-generated line is here.

nijel:
@codex review

nijel:

PR #21587, opened 2026-09-11
feat(api): expose and manage unit screenshots

### LINKED ISSUE #6080: Screenshot link for units (API)
Hello everybody,

I use the Weblate API for the external recording app. In the project, sometimes Voice Over artist wants to see the screenshot before the recording. I am already using the "Screenshot feature" in the Weblate. But we can not take the screenshots links special for the units via API. So, it can be nice to see a `screenshot_link` field for these endpoints; 

```/api/translations/PROJECT/COMPONENT/LANGUAGE/units/```
```/api/units/UNIT_ID```

Thanks!

### PR DESCRIPTION
**Pull request by nicommteam**
Closes #6080

## What changed
Adds a `screenshots` field to the Unit API serializer, listing the
screenshots associated with a translation unit (`GET /api/units/{id}/`
and the nested units under `/api/translations/.../units/`).

Also adds two new actions on `UnitViewSet`, mirroring the existing
`/api/screenshots/{id}/units/` actions from the other side of the
relation:
- `POST /api/units/{id}/screenshots/` — associate a screenshot with
  the unit (`screenshot_id` in the request body)
- `DELETE /api/units/{id}/screenshots/{screenshot_id}/` — remove the
  association

Adding a screenshot validates that it belongs to the **same
translation** (component + language) as the unit — a mismatched
screenshot is rejected with a 400, matching the equivalent validation
already done on the screenshot-side endpoint.

## Why
Per @nijel's comment on #6080: exposing screenshot links lets external
tools (e.g. the API-driven recording workflow described in the issue)
fetch a unit's screenshots directly, and the write endpoints let
integrations manage screenshot↔unit assignments from the unit side
without needing to already know a screenshot's translation context.

## Testing
Added 6 tests in `UnitAPITest` covering: reading the new field,
permission denial (401), invalid screenshot id (400), the
component/language mismatch validation (400), successful add (200 +
Change log entry), and successful delete (204 + Change log entry, plus
404 for a non-associated screenshot id). Full `UnitAPITest` class: 31
passed.

### CODE REVIEW

File: docs/changes.rst [hunk 1]
```diff
@@ -15,7 +15,7 @@ Weblate 2026.10
 * Improved :ref:`repository maintenance <repository-maintenance>` with disabled push controls when push configuration is missing and direct links to component VCS settings.
 * The :ref:`automatic translation add-on <addon-weblate.autotranslate.autotranslate>` can create approved strings, falling back to translated strings when reviews are disabled for the target language.
 * Whitespace characters are now rendered consistently in the source string display and the translation editor, and different kinds of whitespace are now distinguishable from each other.
-* History :guilabel:`View details` and :guilabel:`Revert` actions are larger, more widely spaced, and show a hover and focus background.
```
nijel, line 18: This removed existing changelog entry.

nijel:
@codex review and security review

weblate/api/serializers.py
```diff
@@ -3634,6 +3634,11 @@ class UnitSerializer(serializers.ModelSerializer[Unit]):
     source_unit: serializers.HyperlinkedRelatedField[Unit] = (
         serializers.HyperlinkedRelatedField(read_only=True, view_name="api:unit-detail")
     )
+    screenshots: serializers.HyperlinkedRelatedField[Screenshot] = (
+        serializers.HyperlinkedRelatedField(
+            many=True, read_only=True, view_name="api:screenshot-detail"
+        )
+    )
```
nijel, line 3641: I think it would be more consistent with the POST/DELETE API to return a list of screenshots on the /screenshots/ URL.

weblate/api/views.py
```diff
@@ -4191,6 +4191,57 @@ def translations(self, request: Request, *args, **kwargs):
...
+            )
+        except (TypeError, ValueError) as error:
+            raise invalid_integer_error(field_name) from error
+        except Screenshot.DoesNotExist as error:
+            raise not_found_validation_error(field_name, "Screenshot") from error
+
+        screenshot.add_unit(unit, user=request.user)
+        serializer = UnitSerializer(unit, context={"request": request})
+
+        return Response(serializer.data, status=HTTP_200_OK)
```
nijel, line 4233: Returning a unit here seems confusing. Having list of associated units (as the /screenshots/ URL suggested before) would be more expected.

nicommteam:
Thanks for the review, @nijel and @codex — addressed everything in commit 54295f1 (merged in with 4cfa3a66af after reconciling with the auto-generated openapi.yaml commit):

- **N+1 query (P1)**: Added `"screenshots"` to `UnitQuerySet.prefetch_api()`.
- **Schema falls through to `UnitWriteSerializer` (P2)**: Added `UnitScreenshotAssociationSerializer` (request) and `UnitScreenshotsSerializer` (response), wired via explicit `request=`/`responses=` on `@extend_schema`. Regenerated `docs/specs/openapi.yaml`.
- **False `SCREENSHOT_REMOVED` audit entry on no-op delete (P2)**: `delete_screenshots` now checks the association exists before calling `remove_unit()`.
- **Duplicate `SCREENSHOT_ADDED` on retried POST (P2)**: `screenshots` now checks the association doesn't already exist before calling `add_unit()`.
- **API response shape**: `POST`/the associated response now return `{"screenshots": [...]}` instead of the full `Unit` object, matching the suggested design.
- Fixed a bug the prefetch change introduced: after adding an association, the unit's cached `screenshots` prefetch was stale, so the response showed an empty list. Added `unit.refresh_from_db()` before serializing.
- Restored the History view/revert changelog line that got accidentally dropped from `docs/changes.rst` in an earlier push.
- Added two new tests covering the idempotency fixes (`test_unit_add_screenshot_idempotent`, `test_unit_delete_screenshot_noop_no_audit`).

All 33 tests in `UnitAPITest` pass, and `prek` is clean.

nijel:
@codex security review

nijel:
@codex review

nijel:
I still think the API shape should be different. There can possibly be many screenshots and that could slow down the units API. So instead of exposing `screenshots` there, add `screenshots_url` that will point to `/api/units/<int>/screenshots/` which will provide GET paginated list of associated screenshots. POST to this will work as implemented now and will return the added screenshot using the same serializer as GET.

nicommteam:
Implemented as suggested (commit 5bd23d2):

- `screenshots` field replaced with `screenshots_url` pointing to `/api/units/(id)/screenshots/`
- Added `GET` support on that endpoint — paginated list of associated screenshots, using `ScreenshotSerializer` (same as the `/api/screenshots/` list)
- `POST` now returns the single added screenshot via that same serializer, instead of the full unit object
- Regenerated `docs/specs/openapi.yaml`
- Also fixed `test_prefetch_api`, which asserts an exact DB query count — the new `screenshots` prefetch intentionally adds one query (6 → 7), test updated accordingly

All 34 tests pass (`UnitAPITest` + `test_prefetch_api`), `prek` is clean.

nicommteam:
Fixed (commit 7fd7339): `get_serializer_class()` now explicitly routes the `screenshots` action — `ScreenshotSerializer` for GET, `UnitScreenshotAssociationSerializer` for POST — and `delete_screenshots` to `ScreenshotSerializer`, instead of falling through to `UnitWriteSerializer`. Confirmed the browsable API's OPTIONS/form rendering now matches the documented schema, not just the `@extend_schema` annotation.

All 33 tests in `UnitAPITest` pass, `prek` is clean.

weblate/api/views.py
```diff
@@ -4196,6 +4207,104 @@ def translations(self, request: Request, *args, **kwargs):
...
+        methods=["get"],
+        responses=ScreenshotSerializer(many=True),
+    )
+    @extend_schema(
+        description="Associate screenshot with unit.",
+        methods=["post"],
+        request=UnitScreenshotAssociationSerializer,
+        responses=ScreenshotSerializer,
+    )
+    @action(detail=True, methods=["get", "post"])
```
nijel, line 4223: I thought the review asked for something like this (not tested):

```suggestion
    )
    @action(detail=True, methods=["get", "post"], serializer_class=ScreenshotSerializer)
```

And not for introducing additional logic to handle this just for the OpenAPI.

nicommteam:
Done (commit 00e7860): moved `delete_screenshots` to `serializer_class=ScreenshotSerializer` on the `@action` decorator, and made `get_serializer_class()` respect it via a generic fallback instead of hardcoding the action name — this also fixes the same latent bug for the existing `comments` action (its `serializer_class=CommentSerializer` was silently ignored the same way).

Kept `screenshots` branching explicitly on `request.method`, since GET (list) and POST (associate) genuinely need different serializers and a single `serializer_class=` on the decorator can't express that.

All 33 tests pass, `prek` is clean.

nijel:
@codex review

weblate/api/views.py
```diff
@@ -4196,6 +4209,105 @@ def translations(self, request: Request, *args, **kwargs):
...
+    @transaction.atomic
+    def screenshots(self, request: Request, **kwargs):
+        unit = self.get_object()
+
+        if request.method == "GET":
```
nijel, line 4228: `request.method` can also be `HEAD` here, which will fall back to `POST` handling this way. So better to invert the logic - check for `POST` and do the mutation there and fall back to `GET`.

PR #21596, opened 2026-09-11
feat(gettext): replace contributor add-on with format parameter

nijel:
Offer disabled, gettext, and SPDX contributor comments through PO file format parameters. Migrate existing scoped add-ons while preserving component settings, and cover API, commit, and populated upgrade behavior.

Depends on https://github.com/translate/translate/pull/6570


nijel:
@codex review

PR #21605, opened 2026-09-11
feat(components): add separate browser URL for translation files

### LINKED ISSUE #12308: Allow setting a separate repository browser url for translation files

krisnaparahita:
## Summary

- Add an optional `repoweb_translations` field to `Component`, including migration 0105, forms, and documentation.
- Thread an `is_translation` flag through `get_repoweb_link()` and `try_linkify_filename()` so translation-file links can use the separate browser URL.
- Fall back to the existing `repoweb` setting when the translation-specific URL is unset.
- Add two unit tests covering the translation-specific URL and fallback behavior.

## Testing

The two new unit tests were written but were not executed in this sandbox because PostgreSQL was not available. No successful Django test run is claimed here; CI still needs to verify them.

## Implementation note

PR #19184 was used only as a reference for the earlier approach. It was auto-closed for staleness and was not merged; this implementation was rewritten against current `main`.

Fixes #12308

## Update (2026-09-13)

- Fixed a migration-numbering conflict with `0105_contributor_comments` (renumbered to `0106`, depends on it now) — this was causing every DB-touching CI job to fail outright.
- Added `repoweb_translations` to the `ComponentSerializer` (REST API) and a changelog entry.
- Attempted to also include the field in project backup/restore, but reverted that part: the installed `weblate_schemas` package (separately versioned) doesn't yet recognize this field, so including it breaks schema validation on backup/restore (`Additional properties are not allowed ('repoweb_translations' was unexpected)`). Deferring backup/restore support until `weblate_schemas` ships a release with the new field — happy to follow up with that once this lands, rather than guessing at a fix here.

## Update (2026-09-14)

Addressed all outstanding review feedback:

- **Docs**: added the missing `repoweb_translations` entry to `docs/api.rst`'s response-field list (the OpenAPI schema itself was already picked up by a pre-commit hook).
- **Parse-error notifications**: `ParseErrorNotification` now correctly uses the translations browser URL for intermediate-file parse errors too (previously only target-translation errors were handled; template-file errors correctly still use the source browser).
- **Diagnostics**: `BrokenBrowserURL` now also validates `repoweb_translations` against a real non-source translation's filename, independent of the existing `repoweb` check.
- **Test coverage**: strengthened the `BrokenBrowserURL` tests to assert the full expected URL (including filename), not just the hostname — the prior version would have passed even if the wrong file were linked.
- **Backup/restore, final decision**: kept this deferred (not included). Including it breaks schema validation for *every* project export under the currently-installed `weblate_schemas` package version, not just for this field — confirmed this isn't a narrow, contained failure. `BackupSettingCoverageTest` now has a proper exception mechanism (`pending_fields`) for "form-exposed but intentionally not yet backed up" settings, rather than the ad-hoc exclusion from the last update. Happy to add backup/restore support as a fast follow once `weblate_schemas` has a release with this field — let me know if you'd rather sequence this differently (e.g. hold this PR until that's coordinated).


nijel:
@codex review

docs/specs/schemas/weblate-component.schema.json
```diff
@@ -152,6 +152,12 @@
           "title": "Repository browser",
           "type": "string"
         },
+        "repoweb_translations": {
+          "$id": "#root/component/repoweb_translations",
+          "title": "Repository browser for translations",
+          "type": "string",
+          "default": ""
+        },
```
nijel, line 160: The schema has to be changed in https://github.com/WeblateOrg/weblate_schemas/

nijel:
Can you please rename the migration so that it doesn't conflict? Also, the schema change needs to be submitted as a separate pull request to https://github.com/WeblateOrg/weblate_schemas/.

krisnaparahita:
> Can you please rename the migration so that it doesn't conflict? Also, the schema change needs to be submitted as a separate pull request to https://github.com/WeblateOrg/weblate_schemas/.

okay

nijel:
Thanks, the schemas will update in https://github.com/WeblateOrg/weblate/pull/21789, I think this should be ready to merge then after using the new schema.

weblate/accounts/notifications.py
```diff
@@ -790,8 +790,19 @@ def get_context(
...
         )
         if change and change.component:
+            filename = change.details.get("filename")
+            is_translation = (
+                change.translation is not None and not change.translation.is_source
+            ) or (
+                change.translation is None
+                and bool(change.component.intermediate)
+                and filename == change.component.intermediate
+            )
```
nijel, line 800: Is this logic really desired? Shouldn't the translation URL be used for all the translation files (no matter whether they are source, intermediate, or target)?

weblate/trans/tests/test_backups.py
```diff
@@ -127,10 +129,37 @@ def test_backup_setting_coverage(self) -> None:
...
+                },
+                # repoweb_translations is exposed in the settings form (it's
+                # a real, usable setting) but intentionally NOT yet backed
+                # up: the installed weblate_schemas package doesn't
+                # recognize the field, so including it in backup/restore
+                # breaks schema validation for every project export, not
+                # just this one field. Remove from both sets above once
+                # weblate_schemas ships a release with the field and it's
+                # added back to COMPONENT_BACKUP_FIELDS.
+                {"repoweb_translations"},
```
nijel, line 152: Preserve the new setting in project backups; the schema is now there.

nijel (changes requested):

weblate/accounts/notifications.py
krisnaparahita: You're right, thanks. I removed the source/target distinction. The translation browser URL is now used for all translation files (source, intermediate, template and target), with the fallback to `repoweb` when it is not set. This applies to the parse error notification, the broken browser URL diagnostic, and the editor view, and I updated the tests and docs to match. Done in 08432a1.

weblate/trans/tests/test_backups.py
krisnaparahita: Done. `repoweb_translations` is now part of the project backups (`COMPONENT_BACKUP_FIELDS`), and I removed the temporary exception from `BackupSettingCoverageTest`, so the coverage test now requires it. I also added a backup and restore round trip test for it. This uses the new schema from weblate-schemas 2026.9, and I merged current main into the branch. Done in b873b34.

PR #21624, opened 2026-09-12
feat(checks): add font-monospace flag for the translation editor

### LINKED ISSUE #5859: Monospace font

Some translations are for command-line tools that sometimes require alignment. It would be nice to have an option to switch to a monospace font, so counting spaces is easier :). A simple checkbox below the translation area or as an option in a dropdown menu would work well for this feature. It could be nice to be able to set a default for all languages on specific strings, but I could live without it.

So, I can either try and copy the right number of spaces (counting them), or use the web console to force using a monospace font, as shown below.

**Screenshots**

Screenshot_2021-04-15 guix guix — French(1)

With a monospace font, the alignment is clearer:

Screenshot_2021-04-15 guix guix — French(2)

Same here:

Screenshot_2021-04-15 guix guix — Dutch

and then you discover easily the alignment was incorrect:

Screenshot_2021-04-15 guix guix — Dutch(1)

**Additional context**

nicommteam:
Closes #5859

## What changed
Adds a `font-monospace` flag (see `docs/admin/checks.rst` → *Customizing behavior using flags*) that switches the translation editor's textarea to a monospace font. Set it on the source string via the existing flags UI/API.

## Why
Per nijel's comment on #5859: useful for strings containing command-line/terminal output, where character alignment matters and a proportional font makes misalignment hard to spot.

Deliberately does **not** reuse the existing `font-family`/`font-size`/`font-weight` flags — those configure custom uploaded fonts for the `max-size` rendering check (server-side image rendering). `font-monospace` only ever applies the browser's built-in `monospace` CSS font stack, avoiding the user-uploaded-font security/performance concern nijel raised in the same thread.

## Testing
Added `test_font_monospace_flag` in `BasicViewTest`: asserts the CSS class is absent by default, and present on the rendered editor once the flag is set on the source unit.

### DIFF
--- a/weblate/static/styles/main.css
+++ b/weblate/static/styles/main.css
@@ -291,6 +291,10 @@ textarea[name="explanation"] {
   min-height: 3em;
 }
 
+textarea.font-monospace {
+  font-family: monospace;
+}
+
 textarea#id_comment {
   min-height: 5em;
 }
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -15,6 +15,7 @@ Weblate 2026.10
 * Improved :ref:`repository maintenance <repository-maintenance>` with disabled push controls when push configuration is missing and direct links to component VCS settings.
 * The :ref:`automatic translation add-on <addon-weblate.autotranslate.autotranslate>` can create approved strings, falling back to translated strings when reviews are disabled for the target language.
 * Whitespace characters are now rendered consistently in the source string display and the translation editor, and different kinds of whitespace are now distinguishable from each other.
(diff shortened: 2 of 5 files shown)


weblate/static/styles/main.css
```diff
@@ -299,6 +291,10 @@ textarea[name="explanation"] {
   min-height: 3em;
 }
 
+textarea.font-monospace {
+  font-family: monospace;
```
nijel: Use configured font:

```suggestion
  font-family: var(--bs-font-monospace);
```

nijel:
@codex review

KarenKonou (approved):

nijel:

PR #21643, opened 2026-09-14
fix(machinery): preserve Ollama base URL paths

### LINKED ISSUE #21642: Ollama machinery service discards path component of custom base_url
## Problem

`OllamaTranslation.get_chat_url()` builds the chat endpoint with:

```python
return urljoin(self.settings["base_url"], self.end_point)
```

`end_point` is `"/api/chat"`. Because the path starts with `/`, Python's `urljoin` discards any path prefix configured in `base_url`.

## Reproduction

```python
from urllib.parse import urljoin

urljoin("https://gateway.example/ollama", "/api/chat")
# https://gateway.example/api/chat
```

The expected URL for a path-prefixed Ollama gateway is:

```
https://gateway.example/ollama/api/chat
```

This is the same URL-joining class of problem described in #20992 for the Anthropic machinery integration.

## Expected behavior

The Ollama machinery service should preserve a configured path prefix in `base_url`, while retaining the default Ollama URL behavior.

## Suggested direction

Use the shared base/path joining helper in `BaseLLMTranslation` (or an equivalent normalized join with a relative endpoint) and add regression coverage for:

- a standard base URL
- a custom base URL with a path prefix
- a base URL with a trailing slash

I am happy to prepare a focused PR with tests if this direction is acceptable.

avgon:
## Summary

- preserve path prefixes in custom Ollama `base_url` values
- join the Ollama endpoint as a relative path
- add a regression test for `http://localhost:11434/ollama`

Fixes #21642.

## Validation

- `python -m py_compile weblate/machinery/ollama.py weblate/machinery/tests.py`
- URL regression assertions for default and path-prefixed bases

I attempted `uv run pytest weblate/machinery/tests.py -k OllamaTranslationTest -q`, but the Windows environment could not build the repository's `tesserocr` dependency because the system Tesseract library is unavailable. The targeted test is included for CI to run in the project environment.


nijel:
@codex review

nijel:

PR #21685, opened 2026-09-16
feat: @mention autocomplete in translation comments

### LINKED ISSUE #21143: Suggest contributing users first when starting a mention
### Describe the problem

As a maintainer of translations, I sometimes want to ask contributors about their contribution. That can be done through the Comments tab. At the moment when I start typing a `@` symbol for the mention, no suggestions appear. I Suggestions appear when I start typing, but that seems to search through the whole list of known Weblate users. This means that when I'm on the History tab and see a user I want to discuss with, I have to take note of the entire name, because I can't view it once I moved over to the Comments tab to start to create a comment with a mention.

### Solution brainstorm

I'd like to have suggestions based on my project, the language and the string. There are about 50 people that contributed to the whole translation project, 5 that contributed to that language and maybe less that worked on that string. It would make sense to highlight those users first.

This is also the behavior I've become accustomed to from other collaborative software like git hosting and chat applications.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

mmustafasenoglu:
Add @mention autocomplete to translation comment forms.

- templates/translate.html: data-unit-id for unit-filtered API
- static/loader-bootstrap.js: Vanilla JS autocomplete (fetch, dropdown, keyboard nav, caret positioning)
- api/views.py: order_by_contributions queryset method
- api/tests.py: test_list_contributors_first test

Results: test_list_contributors_first passed; UserAPITest: 57 passed + 4 subtests; Biome: 0 errors; Ruff: all checks pass

Policy: Feature request - maintainer approval needed before merge


nijel:
@codex review

weblate/static/loader-bootstrap.js
```diff
@@ -2713,3 +2710,146 @@ onReady(() => {
     updateTarget();
   });
 });
+
+/* Mention autocomplete in comment forms */
+document.querySelectorAll("#comment-form textarea").forEach((commentInput) => {
```
KarenKonou: This conflicts with the existing @ mention autocomplete added in https://github.com/WeblateOrg/weblate/commit/03d0e1c1db

KarenKonou (changes requested):

weblate/static/loader-bootstrap.js
mmustafasenoglu: Good catch, thanks — you're right. The comment textarea already carries the `markdown-editor` class (via `MarkdownTextarea`), so the autoComplete-based mention from 03d0e1c1db attaches to it and my hand-rolled dropdown was pure duplication.

Fixed in 66b67c6: dropped the custom 140-line implementation and hooked the `unit` id into the existing lookup instead, so contributor ranking still applies. The backend (`order_by_contributions`) and its test are unchanged.

KarenKonou (approved):

mmustafasenoglu:
Good catches, all three addressed: anonymous requests now return early before touching the unit lookup, the unit parameter is in the OpenAPI schema (spec regenerated), and the test now covers the translation-wide tier too. UserAPITest is green locally (54 passed).

nijel:

PR #21753, opened 2026-09-21
feat(addons): add declarative workflow automation

### LINKED ISSUE #3315: Add-on ordering
In some situations it might be useful to define order to execute the addons. Typical use case might be automatic translation addons, but it might be useful elsewhere.

UI and backend to define ordering of the addons.

For machine translation services, there is hardcoded ordering existing, so for this particular use case the solution exists - install one addon with all desired backends and they will be executed in order.

### LINKED ISSUE #5770: Set priority for Machine translation engines in Automatic Translation Addon.
When the Automatic Translation Addon is configured for a component multiple engines can be activated (chosen).
In the scenario skiping suggestions and translating all Messages "needing action" right away it would be great to first look them up in the local Translation Memory and in a second step, all other Messages still needing action get a translation from another engine.

Using e.G. deepl and the Weblate Translation Memory there seem to be no way to change the order or priority.
I would want to make sure  the local Memory is used before the remote API is asked for a translation.
If Weblate by dedault considers local Memory at first anyway, this should be visible in the Addon's settings.

A good solution would be drag and drop capability for the list elements and some visible numbers, that indicate priority.

nijel:
Localization workflows need to respond to events, apply project-specific conditions, and coordinate operations in a predictable order. Declarative automation lets administrators express these workflows without maintaining custom Python add-ons.
    
This also enables explicit translation priorities, such as applying local translation memory before sending remaining strings to remote services.

Fixes #3315
Fixes #5770


nijel:
@codex security review

nijel:
@codex security review

nijel:
@codex review

nijel:
@codex please respond

PR #21766, opened 2026-09-22
fix(backups): clarify self-hosted backup purchasing

nijel:
Add a direct purchase link and setup guidance so self-hosted administrators can find the standalone backup service. Hide the offer on dedicated instances and when provisioned backup storage is already linked.


nijel:
@codex review

